PR #1, opened 2014-07-07
[ADD] standard files

hbrunn:


.coveragerc
```diff
@@ -0,0 +1,19 @@
...
+
+[report]
+omit =
+    /usr/*
+    */bin/*
+    */lib/*
+    */odoo/*
+    */openerp/*
+    */tests/*
+    *__init__.py
```
bwrsandman, line 8: A better regex for filenames is:

```
[report]
include =
    */OCA/web/*

omit =
    */tests/*
    *__init__.py
```

hbrunn:
how about this one? Am I supposed to merge this initial stuff myself or are you waiting until the pyPDF dependency is fixed?

.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+
+# Need coveralls for coverage reports
+# Need flake8 for pep8 testing
+# Manually get PyChart
+# Install tested version of odoo (official or ocb)
+# Get modules from other repos which have dependencies (in this case travel requires modules from lp:partner-contact-management and lp:openerp-hr
+install: 
+    - pip install coveralls flake8
+    - pip install http://download.gna.org/pychart/PyChart-1.39.tar.gz
+    - pip install ${ODOO}
```
bwrsandman, line 27:
```
    - pip install pypdf
```

hbrunn:
Thanks! Looks like web commits (I don't have a shell at hand currently) don't trigger a travis build. Maybe one of you guys could do that?

hbrunn:
yay! Please merge this and #13 fast so that the other PRs can proceed

.travis.yml
```diff
@@ -0,0 +1,46 @@
...
+
+# Test with flake, ignore F401 for __init__.py files, use a max length of 120
+# Run tests with coverage
+# Only test modules in repo (list populated by directories in repo)
+# Preload modules before testing to only run tests of repo's modules
+# Include current directory and dependent repos in addons-path as well as official addons
+script:
+    - flake8 . --max-line-length=120 --exclude=__unported__ --filename=__init__.py --ignore=F401
+    - flake8 . --max-line-length=120 --exclude=__unported__,__init__.py
+    - odoo.py -d test --stop-after-init --init=$(python -c 'import os; print(",".join(x for x in os.listdir(".") if os.path.isdir(x) and not x.startswith(".") and x != "__unported__"))') --addons-path=$(pwd),`python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"`/addons
```
red15, line 42: Would it not be better to have `--init all` instead of doing an inline python program?

hbrunn, line 42: This would cause all tests to run, but we only want the tests from the current directory (= addons in this branch)

bwrsandman, line 42: Precicely, however, this line does not run tests, it is the following one which has `--test-enable`.
The reason for this is to avoid testing dependency modules when they install.

bwrsandman:
LGTM
:+1:

.travis.yml
red15, line 42: @hbrunn running all tests is not a luxury, it's almost a requirement to ensure the basic functionality is still working as it should (remember how odoo cannot block incompatible addons)

@bwrsandman you've noticed something I didn't: isn't that line suppose to contain the `--test-enable` option? I can't seem to find it.

bwrsandman, line 42: @red15 we let odoo run their test on their own addons, eventually we will have an automated travis for OCB, but for now the purpose of travis in **an OCA addon repo** is to test the addons comprised inside. 

We do not test and **should not** test every addon in Odoo.

`--test-enable` is not included in the line you commented on because that line is the pre-loading line. It initializes the database and will fail if modules have missing dependencies, bad views, and more generic errors. This saves time and avoids testing modules which are dependencies of the modules in web.

The following line (L43) tests using coverage in order to publish which lines are run in **unittests** not during initialization. It will test using only the addons in the repo using the `python -c ...`.

Again, **running all tests is not the point of travis in web or any OCB addon repo**. We don't care if the unittests of an unrelated module fail. We don't want other repos which fail their own unittest to make modules in my PR fail. We don't want to wait 20 extra minutes to install every module under the sun every time we fix the view of one module in the OCB web addons module.

This is not even mentioning incompatible modules. That's a whole other problem.

hbrunn, line 42: I guess the concern is that it makes sense to run all tests in case some addon 
breaks core functionality (poorly done monkey patches, fuckups in BaseModel 
etc). That's a valid point I think, but that should not be covered by every 
single commit test.

The good approach for this would be an OCA-wide test run so that we can be 
sure that all OCA modules play nice with each other and with (possibly OCB) 
core. Something like an OCA-runbot.

bwrsandman, line 42: Unless dependencies are not properly set, running unittests on all modules won't pick up these errors since the tests are run from the base dependency up to the top if they are not yet installed.

In that case, the initialization should fail based on the fact we are only installing modules from the repo, and not the dependency. An edge case would be if some modules tested before do install it.

red15, line 42: @hbrunn I'm glad you got my point about poor monkey patches etc but it goes even further than that, if you don't run the --test-enable to setup your skeleton database you won't have demo data which typically is needed for some test scenarios.

I'm not trying to play hardball here, I'm just trying to point out the dangers that running a test scenario in another way than odoo runs there thus yielding different possible results.

bwrsandman, line 42: @red15 that is simply not true about the demo data.

`--without-demo=True` is the command which will load your db without demo data

As you can see here, the demo is loaded:

https://travis-ci.org/OCA/web/builds/29399441#L1149

hbrunn, line 42: I think this is an important discussion to have. But one we should have on the mailing list, as it is a general issue that is in no way specific to this branch.

So let's have those concerns expressed on the mailing list and get on with this PR, as it is the last one to have the migration for web-addons on github complete.

red15, line 42: @bwrsandman I was indeed mistaken on this.
@hbrunn Agreed

bwrsandman, line 42: Agreed

@red15 Thank you for voicing your opinion, Niels. Even if we disagree on certain points, these are indeed things which should be open to discourse and debate.

rvalyi:
:+1:

pedrobaeza:
:+1:

PR #9, opened 2014-07-07
[ADD] web_relativedelta

hbrunn:
Reimplementation of relativedelta for use in client side domains in the hope that it sucks less than the one provided in upstream.

This uses http://momentjs.com which was committed separately to simplify the review.


hbrunn:
superseded by #15

PR #16, opened 2014-07-14
[ADD] web_tree_many2one_clickable

hbrunn:
A widget restoring the old behavior of many2one fields in tree views


hbrunn:
Original reviews here: https://code.launchpad.net/~therp-nl/web-addons/7.0-web_tree_many2one_clickable/+merge/217655

guewen:
Has been accepted on Launchpad, I proceed with the merge.

pedrobaeza:
I'm not too confortable about the name of the module, because it can reflect that this module replaces current behaviour, which is not accurate, because it adds a widget that restore the old behaviour.  I think renaming to _web_tree_many2one_clickable_widget_ is more appropiate.

I would like also to have a module another module _web_tree_many2one_clickable_ that indeed replaces the behaviour without modifying any view.

@hbrunn What do you think?

PR #18, opened 2014-07-15
Migrate web_easy_switch_company to the last v8

AdilHoumadi:
- change route in controller
- adapt the JS to call the backend
- modify the css to graft the widget in the toolbar


hbrunn:
please fix the minor flake8 errors as indicated above
[edit] they were above my input field, but you'll see them under my comment now

hbrunn:
Thank! Two last requests: Please remove the module from **unported** and also remove placeholder, which only served to make travis run (it chokes on empty addons paths)

hbrunn:
:+1:

pedrobaeza:
You have to change icon path from /static/src/img to /static/description in order to get displayed.

pedrobaeza:
The rest of the module LGTM. We still have logo from other companies issue, but I can't imagine any solution. I will push my thumb up when you do this change.

Regards.

pedrobaeza:
:+1: 

Thanks

rvalyi:
:+1:

PR #19, opened 2014-07-16
[ADD] web_compute_domain_x2many

hbrunn:
When using `attrs="..."`, evaluation of x2many fields nearly always goes
wrong in ways not to be expected when being used to server side domains in
Model.search().

This addon fixes those cases while keeping backwards compatibility for cases
where you might have checks in a somewhat hacky way, ie. `attrs="{'invisible':
[('category_id', '=', [[6, False, []]])]}"`

Why no PR in odoo? For 7.0, it's neither going to make it in upstream nor ocb, and for 8.0 I think the code can be tested by more people before putting it there. This will be simple then anyways.


StefanRijnhart:
:+1:, thanks!

pedrobaeza:
:+1:

bwrsandman:
Tests failing due to pep8 in another module.
:+1:

PR #25, opened 2014-07-22
Web module to remove commercial support bar

ovnicraft:
For community support dont need the bar after each login


hbrunn:
https://github.com/OCA/server-tools/pull/26 is the port of the canonical module for that used since 6.1 - it also removes some less obvious  online-annoyances

PR #28, opened 2014-07-26
[FIX] Remove obsolete instruct (JS) in openerp.py

AdilHoumadi:
Remove obsolete instruct (JS) in **openerp**.py
In the new version of Odoo the JS is appended using template and xpath
Check this file : view/res_users_view

### DIFF
--- a/web_easy_switch_company/view/res_users_view.xml
+++ b/web_easy_switch_company/view/res_users_view.xml
@@ -14,7 +14,7 @@
             </field>
         </record>
 
-        <template id="assets_backend" name="web_tests assets" inherit_id="web.assets_backend">
+        <template id="assets_backend" name="web_easy_switch_company assets" inherit_id="web.assets_backend">
             <xpath expr="." position="inside">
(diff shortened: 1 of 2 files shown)


web_easy_switch_company/view/res_users_view.xml
```diff
@@ -14,7 +14,7 @@
             </field>
         </record>
 
-        <template id="assets_backend" name="web_tests assets" inherit_id="web.assets_backend">
+        <template id="assets_backend" name="web_easy_switch_company assets" inherit_id="web.assets_backend">
```
pedrobaeza, line 17: Why do you rename it?

AdilHoumadi, line 17: It's a copy paste from another module, I forget to change the name
check here : https://github.com/odoo/odoo/blob/master/addons/web_tests/views/web_tests.xml
I took inspiration from the web_tests module to inject javascript.

hbrunn:
:+1:

pedrobaeza:
:+1: (code review)

PR #29, opened 2014-07-28
[ADD] advanced_filters

hbrunn:


advanced_filters/model/ir_filters.py
```diff
@@ -0,0 +1,150 @@
...
+        assert len(ids) == 1
+        this = self.browse(cr, uid, ids[0], context=context)
+        return self.pool[this.model_id].search(
+            cr, uid, safe_eval(this.domain), context=safe_eval(this.context))
+
+    def button_save(self, cr, uid, ids, context=None):
+        return {'type': 'ir.actions.act_window.close'}
+
+    def button_freeze(self, cr, uid, ids, context=None):
+        '''evaluate the filter and write a fixed [('ids', 'in', [])] domain'''
```
guewen: Comment should be `[('id', 'in', [])]`

advanced_filters/model/ir_filters.py
```diff
@@ -0,0 +1,150 @@
...
+                        and not exists (select right_filter_id
+                            from ir_filters_union_rel where left_filter_id=id)
+                        and not exists (select right_filter_id
+                            from ir_filters_complement_rel where
+                            left_filter_id=id)
+                        ''',
+                       (tuple(removed_filter_ids),))
+
+    def button_test(self, cr, uid, ids, context=None):
+        for this in self.browse(cr, uid, ids, context=None):
```
guewen: `context=None`

guewen:
Great module. One UI issue: when I switch from the list view to the form or kanban view and back, the "Advanced filters" button disappears.

hbrunn:
Thanks, your points are fixed now

guewen:
Thanks! :+1:

hbrunn:
@yvaucher can we lift the 'needs fixing' label?

guewen:
@yvaucher is away 3 weeks so I remove the flag.

legalsylvain:
Hi holger, 
Is there a reason that you didnt named your module  'web_...' ?

hbrunn:
not really, it would be better without this prefix I think. Do you want me to rename it?

legalsylvain:
Hi, 
I'd rather change the name: 
- Specific name avoids futur collision between many modules present in many repositories;
- I also think that your module is "mainly" web, even if it changes data model a little;

But, it is a kind of matter of taste. Not a blocking point for me.

Regards.

max3903:
:+1:

pedrobaeza:
I have tried, and functionally speaking, I miss the possiblity to add the current set to a new filter. It seems redundant at first sight, because you already have this functionality on the standard, but as this functionality is split across dropdown lists, we should put together on the menu _Advanced filters_.

Another question is about name: why don't we call it _web_advanced_filters_?

hbrunn:
@pedrobaeza thanks for your input, I added your suggestion and a few other usability enhancements

pedrobaeza:
Thank you very much for the changes.

:+1:

PR #31, opened 2014-07-28
color theme module added

karim-omran:


hbrunn:
all your files live in **unported**, I don't believe that's what you want

karim-omran:
i don't know!
should i move them to another directory?

karim-omran:
i moved it outside **unported** and made a new pull request
thank you

hbrunn:
well, **unported** is meant for modules that were made for version 7.0 but need changes to work for version 8. If your module is meant for version 8, move your module to the root of the branch. If your module is meant for 7, make a pull request against branch 7.0 and also move it to the root of the branch

karim-omran:
thank you
i moved it

color_theme/__openerp__.py
```diff
@@ -0,0 +1,36 @@
+{
+    'name' : 'Color Theme',
+    'version' : '0.1',
+    'sequence' : 1,
+    'category' : 'Tools',
+    'author': 'K-Omran ISFP Egypt',
+    'description': """
+This Module Allows User To Easily Change OpenERP Color Theme.
+=============================================================
+This Module Allows User To Define New Color Theme And Save It For Future Use.
```
max3903: No need to use upper cases at the beginning of each word.

rvalyi:
Hello @karim-omran,

Your module seems nice, but as for myself, I have a problem accepting it in OCA because:
- OCA reviewers use to focus more on server side then client side, client side skill and effort is marginal today
- module author doesn't use to review code (yet?)
- no test

At the end, the module sounds cool, but feature doesn't seem essential and I fear sooner or later it will break and not be maintained. Then while it's probably a cool contribution, may be OCA is not the right place for it now because we would make users expect it will be a stable feature and may be it will not be.

What do you think?

max3903:
@rvalyi 
We are interested by this module and willing to support it as it allows us to differentiate a production and test server very easily.

rvalyi:
@max3903 hum ok good to know, this is important indeed. Are you committed to do some maintenance of it? Else I would approve only with a few tests. But given your interest I'm not voting against either.

max3903:
@rvalyi yes we are

@karim-omran Can you add some tests in your module ? Otherwise we will fork your branch, add the tests and ask for the pull request.

karim-omran:
@max3903 i am sorry i will not be able to make any changes until next sunday, because i am on vacation. so please if you can make the test i will be grateful.
it works for version 7 but need some modification for 8 because of changes in base style classes naming but it is easy to be made.
thank you

max3903:
ping @bwrsandman 

This is one of the module we want to install by default. It is missing some tests and this is far beyond my skills :)

bwrsandman:
Module needs to be fixed for pep8.
As for unittests, it doesn't have that much python code, it would maybe need phantomjs tests and I have no idea how to implement those.

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
+from openerp.osv import osv, fields
+ 
+ #============================Color Theme Model==============================
+
+class color_theme(osv.osv):
```
bwrsandman: Please use `orm.Model` or even better: `openerp.models.Model`

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+                            'Link Color',
+                            size=6,
+                            ),
+                
+                }
+    
+    _sql_constraints = [('unique_theme_name', 'unique(theme_name)', 'Theme With The Same Name Already Exists'),]
+    
+         
+color_theme()
```
bwrsandman: Not necessary

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+#==============================================================================================
+
+
+#===============================User Theme Model===============================================
+class user_theme(osv.osv):
```
bwrsandman: Please put this class in its own file named `user_theme.py`
Same thing about `osv.osv`

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+                            'color.theme',
+                            'Color Theme',
+                            required=True,
+                            ondelete='cascade',
+                            ),
+                }
+    
+    def _check_user(self, cr, uid, ids):
+        user_theme = self.search(cr, uid, ['&', ('user', '=', uid), ('id', '!=', ids[0]),])
+        print user_theme
```
hbrunn: this looks like debugging code

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+        if not user_theme:
+            return True
+        else:
+            return False
```
hbrunn: replace all of the above by return bool(user_theme)

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+               'button_gradient_1' : user_theme.color_theme.button_gradient_1,
+               'button_gradient_2' : user_theme.color_theme.button_gradient_2,
+               'left_bar' : user_theme.color_theme.left_bar,
+               'body_font' : user_theme.color_theme.body_font,
+               'main_menu_font' : user_theme.color_theme.main_menu_font,
+               'sub_menu_font' : user_theme.color_theme.sub_menu_font,
+               'top_bar_menu_font' : user_theme.color_theme.top_bar_menu_font,
+               'tab_string' : user_theme.color_theme.tab_string,
+               'many2one_font' : user_theme.color_theme.many2one_font,
+               }
```
hbrunn: this asks for being forgotten to update when new properties are added. better
return self.pool['color.theme'].read(cr, uid, user_theme.color_theme.id, context=context)

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+               'user':user,
+               'color_theme':vals['color_theme']
+               }
+          
+        return super(user_theme, self).create(cr, user, rec)
+          
+        
+    
+    def get_user_theme(self, cr, uid, context=None):
+        user_theme_id = self.pool.get('user.theme').search(cr, uid, [('user', '=', uid),])
```
hbrunn: pass context=context

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+        
+        if user_theme_id:
+            user_theme = self.pool.get('user.theme').browse(cr, uid, user_theme_id[0])
```
hbrunn: context

color_theme/color_theme.py
```diff
@@ -0,0 +1,156 @@
...
+               'tab_string' : user_theme.color_theme.tab_string,
+               'many2one_font' : user_theme.color_theme.many2one_font,
+               }
+            
+            return res
+        
+        else:
+            return None
+    
+user_theme()
```
hbrunn: not necessary

color_theme/static/src/colorpicker/js/colorpicker.js
```diff
@@ -0,0 +1,484 @@
+/**
```
hbrunn: libraries belong in /static/lib

hbrunn:
:+1:

color_theme/__openerp__.py
```diff
@@ -0,0 +1,29 @@
+{
+    'name': 'Color Theme',
+    'version': '0.1',
```
bwrsandman: How about 1.0

color_theme/user_theme.py
```diff
@@ -0,0 +1,44 @@
+from openerp.osv import orm, fields
+
+
+# ===============================User Theme Model===============================================
```
bwrsandman: These lines are not necessary, could you please remove them.

bwrsandman:
Take care not to include compiled python files (*.pyc) in your PR
- color_theme/__init__.pyc
- color_theme/color_theme.pyc
- color_theme/user_theme.pyc

gurneyalex:
The module looks nice. However I'm about to delete the master branch in favor of the 8.0 branch. @karim-omran can you please resubmit on the correct branch?

PR #34, opened 2014-08-27
7.0 web field masks

menecio:
Added web_field_masks module

### DIFF
--- a/web_fields_masks/static/src/js/main.js
+++ b/web_fields_masks/static/src/js/main.js
@@ -1,0 +1,27 @@
+openerp.web_fields_masks = function (instance) {
+    instance.web.form.FieldChar.include({
+        mask: '',
+        init: function(field_manager, node) {
+            this._super(field_manager, node);
+            if (_.has(this.node.attrs, 'data-inputmask')) {
+                this.mask = this.node.attrs['data-inputmask'];
+            }
+        },
+
+        render_value: function () {
+            var self = this;
+            this._super();
+            if (this.mask !== '') {
+                this.$('input').attr('data-inputmask', this.mask);
+                this.$('input').inputmask(undefined, {
+                    onincomplete: function () {
+                        self.$el.addClass('oe_form_invalid');
+                    },
+                    oncomplete: function () {
+                        self.$el.removeClass('oe_form_invalid');
+                    },
+                });
+            }
+        },
+    });
+}
--- a/web_fields_masks/__openerp__.py
+++ b/web_fields_masks/__openerp__.py
@@ -1,0 +1,55 @@
+{
+    'name': 'Fields masks',
+    'version': '0.2',
+    'description': """
+Fields masks
+================================================================================
+
+Based on jquery.inputmask 3.x (https://github.com/RobinHerbots/jquery.inputmask)
+
+An inputmask helps the user with the input by ensuring a predefined format.
+This can be useful for dates, numerics, phone numbers, ...
+
+
+
+Instructions:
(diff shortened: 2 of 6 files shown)


hbrunn:
./web_fields_masks/**openerp**.py:8:80: E501 line too long (80 > 79 characters)

hbrunn:
:+1:

legalsylvain:
Hi, 
I tried this module that works pretty good. Two points:
1/ It doesn't work on a tree view editable:

```
    <record id="view_id" model="ir.ui.view">
            <tree string="My Model" editable="bottom">
                <field name="description" data-inputmask="'mask': '99:99'"/>
            </tree>
        </field>
    </record>
```

This code return, when updating my module:

```
2014-10-14 ERROR DBNAME openerp.addons.base.ir.ir_ui_view: <string>:7:0:ERROR:RELAXNGV:RELAXNG_ERR_NOELEM: Expecting an element diagram, got nothing
2014-10-14 ERROR DBNAME openerp.addons.base.ir.ir_ui_view: <string>:7:0:ERROR:RELAXNGV:RELAXNG_ERR_INVALIDATTR: Invalid attribute data-inputmask for element field
2014-10-14 ERROR DBNAME openerp.addons.base.ir.ir_ui_view: <string>:2:0:ERROR:RELAXNGV:RELAXNG_ERR_EXTRACONTENT: Element tree has extra content: field
2014-10-14 23:24:18,080 25427 ERROR DBNAME openerp.tools.convert: Parse error in /my_path/view/view.xml:146: 
```

-> I don't know if it is easy to improve your module to work in editable tree view. Otherwise, Can you write this limitation in the description of the module ?

2/ Some User cases: 
- I have a database with an bad email like 'test@gmail' without ".com" in the email field of a res partner;
- I install your module and write a module to set this field with the 'alias' 'email';
- I edit the partner and change another field, then save;
  ->  The value of email is now 'bob@gmail.__'

More generaly, if the field is not correctly filled, the field become red, but it is possible to save. 
Is it the expected behaviour ?
For exemple, it is possible to write an ip like '123.021.'

Regards.
(functional test, no code review)

web_fields_masks/static/src/js/main.js
```diff
@@ -0,0 +1,27 @@
...
+        },
+
+        render_value: function () {
+            var self = this;
+            this._super();
+            if (this.mask !== '') {
+                this.$('input').attr('data-inputmask', this.mask);
+                this.$('input').inputmask(undefined, {
+                    onincomplete: function () {
+                        self.$el.addClass('oe_form_invalid');
```
hbrunn, line 18: Reacting to https://github.com/OCA/web/pull/34#issuecomment-59136418, it probably would make sense to do this  by overriding is_syntax_valid - then you get the classes for free and the standard mechanism would force you to correct the value

eLBati:
Thanks @menecio , does this support the 'regex-mask'?

menecio:
Sorry I didn't receive any notifications about this in my mail.

About working on editable list views, I only extended view_form.js, so it will need extra work to add compatibility with view_list

legalsylvain:
@menecio : No problem about limitation, but can you only write it in _ _ openerp _ _  file ?

And what about other remarks ?

Regards.

lmignon:
@menecio Thank you for the contribution. Same question as @legalsylvain : 

> If the field is not correctly filled, the field become red, but it is possible to save. 
> Is it the expected behaviour ?

hbrunn:
closing due to inactivity

PR #36, opened 2014-09-03
[IMP] Web Warning Sounds

kirca:
Implementation of Web Warning Sounds module. Please check README.rst file for more details on how it works.


web_warning_sound/static/src/js/sound_extend.js
```diff
@@ -0,0 +1,61 @@
+openerp.web_warning_sound = function(instance) {
+    var QWeb = instance.web.qweb;
+    var _t = instance.web._t;
+
+    instance.web.FormView = instance.web.FormView.extend({
+        on_processed_onchange: function(result, processed) {
```
hbrunn, line 6: Is there any reason to break inheritance here?

web_warning_sound/static/src/js/sound_extend.js
```diff
@@ -0,0 +1,61 @@
...
+            } catch(e) {
+                console.error(e);
+                instance.webclient.crashmanager.show_message(e);
+                return $.Deferred().reject();
+            }
+        },
+    });
+
+    instance.web.CrashManager = instance.web.CrashManager.extend({
+        show_warning: function(error) {
```
hbrunn, line 20: Isn't there a possibility to extend except_orm to include a sound field and have that passed to the web client? This solution seems pretty hacky to me

web_warning_sound/static/src/js/sound_extend.js
kirca, line 6: It seems like the simplest approach, that is overwrite original method and add if block (@lines 8-11) to handle playing the audio.

web_warning_sound/static/src/js/sound_extend.js
kirca, line 20: I think that this is good enough since framework didn't intend and make any easier way to add audio to warning message. As for onchange here also the original method is overwritten and lines 46-52 are added. I'm not sure what you mean though by passing sound field to web client? Since JS Audio API requires only the resource location, I think it is a good approach just to parse that from the warning message and use it to play the audio.

web_warning_sound/static/src/js/sound_extend.js
hbrunn, line 6: but that means you break every other module that overrides this function. Don't do that unless absolutely necessary.

kirca, line 6: sure overwrite is not the safest but as I mentioned in other comment the framework didn't give any easy way to hook this in. Any suggestion on better solution is appreciated :)

hbrunn, line 6:
```
on_processed_onchange: function(result, processed) {
    if (!_.isEmpty(result.sound)) {
                    var audio = new Audio(result.sound);
                    audio.play();
    }
```

web_warning_sound/static/src/js/sound_extend.js
hbrunn, line 20: inherit a class except_orm_with_sound from except_orm, give it a third parameter for the audio url that is stored in a separate field on the instance, raise that exception and see if this field is passed to the web client when you raise it

web_warning_sound/static/src/js/sound_extend.js
kirca, line 6: Thanks for suggestion, it is implemented

web_warning_sound/static/src/js/sound_extend.js
kirca, line 20: Unfortunately what you are suggesting is not possible without modifying web core:
https://github.com/odoo/odoo/blob/7.0/addons/web/http.py#L220

Modified however so that inheritance chain is not broken.

web_warning_sound/README.rst
```diff
@@ -0,0 +1,28 @@
+This plugin allows to create modules that can play sound files.
```
legalsylvain: This file content the same text as openerp file description Is it necessary ?

legalsylvain:
Hi, thanks for sharing your work.
About the module, I think that overloading the onchange, adding a wav is a very good feature that can be usefull. About the exception, I would have preferred something more generic. (Like all orm_exception generate that sounds...). But OK.

Except the remarks about the readme file:
:+1: (functional test)

web_warning_sound/README.rst
kirca: You are right,  it is not necessary. But it is common for opensource projects to have such file because newcomers in odoo wouldn't know to check **openerp**.py for the description. At least that is my reasoning :). Going to remove it for the sake of consistency since no other module in this repo has it.

kirca:
Hi Sylvain,

Thanks for your review. I'm only familiar with except_orm though so yes it will work only on that.

foutoucour:
:+1:

PR #37, opened 2014-09-09
Color Themes

karim-omran:


color_theme/__openerp__.py
```diff
@@ -0,0 +1,28 @@
+{
+    'name': 'Color Theme',
+    'version': '1.0',
+    'category': 'Tools',
+    'author': 'K-Omran',
+    'description': """
+This module allows user to easily change OpenERP color theme.
+=============================================================
+This module allows user To define new color theme and save it for future use.\n
```
pedrobaeza: You don't need `\n`

color_theme/__openerp__.py
```diff
@@ -0,0 +1,28 @@
...
+Each user can select any stored color theme as his theme\n
+and also can make his own theme.\n
+""",
+
+    'depends': ['base', 'web', ],
+    'data': ['security/ir.model.access.csv',
+             'views/color_theme_view.xml', ],
+    'js': ['static/lib/colorpicker/js/colorpicker.js',
```
pedrobaeza, line 17: There is already a color picker widget on OCA (https://github.com/OCA/web/tree/7.0/web_color). Why don't use it?

color_theme/color_theme.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class color_theme(orm.Model):
+    """This Model To Store Color Themes To be Selectd Later"""
+
+    _name = 'color.theme'
+
+    _rec_name = 'theme_name'
+
+    _columns = {'theme_name': fields.char('Theme Name', size=128, required=True, ),
```
pedrobaeza: Respect PEP8 80 cols limit

karim-omran:
@pedrobaeza 
fixed pep8 cols limit 80
i saw the color picker widget on OCA and i think the color picker i used is a better one and gives all color spectrum

pedrobaeza:
@ karim-omran, thanks for the changes! And don't you think it's better to update the other widget with your library and base this module on that, that duplicating color picker libs?

legalsylvain:
@karim-omran Agree with @pedrobaeza about avoiding to duplicate library.

legalsylvain:
I close this PR, because 8.0/web_widget_color do the same thing.

PR #40, opened 2014-09-19
[ADD] Module  help_online

lmignon:
# User documentation using the website module

_This PR follows the positive feedback from members of the community in response to a [post](https://www.odoo.com/groups/community-59/community-8982559?mode=thread&date_begin=&date_end=) of Stéphane Bidoul on the mailing list._

> People interested in providing user documentation in their projects may want to look into a module we developed to create online help.
> 
> The principle is rather simple: a help icon appears on form and list views. The icon shows a help page for the model if it exists or let the user with appropriate rights create a new help page themselves. Help pages are created with the v8 website editor.
> 
> We have a few ideas to improve it, but the module is already functional in its current state.


help_online/__init__.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Nemry Jonathan
+#    Copyright (c) 2014 Acsone SA/NV (http://www.acsone.eu)
+#    All Rights Reserved
```
guewen: What means 'All Rights Reserved' here?
The phrase is obsolete (see http://en.wikipedia.org/wiki/All_rights_reserved#Obsolescence) and feels like a contradiction to the license below.

help_online/controllers/help_online_controllers.py
```diff
@@ -0,0 +1,40 @@
...
+from openerp.http import request
+
+
+class HelpOnlineController(http.Controller):
+
+    @http.route('/help_online/build_url', type='json', auth='user')
+    def build_url(self, model, view_type, domain=None, context=None):
+        help_online_model = request.env['help.online']
+        return help_online_model.get_page_url(
+            model, view_type, domain=None, context=None)
```
guewen: The domain and context should probably be passed to get_page_url()

By the way, are they necessary with the new API?

lmignon:
@guewen 
....copy/paste... Sorry. I'll fix

help_online/models/help_online.py
```diff
@@ -0,0 +1,68 @@
...
+        return website_model.page_exists(name)
+
+    def get_page_url(self, model, view_type, domain=None, context=None):
+        user_model = self.env['res.users']
+        if not user_model.has_group('help_online.help_online_group_reader'):
+            return {}
+        ir_model = self.env['ir.model']
+        description = self.env[model]._description
+        res = ir_model.name_search(model, operator='=')
+        if(res):
```
guewen: `if res:`

help_online/static/src/css/help_online.css
```diff
@@ -0,0 +1,12 @@
+li.oe_help_online_not_found {
+  background-color: red;
```
guewen: The 'red' color for the button is a bit harsh, maybe using a red nuance used elsewhere in the interface would be smoother? The gradient of the red buttons goes from #df3f3f to #a21a1a, so maybe one of these?

guewen:
Looks great! I just left some minor comments. :+1:

lmignon:
@guewen Thank you for the review. The code is fixed.

guewen:
Thanks! :+1:

pedrobaeza:
One dumb question without reviewing the module and knowing nothing about it at all: can we create xml data in another module depending on this one that contains some help pages and be loaded installing it?

lmignon:
@pedrobaeza, yes you can since all pages are pages from the website. Therefore, you can create xml data in one module providing 'template' elements for your documentation. The id/name of the element  must formatted as expected by the module.  (see https://github.com/OCA/web/pull/40/files#diff-d938b0a2d87ef99367e37e034250bd7eR28)
For one of our client we've developed a wizard to export the documentation written with the website. The wizard is really helpful since it export pages and linked images in an xml file. It also fixes a lot of details in the xml to make the pages 'portable' from one instance to another. (replace absolute urls by relative urls, etc). However, the code still needs to be cleaned to make it generic.

pedrobaeza:
Wow, that's very cool! I'm starting to see this as a solution for documenting functionally OCA modules. The module you mentioned it's also a very hepful tool, so please complete it.

For now, I proceed with the merge of this one. :+1:

jjscarafia:
Very nice module!!! congratulations and thanks!

max3903:
Is the documentation publicly available or can it be restricted to the Employee group ?

lmignon:
@max3903 The documentation is publicly available. We are aware that this can be a problem when the web module is also used for the management of a site next to the doc. To solve this problem, we thought of two ideas.
- create a dedicated site for the documentation
- create dedicated record rules (name of the pages for the documentation always starts with 'help-' and are 'ir.ui.view')
  The first one is no possible since multi site is not supported by Odoo in v8 (but will be supported in v9). The second one seems also difficult to put in place. In fact, we have not tried to solve this limitation because it is not a problem for our customer.

pedrobaeza:
But if you don't add any menu link, they will not be accesible, doesn't it?

lmignon:
@pedrobaeza 
Yes and no; the pages are not available via menus but if you write the right url, you can access it.

pedrobaeza:
Yeah, but I think it's a minor issue for general cases, where website visitors doesn't know that URL, and also documentation pages don't contain sensible information I think.

lmignon:
It's also my point of view

elicoidal:
U
Eric Caudal (From his mobile)

Cedric-Pigeon:
For information, I am currently preparing the addition of the wizard to export the documentation.

Coming soon ...

max3903:
@lmignon :( So I can't use them as a replacement for document.page, which I used to document companies procedures and work instructions...

Good to know it exists though. Good work!

sbidoul:
@max3903 @lmignon is it difficult to adapt the web controller to restrict access to web pages starting with help- to authenticated users only?

lmignon:
@sbidoul This does not seem difficult. Just a few lines to write. But this requires to override the website controller.

Olivier-LAURENT:
@max3903 with the 2 following rules you can implement this level of security:

```
    <record id="online_help_confidential_rule" model="ir.rule">
        <field name="name">Online Help Hidden by Default</field>
        <field name="model_id" ref="base.model_ir_ui_view"/>
        <field name="domain_force">[
            '|',
            ('type', '!=', 'qweb'),
            ('name','not like','help-%'),
        ]</field>
        <field name="groups" eval="[(6, 0, [
            ref('base.group_portal'),
            ref('base.group_public'),
            ref('base.group_user'),
        ])]"/>
        <field name="perm_read" eval="1"/><field name="perm_write" eval="0"/>
        <field name="perm_create" eval="0"/><field name="perm_unlink" eval="0"/>
    </record>

    <record id="online_help_reader_rule" model="ir.rule">
        <field name="name">Online Help for Help Reader</field>
        <field name="model_id" ref="base.model_ir_ui_view"/>
        <field name="domain_force">[(1, '=', 1)]</field>
        <field name="groups" eval="[(6, 0, [
            ref('help_online.help_online_group_reader'),
        ])]"/>
        <field name="perm_read" eval="1"/><field name="perm_write" eval="0"/>
        <field name="perm_create" eval="0"/><field name="perm_unlink" eval="0"/>
    </record>
```

Such rules will be included in our upcoming PR on this module.

Cedric-Pigeon:
Import and export pages features are proposed in PR #46

PR #41, opened 2014-09-22
add "web_recipients_uncheck" module

OSguard:
Uncheck additional recipients and followers when
+sending a message at the customer view.

### DIFF
--- a/web_recipients_uncheck/__init__.py
+++ b/web_recipients_uncheck/__init__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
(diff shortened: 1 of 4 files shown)


hbrunn:
This relies on the bind_events() method only being called when the recipients are checked. Currently, I can't see when this would not be the case, but wouldn't it be safer to override the xml template instead to always render the checkboxes as inactive?

max3903:
Can you add the translation file so the description can be translated ?

Otherwise LGTM

OSguard:
Translation added

@hbrunn maybe modify xml work as well. But this solution works for us and is tested and live, so ne reason to find a other one.

PR #42, opened 2014-09-22
add "web_list_html_widget" module

OSguard:
This module allows to use a widget="html" for fields in
list(tree) views.

### DIFF
--- a/web_list_html_widget/__init__.py
+++ b/web_list_html_widget/__init__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Some questions:
- It is possible/advisable to sanitize first HTML?
- Have you tried to propose this to core?

Regards

OSguard:
It is the same behavior as HTML.fields in form view
And the module is from the pre Github times, as our MR was ignored, so we build modules if it is possible to ship it ;) - And yet not retry again.

hbrunn:
I think the warning in the manifest suffices. If you deal with anything the remotely comes from external input, you'll wrap this in a function field that does the sanitizing I'd say. :+1:

max3903:
Can you generate the translation file so the description could be translated ?

Otherwise LGTM

OSguard:
done

PR #45, opened 2014-10-06
[ADD] web_pytz

hbrunn:
## Introduction

This module allows complex timezone operations in domains mimicing python's
pytz. The heavy lifting is done by http://momentjs.com/timezone.

It is meant to allow correct filters for 'Today', 'Yesterday' etc.

In addition to implementing a subset of `pytz.tzinfo` and
`datetime.astimezone`, there's a shortcut called `utc_today()` which returns
the beginning of the day in the current user's time zone translated to UTC,
this is equivalent to::

  pytz.timezone(tz).localize(datetime.datetime.now().replace(hour=0, minute=0,
      second=0)).astimezone(pytz.utc)

in python.
## Usage

Depend on this module and use filters like::

  [('write_date', '>=', utc_today().strftime('%Y-%m-%d %H:%M:%S'))]

which displays records changed in the user's conception of today


max3903:
Can you add the translation file to translate the module description ?

hbrunn:
Done

lmignon:
:+1:  Thank you for the module AND the associated tests!

pedrobaeza:
:+1:

digitalsatori:
:+1:

PR #52, opened 2014-11-04
Help Online: export improvement

Cedric-Pigeon:
If the same image is used on multiple pages, the website module creates only one ir.attachment referenced several times which is optimal.

For help pages, the export feature generates as many xml_id as images. But finally all these xml_id will be linked to the same ir.attachment.
By this way you get several ir.model.data records referencing the same res_id.
This is really unsafe because if you delete one of these images, the odoo update process will delete the ir.attachment and all ir_model_data records linked. So the image will disappear from all pages.

I resolved the problem by using a single xml_id for all duplicates images.

In addition the autobackup feature is not called anymore. This is useless because it should be done in a inherited module. Otherwise it will be executed twice.


foutoucour:
:+1:

pedrobaeza:
:+1: Thanks for the improvement.

@Cedric-Pigeon, have you seen the debate at #55 ?

Cedric-Pigeon:
@pedrobaeza 
Thanks for merging.
I just posted a comment on #55.

PR #54, opened 2014-11-14
first release of web_dashboard_tile

OSguard:
This module is a PoC to have a new modern dashboard for OpenERP.

See example here https://twitter.com/OSguard/status/513041022906298368


legalsylvain:
Travis can not install the module. (xml_id bug).
Regards.

OSguard:
@legalsylvain it is now fix, so review need i guess

legalsylvain:
Sorry, I didn't see your last commit.

web_dashboard_tile/__openerp__.py
```diff
@@ -0,0 +1,46 @@
...
+{
+    "name": "Dashboard Tile",
+    "version": "0.3",
+    "depends": ['web', 'board', 'mail'],
+    'author': 'initOS GmbH & Co. KG',
+    "category": "",
+    'license': 'AGPL-3',
+    "description": """
+
+    """,
```
hbrunn, line 47: That's not exactly an extensive description

OSguard:
documentation video: https://www.youtube.com/watch?v=H5eGWHPvYe4

web_dashboard_tile/security/rules.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data noupdate="1">
+
+    <record id="payment_method_rule" model="ir.rule">
```
LeartS: `payment_method_rule` ? Remember to change the id when you use another module as template!

web_dashboard_tile/static/src/xml/custom_xml.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates id="template" xml:space="preserve">
+    <t t-extend="SearchView.addtodashboard">
+        <t t-jquery="form" t-operation="after">
+            <div>
+                <label for="dashboard_tile_new_name">Tile:</label>
+                <input id="dashboard_tile_new_name" />
+                <button id="add_dashboard_tile">Okay</button>
```
legalsylvain: Hi, Is it possible to change the title of the button ? "Create" or "Create Tile" could be better than "Okay".

OSguard: done

web_dashboard_tile/tile.py
```diff
@@ -0,0 +1,97 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm, fields
+import random
+
+
+class tile(orm.Model):
+    _name = 'tile'
```
legalsylvain: Hi, 
'tile' is a very short name for a model that doesn't mentions correctly what is is. What about 'dashboard.tile', 'web.tile', or 'web.dashboard.tile' ? 
Just a remark. What do you think  ?

web_dashboard_tile/tile.xml
```diff
@@ -0,0 +1,111 @@
...
+        <record model="ir.actions.act_window" id="action_kanban_dashboard_tile">
+            <field name="name">Dashboard</field>
+            <field name="res_model">tile</field>
+            <field name="view_type">form</field>
+            <field name="view_mode">kanban</field>
+            <field name="domain">['|',('user_id','=',False),('user_id','=',uid)]</field>
+            <field name="view_id" ref="dashboard_tile_tree_view"/>
+        </record>
+
+        <record id="mail_dashbaord" model="ir.ui.menu">
```
legalsylvain: typo: 'dashboard'

web_dashboard_tile/tile.xml
```diff
@@ -0,0 +1,111 @@
...
+                    <field name="color"/>
+                    <templates>
+                        <t t-name="kanban-box">
+                            <div t-attf-class="oe_tile_color_#{kanban_getcolor(record.color.raw_value)} oe_dashbaord_tile oe_kanban_global_click" >
+                                <div class="oe_dropdown_toggle oe_dropdown_kanban">
+                                    <span class="oe_e">í</span>
+                                    <ul class="oe_dropdown_menu">
+                                        <t t-if="widget.view.is_action_enabled('edit')"><li><a type="edit">Edit...</a></li></t>
+                                        <t t-if="widget.view.is_action_enabled('delete')"><li><a type="delete">Delete</a></li></t>
+                                        <li><ul class="oe_kanban_colorpicker" data-field="color"/></li>
```
legalsylvain, line 56: :question: 
This 3 lines doesn't display menu item to edit / delete the 'tile' and change colors.
(Used with odoo 7.0 et tested on Firefox 35.0 & chromium.)
Do you have the same problem ?
Regards.

legalsylvain:
Except minor remarks, that's a great module. Thanks a lot !

Feature remarks (non blocking point): 
Why did you create two actions for 'tile' and two menu items. It could be more simple to have a uniq action with "kanban,tree,form" value in view_mode field ?

dreispt:
@OSguard 
As of v8 Kanban views can also provide dashboards.
What would be the main differences between this approch vs using kanban views?

OSguard:
@dreispt
Kanban views are always for one model + static defined KPI values.
so my design goal is to have a dashboard where you can add every object you want. The second goal is that the user can define how the KPI is calculated without asking a developer to change the (kanban) view

web_dashboard_tile/tile.py
OSguard: I was thinking on that, following odoo strict naming the correct name should be web_dashboard_tile.tile but this seams strange. So i decide to keep it simple.

web_dashboard_tile/tile.xml
OSguard, line 56: I know and the function of the menus is not always working correct. For example you can not edit if you have no form view.
maybe i should write a Fixme in the code

OSguard:
I create two action to have a dashboard view without tree and form view in messages. And have a second where i can edit in settings menu.

web_dashboard_tile/tile.xml
legalsylvain, line 56: All right, 
So I guess writing a little "FIXME" will be helpfull for developpers.

OSguard, line 56: done

legalsylvain:
Thanks ! :+1:

web_dashboard_tile/tile.py
```diff
@@ -0,0 +1,97 @@
...
+        'color': 0,
+    }
+
+    def open_link(self, cr, uid, ids, context=None):
+
+        tile_id = ids[0]
+        tile_object = self.browse(cr, uid, tile_id, context=context)
+
+        if not context:
+            context = {}
```
dreispt: Your logic is not depending on `context` so it's OK to remove this check.

dreispt:
@OSguard OK with your explanation.
Other than my comment on the too short model name, it's :+1:

legalsylvain:
@dreispt : which comment ?

web_dashboard_tile/tile.py
dreispt: I agree with @legalsylvain : model names share a global namespace, so they should use a main module prefix  to name avoid collisions. I suggest using `tile.tile` for this model.

dreispt:
@legalsylvain missed a click - done now, thanks.

OSguard:
Now choose tile.tile as new name

lmignon:
@OSguard Thank you for sharing this great module. :+1:

legalsylvain:
Hi, 
For me, this PR is ready to merge, but I have some doubt about Runbot fail. (It is OK for travis). I can not see the origin of the problem. Any Idea someone ?

Regards.

web_dashboard_tile/tile.xml
```diff
@@ -0,0 +1,112 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+       <record model="ir.ui.view" id="dashboard_tile_tile_tree_view">
+            <field name="name">tile.tile.tree</field>
+            <field name="model">tile.tile</field>
+            <field name="type">tree</field>
```
legalsylvain: This line is useless. (and 2 other below)
Cf https://runbot.odoo-community.org/runbot/static/build/2924792-54-6f0bec/logs/job_20_test_all.txt.

OSguard: i just can remove it?
This is just a warning, so the reason why runbot fail?

legalsylvain: I don't know why runbot fails. (I personaly don't use it) 
I just checked log and saw that three warnings...
Please, yes remove this 3 lines. (It will be done, because it's an obsolete field).

OSguard:
so i remove the tree lines.
In the last Runbot instance i was able to login and everything works correct:
http://runbot.odoo-community.org:8330/?db=2924792-54-6f0bec-all#view_type=kanban&model=tile.tile&menu_id=122&action=116
Maybe this time the test returns correct, we will see soon

gurneyalex:
runbot can be ignored for now. I am configuring it these days. Sorry for the inconvenience. 

the way to get a run with warnings to show up as green instead of red has changed recently and I need to adapt the patch.

legalsylvain:
OK. Thanks @gurneyalex for the information.

PR #61, opened 2014-12-20
add form_sheet_full_screen

luc-demeyer:
Two small modules:
1) form_sheet_full_screen: adds css  class to change a Form View into full screen via inherited view.
2) form_sheet_full_screen_all : changes all Form View into full screen

### DIFF
--- a/form_sheet_full_screen/__init__.py
+++ b/form_sheet_full_screen/__init__.py
@@ -1,0 +1,21 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#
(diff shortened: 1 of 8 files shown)


legalsylvain:
:question: This module seems to do the same that 'web_sheet_full_with'. Doesn't it ?

luc-demeyer:
@legalsylvain 
Sylvain, you are right.. The 'form_sheet_full_screen_all' does exactly the same as the 'web_sheet_full_width' module (which didn't exist yet when I made mine). I'll drop my contribution in favor of this already merged module. 
I believe there is some value in my approach of two separate modules: one which turns all forms into full screen and a second module that enables a more selective approach. 
I'll make a new PR to add a 'web_sheet_full_width_selective' module as a replacement for the selective module in this PR..

PR #63, opened 2014-12-20
[WIP] add web_disable_export

luc-demeyer:
this module removes the 'Export' option from the 'More' menu in the list view except for the admin user.

### DIFF
--- a/web_disable_export/__openerp__.py
+++ b/web_disable_export/__openerp__.py
@@ -1,0 +1,41 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#
+#    Copyright (c) 2014 Noviat nv/sa (www.noviat.com). All rights reserved.
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Web Disable Export',
+    'version': '0.1',
+    'license': 'AGPL-3',
+    'author': 'Noviat',
+    'website': 'http://www.noviat.com',
+    'category': 'Web',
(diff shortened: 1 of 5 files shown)


web_disable_export/static/src/js/disable_export.js
```diff
@@ -0,0 +1,57 @@
...
+openerp.web_disable_export = function(instance) {
+
+    var _t = instance.web._t;
+    var SUPERUSER_ID = 1;
+
+    instance.web.Sidebar.include({
+
+        add_items: function(section_code, items) {
+            // allow Export for admin user
+            if (this.session.uid == SUPERUSER_ID) {
```
nhomar, line 33: It should be great if it is checked by ACL and not only by SUPERUSER_ID, don't you think?

luc-demeyer, line 33: Nhomar,

You are right but the ‘sponsoring’ customer only wanted to pay a couple of hours to disable the export and hence we implemented this. 

Regards,

Luc

 http://www.noviat.com/ www.noviat.com

Rusatiralaan 1, 1083 Brussel

+32 2 808 86 38

pedrobaeza:
Extract the description to a README.rst file and please put what Nhomar comments in a TODO section of this description.

pedrobaeza:
Please remove description key from manifest file.

Alembo:
This can also be done with http://change-openerp-web.blogspot.com/2013/08/hide-export-in-more-option-in-openerp-7.html

Yours would be better if we could also add a expect group instead of only user admin.

web_disable_export/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+    'license': 'AGPL-3',
+    'author': 'Noviat',
+    'website': 'http://www.noviat.com',
+    'category': 'Web',
+    'description': """
+
+Remove the 'Export' option from the 'More' menu
+in the list view except for the admin user
+
+""",
```
dreispt, line 35: The description key can be removed since the README will be automatically used for it

hbrunn:
I think this should happen on template level and use a group that is allowed to export to be useful :-1:

luc-demeyer:
@hbrunn 
I agree that your suggestion would make it better, but that goes beyond what the sponsoring customer wanted to pay for.

legalsylvain:
Hi, 
I understand @luc-demeyer PoV. 
By the other side, remarks of @hbrunn and @nhomar are very relevant.

Maybe we can mark this PR as 'WIP' and 'help Wanted' and wait some other contributors to finish this module ?
Regards.

luc-demeyer:
> > > Maybe we can mark this PR as 'WIP' and 'help Wanted' 
> > > Sounds ok to me. In the meanwhile I'll publish it via our apps.odoo.com branch (and try not to forget to remove it once it becomes available via the OCA).

hbrunn:
closing due to inactivity

PR #64, opened 2014-12-21
add web_sheet_full_width_selective

luc-demeyer:


pedrobaeza:
Please extract the module description in a separated README.rst file to be used by GitHub and Odoo and the same time. You can also put a reference to the other module indicating that if user want to replace the default behaviour in all sheets, they should install the other one.

pedrobaeza:
Thanks for the change, Luc, but you don't need to keep the description in **openerp**.py, because it's also parsed by Odoo from README.rst file.

luc-demeyer:
@pedrobaeza 
Pedro, when I remove the openerp.py description than the content of the README.rst becomes visible to all end-users who have access to the module list. As a result, the 'TODO' that is in the README.rst becomes visible to a non-intended audience. I think we should consider another convention (targeted towards the OCA contributors not the end-users) to put module TODO's (e.g. like the TODO  comments used by the odoo developers in the .py files).

pedrobaeza:
I prefer to see the TODO even at user space, because you can know what are the intended improvements, and also defines what is the current scope.

luc-demeyer:
I don't think this PR is the best place to have this discussion since it's about a convention that affects all OCA contributors.

pedrobaeza:
Ok, I don't see the connection but I'm not going to block the PR.

:+1:

adrienpeiffer:
:+1: ( Code review )

web_sheet_full_width_selective/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Show selected sheets with full width',
+    'version': '0.1',
+    'license': 'AGPL-3',
+    'author': 'Noviat',
```
legalsylvain: Could you add OCA in the author, as it's now the general rules for all OCA modules ?
Regards.
Ref : https://github.com/OCA/web/blob/8.0/help_online/__openerp__.py#L24

hbrunn:
:+1:

PR #68, opened 2015-01-01
new:[web_sheet_full_width] give openchatter full width too.

### LINKED ISSUE #67: web_sheet_full_width should also expand chatter
I think that we should also resize chatter area to extend it to full width area. What do you think?

njeudy:
I update css to have openchatter full width too.
- fixes #67


pedrobaeza:
:+1: 

Thanks.

hbrunn:
:+1:

anajuaristi:
:+1: Great contribution full width. It should be included on Core. I don't understand why they included fixed width
Thank you!!

PR #77, opened 2015-01-15
[ADD] New module website_blog_mgmt

lmignon:
- This module lets you specify a date for the post at which it will be published.
- Posts on the website are ordered by publication date.
- The order of posts is preserved when navigating between posts
- Since it's HARD/not possible to extend controller without rewriting completly
  the overrided functions, the module also fixes https://github.com/odoo/odoo/pull/3097
  (loop issue when pressing the next button on a blog page)

### DIFF
--- a/website_blog_mgmt/__init__.py
+++ b/website_blog_mgmt/__init__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Laurent Mignon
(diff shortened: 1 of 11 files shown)


sbidoul:
I think this PR should go to OCA/website instead.

pedrobaeza:
Me too

lmignon:
You are right. I'll relocale my proposal on 'website'

lmignon:
new PR on OCA/website https://github.com/OCA/website/pull/8

PR #82, opened 2015-02-07
Fix parsing of integers with separator

bringsvor:
Small fix to allow integers to be parsed correctly when a locate with a thousand separator is used.

### DIFF
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -85,7 +85,8 @@ openerp.web_export_view = function (instance) {
                                 }
                             }
                             else if (cell.classList.contains("oe_list_field_integer")) {
-                                export_row.push(parseInt(text));
+                                tmp2 = text.replace(instance.web._t.database.parameters.thousands_sep, "");
+                                export_row.push(parseInt(tmp2));
                             }
                             else {
                                 export_row.push(text.trim());
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -85,7 +85,12 @@ openerp.web_export_view = function (instance) {
                                 }
                             }
                             else if (cell.classList.contains("oe_list_field_integer")) {
-                                tmp2 = text.replace(instance.web._t.database.parameters.thousands_sep, "");
+                                var tmp2 = text;
+                                do {
+                                    tmp = tmp2;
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Thousand separator is going to be handled correctly depending user language?

bringsvor:
Hi

Yes, it uses the same way to get the thousands separator that is used for parsing floats.

Don't know why the build failed, but I don't think it can relate to this change...

web_export_view/static/src/js/web_export_view.js
```diff
@@ -85,7 +85,8 @@ openerp.web_export_view = function (instance) {
                                 }
                             }
                             else if (cell.classList.contains("oe_list_field_integer")) {
-                                export_row.push(parseInt(text));
+                                tmp2 = text.replace(instance.web._t.database.parameters.thousands_sep, "");
```
StefanRijnhart: I think you need to use the **var** keyword to scope the variable tmp2 properly. Alternatively, don't use a temporary variable but perform the operation inline instead.

hbrunn:
:+1:

web_export_view/static/src/js/web_export_view.js
hbrunn: all the js code in this file is a mess (using global scope for all variables), so this makes it only a bit worse i think. but a PR that fixes all global variables would be a good thing of course

StefanRijnhart: Thanks for the clarification. I can agree with that, but I did check if this could then cause any clashes with other tmp2 variables in the core code. I did not find any, but I did find another example of thousands separator replacing, and it occurred to me that replace() in Javascript only replaces the first occurrence so this code is better:

https://github.com/odoo/odoo/blob/8.0/addons/web/static/src/js/formats.js#L245-L249

I suggest @bringsvor that you copy this code verbatim here to cover the case with more than one thousands separator.

bringsvor:
Ok @StefanRijnhart  your wish is my command. Code updated.

StefanRijnhart:
Neat, thanks! :+1: then. Is an upstream proposal?

bringsvor:
Sorry @StefanRijnhart , I don't quite know what that means. I'm rather new to the world of git.

StefanRijnhart:
Sorry, my mistake. I thought we were in OCB. So :+1: regardless.

pedrobaeza:
:+1: then

gurneyalex:
@bringsvor I linked your github login to the OCA database but could not find a CLA for you or your company.

PR #83, opened 2015-02-10
7.0 advanced filters

hbrunn:
Advanced filters suffer from the year old problem that odoo handles negative operators on x2many fields in a weird way: https://bugs.launchpad.net/therp-backports/+bug/1062795

With this commit, I try to detect if we're in a problematic case (when doing a filter's complement) and then evaluate the filter and take this id list's complement. Seems to work nicely.

The other two commits are uninteresting small improvements


pedrobaeza:
I have just found the same bug yesterday with an advanced search in HR categories. Thanks for pointing to the problem. Do you think a patch can be OCB candidate?

About this, :+1:

hbrunn:
from Stefan's MP, you can see how important upstream judges this problem. I think there's no precise right or wrong here, logically, I think we deal with http://www.logicmuseum.com/cantor/Eximport.htm which has several possible solutions. But of course, there's the expected solution. And there's the year old reliance on how it works currently, especially for `ir.rules`.

I think we need a set of prefixed operators that indicate 'we want the exclusive and non-existential' version of `!=`, `not like` etc

web_advanced_filters/model/ir_filters.py
```diff
@@ -65,17 +66,63 @@ def eval_n(domain):
...
+                        domain,
+                        [
+                            [
+                                'id', 'not in',
+                                self.pool[c['model_id']].search(
+                                    cr, uid, eval_n(c['domain']),
+                                    context=context)
+                            ]
+                        ]])
+                else:
```
StefanRijnhart, line 80: ll.73..82 would be more readable if you stored the result of the call to search() in a variable and replace these 10 lines by 

```
domain = expression.AND([domain, [('id', 'not in', neg_ids)]])
```

StefanRijnhart:
Thanks for the workaround. I think a patch on Odoo master would still be valid (maybe not OCB). The impact should be limited: there are no ir.rule in the standard addons of 8.0 that rely on the current semantics.

web_advanced_filters/model/ir_filters.py
hbrunn, line 80: thanks, done

StefanRijnhart:
Thanks! :+1:

PR #85, opened 2015-02-16
[ADD] : port web color to 8.0

AdilHoumadi:
Port web_color to odoo 8.0
Update JS library with the one which used in https://bitbucket.org/anybox/widget_color/


pedrobaeza:
Another great contribution for your part, Adil!

Two things that I think we should do:
- Rename the module to _web_widget_color_, to follow conventions about widgets.
- Remove special type, and bind it to a regular field type, because it loads a lot this module (and it's also non a pure UI module then).

Can you do it, please?

AdilHoumadi:
@pedrobaeza, Thanks for your feedback :+1: 
Yeah, It will be my pleasure to do it.

legalsylvain:
Hi @ah-taktik, 
Thanks for porting this module. I have a question about another PR in V8. #37 . In that PR, the author sayed that his library was better. So @pedrobaeza proposed to use this one and to change 'web_color' to use it. 

I see that you use a third one, no ?
I personaly don't know the different libraries. Don't you ? What do you think about the color_picker used in #37 ?

Other point, minor flake8 / pylint bug to fix, to avoid to crash travis.
https://travis-ci.org/OCA/web/jobs/50990421

Regards.

AdilHoumadi:
Hi @legalsylvain,

Thanks for your remarks, It's clear that the old one offer less choice.
The one which is proposed in this PR offer many choice and less JS files.
For me, it does not matter what library we will choose, I am open to any proposal.
we can agree on any library and I will update the code.

Thanks.

web_widget_color/controller.py
```diff
@@ -0,0 +1,61 @@
...
+##############################################################################
+import openerp.http as http
+from openerp.http import request
+from os.path import join, abspath, exists
+import mimetypes
+
+
+class JsColor(http.Controller):
+    @http.route("/jscolor/<string:image>", type="http", auth="user")
+    def jscolor(self, image):
```
hbrunn: this is needed because detectDir in https://github.com/OCA/web/pull/85/files#diff-8e64e0af10a444a4fcae715f63991980R39 fails, right? Wouldn't it be better then to patch this function to default to '/web_widget_color/static/lib/jscolor'?

AdilHoumadi: @hbrunn, Yeah, sure we can override this function to do that and remove the `/jscolor`route

Thanks a lot.

hbrunn:
:+1:

pedrobaeza:
@ah-taktik, did you tried to change the field type to one of the existing instead of creating a new field type?

hbrunn:
@pedrobaeza it's just a widget with the underlying db type char, or am I missing something?

AdilHoumadi:
Hi @pedrobaeza,
Yeah, In this commit, I have removed the Color field type and i have replaced it with Char 
https://github.com/taktik/web/commit/f3f52bb82c5129156db691640dd19ce2459b259a
Please correct me if I'm missing something ?
Thanks.

pedrobaeza:
OK, sorry for not seeing it. You will need to update then README file and instructions to adequate to this new condition.

AdilHoumadi:
@pedrobaeza, no problem, already done in the same commit : 
https://github.com/taktik/web/commit/f3f52bb82c5129156db691640dd19ce2459b259a?short_path=961331a#diff-961331a2febb328b04f4fffc1e82fc4b

pedrobaeza:
Thanks for the changes, but there's still two changes to do:

Replace:
`It adds a new data type color. To apply it, simply change the field declaration as following:`
with:
`You need to declare a char field of at least size 7:`

and:
`In the view declaration there is nothing especial to do, add the field as any other one. Here is a part of tree view example:`
with:
`In the view declaration, put widget='color' attribute in the field tag:`

About the size, I would recommend not to use it (because Odoo allows it).

One last thing: please change OpenERP by Odoo.

Thanks for all.

AdilHoumadi:
@pedrobaeza, It's done.
Thanks for your remarks, I appreciate that.

pedrobaeza:
Thank you. I proceed with the merge.

PR #93, opened 2015-03-04
[IMP] monkey patch lxml's html cleaner

hbrunn:
in order to allow image data urls in html code (this widget's image upload feature uses this)

I made a PR on lxml for that too: https://github.com/lxml/lxml/pull/162

### DIFF
--- a/web_ckeditor4/models/__init__.py
+++ b/web_ckeditor4/models/__init__.py
@@ -1,9 +1,8 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
 ##############################################################################
 #
 #    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2013 Therp BV (<http://therp.nl>)
-#    All Rights Reserved
+#    This module copyright (C) 2015 Therp BV (<http://therp.nl>).
 #
 #    This program is free software: you can redistribute it and/or modify
 #    it under the terms of the GNU Affero General Public License as
@@ -19,4 +18,4 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from . import models
+from . import ckeditor_monkeypatch
--- a/web_ckeditor4/models/ckeditor_monkeypatch.py
+++ b/web_ckeditor4/models/ckeditor_monkeypatch.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2015 Therp BV (<http://therp.nl>).
+#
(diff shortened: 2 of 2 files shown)


lmignon:
This kind of tags is not allowed by for example gmail... 

[EDIT] Removed thumb down for clarity when checking reviewing state.

hbrunn:
and gmail doing something means what for us?

lmignon:
@hbrunn  For one of our client we have done the same kind of changes to allow sending mails with embedded images. What we have observed is that this kind of tags are not rendered by outlook or gmail. IMO I would prefer to restrict this kind of tags to avoid unpredictable rendering issues due to security reasons. It's just an opinion...

hbrunn:
@lmignon please note that this is about html sanitizing in general. Indeed, a side effect of this is that you can send data uris in html emails, but that's not the reason I made this patch. Data uris pointing to images can be considered safe in my opinion, so no reason to remove them from a security pov.

The reason for the patch: Make a note (or any other html field for that matter) and 'upload' an image with ckeditor. It will be filtered, which is bad for my use case. Here, we're in a context where no emails are involved at all.

With your argument, we should also filter out many style attributes and whatever else most email clients filter out, but I don't think we should

lmignon:
@hbrunn I understand your POV. So you have my :+1:

pedrobaeza:
Is this patch going to solve the problem editing mako/mail templates?

hbrunn:
@pedrobaeza you mean that ckeditor messes up quotes within mako expressions? No. That's something the client side does, the backend receives it already broken. To fix this, you'll have to dig into the code where ckeditor does html sanitizing

pedrobaeza:
OK, well, at least this fix serves for another things.

:+1:

hbrunn:
@pedrobaeza @lmignon https://github.com/lxml/lxml/pull/162 was just merged, so for updated lxmls, we don't need this patch. But reverting this at once will break installations with lxmls without the patch. What do you think is the best approach here?

I'd propose to wait until the first lxml release containing the patch, checking for this version and only apply the monkey patch for older libs, then throw it out completely in a year or so. Agreed? We still can't have versioned external dependencies, right?

Another option would be to throw out the patch a once and raise if lxml's version is too low. Maybe that's cleaner? (but quite radical)

hbrunn:
or also check if `cleaner` has `_is_image_dataurl` in https://github.com/OCA/web/pull/93/files#diff-088933dcd898c94902c0afc388e1b103R34, this way we check if https://github.com/lxml/lxml/pull/162 is applied. That actually seems to be the better alternative to me

pedrobaeza:
Yeah, I like also that last one

hbrunn:
@pedrobaeza agreed, here you go: #215

PR #95, opened 2015-03-13
8.0 web_last_viewed_records

yelizariev:
quick review of the module: http://yelizariev.github.io/odoo/module/2015/02/18/last-viewed-records.html

### DIFF
--- a/web_last_viewed_records/README.rst
+++ b/web_last_viewed_records/README.rst
@@ -1,0 +1,9 @@
+The idea is taken from SugarCRM's "Last viewed" feature.
+
+This module doesn't affect on server performance, because it uses browser's localStorage to save history. But dissadvantage is that history is not synced accross browsers.
+
+FIXME: doesn't work in a res.config view
+
+Tested on 8.0 ab7b5d7732a7c222a0aea45bd173742acd47242d.
+
+Further information and discussion: https://yelizariev.github.io/odoo/module/2015/02/18/last-viewed-records.html
--- a/web_last_viewed_records/README.rst
+++ b/web_last_viewed_records/README.rst
@@ -1,9 +1,46 @@
-The idea is taken from SugarCRM's "Last viewed" feature.
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
 
-This module doesn't affect on server performance, because it uses browser's localStorage to save history. But dissadvantage is that history is not synced accross browsers.
+Last viewed records
+===================
 
-FIXME: doesn't work in a res.config view
+The module shows links to last viewed records and pages.
 
-Tested on 8.0 ab7b5d7732a7c222a0aea45bd173742acd47242d.
+The module doesn't affect on server performance, because it uses
+browser's localStorage to save history. But dissadvantage is that
+history is not synced accross browsers.
 
-Further information and discussion: https://yelizariev.github.io/odoo/module/2015/02/18/last-viewed-records.html
+Usage
+=====
+
+There are no any settings to configure. Just install module and continue to use odoo in a usual way. You will see links to last visited pages.
+
+For further information, please visit:
+
+ * https://yelizariev.github.io/odoo/module/2015/02/18/last-viewed-records.html
+
+Known issues / Roadmap
+======================
+
+ * FIXME: doesn't work in a res.config view
+
+Credits
+=======
+
+Contributors
+------------
+
+* Ivan Yelizariev <yelizariev@it-projects.info>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/web_last_viewed_records/__openerp__.py
+++ b/web_last_viewed_records/__openerp__.py
@@ -1,7 +1,7 @@
 {
     'name' : 'Last viewed records',
     'version' : '1.0.0',
-    'author' : 'Ivan Yelizariev',
+    'author' : 'Ivan Yelizariev, Odoo Community Association (OCA)',
     'category' : 'Base',
     'website' : 'https://yelizariev.github.io',
     'depends' : ['web', 'mail'],
--- a/web_last_viewed_records/__openerp__.py
+++ b/web_last_viewed_records/__openerp__.py
@@ -1,0 +1,26 @@
+{
+    'name' : 'Last viewed records',
+    'version' : '1.0.0',
+    'author' : 'Ivan Yelizariev',
+    'category' : 'Base',
+    'website' : 'https://yelizariev.github.io',
+    'description': """
+The idea is taken from SugarCRM's "Last viewed" feature.
+
+This module doesn't affect on server performance, because it uses browser's localStorage to save history. But dissadvantage is that history is not synced accross browsers.
+
+FIXME: doesn't work in a res.config view
+
+Tested on 8.0 ab7b5d7732a7c222a0aea45bd173742acd47242d.
+
+Further information and discussion: https://yelizariev.github.io/odoo/module/2015/02/18/last-viewed-records.html
+    """,
+    'depends' : ['web', 'mail'],
+    'data':[
+        'views.xml',
+        ],
+    'qweb' : [
+        "static/src/xml/*.xml",
+    ],
+    'installable': True
+}
--- a/web_last_viewed_records/__openerp__.py
+++ b/web_last_viewed_records/__openerp__.py
@@ -1,14 +1,14 @@
 {
-    'name' : 'Last viewed records',
-    'version' : '1.0.0',
-    'author' : 'Ivan Yelizariev, Odoo Community Association (OCA)',
-    'category' : 'Base',
-    'website' : 'https://yelizariev.github.io',
-    'depends' : ['web', 'mail'],
-    'data':[
+    'name': 'Last viewed records',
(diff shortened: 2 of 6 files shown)


bealdav:
Really really nice. Thanks a lot

rvalyi:
looks very cool indeed. However I'm not sure if OCA can accept such a module without tests (might break silently and OCA fail its promise of a stable world).

bealdav:
@rvalyi This module operates client side only. Which test could we suggest ? Is there an example ?

web_last_viewed_records/README.rst
```diff
@@ -0,0 +1,9 @@
+The idea is taken from SugarCRM's "Last viewed" feature.
```
pedrobaeza: Please use OCA's README template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

web_last_viewed_records/README.rst
```diff
@@ -0,0 +1,9 @@
...
+
+This module doesn't affect on server performance, because it uses browser's localStorage to save history. But dissadvantage is that history is not synced accross browsers.
+
+FIXME: doesn't work in a res.config view
+
+Tested on 8.0 ab7b5d7732a7c222a0aea45bd173742acd47242d.
+
+Further information and discussion: https://yelizariev.github.io/odoo/module/2015/02/18/last-viewed-records.html
```
pedrobaeza: You can copy/paste this page into here or static/description/index.html, like in https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/index.html

web_last_viewed_records/__openerp__.py
```diff
@@ -0,0 +1,15 @@
+{
+    'name': 'Last viewed records',
+    'version': '1.0.0',
+    'author': 'Ivan Yelizariev, Odoo Community Association (OCA)',
+    'category': 'Base',
+    'website': 'https://yelizariev.github.io',
+    'depends': ['web', 'mail'],
```
pedrobaeza, line 7: Why do you need mail?

web_last_viewed_records/__openerp__.py
```diff
@@ -0,0 +1,15 @@
...
+    'data': [
+        'views.xml',
+        ],
+    'qweb': [
+        "static/src/xml/*.xml",
```
pedrobaeza: Please better put an explicit line for each xml.

web_last_viewed_records/static/src/js/main.js
```diff
@@ -0,0 +1,174 @@
...
+        },
+        load_last_viewed_history: function(){
+            var data = localStorage[this.last_viewed_history_var] || '{}';
+            return JSON.parse(data);
+        },
+        save_last_viewed_history: function(data){
+            localStorage[this.last_viewed_history_var] = JSON.stringify(data);
+        },
+        // icon map: http://bistro.convergencecms.co/entypo
+        _model2icon: {
```
pedrobaeza, line 53: Can this be extracted someway to allow easy extending for more models?

web_last_viewed_records/static/src/js/main.js
```diff
@@ -0,0 +1,174 @@
...
+            'project.project':'t',
+            'project.task':'W',
+            'account.invoice':'h',
+            'ir.module.module':'Z',
+            'hr_timesheet_sheet.sheet': 'N',
+            'res.groups': ',',
+            'res.company': '_',
+            'res.user': 'ó',
+            'gamification.challenge':'è',
+            'gamification.badge':'8',
```
pedrobaeza: Another great option is if you can support Font awesome here: http://fortawesome.github.io/Font-Awesome/icons/ It's already used in Odoo in smart-buttons.

pedrobaeza:
Great module! I have put it some remarks inline.

Thanks for your work and this first contribution. I hope this is the beginning of a large number of contributions.

web_last_viewed_records/__openerp__.py
yelizariev, line 7: There is a reference to mail at static/src/xml/main.xml
Mail inbox page use own template which has to be extended in a different way.

Option to exclude this dependencies is to make another module
**web_last_viewed_records_mail**
with a settings

```
'depends': ['mail', 'web_last_viewed_records'],
'auto_install': True
```

and that module would containt only this xml:

```
<templates>
    <t t-extend="mail.wall">
        <t t-jquery="tr.oe_header_row_top" t-operation="after">
            <t t-call="web_last_viewed_records.header">
                <t t-set="colspan" t-value="2"/>
            </t>
        </t>
    </t>
</templates>
```

web_last_viewed_records/static/src/js/main.js
yelizariev: good idea!

web_last_viewed_records/static/src/js/main.js
yelizariev, line 53: I think icons for main models can be hardcoded at the module.

For extending one can create a custom module to redefine _model2icon variable.

yelizariev:
@pedrobaeza thank you for review. 
I've made some updates and posted comments for the questions

bealdav:
I have tried it, all is ok
:+1:

rvalyi:
@bealdav javascript code can still be tested https://github.com/odoo/odoo/tree/8.0/addons/web/static/test
JS tests can be launched manually from the web client and some JS tests (like the website tour) are launched using PhantomJS which is integrated in the test suite.
In fact I feel that OCA reviewers aren't usually very good at Javascript so sadly unested Javascript code is especially likely to break silently.

legalsylvain:
Hi @rvalyi , Very interesting point.
Do you have some sample of JS Test integrated with Travis ? (Odoo modules you wrote or you saw).
Regards.

bealdav:
@rvalyi, I think you're right but for a first contribution, maybe we can accept this PR as it ? No damage can be done to the data. If problem, just uninstall the module.

legalsylvain:
@bealdav, Of course we can ! There are actually 15 modules in V8 series in OCA / web, and no module have any JS tests. For the time being, JS test is not an OCA standard. It's not a matter of first contribution.
@rvalyi, I propose you to open a general issue in OCA / web, with description of how we could integrate JS test in Odoo with OCA CIS (Travis + runbot). Concrete examples would be welcome. This point is out of the scope of this PR.

Regards.

web_last_viewed_records/static/src/xml/main.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates>
```
hbrunn, line 2: please format your xml in some coherent way

web_last_viewed_records/static/src/js/main.js
```diff
@@ -0,0 +1,178 @@
+openerp.web_last_viewed_records = function(instance){
+    var QWeb = instance.web.qweb;
+    var _t = instance.web._t;
```
hbrunn: those two are unused

web_last_viewed_records/static/src/js/main.js
```diff
@@ -0,0 +1,178 @@
+openerp.web_last_viewed_records = function(instance){
```
hbrunn, line 1: I think you should test if the localStorage object exists here and continue only if so

hbrunn:
I added some tests in https://github.com/OCA/web/blob/7.0/web_pytz/static/test/web_pytz.js and https://github.com/OCA/web/blob/7.0/web_relativedelta/static/test/web_relativedelta.js for 7.0 - but I didn't have a look yet how this has changed for 8.0

In general, adding tests is a good thing, but to be honest, I wouldn't really know what to test here as the code will be tested as a side effect by any client side view test. So no, I don't think it's fair to block this contribution for lack of tests. http://odoo-community.org/page/how-to says tests are much encouraged, which I understand as them not being mandatory

web_last_viewed_records/static/src/js/main.js
yelizariev, line 1: I think it should be discussed. Odoo's point_of_sale doesn't test existence of localStorage. https://github.com/odoo/odoo/blob/master/addons/point_of_sale/static/src/js/db.js
Is it a bug in odoo or there are some logic in it?

hbrunn, line 1: odoo's pos is an isolated piece of software that doesn't break other things. If you have a browser that doesn't support localStorage, you can't use POS, but the rest of odoo. If I install this module, everything breaks on an old browser

yelizariev, line 1: You are right. I've added test for localStorage

hbrunn:
:+1:

web_last_viewed_records/static/src/xml/main.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<templates>
+    <t t-name="web_last_viewed_records.header">
+        <tr class="oe_header_row">
+            <td t-att-colspan="colspan or '3'">
+                <div class="oe_view_manager_last_viewed"/>
+            </td>
+            <td></td>
```
legalsylvain, line 8: Hi, 
Excellent feature :+1:  Thanks for the localstorage test. 
One thing :
This line (empty td) generate an empty space useless. Is it usefull ? 

I tried the following code and it works better. But I don't know the impact on mail.wall...

```
t t-name="web_last_viewed_records.header"
    tr class="oe_header_row"
        td t-att-colspan="4"
            div class="oe_view_manager_last_viewed"/
        /td
    /tr
/t
```

capture du 2015-03-16 15_34_27

yelizariev, line 8: As I remember, I added extra **td** to put there some control buttons in the future.

legalsylvain, line 8: OK.

pedrobaeza:
:+1: for me too. I proceed with the merge.

PR #96, opened 2015-03-13
8.0 web widget x2many 2d matrix

hbrunn:
# 2D matrix for x2many fields

This module allows to show an x2many field with 3-tuples
($x_value, $y_value, $value) in a table

|  | $x_value1 | $x_value2 |
| --- | --- | --- |
| $y_value1 | $value1/1 | $value2/1 |
| $y_value2 | $value1/2 | $value2/2 |

where `value(n/n)` is editable.

An example use case would be: Select some projects and some employees so that a manager can easily fill in the planned_hours for one task per employee. The result could look like this:

screenshot
# Usage

Use this widget by saying

``` xml
<field name="my_field" widget="x2many_2d_matrix" />
```

This assumes that my_field refers to a model with the fields `x`, `y` and
`value`. If your fields are named differently, pass the correct names as
attributes

``` xml
<field name="my_field" widget="x2many_2d_matrix"
 field_x_axis="my_field1" field_y_axis="my_field2" field_value="my_field3" />
```

You can pass the following parameters:

field_x_axis
    The field that indicates the x value of a point
field_y_axis
    The field that indicates the y value of a point
field_label_x_axis
    Use another field to display in the table header
field_label_y_axis
    Use another field to display in the table header
field_value
    Show this field as value
show_row_totals
    If field_value is a numeric field, calculate row totals
show_column_totals
    If field_value is a numeric field, calculate column totals
# Known issues / Roadmap
- it would be better to instantiate the proper field widget and let it render the input


web_widget_x2many_2d_matrix/static/src/xml/web_widget_x2many_2d_matrix.xml
```diff
@@ -0,0 +1,36 @@
...
+                        </th>
+                        <th t-if="widget.show_row_totals">Total</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr t-foreach="widget.get_y_axis_values()" t-as="y" t-att-data-y="y">
+                        <th><t t-esc="widget.get_y_axis_label(y)" /></th>
+                        <td t-foreach="widget.get_x_axis_values()" t-as="x" t-att-class="'oe_list_field_cell' + (widget.is_numeric ? ' oe_number' : '')">
+                            <span t-att-class="widget.get_xy_value_class()">
+                                <input style="display: none" t-att-data-x="x" t-att-data-y="y" t-att-data-id="widget.get_xy_id(x, y)" t-att-value="widget.format_xy_value(widget.get_xy_value(x, y))" />
```
hbrunn: TODO to self: it doesn't make sense to fill in x and y on the input. Better put data-x in the td, and leave data-y in tr, then we can select everything by using `tr[data-y="y"] td[data-x="x"]`

pedrobaeza:
Very useful!

:+1:

sebastienbeau:
Reeaaaaaallllllyyyyy coolll :100: :+1:

PR #98, opened 2015-03-15
[ADD] web_tree_many2one_clickable

pedrobaeza:
# Clickable many2one fields for tree views

This addon provides a separate widget to allow many2one fields in a tree view
open the linked resource when clicking on their name.

You can also define a system parameter to have this behaviour for all the
existing many2one fields in tree views.
# Installation

Install it the regular way.
# Configuration

If you want to have all many2one fields clickable by default, you have to
define in _Configuration > Technical > Parameters > System parameters_, a new
parameter with name `web_tree_many2one_clickable.default` and with value
`true`.
# Usage

For the widget option, you need to add `widget="many2one_clickable"` attribute
in the XML field definition in the tree view.

For example:

`<field name="partner_id" widget="many2one_clickable" />`

will open the linked partner in a form view.
# Known issues / Roadmap
- You cannot deactivate clickable behaviour for an specific many2one field if
  you configure the system parameter.
- The value of the system parameter is retrieved for each many2one field
  present in the view instead of only once.


web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -0,0 +1,105 @@
...
+        */
+        init: function(id, tag, attrs) {
+            this._super(id, tag, attrs);
+            if (this.type == 'many2one') {
+                this.get_options();
+            }
+        },
+
+        get_options: function() {
+            if (_.isUndefined(this.ir_options_loaded)) {
```
hbrunn: I'd choose a more specific name here so that this doesn't clash with some other marker variable

pedrobaeza: Done!

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -0,0 +1,105 @@
...
+                        self.ir_options_loaded.resolve();
+                    });
+                return this.ir_options_loaded;
+            }
+            return $.when();
+        },
+
+        _format: function (row_data, options)
+        {
+            if (this.use_many2one_clickable) {
```
hbrunn, line 60: don't you need to check for the field's type here and do this only if it's a many2one field?

pedrobaeza, line 60: No, because in case the field type != many2one, this contains an undefined value, so evaluation is false.

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -0,0 +1,105 @@
...
+                return _.str.sprintf('<a class="oe_form_uri" data-many2one-clickable-model="%s" data-many2one-clickable-id="%s">%s</a>',
```
hbrunn, line 61: how about a refactoring so that this and instance.web_tree_many2one_clickable.Many2OneClickable._format call the same function on `instance.web.list.Column`?

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
hbrunn, line 60: clear, thanks

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
pedrobaeza, line 61: Yeah, thanks for the suggestion. I have done it other way. Please check it.

hbrunn, line 61: nice

hbrunn:
:+1:

guewen:
:+1:

antespi:
:+1:

PR #100, opened 2015-03-18
[FIX] evaluate domains containing _auto_join fields

hbrunn:
before taking unions


hbrunn:
This is necessary because _auto_join fields behave like a join (big surprise). So if we combine a filter with another one that uses auto_joined fields, we'll only have results returned from the join

guewen:
:+1:

StefanRijnhart:
:+1:

PR #103, opened 2015-03-24
8.0 [ADD] web_widget_datepicker_options

veloutin:
# 
# Allow passing options to datepicker widgets

This will set all options specified in the "datepicker" option of datetime
fields to the datepicker.

See http://api.jqueryui.com/datepicker/ for options
## Example:

 `<field name="birthdate" options="{'datepicker':{'yearRange': 'c-100:c+0'}}"/>`


web_widget_datepicker_options/static/src/js/datepicker.js
```diff
@@ -0,0 +1,53 @@
...
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+/*global openerp, _, $ */
+
+openerp.web_datetime_options = function (instance) {
+
+    "use strict";
+
+    instance.web.form.FieldDatetime = instance.web.form.FieldDatetime.extend({
```
hbrunn: better say

```
instance.web.form.FieldDatetime.include({
    //your stuff
});
```

because otherwise, you replace the reference to the field by the reference to another object. This can cause subtle bugs that are very difficult to debug

web_widget_datepicker_options/static/src/js/datepicker.js
```diff
@@ -0,0 +1,53 @@
...
+                if (typeof this.options.datepicker === 'object') {
+                    $.map(this.options.datepicker, function(value, key) {
+                        self.datewidget.picker('option', key, value);
+                    });
+                }
+            }
+        }
+    });
+
+    instance.web.form.FieldDate = instance.web.form.FieldDate.extend({
```
hbrunn: see above

veloutin:
@hbrunn Fixed in both PRs

web_widget_datepicker_options/__openerp__.py
```diff
@@ -0,0 +1,57 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": 'web_widget_datepicker_options',
+    "version": "1.0",
+    "description": """
+===========================================
```
pedrobaeza: Please use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

web_widget_datepicker_options/__openerp__.py
```diff
@@ -0,0 +1,55 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    "name": 'web_widget_datepicker_options',
+    "version": "1.0",
+    "description": """
```
yvaucher: As you created `README.rst`, please remove `description` thus `README.rst` content will be displayed on app.odoo.com and in odoo.

yvaucher:
Thanks
:+1: 

If you can also add the OCA icon and the AGPL badge it would be perfect.
https://github.com/OCA/maintainer-tools/tree/master/template/module

web_widget_datepicker_options/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+
+{
+    "name": 'web_widget_datepicker_options',
+    "version": "1.0",
+    "depends": [
+        'base',
+        'web',
+    ],
+    "js": [
+        'static/src/js/datepicker.js',
```
anthony-muschang: datepicker.js will never be loaded with this type of reference in 8.0.

In 8.0 the javascript files must be referenced in a qweb template like in: https://github.com/OCA/web/blob/8.0/web_dialog_size/view/qweb.xml#L6

veloutin:
Thanks @anthony-muschang , I've changed it according to your suggestion, and it worked.

StefanRijnhart:
:+1:

PR #104, opened 2015-03-24
[ADD] backport web_widget_color

### LINKED ISSUE #51: 'web_color' module doesn't work
Hi all, 

I tried today to use 'web_color' module, but it failed. Description of the module is very small. Are some people that use it in some custom modules ?
Regards.

My code: 

[From Model]

```
class MyModel(Model):
    _columns = {
        'color': fields.char(
            'Color', size=7,
            help="Color of the Price Tag by default. Format #RRGGBB"),
```

[From view]

```
 field name="color" widget="color"
```

[from openerp file]

```
depends = ['web_color']
```

At the end, i have a Javascript error. (Mozilla / Chrome)

```
TypeError: obj is not a constructor
```

hbrunn:
and replace web_color with it. Fixes #51


hbrunn:
This module actually works in form views and editable tree views

legalsylvain:
Hi @hbrunn,
:question: 
For me, it is quite problematic to remove a module in a 'stable and used in production' repository. module renaming or replacement should be only possible when we create a new milestone for a project, and when no module is ported.
I imagine the case when an admin do a regular update of the repository and break his instance because the old module is not available and the new one is uninstalled.
What do you think about that point ?
But I can imagine to allow backporting 'web_widget_color' 8.0 -> 7.but to let the old module in the repository with a advice in the openerp file to indicate that the old module is not the best one.
Regards.

hbrunn:
@legalsylvain good point. In this specific case, web_widget_color can be used as a drop-in replacement for web_color, so i think a symlink from web_color to web_widget_color would be sufficient. Any objections?

pedrobaeza:
Yeah, for me it's enough.

legalsylvain:
@hbrunn, oh I see. Symlink is a good idea ! Will it work for all platform ? (windows / linux / ...) 
If yes, I have no objection.
In that case, could you increase version to 1.1 for the new module (for the time being 1.0 in both module) ? So, it will indicate clearly that an update of the module is necessary. (because for exemple, there is xml to update (lib.xml to unload and widget.xml to load).
Regards.

legalsylvain:
@hbrunn any news about my question ?

legalsylvain:
Hi, 
No news about the symlink compatibility on various system ? If not, @hbrunn, could you imagine to revert the 'web_color' deletion in this PR ? So we could merge this PR and use this great module.
('web_color' deletion is another point for me and could be treated in an other PR.)

hbrunn:
@legalsylvain the news is that I was on vacation ;-)

I replaced the branch by one only adding web_widget_color, then we can discuss everything else separately.

Symlinks with git on windows seem to be a problem: http://stackoverflow.com/questions/5917249/git-symlinks-in-windows (which surprises me, because NTFS can do that, even with some features standard linux links miss)

legalsylvain:
Hi @hbrunn , sorry for the harassment. I hope your hollidays was great !
:+1: (Code review + Test) Great module, thanks a lot.

pedrobaeza:
:+1:

PR #109, opened 2015-04-08
[ADD] Added web_remove_developer_mode_link.

alejandrosantana:
As some customers want to restrict 'Developer Mode' for most users, this is simple module that removes the link in "About" dialog that activates Developer Mode.
It still allows to type ?debug in URL for those technical users that actually need this.

As a sidenote, first I wanted to make it optional and be able to show it only to a certain group of users, but it seems impossible to apply "groups" policy in qweb-json temapletes (see https://github.com/odoo/odoo/issues/5058 and https://github.com/odoo/odoo/issues/6190). Or, at least, I do not know how to do it.
As other option I thought in allowing only Administrator user (uid == 1), but I do not know how to get user id in such qweb template.
I tried this with out any luck (nor 'groups' nor uid):

```
    <t t-extend="UserMenu.about">
        <t t-jquery=".oe_activate_debug_mode" t-operation="replace">
            <t t-if="uid == 1">
                <a class="oe_activate_debug_mode oe_right label label-primary" href="?debug" groups="base.group_no_one">Activate the developer mode</a>
            </t>
        </t>
    </t>
```

I am open to ideas to enhace this.

### DIFF
--- a/web_remove_developer_mode_link/static/src/xml/base.xml
+++ b/web_remove_developer_mode_link/static/src/xml/base.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<templates id="template" xml:space="preserve">
+
+    <t t-extend="UserMenu.about">
+        <t t-jquery=".oe_activate_debug_mode" t-operation="replace"></t>
+    </t>
+
+</templates>
--- a/web_remove_developer_mode_link/static/src/xml/base.xml
+++ b/web_remove_developer_mode_link/static/src/xml/base.xml
@@ -3,7 +3,9 @@
 <templates id="template" xml:space="preserve">
 
     <t t-extend="UserMenu.about">
-        <t t-jquery=".oe_activate_debug_mode" t-operation="replace"></t>
+        <t t-jquery=".oe_activate_debug_mode" t-operation="replace">
+            <a class="oe_activate_debug_mode oe_right label label-primary hidden" href="?debug" groups="base.group_no_one">Activate the developer mode</a>
+        </t>
     </t>
 
 </templates>
--- a/web_remove_developer_mode_link/README.rst
+++ b/web_remove_developer_mode_link/README.rst
@@ -1,0 +1,36 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Demove Developer Mode link
+==========================
+
+This module removes the link that activates the Developer Mode.
+
+
+Known issues / Roadmap
+======================
+
+* It would be desirable to make the link available to a certain group of users.
+
+Credits
+=======
+
+Contributors
+------------
+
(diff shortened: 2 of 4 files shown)


alejandrosantana:
Mmm... oca-clabot does not recognize me, yet I have signed it and I am current with fees.
Anything I can do?

pedrobaeza:
@gurneyalex, can you please check the CLA of Alejandro?

gurneyalex:
I updated @alejandrosantana information in the OCA database.

alejandrosantana:
Do I need to do anything?

gurneyalex:
@alejandrosantana no it's ok. If you push again on this PR (e.g. after a commit --amend), there will be a message saying it's ok, but it is not required. I confirm your CLA is ok

web_remove_developer_mode_link/static/src/xml/base.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<templates id="template" xml:space="preserve">
+
+    <t t-extend="UserMenu.about">
+        <t t-jquery=".oe_activate_debug_mode" t-operation="replace"></t>
```
pedrobaeza: You can add style hidden to avoid removing entirely.

legalsylvain:
I reopen this PR and close #110 to keep comments.
Regards.

legalsylvain:
Hi @alejandrosantana Thanks for this module.

I have no idea to restrict acces for only a category of users. (maybe getting context or session('uid') in js. Sample (works in 7.0, untested in 8.0): 
In a JS file, you can extend an object that inherit from instance.web.Widget.

```
myWidget =  instance.web.Widget.extend({
    init: function(parent, ..., ...) {
       // TODO add some tests to avoid error if not defined
        console.log(parent.session.uid);
       // TODO process depending of the value 'uid'
    },
});
```

But It's a non blocking point and it doesn't mind because the admin know the behaviour of this module and write ?debug in the url or click on "About OpenERP" + "debug mode" take the same time...
:+1: (Code review, No test).

web_remove_developer_mode_link/static/src/xml/base.xml
alejandrosantana: Agree. New commit in a minute.

alejandrosantana:
@legalsylvain Yes, I have to research on that (as it would be useful in other cases).
I have no clue on JS so far, but it seems there is more and more JS involved in Odoo as time goes on (not a thing I like at all, but that's life). I will check what you propose.

But for now, as you say, it is not blocking and I would like to give a raw functionality now, and enhance it later.

alejandrosantana:
@legalsylvain I have tried with JS and I managed to make the behaviour group-based, but with some weird stuff.

I close this PR, as new PR https://github.com/OCA/web/pull/128 has a more advanced behaviour.

PR #110, opened 2015-04-08
[ADD] Added web_remove_developer_mode_link.

alejandrosantana:
As some customers want to restrict 'Developer Mode' for most users, this is simple module that removes the link in "About" dialog that activates Developer Mode.
It still allows to type ?debug in URL for those technical users that actually need this.

As a sidenote, first I wanted to make it optional and be able to show it only to a certain group of users, but it seems impossible to apply "groups" policy in qweb-json temapletes (see https://github.com/odoo/odoo/issues/5058 and https://github.com/odoo/odoo/issues/6190). Or, at least, I do not know how to do it.
As other option I thought in allowing only Administrator user (uid == 1), but I do not know how to get user id in such qweb template.
I tried this with out any luck (nor 'groups' nor uid):

```
    <t t-extend="UserMenu.about">
        <t t-jquery=".oe_activate_debug_mode" t-operation="replace">
            <t t-if="uid == 1">
                <a class="oe_activate_debug_mode oe_right label label-primary" href="?debug" groups="base.group_no_one">Activate the developer mode</a>
            </t>
        </t>
    </t>
```

I am open to ideas to enhace this.
Same PR as https://github.com/OCA/web/pull/109 , but sent again due to OCA CLA verification.

### DIFF
--- a/web_remove_developer_mode_link/README.rst
+++ b/web_remove_developer_mode_link/README.rst
@@ -1,0 +1,36 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Demove Developer Mode link
+==========================
+
+This module removes the link that activates the Developer Mode.
+
+
+Known issues / Roadmap
+======================
+
+* It would be desirable to make the link available to a certain group of users.
+
+Credits
+=======
+
+Contributors
+------------
+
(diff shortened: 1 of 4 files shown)


legalsylvain:
Duplicate ref #109

PR #111, opened 2015-04-09
[7.0] web_dashboard_tile : Fixes & Improvements

legalsylvain:
This PR:
- Fix a bug, if you create a 'global tile' (without user) and some of your users doesn't belong to a group. Sample : you want to display pos.order count, but some users doesn't belong to 'point_of_sale.user' ;
  - Before : error, displaying the dashboard for the user; (access denied)
  - Now : the tile is hidden;
- Add the possibility to compute a function on a field of the current model. Function can be 'sum' / 'avg' / 'min' / 'max' / 'median' : 
  screenshot_dashboard
- Use the new module web_widget_color (BTW, thanks to @hbrunn for backporting this great module) 
  - possiblity to select a background color (before, it was a random color between 0 and 9 with defined color set in a CSS file) ; 
  - script migration provided to change integer color (v 0.4) to RGB color (v 1.0) ;
  - possiblity to select a font color (before, static value) ; 
- Some OCA refactoring: 
  - demo data;
  - 'view' folder;

CC: @OSguard


OSguard:
Wow greate stuff, i will test and review soon

legalsylvain:
@OSguard Thanks !
Please say me if the migration script works correctly for your database. (I guess you use this module on production databases). Theoriticaly, your dashboard must not have color changes after the update.

bealdav:
Impressive work, all the roadmap done.

legalsylvain:
Hi @OSguard , did you have the time to review this PR ? 
Kind regards.

OSguard:
Sorry not yet. My suggestion what i think about is, that if we compute a field (like sum) we should replace the the count value with the sum. This is in my opinion the more useful information. Then remove the count or make it small (or only show on mouseover) "based on #3" - 3 means count

legalsylvain:
Hi @OSguard, I tested your suggestion. It doesn't appear to be most readable : 
- If there are some tiles with functions and other without, in the current commit, the count number is always in the same place, that is more simple to have a quick overview;
- I can not increase a lot the size of the functional result, because the number (total) can be very high; (billions...) ;
- I understand your PoV about "the total is the most important", but we can imagine too that "the count is the most important". In fact, the count is the quantity of things the users have to do. (3 Delivery Orders to confirm, 4 Invoices without payment...). IE : I see that module as a todo list for end users. But this a matter of tastes and colours.

Could we imagine to approve this functionality as it, and maybe to propose improvements in other PR, or when we will port this module to 8.0 serie.

What do you think ?

Regards.

OSguard:
My idea is to develop it also to a kpi dashboard. 
- So you login and get a overview: Today revenue, How many orders need shipped. ... So then only one number is for the user interesting. As he self defined the dashboard he pretty know what the meaning is
- for bigger numbers i would include shortes: 104K for 104.345 and 1.2M fo 1.234.245 .. and so on. So that you get the important information. On hover you can get the exact detail.
- You are right for todo the count matters, for manager the sum matters so it should be configurable what you need and this is the big number and fit both

@legalsylvain i will be join the next OCA code sprint you are their?

web_dashboard_tile/view/tile.xml
```diff
@@ -57,16 +66,27 @@
                                     </ul>
```
OSguard: this is not working any more, then we just need to remove it

OSguard:
Test & Review:
- Migration of old colors works well
- Not possible to add a function + field from the dashboard dropdown

web_dashboard_tile/static/src/css/tile.css
```diff
@@ -1,75 +1,44 @@
 .openerp .oe_kanban_view .oe_dashbaord_tile{
     width: 150px;
     height: 150px;
-    border: 0;
+    border: 1px solid;
```
OSguard: i don't see any reason to add a border. As 'flat design' is alway without ;)

legalsylvain:
> Not possible to add a function + field from the dashboard dropdown

yes, it is written in limit section of the manifest file.

Other points : fixed.

@OSguard : yes I'll be in OCA Code sprint. We can see us to do some improvements !
Are you OK to approve this PR, or are there some blocking points ?

Regards.

OSguard:
:+1: this makes some nice new features and is the next step

bealdav:
:+1:  Thank for your job.

PR #112, opened 2015-04-09
Added new module web_widget_options

llacroix:
This module allow the user to set some rules that will control the visibility of the "Add item" on a `one2many` field. This module can be extended to support more than the `create` attributes.

It is possible to extend the set of rules. 

It currently can test for the action_id and if the user has all_groups defined in `groups`.

A rule looks like this on a tree object.

```
<tree create_rule='{"groups": [1,2], "action_id": 34}' ...>
```

Here's a more complex example of how to use:

```
<xpath expr="//tree" position="attributes">
      <attribute name="create_rule">
        {
          "groups": [%(some_group_id)d],
          "action_id": %(some_action_group)d
        }
      </attribute>
  </xpath>
```

### DIFF
--- a/web_widget_options/static/src/js/web_widgets.js
+++ b/web_widget_options/static/src/js/web_widgets.js
@@ -1,0 +1,74 @@
+openerp.web_widget_options = function(instance, m) {
+  var _t = instance.web._t,
+      QWeb = instance.web.qweb;
+  var wwo = instance.wwo = {};
+  var o2m = instance.wwo.o2m = {}
+
+  o2m.rules = {
+    groups: function (rule, context) {
+      var user_id = instance.session.uid,
+          users = new instance.web.Model('res.users'),
+          query = users.query(['name', 'groups_id']).filter([['id', '=', user_id]]);
+
+      return query.all().then(function (res) {
+               var user = res[0];
+
+               return rule.reduce(function (accum, val) {
+                 return accum && user.groups_id.indexOf(val) >= 0;
+               }, true);
+             });
+    },
+
+    action_id: function (rule, context) {
+      var action_id = context.view.options.action.id;
+      return rule == action_id;
+    }
+  };
+
+  function and(a, b) {
+    return a && b;
+  };
+
+  function can_add_items(view_root, field_name, rules) {
+    var def = $.Deferred(),
+        context = {
+          view: view_root,
+          field_name: field_name,
+          rule: rules
+        },
+        pending = Object.keys(rules).map(function (rule_name) {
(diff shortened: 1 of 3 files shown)


legalsylvain:
:question: 
Hi, In your sample, is it possible to write "groups": base_group_no_one", (for exemple), or do we have to write the id of the group. If it's not possible, it will be great do add this feature.

llacroix:
@legalsylvain in the xml file you can always use this syntax:

```
<tree create_rule='{"groups": [%(base.group_system)d], "action_id": %(some_action)d}' ...>
```

The same thing apply for the action_id.

llacroix:
For now, since I'm using internally JSON.parse the content in the attribute has to be a valid json string. Which means that is invalid:

```
{'groups': [2]}
```

But this is valid:

```
{"groups": [2]}
```

@legalsylvain If you had in mind of lazy loading ids, it doesn't do that now. It could be added in the future if the type is a string instead of a number but I believe it is better to have numbers only to reduce the amount of traffic needed to handle this.

web_widget_options/static/src/js/web_widgets.js
```diff
@@ -0,0 +1,73 @@
...
+  o2m.rules = {
+    groups: function (rule, context) {
+      var user_id = instance.session.uid,
+          users = new instance.web.Model('res.users'),
+          query = users.query(['name', 'groups_id']).filter([['id', '=', user_id]]);
+
+      return query.all().then(function (res) {
+               var user = res[0];
+
+               return rule.reduce(function (accum, val) {
```
hbrunn, line 16: wouldn't `_.intersection()` be more convenient here?

hbrunn:
It would be good to put your first comment into the manifest's description, that makes it a lot clearer what this module does.

Did you consider to make the attributes `create`, `edit` and `delete` smarter instead? I think about (py/js)evaling (currently, it's js I think) them on view load with the user record and the context. This way, you could do a lot more for the same price

web_widget_options/static/src/js/web_widgets.js
llacroix, line 16: Something like this? 

```
return _.intersection(rule, user.groups_id).length == rule.length
```

llacroix:
@hbrunn  That would be cool to have python eval yet I'm not yet aware of how to do that.

hbrunn:
you can find an example here: https://github.com/OCA/web/blob/8.0/web_context_in_colors/static/src/js/web_context_in_colors.js#L34

hbrunn:
closing due to inactivity

PR #113, opened 2015-04-10
[FIX] web_tree_many2one_clickable: Don't complain if there is no parameter defined

pedrobaeza:
To avoid an ugly JS error if you don't define the system parameter (nevertheless the value).

### DIFF
--- a/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
+++ b/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
@@ -47,7 +47,9 @@ openerp.web_tree_many2one_clickable = function(instance, local)
                     .filter([['key', '=', 'web_tree_many2one_clickable.default']])
                     .first()
                     .then(function(param) {
-                        self.use_many2one_clickable = (param.value == 'true');
+                        if (param) {
+                            self.use_many2one_clickable = (param.value == 'true');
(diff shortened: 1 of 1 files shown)


bwrsandman:
:+1:

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -47,7 +47,9 @@ openerp.web_tree_many2one_clickable = function(instance, local)
                     .filter([['key', '=', 'web_tree_many2one_clickable.default']])
                     .first()
                     .then(function(param) {
-                        self.use_many2one_clickable = (param.value == 'true');
+                        if (param) {
+                            self.use_many2one_clickable = (param.value == 'true');
+                        }
```
dreispt, line 52: Far from being an expert at Js, but could this work?:

```
self.use_many2one_clickable = (param and param.value);
```

pedrobaeza, line 52: I don't think so. The evaluation is not done in the same way as in Python

dreispt, line 52: OK, ignore it then.

dreispt:
:+1:

PR #114, opened 2015-04-13
[IMP] web_m2x_options

JordiBForgeFlow:
This PR is a backport from the 8.0, that adds the feature that makes it possible to disable "Create..." and "Create and Edit..." entry for all widgets in the odoo instance.

I could not test the option 'm2o_dialog'. I don't really understand what it means. By inspecting the code it seems that this option has no effect in the application.

Regards,
Jordi.

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -8,7 +8,34 @@ openerp.web_m2x_options = function (instance) {
         _t  = instance.web._t,
         _lt = instance.web._lt;
 
+    var OPTIONS = ['web_m2x_options.create',
+                   'web_m2x_options.create_edit',
+		           'web_m2x_options.limit',];
+
     instance.web.form.FieldMany2One.include({
+
+        start: function() {
+            this._super.apply(this, arguments);
+            return this.get_options();
+        },
+
+        get_options: function() {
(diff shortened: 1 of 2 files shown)


web_m2x_options/static/src/js/form.js
```diff
@@ -8,8 +8,35 @@ openerp.web_m2x_options = function (instance) {
         _t  = instance.web._t,
         _lt = instance.web._lt;
 
+    var OPTIONS = ['web_m2x_options.create',
+                   'web_m2x_options.create_edit',
+		           'web_m2x_options.limit',];
```
pedrobaeza: This isn't aligned.

pedrobaeza:
A little remark, but seems OK at first sight.

JordiBForgeFlow:
Done. Thanks for your review @pedrobaeza

pedrobaeza:
Tried on runbot (in _base_ database installing modules, because _all_ doesn't work) and the module works perfectly.

:+1: Thanks.

hbrunn:
:+1:

pedrobaeza:
I merge so that #120 can implement such functionality.

JordiBForgeFlow:
Thanks!!

PR #115, opened 2015-04-14
[MIG] migrate web_widget_digital_signature for 8.0

fva-idealis:


web_widget_digital_signature/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+##############################################################################
+
+{
+    "name": "Web Widget Digital Signature",
+    "version": "1.0",
+    "author": "Serpent Consulting Services Pvt. Ltd.,Taktik,"
+              "Odoo Community Association (OCA)",
+    "category": '',
+    'complexity': "easy",
```
hbrunn, line 29: this key is deprecated

web_widget_digital_signature/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+    'depends': ['base', 'web'],
+    "description": """This module provides the functionality to store digital
+        signature image for a record. The example can be seen into the
+        Partner's form view where we have added page 'Signature'.
+    """,
```
hbrunn, line 34: better put this into `README.rst`

web_widget_digital_signature/static/src/lib/js/excanvas.js
```diff
@@ -0,0 +1,924 @@
+// Copyright 2006 Google Inc.
```
hbrunn, line 1: libraries belong into `/static/lib`

web_widget_digital_signature/partner_view.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" ?>
+<openerp>
+    <data>
+
+        <template id="assets_backend" name="web_widget_color assets"
```
hbrunn, line 5: it's a bit misleading to have this in a file called `partner_view.xml`

web_widget_digital_signature/static/src/js/digital_sign.js
```diff
@@ -0,0 +1,131 @@
...
+            }else if(this.view.get("actual_mode") == 'edit' ){
+                this.$el.find('> img').remove();
+                this.$el.find('> canvas').remove();
+                if(! this.get('value')){
+                    this.$el.find('> img').remove();
+                    $(this.$el[0]).find(".signature").signature();
+                }else if(this.get('value')){
+                    this.$el.prepend($img);
+                }
+            }else if( this.view.get("actual_mode") == 'create'){
```
hbrunn, line 52: what's the difference between this and the `this.view.get("actual_mode") == 'edit'` branch above?

web_widget_digital_signature/static/src/js/digital_sign.js
```diff
@@ -0,0 +1,131 @@
+openerp.web_widget_digital_signature = function(instance) {
+    var _t = instance.web._t;
+    var QWeb = instance.web.qweb;
+    var images = {}
+    instance.web.form.widgets.add('signature', 'instance.web.form.FieldSignature');
```
hbrunn, line 5: all through this file, you sometime declare your variables and sometimes not. consider using `"use strict";`

web_widget_digital_signature/static/src/js/digital_sign.js
```diff
@@ -0,0 +1,131 @@
...
+                this.$el.find('> img').remove();
+                this.$el.find('> canvas').remove();
+                if(! this.get('value')){
+                    this.$el.find('> img').remove();
+                    $(this.$el[0]).find(".signature").signature();
+                }else if(this.get('value')){
+                    this.$el.prepend($img);
+                }
+            }
+            $(this.$el[0]).find('.clear_sign').click(function(){
```
hbrunn, line 63: `this.$el` already is a jquery object, and it's only one, so it's enough to say that

web_widget_digital_signature/static/src/js/digital_sign.js
```diff
@@ -0,0 +1,131 @@
...
+                $img.css("margin-top", "" + (self.options.size[1] - $img.height()) / 2 + "px");
+            });
+            $img.on('error', function() {
+                $img.attr('src', self.placeholder);
+                instance.webclient.notification.warn(_t("Image"), _t("Could not display the selected image."));
+            });
+        },
+    });
+    instance.web.FormView.include({
+        save: function(prepend_on_create) {
```
hbrunn, line 117: this seems quite wrong to me. You break modularity here. Do whatever you need to do and call super() then.

But I don't believe you have to do anything here, you can also do that in Field.get_value()

hbrunn:
Generally, I think adding the signature to the partner model shouldn't be part of this module, but another one depending on this

Cedric-Pigeon:
I agree with @hbrunn adding signature to partner model should not be done in a widget module.
@fv-taktik This is an interesting PR. Thanks. 
 In order to be merged, do you plan to fix it soon ?

dreispt:
I also agree this should be split in two: the widget _per se_ and it's usage on Partners.
I personally don't see the point of having this on Partners, but I'm interested in using it on Project Tasks.

I might even consider taking on from where @fv-taktik left off (if I can get some help with the JS parts).
It would be nice to also add TouchPunch to [enable it in touchscreens](http://keith-wood.name/signature.html).

nhomar:
Hello.

In my opinion this module be splitted can take a job and having it in partners meke it testeable by default and do not add any problem in my opinion.

I give a :+1: to this PR and may be split fo v9.0.

Opinions? it was sttoped and we should have it as it is already merged :-) and it has months sttoped BTW: then it can means that the author is not interested on make more improvements?

pedrobaeza:
@nhomar, we shouldn't merge things that conceptually don't fit correctly, because we have to deal with migration paths or solve these problems in the future.

dreispt:
I agree with @pedrobaeza.
AFAICT the split is fairly simple, if someone wants to take over the PR...

nhomar:
ok, then let's do the split first. I am not against of that position totally, let me put on the roadmap.

eLBati:
Hi, I would close this and continue on https://github.com/OCA/web/pull/267

eLBati:
https://github.com/OCA/web/pull/267 is merged.
This can be closed

Issue #116, opened 2015-04-15
web_easy_switch_company: menu items wrapping


thomaspaulb:
I've noticed an issue that when the "easy switch company" menu item is inserted into the top bar, the other menu items don't "contract" to make space for this new item, and thus the new item wraps to the next line. 

However, on screen resize the "menu_bound" event is executed and the size of the section with normal menu items is recalculated, some items are shifted to the "More" menu, and it all fits again.

I've found a workaround which is to add the following code to instance.web.UserMenu.init in switch_company.js,

```
        $(document).ready(function(ev) {
            openerp.web.bus.trigger('resize');
        });
```

It doesn't seem to be the cleanest solution, but at least it works. Anybody else has noticed the same?

legalsylvain:
Hi, thanks for reporting this bug. Can you indicate the version you use ? 7.0 / 8.0 ?
- i use this module in 7.0, and I don't have any problem, but I'll take a look. 
- I didn't test the  module ported in V8.0.

sbidoul:
Hi, we are using the module on 8.0 and I did not notice anything wrong. Can you post a small screenshot or video illustrating the issue?

pedrobaeza:
Hi, I confirm that this happens sometimes when you update browser page, switching for example the company, but I don't have an exact trace.

antiflu:
![screenshot from 2015-04-15 17 58 48](https://cloud.githubusercontent.com/assets/1466356/7163255/56a5515e-e399-11e4-8cd3-e92e5a112c67.png)

thomaspaulb:
![screenshot](https://cloud.githubusercontent.com/assets/1466356/7163338/b8f50b7e-e399-11e4-9668-0dc855ef8f23.png)

legalsylvain:
@antiflu : Could you realize a PR with your fix ?

thomaspaulb:
It feels like an ugly workaround but at least it works for my client. In my experience this only occurs in 8.0, in Chrome, and only when there are a lot of menu items obviously. Also, the problem disappears when you somehow resize the screen (eg by opening the Javascript console or moving the window). That's also the nature of this workaround it triggers a resize event after things are rendered

pedrobaeza:
Was this fixed in OCA?

pedrobaeza:
No news, so I close.

PR #118, opened 2015-04-16
Fix for #116 menu elements wrapping in v8.0

thomaspaulb:


legalsylvain:
@pedrobaeza, could you check if this PR fixes your bug ?
Regards.

pedrobaeza:
I tried on runbot, messing up menus to have a lot of upper entries, and it seems that the problem is gone, so :+1:

thomaspaulb:
Hmm I've been working with it now for several days but it seems that the document ready event still comes too early sometimes..... we need another Javascript event that really fires after the menu item is rendered

pedrobaeza:
Well, I haven't detected the problem in the tests I made. Do you have any method to get the problem?

thomaspaulb:
I normally get the problem in Chrome, whenever I rightclick a top menu item
such as Sales or Settings and open it in a new tab. When I refresh the
page, the problem used to still be always there, now, sometimes it is gone,
sometimes it is still there, IMO this depends on the timing: when docready
occurs "too early" and calls the resize event it will not solve the
problem. But after that when I call the Javascript function from the
console, or resize the page, it will solve the problem.

Possible your Odoo server/client computer is faster than mine so your
timing gets lucky, or the configuration of your menus is different and the
wrapping gets lucky, and so you cannot reproduce?

How I would like to proceed in this is to discuss with you guys about if
there are some Javascript events that I can hook in order to call the
resize event at the appropriate time.... have you guys some experience with
8's Javascript stuff? Otherwise I'll just have to dig in deeper myself...

yvaucher:
BTW this commit is authored by "root". You can fix it with a git commit --amend --author='name <email>'

thomaspaulb:
I think I changed the name now. I also pushed a fix with my lame setTimeout
solution as a workaround for the problems I mentioned before.

yvaucher:
@antiflu First commit is still wrong now that you added commit you need to rebase and amend
using `git rebase -i`

Then do a `git push -f`to apply your changes here

hbrunn:
closing due to inactivity

thomaspaulb:
Have rebased finally, so can we reopen this?

pedrobaeza:
The rebase is not correct. Closing again.

PR #120, opened 2015-04-17
Add two new functionalities to m2x dialogs

ecino:
Add two new options in module web_m2x_options :
- search_more : option to always show the "Search more..." in one2many dialogs, even if all search results are displayed.
- field_color : option for changing the color of listed records based on field's values

Other changes :
- Added option m2o_dialog in system parameters
- Added function to check either for "True" or "true" values

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -46,6 +46,18 @@ New option
 ``limit`` *int* (Default: openerp default value is ``7``)
 
   Number of displayed record in drop-down panel
+  
+``field_color`` *string*
+
+  A string to define the field used to define color.
+  This option has to be used with colors.
+  
+``colors`` *dictionary*
+
+  A dictionary to link field value with a HTML color.
+  This option has to be used with field_color.
+  
+
 
 
 ir.config_parameter options
@@ -79,7 +91,7 @@ Example
 Your XML form view definition could contain::
 
     ...
-    <field name="partner_id" options="{'limit': 10, 'create': false, 'create_edit': false}"/>
+    <field name="partner_id" options="{'limit': 10, 'create': false, 'create_edit': false, 'field_color':'state', 'colors':{'active':'green'}}"/>
     ...
 
 Note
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Interesting changes! Can you add also a system parameter to make "Search more..." default for all m2x fields?

Do you plan to migrate these changes to 8.0?

ecino:
Currently, branch 7.0 does not support system parameters so we didn't include this option. It could be added after #114 is merged.

It would be great to migrate for version 8.0 but it is not our priority at the moment.

pedrobaeza:
OK, I have merged #114, so you need to rebase and adjust conflicts (sorry for that), and you can implement the feature I mention. About 8.0 migration of new features, I have added an issue to track and remember this: https://github.com/OCA/web/issues/121.

ecino:
@pedrobaeza Added system parameter option.

I noticed that parameters format are not consistent between views and system parameters. In the xml, a boolean option is set with lowercase (true/false), and in system parameters it is set with a capital letter (True/False). I followed this convention but wouldn't it be better to standardize the format ?

pedrobaeza:
Yeah, you're right. Some standarization or cast to boolean would be desirable.

ecino:
I added a function to check for "True"/"true" values in javascript.
I also added the missing m2o_dialog option in system parameters.

pedrobaeza:
I'm afraid the rebase is not correct. You have changes in web_advanced_filters module an in .travis.yml.

ecino:
Ok, I corrected the pull request.

pedrobaeza:
It seems good now, but I was not able to try on runbot, so you have my :+1: (code review).

ecino:
Anyone else for a review ? We may soon start porting this for V8.

njeudy:
+1 I just test it on V7 :) thanks for improvement.

I will try to port this to v8  :)

njeudy:
:+1:

ecino:
We are currently porting it to v8. What is the status of this one, could we merge it ?

pedrobaeza:
Yeah, it can be merged with the 2 approvals and more than 5 days.

PR #126, opened 2015-04-24
[7.0] New module 'web_vendor_info'

sebalix:
This module replaces the _Powered by OpenERP_ at the bottom of the Web interface by the vendor's name, and optionally the release number.

These new configuration parameters are available:
- `web_vendor_name`
- `web_vendor_release`

### DIFF
--- a/web_vendor_info/README.rst
+++ b/web_vendor_info/README.rst
@@ -24,3 +24,18 @@ Contributors
 ------------
 
 * Sébastien Alix <sebastien.alix@osiell.com>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/web_vendor_info/README.rst
+++ b/web_vendor_info/README.rst
@@ -1,0 +1,26 @@
+
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Display the vendor's name and the current release number
+========================================================
+
+This module replaces the `Powered by OpenERP` at the bottom of the Web
+interface by the vendor's name, and optionally the release number.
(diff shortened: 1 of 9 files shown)


StefanRijnhart:
Thanks. The function fields on the company are misleading because the information is always the same. Compare our non-OCA module for this, where the config parameter is read out directly from the qweb-template: https://github.com/Therp/Therp-Addons/blob/8.0/support_branding/views/qweb.xml#L12. Would you consider dropping the new fields in favour of this construction?

Functionally, would you consider keeping Odoo and _add_ the vendor name instead of replacing it? There is a political dimension to this question ;-)

sebalix:
Hi,
I am not comfortable with these function fields too, thanks for the trick!
But your module seems quite complete (I was not aware of it, and it is already ready for 8.0), why not pushing yours instead? I could complete it with the possibility to set an optional project release number.

StefanRijnhart:
Well, we were still experimenting with it a little bit and never got round to pushing it to the OCA. By now, we'd be happy to push it if that would work for you. Oh, and I now see that the version I linked to is for 8.0, but we have a version for 7.0 as well.

OSguard:
@StefanRijnhart i saw in one of your version, that you have a email link for support. Greate idea.
The default values should be OCA ;)

sebalix:
@StefanRijnhart I tested your module on a 8.0, it works perflectly for us. Maybe 2 things I would prefer:
- Maybe put the "Supported by Y" on a new line?
- Set the config parameters with a noupdate
  What do you think?

If it is pushed on OCA, I could add a new - but optional - `support_branding.release` parameter.
Also, `OCA/server-tools` is maybe a better place for this module?

hbrunn:
@sebalix @OSguard @StefanRijnhart I pushed a cleaned up version of support_branding for 8.0 to #161 - this also contains the release.

What I also changed is that the email link sends a mail via odoo, and only resorts to the local mail client if this call fails for whatever reason.

sebalix:
@hbrunn Great work, thank you. I think I can close this PR, the #161 is the best candidate!

hbrunn:
@sebalix mind you that #161 is for 8.0, backporting my additions to the 7.0 version shouldn't be difficult, but is not on my agenda currently

sebalix:
@hbrunn no problem, I can take charge of the backport to the 7.0.

PR #128, opened 2015-04-28
[ADD] web_developer_mode_optional: Added module to allow access to devel...

alejandrosantana:
...oper mode, based on user groups.

Consider this as a substitute and improvement of PR https://github.com/OCA/web/pull/109

Now this allows to show "developer mode" link only to those users who are in a certain group.

~~**NOTE:**~~
- ~~To make the js work, I had to duplicate all code within `instance.web.Widget.extend({`, even though I only needed `on_menu_about`. But if I removed any of the rest, it didn't work as expected.~~
- ~~I have tried with _extend_ and _include_, with same results.~~

~~Have said that...~~
- ~~How do I properly extend a nameless function? (See chrome.js first line)~~

UPDATE: All strikethrough text above has been solved.


alejandrosantana:
OK, my problems with JS have been solved already. All spaguetti code have been removed.

anthony-muschang:
Tested :+1:

web_developer_mode_optional/static/description/icon.svg
```diff
@@ -0,0 +1,184 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
legalsylvain, line 1: Why did you add two version of the icon ? the svg file seems to me unnecessary ...

pedrobaeza, line 1: Yeah, it should be only needed if you make any modifications, so that you allow to get the editable one available.

alejandrosantana, line 1: @legalsylvain @pedrobaeza I just thought it could be useful for anyone that could modify the module to have the vectorial source of the image. But I can delete, if needed, of course. Should I take the "no .svg" as an OCA policy for any future module?

legalsylvain, line 1: I can not confirm that it is an OCA policy, but I can say that it's the first time I see a vectorial icon in OCA modules. (But I don't know all of them)
@jgrandguillaume, @gurneyalex : What'is your PoV about that point ?

My _personal_ PoV is : 
- change icon is not a real need ;
- add svg file make PR on github and git diff less readable ;

But it's not a blocking point for me.

pedrobaeza, line 1: You already have the source in https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg, that's why I ask not to duplicate the file unless you make any modification.

legalsylvain, line 1: @pedrobaeza It's not the same icon ! https://github.com/OCA/web/pull/128/files?short_path=e340032#diff-e34003280f7857fb9036ec1c6f49edc0

regards.

pedrobaeza, line 1: Ah, sorry, browser doesn't show me the preview when I review it and I think it was the template one. Then I totally agree with Alejandro that the SVG must be there as en editable source. It's the same as if you include the compiled python pyc, but not the py :wink:

alejandrosantana, line 1: @jgrandguillaume, @gurneyalex : Then, what do I proceed with OCA modules icons?
I think any module could have its own icon. In that case, for any future modification, it should include source (in this case an .svg file), even if it makes a commit less readable.
If no particular icon is used, then default to OCA .png-only icon, as it is on templates.

pedrobaeza, line 1: @alejandrosantana, you have made it right. Include the source SVG when the icon is custom

alejandrosantana:
@jgrandguillaume, @gurneyalex Anything on this? If needed, I can remove the svg and dicuss its presence, yet I believe icon .svg source should be there for others to modify in the future (as I give permission to that).

pedrobaeza:
Alejandro, there's no problem with the SVG. The problem is having the time to review...

alejandrosantana:
For some reason, OCA CLAbot is refusing my CLA, but I have signed, as I had contributed before.
Also: https://odoo-community.org/members?name=Santana

pedrobaeza:
Not exactly, Alejandro. It's talking about other comits in your PR. You have made the rebase in a strange way that have mixed older commits (you count 43 in this PR). Please fix this.

alejandrosantana:
OK, time to learn a bit more git.

alejandrosantana:
Just gone the easy way: make a clean fork, new branch and new clean PR in #248 
Sorry for inconveniences, but this way is way cleaner, I think.

PR #129, opened 2015-04-29
8 add checkbox in m2m tree view

bealdav:
WIP  
m2mcb

Issue:
In a many2many_tags  widget when a lot of entries should be selected it's fastidious to select 80% of attributes. Then you may click on 'search more', but impossible to select several attributes at once

Resolution:
this module add a checkbox to this list

by @sylvainc


sylvainc:
improvement of the "many2many_tags" widget

pedrobaeza:
I don't see the relation between module name/code and the description of the PR.

sylvainc:
Hello @pedrobaeza,

I have renamed the module 'web_m2m_tags_multiple_selection'.
This module allows the user to select multiple options through the search box in the case of the "many2many_tags" widget (this isn't possible with the native widget)

Best regards,

pedrobaeza:
OK, I see. Thanks for the explanation.

I prefer then the name _web_widget_many2many_tags_multi_selection_. It's longer, but it contains the current conventions for these modules (widget word, full name of the widget, and the function it does). In contrary, "multiple" can be short to "multi" without losing meaning.

dreispt:
@pedrobaeza I would put less wheels on the truck :smile: 
What about `web_widget_m2mtags_multi`

bealdav:
as you know @pedrobaeza, I prefer any name but without plural ;-)
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.rst#module

pedrobaeza:
Yeah, but in this case is not a question of a model name, but a widget name, that is _many2many_tags_, so you shouldn't change it.

bealdav:
Yes Pedro, I understand your logic, but in practice, to make understandable a rule to everybody (and avoid long debates), it could easier to have the less exceptions as possible. If the rule is complex (if model no 's', if widget 's', if toto no 's', etc.) we all loose a lot of time, for any rule.
Think about new odoo developers, there are so many rules in code, views, than easiest conventions will benefits to all oca finally
my 2 cents

pedrobaeza:
Well, I don't see this as a problem, because the name of the widget is as is, but I won't block if you put it in singular. It will feel strange anyways: it's the same as if you rename _res.groups_ model to _res.group_ in one part because it doesn't follow the conventions, but the rest of the parts still point to the plural form.

web_widget_many2many_tags_multi_selection/README.rst
```diff
@@ -0,0 +1,33 @@
+Allows multiple selection on many2many_tags widget
+==================================================
+
+This module allows the user to select multiple entries through the search box in the case of the "many2many_tags" widget.
```
guewen: The text you put on the PR (I changed it a bit):

```
In a many2many_tags widget when a lot of entries should be selected it's fastidious to select 80% of them. Then you may click on 'search more', but impossible to select several attributes at once.

This module adds a checkbox to this list so multiples entries can be selected at once.
```

would be useful in this file (more explicative than the current one I find).

web_widget_many2many_tags_multi_selection/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'web_widget_many2many_tags_multi_selection',
```
anthony-muschang: A non-technical name like "Tags multiple selection" would be better.

anthony-muschang:
Tested. Work as excepted :+1: 

But the module name in the module list should be improved.

legalsylvain:
Tested. :+1:.
Thanks.

sylvainc:
Thanks !

PR #130, opened 2015-05-01
add web_searchbar_full_width

hbrunn:
# Show searchbar over full width

Odoo's default search bar is rather narrow, which causes it to grow vertically when applying a lot of filters. For small screens this can be a problem as it wastes a lot of space. This addon addresses this issue by growing the search bar to the whole width of the screen, moving it below the breadcrumb. A side effect of this is that the breadcrump can also use the full screen width, which makes it reasonable not to cut off titles after 7 characters.

The result looks like this:

preview


web_searchbar_full_width/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+##############################################################################
+{
+    "name": "Full width searchbar",
+    "version": "1.0",
+    "author": "Therp BV",
+    "license": "AGPL-3",
+    "category": "Tools",
+    "summary": "Show search bar in full screen width",
+    "depends": [
+    ],
```
pedrobaeza, line 30: You have missed 'web' dependency.

pedrobaeza:
Another essential!

Please fix my comment.

hbrunn:
@pedrobaeza thanks! It's done now

pedrobaeza:
:+1:

lfreeke:
:+1: (test)

guewen:
:+1:

PR #131, opened 2015-05-01
[ADD] web_menu_autohide

hbrunn:
# Hide menus

This addon hides the top menu bar and the left menu bar. They become visible again when you move the mouse towards the border of the screen (the distance is configurable). Moving the mouse out of a menu causes it to disappear after a configurable delay.
# Configuration

To configure this module, you need to:
- go to Settings / Technical / Parameters / System Parameters
- adjust `web_hide_menu.show_bar_treshold` to set the distance in pixels from the border from which the menu appears
- adjust `web_hide_menu.hide_delay` to set the amount of milliseconds after which the menu disappears if the mouse left it
# Known issues / Roadmap
- This module heavily relies on mouse events, so it probably messes things up for touchscreens


web_hide_menu/README.rst
```diff
@@ -0,0 +1,45 @@
+Hide menus
+==========
+
+This addon hides the top menu bar and the left menu bar. They become visible again when you move the mouse towards the border of the screen (the distance is configurable). Moving the mouse out of a menu causes it to disappear after a configurable delay.
```
pedrobaeza: s/top menu bar and the left menu bar/top and left menu bars

web_hide_menu/README.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* go to Settings / Technical / Parameters / System Parameters
+* adjust `web_hide_menu.show_bar_treshold` to set the distance in pixels from the border from which the menu appears
```
pedrobaeza: Mention the default value of the parameter

web_hide_menu/README.rst
```diff
@@ -0,0 +1,45 @@
...
+* adjust `web_hide_menu.hide_delay` to set the amount of milliseconds after which the menu disappears if the mouse left it
```
pedrobaeza: Mention the default value of the parameter

pedrobaeza:
I think a better name would be _web_menu_autohide_.

pedrobaeza:
Please protect the parameter reading against the removal of them, defaulting to a value when not found.

pedrobaeza:
I have tried the module, and it behaves weird on the left menu bar: the menu is not hidden by default, and you only get it when clicking in one of the menus. Then, if you go again to the threshold area, the menu is displayed and doesn't dissapear when you go out of that area.

Another improvement would be to make a smooth fold and unfold animation to make less abrupt.

Anyways, very good module, as always.

hbrunn:
@pedrobaeza thanks for your points, I'll pick them up tomorrow. Also the left menu is supposed to disappear after the timeout, which is 10 seconds by default. Doesn't that work for you?

There are already default values for the parameters, if you delete them, he'll never reach https://github.com/OCA/web/pull/131/files#diff-85d5d7b487a6a94614afe61b7a7eafd1R45 and stick to the defaults in https://github.com/OCA/web/pull/131/files#diff-85d5d7b487a6a94614afe61b7a7eafd1R25

For the animation: This module was written for a customer using very weak workstations which connect to a nomachine (https://www.nomachine.com/) server. In this environment, animations always look ugly, so for them, animations are a no go. But if someone adds a parameter to turn on animations, I'd certainly not object

pedrobaeza:
Thanks for the explanations. I expected that the left menu behaves the same as the top one, hiding when you go out the threshold area. Anyways, 10 s seems a lot, doesn't it?

hbrunn:
@pedrobaeza now I picked up your suggestions. The delay is rather long, but after some playing with it I noticed that when actually working with odoo, you click a menu connected to an action anyway, and then the left bar hides instantly. The long delay then is helpful for not needing to hurry. Think about clicking the rightmost entry in the main menu. Then you move your mouse to the left to click what you're interested in, but if the delay is too short, the menu will already be gone before you reach it.

pedrobaeza:
:+1: Thanks

web_menu_autohide/static/src/js/web_hide_menu.js
```diff
@@ -0,0 +1,159 @@
...
+        hide_delay: 10000,
+        leftbar_hide_timeout_id: null,
+        main_menu_hide_timeout_id: null,
+        start: function()
+        {
+            var self = this;
+            return this._super.apply(this, arguments)
+            .then(function()
+            {
+                var addon_name = 'web_hide_menu',
```
anthony-muschang: "web_hide_menu" should be renamed to "web_menu_autohide"

web_menu_autohide/views/templates.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <template id="assets_backend" name="web_hide_menu assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/web_hide_menu/static/src/js/web_hide_menu.js"></script>
```
anthony-muschang: Samething, rename "web_hide_menu" to "web_menu_autohide"

hbrunn:
@anthony-muschang thanks, fixed now

web_menu_autohide/static/src/js/web_menu_autohide.js
```diff
@@ -0,0 +1,159 @@
...
+//   but WITHOUT ANY WARRANTY; without even the implied warranty of
+//   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//   GNU Affero General Public License for more details.
+//
+//   You should have received a copy of the GNU Affero General Public License
+//   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+//############################################################################
+
+openerp.web_hide_menu = function(instance)
```
guewen: s/web_hide_menu/web_menu_autohide/

otherwise this code is not loaded

hbrunn:
thanks @guewen In case I never mentioned that: I hate renaming modules

guewen:
> thanks @guewen In case I never mentioned that: I hate renaming modules

We are 2.

Thanks!

:+1:

anthony-muschang:
Work well :+1:

guewen:
When a left menu is too long for the available height, it do no longer show a scrollbar, so the bottom items cannot be accessed.

anthony-muschang:
@guewen I do not have this problem, the scrollbar is correctly displayed (on chrome, linux). Maybe a problem with the browser cache?

hbrunn:
@guewen adjustments for this happen in css. Did you update your database after the module rename?

guewen:
@hbrunn I installed it after the rename

I'm using Firefox 37.0.1. 
Here is an example:
2015-05-04-134631_1673x916_scrot

hbrunn:
@guewen how about now?

guewen:
> how about now?

it works nice now!

legalsylvain:
Hi, 
I installed this module and tested it. My PoV is that is not very user friendly to have menus that hides content of the page. Did you imagine to adjust the right content, reducing the size, when the menu is displayed ?

Regards.

hbrunn:
@legalsylvain that was my first version, but then the content window flickers a lot due to the necessary relayout. You'll see that if you remove the css code, then you have everything jumping every time the menu opens.

pedrobaeza:
Why not make it as a rollup menu like old 6.1, Odoo material design concept or bootstrap website menu?

pedrobaeza:
It seems quite better rjght now.

Thanks. :+1:

OSguard:
i think this is a little confusing. Two points:
- going to the left top corner the left panel overlaps the top panels and this looks broken
- maybe we can add a menu icon left to caption, to open both with touch screen

what do you think?

nbessi:
@hbrunn @pedrobaeza I want to publish a small JS development that add a button to manually toggle left menu https://gist.github.com/nbessi/00dbb32768f9bc9e677d#file-toggle_menu-js. But before making a module and a PR I wanted to know if both addons can be in the repo or if they are going to be considered as too redundant?

pedrobaeza:
I think both can be hosted, as they are different behaviours that aren't exclusive. Ideal option would be to integrate both in a module web_hide_menu, and via a system parameter, choose which method for hiding the menu, but maybe this is too much work.

hbrunn:
agreed with Pedro. Please don't hesitate to make a PR!

PR #133, opened 2015-05-06
Add Web Sticky Header module

nbessi:


web_sticky_header/__openerp__.py
```diff
@@ -0,0 +1,27 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{"name": "Sticky list headers",
+ "category": "Hidden",
+ "author": "Camptocamp, Odoo Community Association (OCA)",
+ "version": "0.1",
+ "depends": ['web', 'base_import'],
```
legalsylvain, line 25: Why is 'base_import' necessary ?

legalsylvain:
No code review. Tested. Some bugs found.
- In a form view, scroll : a useless bar is displayed;
  (user form)
  sticky-bug-form
  (partner form)
  bug-form-partner
- In the tree view I have some troubles if I try to scroll using the mouse and clicking on the vertical bar, do you have the same problem ? (only on firefox)

(Ubuntu 14.04 / Firefox v37 + Ubuntu 14.04 / chromium v41)

legalsylvain:
Otherwise, nice and really usefull feature. thanks.

nbessi:
Thanks for the tests :). For the base_import I'm not sure, I have to test. It seems I forgot to push a commit where I unset the event listener on the do_hide. I'll do it on Friday

sbidoul:
@nbessi do you know if it would be hard to make the footer (with totals) sticky too?

legalsylvain:
@sbidoul : :100:

alejandrosantana:
I have tested it and found it worked quite weel, but I found little things:
1. When you are at the top, the headres are one style, and once you scroll down style turns different, like flattened.
1. When you are at the top, scrollbar includes the headers too, but if you scroll down, the scrollbar excludes the headers. And there is a jumpy behaviour between these two states.
1. If you drag the scrollbar just a little bit, a weird thing happens, as there appears a "phantom row" at the top, just between the headers and the first row. I attach an image on this.
sticky_list_header_

Kubuntu 14.04LTS + Chromium (41.0.2272.76 Ubuntu 14.04 (64-bit))

nbessi:
@alejandrosantana 
For the style it is on purpose, it is to show that the head became passive. 

The white bar and the jump effect are known issues that can not be fixed with actual webclient state.
When scrolling I hide the original header. But I can not use display None else it will breaks columns width So I use visibility hidden. I try to add col groups, fixed head, etc but nothing works without altering editable lists.  

This module is mostly a hack as the webclient structure is quite fragile in is conception.

nbessi:
@sbidoul sticky footer may be possible. If we are lucky we can put them in a fixed position and it should works if not we have to use the same trick we used with the headers

nbessi:
@legalsylvain Fixed pushed it seems behaviour only happens on addons list.

pedrobaeza:
Trying the module at runbot, I get a JS error: `TypeError: openerp.init is not a constructor` that prevents to use it.

nbessi:
@pedrobaeza I missed my rebased fixed

pedrobaeza:
I have tried you module in runbot (thanks @gurneyalex for including the force rebuild button), and I found following issue:
- one2many listviews are not displayed.
- When you switch from form view to list view, sticky header is lost.
- I think there's a possible memory glitch, because my memory consumption has arisen a lot making the tests.
- As others have mentioned, if would be great to have also sticky footer when you have totals.
- A better name would be _web_listview_sticky_header_.

nbessi:
@pedrobaeza thanks for your review. 
I was able to address all reported issues but it means a little change in visual design.

For the footer they will came in a next release.

For the name I was thinking to also alter form view buttons and workflow status in a future release, but maybe we should make two addons

nbessi:
wow it seems I git shot myself I will redo the PR

nbessi:
Rebased and repushed

pedrobaeza:
@nbessi, thanks for fixing the issues. I have tried again and now I only see one remaining thing, and it's that now a grey row that corresponds to the sticky header stays when you're at the top of the list and the sticky header is not displayed. Can you solve this? It also seems than the new sticky header height is larger. Is that intended?

nbessi:
@pedrobaeza that grey row is here on purpose. On large list it avoids the table div to move and be recomputed that make a nasty lag effect and impact negatively usability. If you have an other idea it is welcome because I'm out of inspiration here.

pedrobaeza:
Well, I'm not CSS/HTML expert, so I can't help, but if you see no other solution, please make it the minimal, because now it seems a little larger than needed (it's higher than the normal header).

nbessi:
It is 10 pixel higher that the original header this is the minimum value to prevent div trembling.

pedrobaeza:
OK, as I say, I'm not an expert in this area, so I don't oppose more, but as a module that impacts all the views, I'm not satisfied enough with the result, so I'm not approving it either.

nbessi:
Mabe we can add an option to decide to auto hide show the grey block or show. So the end user can choose knowing the pro and con of each usecase. 

Honestly here I'm stuck with the Odoo webclient performance on large list and it is a lot worse if the list is editable. 

At least if it is merged some one with stronger JS/CSS knowledge can propose an improvement.

pedrobaeza:
Well, I would like to see the other option working to see if it's tolerable.

P.S.: I have to improve also my JS skills...

nbessi:
I mosty hope this will be a native function in next webclient. If we agreed on the concept I'll try to push it on the master core.

pedrobaeza:
I hope so.

gurneyalex:
CLA ok.

gurneyalex:
CLA ok

mdietrichc2c:
Code review: :+1: LGTM

web_sticky_header/static/src/js/custom_list.js
```diff
@@ -0,0 +1,149 @@
...
+    self._super(parent, dataset, view_id, options);
+    this.scroll_start = false;
+  },
+
+  do_show: function(){
+    /**
+     * Move search div to a sticky location.
+     * Add scroll event listener and remove leftover headers
+     **/
+     var self = this;
```
OSguard: wrong line ident

web_sticky_header/README.rst
```diff
@@ -0,0 +1,45 @@
...
+============
+
+Follow standard addons installation steps
+
+Configuration
+=============
+
+No configuration is needed
+
+* ...
```
OSguard: remove empty list

OSguard:
Please hide the checkbox as it has no function.

web_sticky_header/__openerp__.py
nbessi, line 25: yes as we want the layout to be correctly initialized

angelmoya:
Could you check to use this javascript lib

http://mkoryak.github.io/floatThead/

I tested this code in javascript console and works fine.

I'm working in this PR #247 , but I'm not javascript developer, and I don't know how to call it.

pedrobaeza:
Closing as the user is no more active and there's more or less a replacement in #247

PR #134, opened 2015-05-06
[ADD] web_advanced_search_x2x

hbrunn:
# Search for x2x records in advanced search

Standard behavior in advanced search for one2many, many2many and many2one fields is to do a `name_search`. This often is not satisfactionary as you might want to search for other properties. There might also be cases where you don't exactly know what you're searching for, then a list of possible options is necessary too. This module enables you to have a full search view to select the record in question, and either select specific records or select them using a search query of its own.
# Usage

- open the advanced search options in a search view
- select a one2many, many2many or many2one field
- select operator `is equal to` or `is not equal to`
- the textfield changes to a many2one selection field where you can search for the record in question

To search for properties of linked records (ie invoices for customers with a credit limit higher than X):
- open the advanced search options in a search view
- select a one2many, many2many or many2one field
- select operator `is in selection`
- in the search view that pops up, select the criteria
- click `Use criteria`
- if you're only interested in certain records, mark them en click `Select`
- if you want to change your selection afterwards, click the search symbol right of the selection term

In both cases, don't forget to click `Apply` to actually execute the search.

Note that you can stack searching for properties: Simply add another advanced search in the selection search window. You can do this indefinetely, so it is possible to search for moves belonging to a journal which has a user who is member of a certain group etc.


legalsylvain:
Very usefull feature. Tested. Works like a charm as always.
Thanks a lot. :+1:

pedrobaeza:
Isn't better to add a new option in the list "With the field..." or something similar instead of using "is equal to"?

hbrunn:
@pedrobaeza I'd like to avoid that. First, I wouldn't know how to explain the difference to a nontechnical user and second, I can't really think of a use case where you need both. The only thing I can imagine is a `name_search` that returns multiple records with the `=` operator, but save for partners, I don't think that happens a lot. And even there, if you want to select you and your twin, you'd simple search two times for 'Pedro Baeza' and select a different one each time. For this case, the `use condition` operator from the roadmap would do what you want and a lot more.

anthony-muschang:
Tested :+1:

pedrobaeza:
I have tested now the module, and I misinterpreted the use of it: I understood that you can filter by any condition on child fields of the many2one. For example: filter all the sales orders whose partner is from Spain. That would be very useful.

The module as is, it's working, but I don't see it very useful, because at the end, you can only put one record as criteria. 

Do you think it's possible to do what I'm asking for?

hbrunn:
@pedrobaeza that's what I write about in the roadmap, I think it's only a couple of hours of work.

As it is, it's useful for our customer because (a) they don't always know what the possible values of an x2x field are, so the textbox doesn't help them, (b) some models have a very broken `name_search`, (c) they sometimes know roughly what they're searching for, but need the model's search view to actually find this record and (d) often mistype a name.

You can fill in multiple records by simply adding another condition on the same field (see last bullet point in _Usage_).

I agree that what you describe and what I put in the roadmap is a way cooler feature (because you can do this recursively, you should be able then to search for sale orders whose partner has an analytic account which belongs to a certain journal etc.pp.), but this can be added afterwards I think as soon as somebody provides funding

nhomar:
+1 like a charm.

pedrobaeza:
@hbrunn, please contact me for talking about funding the feature I comment.

LeartS:
I actually needed this feature as it is now multiple times, and I ended up hacking it by overriding `fields_get` and the advanced search widget so that many2one fields were actually passed as selection fields to the advanced search.
This is definitely better. Good job!

hbrunn:
@pedrobaeza I pushed some commits you might be interested in :-)

pedrobaeza:
Hi, Holger, thank you very much for looking into this. It's a huge step in the usability and features of the system.

I have tried and see the concept used, but I can't get it working.

In the selection popup, I get when I enter a criteria:

```
ReferenceError: _t is not defined
```

When I apply criteria, I get this error:

```
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/models.py", line 5129, in search_read
    record_ids = self.search(cr, uid, domain or [], offset=offset, limit=limit, order=order, context=context)
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/api.py", line 241, in wrapper
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/models.py", line 1642, in search
    return self._search(cr, user, args, offset=offset, limit=limit, order=order, context=context, count=count)
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/api.py", line 241, in wrapper
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/models.py", line 4638, in _search
    query = self._where_calc(cr, user, args, context=context)
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/api.py", line 241, in wrapper
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/models.py", line 4461, in _where_calc
    e = expression.expression(cr, user, domain, self, context)
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/osv/expression.py", line 643, in __init__
    self.expression = distribute_not(normalize_domain(exp))
  File "/Datos/Installs/ERP/odoo/8.0/OCB/openerp/osv/expression.py", line 200, in normalize_domain
    expected += op_arity.get(token, 0) - 1
TypeError: unhashable type: 'dict'
```

Can you please check it?

Anyways, gorgeous work!

hbrunn:
@pedrobaeza thanks for testing, could you test after this commit (don't forget to refresh for the client side change)? The python error seems to be a followup error from the js part

pedrobaeza:
Thanks for the fix. Now the search is working properly, but there's some visual glitches with the overlap of the search box, the close button and the advanced search controls. Can you please check it and the module will be perfect?

hbrunn:
I fear what you describe is a general problem with search popups, and if it's what I mean, then we'd better fix it globally. could you post a screenshot? and compare that with the popup you get with any x2x form field?

pedrobaeza:
@hbrunn, it seems to be something with Firefox, but not with your module, but a general issue (reproducible on runbot). For your curiosity:

seleccion_001

I'm going to put an issue.

pedrobaeza:
For this PR, :+1: 

Many thank you

gfcapalbo:
:+1:

rafaelbn:
:+1:

PR #144, opened 2015-06-05
[ADD] add module 'web_warning_on_save'

damdam-s:
allow to display a message concerning the current record (create or write) after it has been saved.


web_warning_on_save/README.rst
```diff
@@ -0,0 +1,86 @@
...
+    :alt: License: AGPL-3
+
+Module name
+===========
+
+This module was written to extend the functionality of saving a record in the web interface.
+/!\/!\/!\ In no way this module stops the save of the record. You must consider this as a warning displayed 
+to the user AFTER save completed. /!\/!\/!\
+
+Installation
```
pedrobaeza: Remove sections that doesn't say anything.

web_warning_on_save/README.rst
```diff
@@ -0,0 +1,86 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Module name
+===========
+
+This module was written to extend the functionality of saving a record in the web interface.
+/!\/!\/!\ In no way this module stops the save of the record. You must consider this as a warning displayed 
+to the user AFTER save completed. /!\/!\/!\
```
pedrobaeza: Mention that for that purpose, you already have the constraints.

web_warning_on_save/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import controller
```
pedrobaeza: Use relative import: "from . import ...."

web_warning_on_save/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Web warning on save",
+    "version": "1.0",
+    "depends": ['web'],
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
```
pedrobaeza, line 26: In v7, you also need to put the description key, not only the README.

web_warning_on_save/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+        'static/src/js/web_warning_on_save.js',
+    ],
+    'css': [
+    ],
+    'qweb': [
+    ],
+    'installable': True,
+    'auto_install': False,
+}
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
pedrobaeza: Remove these vim specific lines.

web_warning_on_save/README.rst
damdam-s: Hi, sorry but i don't understand what you mean there.

pedrobaeza: I mean that if you want to prevent the save, you can use the constraints.

damdam-s: Ah ok.
But i don't want to prevent the save.
I want the save to complete and eventually display a simple text message to the user.
The same as a warning but when i press "Save" button.

pedrobaeza: OK, that's the main reason of the module indeed. I only ask you to mention what I said.

damdam-s:
Corrections according to  pedrobaeza

web_warning_on_save/static/src/js/web_warning_on_save.js
```diff
@@ -0,0 +1,23 @@
...
+				.done(function(res) {
+					if (res != false){
+						alert(res);
+					}
+				});
+			})
+	    },
+	});
+
+};
```
guewen, line 28: The indentation usually used in javascript in Odoo is using spaces.

web_warning_on_save/static/src/js/web_warning_on_save.js
```diff
@@ -0,0 +1,23 @@
...
+			var id = 0;
+			var self = this;
+			self._super.apply(this, arguments).done(function(id){
+				var model = self.model;
+				var param = {'model': model, 
+						'id': self.datarecord.id ? self.datarecord.id : id};
+				self.rpc('/web/warning_on_save/check_warning_on_save', param)
+				.done(function(res) {
+					if (res != false){
+						alert(res);
```
hbrunn: I think you'd better use `openerp.web.Dialog` instead of `alert`

web_warning_on_save/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+web_warning_on_save
```
pedrobaeza: This must be the module title, but in a legible way. For example: `Raise warnings when saving`. The underlining must be of the same size of the title.

web_warning_on_save/static/src/js/web_warning_on_save.js
```diff
@@ -0,0 +1,23 @@
+openerp.web_warning_on_save = function (instance) {
+
+	instance.web.FormView.include({
+		on_button_save: function(e) {
+//			inherit method to be able to display an alert if
+//			method called returns a string
+			var id = 0;
+			var self = this;
+			self._super.apply(this, arguments).done(function(id){
```
hbrunn: return that, otherwise you lose the deferred object for other functions inheriting `on_button_save`

web_warning_on_save/static/src/js/web_warning_on_save.js
```diff
@@ -0,0 +1,23 @@
+openerp.web_warning_on_save = function (instance) {
+
+	instance.web.FormView.include({
+		on_button_save: function(e) {
+//			inherit method to be able to display an alert if
+//			method called returns a string
+			var id = 0;
```
guewen: Why this line?

web_warning_on_save/controller.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @openerp.addons.web.http.jsonrequest
+    def check_warning_on_save(self, req, model, id):
+        """
+         try to call a method on the model given in parameter
+         if method does not exist in the model, do nothing
+        """
+        m = req.session.model(model)
+        try:
+            return getattr(m, 'check_warning_on_save')(id, req.context)
```
hbrunn: this swallows all exceptions. Why not `hasattr`?

web_warning_on_save/controller.py
```diff
@@ -0,0 +1,39 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import openerp
+
+
+class WarningOnSaveController(openerp.addons.web.http.Controller):
+    _cp_path = "/web/warning_on_save"
```
guewen: the controller should use the namespace of your module, so `/web_warning_on_save/warning_on_save` or better: `/web_warning_on_save`

gurneyalex:
CLA ok

./web_warning_on_save/controller.py:28:1: E303 too many blank lines (3)

./web_warning_on_save/controller.py:45:71: W291 trailing whitespace

./web_warning_on_save/controller.py:46:1: W391 blank line at end of file

mdietrichc2c:
FYI: displaying the module in the Web interface raises an error:

`SystemMessage: <string>:9: (SEVERE/4) Unexpected section title.

Usage
=====`

pedrobaeza:
Something similar happened to @alejandrosantana in another module. Do you know what's the problem, Alejandro?

web_warning_on_save/README.rst
```diff
@@ -0,0 +1,70 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+* write a method called 'check_warning_on_save' which will make some checks and return a string
+
+example :
+
+	def check_warning_on_save(self, cr, uid, id, context=None):
```
yvaucher: You have mixed tab and spaces thus the rst is not rendered correctly

web_warning_on_save/README.rst
```diff
@@ -0,0 +1,70 @@
...
+If you don't want OpenERP to save the record, you should use constraints.
+
+Usage
+=====
+
+To use this module, you need to:
+
+* write a method called 'check_warning_on_save' which will make some checks and return a string
+
+example :
```
yvaucher, line 25: You might need to write

`example::`

In order to tell following text is code

or maybe even

```
example :

.. code:: python
```

However, I'm not sure it will display well on Odoo

alejandrosantana, line 25: In my case, I believe I had some non-shown character which was causing all the trouble.
I solved it removing the "=" underline in offending title, with previous and following line-feed, and type it again.

yvaucher, line 25: Tested on http://rst.ninjs.org/ In this case it is failing on the """ trying to use it to underline a title thus it fails.

yvaucher:
This module seems to generate errors when `check_warning_on_save` is not present on model:

```
2015-06-23 13:02:15,193 23778 ERROR openerp.netsvc: 'payment.order' object has no attribute 'check_warning_on_save'
```

web_warning_on_save/controller.py
```diff
@@ -0,0 +1,46 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import xmlrpclib
+
+import openerp
+
+
+
```
gurneyalex: ./web_warning_on_save/controller.py:28:1: E303 too many blank lines (3)

web_warning_on_save/controller.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        m = req.session.model(model)
+        try:
+            return getattr(m, 'check_warning_on_save')(id, req.context)
+
+        except xmlrpclib.Fault as e:
+            if 'AttributeError' in e.faultString:
+                return False
+            else:
+                raise openerp.osv.osv.except_osv('Error', e.faultCode)
```
gurneyalex: ./web_warning_on_save/controller.py:45:71: W291 trailing whitespace

web_warning_on_save/controller.py
```diff
@@ -0,0 +1,46 @@
...
+        m = req.session.model(model)
+        try:
+            return getattr(m, 'check_warning_on_save')(id, req.context)
+
+        except xmlrpclib.Fault as e:
+            if 'AttributeError' in e.faultString:
+                return False
+            else:
+                raise openerp.osv.osv.except_osv('Error', e.faultCode) 
+
```
gurneyalex: ./web_warning_on_save/controller.py:46:1: W391 blank line at end of file

web_warning_on_save/controller.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
guewen: By convention should be in a `controllers` package.

web_warning_on_save/controller.py
guewen: hasattr is not possible because the model here is not aware of the methods of the model because it passes over RPC (https://github.com/odoo/odoo/blob/7.0/addons/web/session.py#L40). `getattr` always returns something so it is equivalent to `m.check_warning_on_save(id, req.context)` which should be preferred so.

Remains one big issue: the server issues a traceback with `AttributeError: 'any.model' object has no attribute 'check_warning_on_save'` every time we save a record, and we must find a way to avoid that.

The only poor way I found is by doing something such as:

``` python
from openerp.osv import orm


def check_warning_on_save(self, cr, uid, id, context=None):
    return False

orm.BaseModel.check_warning_on_save = check_warning_on_save
```

gurneyalex:
:+1:

guewen:
:-1: please wait, still an issue I think.

guewen:
This point needs to be fixed: https://github.com/OCA/web/pull/144#discussion_r33660907

damdam-s:
@guewen fix done

guewen:
Thanks, a last thing, could you replace:

``` python
return getattr(m, 'check_warning_on_save')(id, req.context)
```

by

``` python
return m.check_warning_on_save(id, req.context)
```

Which is equivalent but simpler. If you are interested to know why it is equivalent, you can glance over https://github.com/odoo/odoo/blob/7.0/addons/web/session.py#L40-L51

damdam-s:
@guewen change call method done

guewen:
Thanks!
:+1:

guewen:
(waiting for the builds)

mdietrichc2c:
Pending green Travis, :+1:

PR #149, opened 2015-06-17
[ADD] help_contextual_popup

bealdav:


help_contextual_popup/README.rst
```diff
@@ -0,0 +1,35 @@
+Add contextual help popup
+=========================
+
+This module add a contextual html help popup on each model action.
```
pedrobaeza: s/add/adds

help_contextual_popup/README.rst
```diff
@@ -0,0 +1,35 @@
...
+
+Installation
+============
+
+It was tested on Odoo 8.0 branch.
+
+Credits
+=======
+
+Akretion
```
pedrobaeza: Remove the company from here (it is intended for individuals). You already put the company of authors list.

pedrobaeza:
Little PEP8: `./help_contextual_popup/model.py:31:1: W391 blank line at end of file`

The rest is working as expected

help_contextual_popup/README.rst
```diff
@@ -0,0 +1,35 @@
+Add contextual help popup
```
lmignon: PLZ use the latest MT template as base for your readme file. The licence badge is missing as well as the bug tracker section. (https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst)

bealdav:
Fixed, thanks for a final review

hbrunn:
:+1:

dreispt:
Very interesting module. I just tested it and have a few functional-wise concerns:
- It is now clear for me what is "owner help" and "custom help". Do you mean something like "End User Help" and "Configuration Help"?
- I personally dislike the help button. It would be nice to use the same as `help_online`

help_popup/model.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class IrActionsActwindow(models.Model):
+    _inherit = 'ir.actions.act_window'
+
+    owner_help = fields.Html(
+        string="Owner Help",
+        help="Use this field to add custom content for documentation purpose\n"
+             "mainly for the ERP owner")
+    custom_help = fields.Text(
```
dreispt: Shouldn't this be also an Html field?

help_popup/report/help.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<openerp>
+<data>
```
dreispt: `noupdate="1"`would be nice, so that we can safely customize it.

bealdav:
@dreispt, thanks a lot for your review.
owner_help is for erp proprietary : data should be insert in the UI => Html field.
custom_help is for the integrators (developers which make reusable feature): data should be insert by xml data and not customized by the client). Maybe the field name should be other ?

bealdav:
Ok the button is really poor, but some PR can improve it, isn't it ? 
help_online button is rendering by css, there is no img. Rendering with a really clever tips which override my css skill and required time on this project. 
Use the same button will prevent to have both modules.
I'm open to any PR to improve UI

dreispt:
> owner_help is for erp proprietary : data should be insert in the UI => Html field.
> custom_help is for the integrators (developers which make reusable feature): data should be insert by xml data and not customized by the client). Maybe the field name should be other ?

This looks like it's seen from the implementation project manager perspective, not the maintenance / daily usage perspective. The two roles I see here are regular End Users and Configuration / Advanced users. 

Technical documentation for integrators / developers does not need to be here. I'm afraid it will be confusing to end users, and devs can look up those docs in the modules files.

bealdav:
@dreispt the purpose of this module is to provide:
1/ a way for developers to document their development
2/ a way for power users to complete documentation of their instance according to internal users profiles.

I don't understand why you say ?

> Technical documentation for integrators / developers does not need to be here.

dreispt:
This is available to all users right? As an end user I expect to see instruction to help me on daily use. I am afraid that having implementation technical details can confuse them.

I can see technical documentation on the GUI, but it should be available only for some users. Otherwise it's best to net be there at all.

bealdav:
Ok it's a misunderstanding. 
As I see things, developers can write functional content for end users (and should every time) and put this content in their modules.

All these help fields must only keep functionnal content. 
As you said, implementation detail is not managed here

dreispt:
OK. 
I still still think that "Owner Help" and "Custom Help" are not meaningful titles.
You dislike my previous suggestion - "End User Help" and "Configuration / Advanced Help"?

bealdav:
Thanks @dreispt, field names are better now.

Maybe +1 from you ?

dreispt:
@bealdav Of course :+1:

PR #150, opened 2015-06-18
[ADD] New module : Web custom search quantity

legalsylvain:
Give the possibility to select manually a custom value for search quantity, (in tree views), other than the default values. (80 / 200 / 500 / 2000 / Unlimited).
If accepted, I'll port this module in V8.

Kind regards.


web_custom_search_quantity/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
bealdav: Please save the cloud planet of all chars of this file ;-)
No code => no licence required, no coding required (it's only my opinion)

bealdav:
:+1: 
LGTM for me. Nice job as usual

web_custom_search_quantity/__init__.py
legalsylvain: @bealdav Thanks for thinking to the (cloud) planet !
Changes done.

OSguard:
can not login to runbot in this PR.
7.0 is working

legalsylvain:
Yes. In fact, I can not connect to -all, but to -base yes, (and install after my module, it worked). I don't know why.
@OSguard : Do you think I have something to fix ?

Thanks

OSguard:
Can you check if you need to rebase? As 7.0 working.

legalsylvain:
Rebased. It works now on runbot.
Thanks a lot @OSguard.

Regards.

OSguard:
i test it on runbot:
- chrome 43.0.2357.125 i can change the number by typing, but the drop down is not working (show up and remove imediadly, like a blink)
- firefox, change by typing working, dropdown not working correctly

legalsylvain:
Hi @OSguard, thanks a lot for your review.

In fact, it works correctly, regarding the new component. (_input with datalist option_), but maybe you don't like the behaviour of this component.
- With an _input with datalist option_, the available values displayed are filtered depending of the value you began to tip; (like in my image I set, if you tip "2", only "200" and "2000" are available, not "80");
- If you want to have the classical dropdown behaviour, you have to remove first the value set in the input ;
- In Chrom(e/ium), if you have only one value available, it will be hidden, and I agree with you, the blink effect is not very good, but for me it's Crome bug, regarding the component, not a bug of my module);

@OSguard : Could you try the _input with datalist option_ available here : http://www.alsacreations.com/xmedia/tuto/html5/datalist/demo1.html
You'll see this web demo page works as my module. (and with the same bad effect for Chrom(e/ium))

What do you think ?

OSguard:
I try the demo and not exactly the same. The difference is that on odoo it is create on click and than you have this effect. In the demo this not that much reproduced.

For my feeling, this is good idea but a confusing behavior for the end-user and it matches not my personal quality on UX.

I have a look to your JS Code. maybe try to remove the "focus()" then we need to click twice and it will be better to use the drop down.

OSguard:
I try it via browser debugging and i think it is better.

legalsylvain:
I'm not sure to have understood your PoV. The difference between my module and the demo is that there is no default value in the demo (so all options are available), in Odoo, there is the current value (so other option are not available).
I added focus to avoid to loose 1 click for end user, because if user want to change the value, he has to click on the input. (extra click for me)
Of course my module change a little the UI, but when we understand the new behaviour, for me (and the customer), it's ok.
If the problem for you, is that options are not available and the blink effect, maybe I can better try to remove the default value, so all option will be available. 

What do you think ?

Kind regards.

OSguard:
just remove the focus() and try out, then blink effect is gone. I think this is a solution fine for me.

Kind regards

legalsylvain:
- I tried, removing focus() call, and the blink effect is still present on Chromium;
- I removed the default value, (so the field is now empty), and I just added the current value as placeholder value. So the user know what is the current value, and can select between all the options / or set a custom value.

What do you think about that new behaviour ?

Kind regards.

pedrobaeza:
I have tried the module and didn't find the blink artifacts in Firefox and generally works good, but only two remarks:
- When pressing Esc key, you would return to the previous state before editing the quantity.
- The module name shouldn't include the `search` word, because it confuses about its use. I would suggest _web_listview_custom_element_number_

legalsylvain:
Hi @pedrobaeza Thanks for your review ! 
The blink effect occurs only on chrome, before my last change;
- about Esc key, It's not the behaviour of Odoo, by default. I don't know exactly how to do that, without calling reload_content(). If you have any idea... Otherwise, it could be done after, in an other [IMP] PR;
- about renaming : done;

pedrobaeza:
I don't know JS enough for advising you, so for now it's OK for me. Put it in the known issues /Roadmap section as a remainder for anyone that want to improve it.

:+1:

legalsylvain:
@OSguard, did you check again, after my last commit ?
Kind regards.

OSguard:
:+1:

PR #152, opened 2015-06-24
[FIX] Allow set web_tree_many2one_clickable.default as True

antespi:
This is a minimal change compatible with current installations.

With this change we are compliance with other addons options. It's confusing for user configuring this option as `True` but applies as False

### DIFF
--- a/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
+++ b/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
@@ -48,7 +48,7 @@ openerp.web_tree_many2one_clickable = function(instance, local)
                     .first()
                     .then(function(param) {
                         if (param) {
-                            self.use_many2one_clickable = (param.value == 'true');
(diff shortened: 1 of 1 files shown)


rafaelbn:
Like other standard config parameter like:

auth_signup.reset_password - True

:+1:

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -48,7 +48,7 @@ openerp.web_tree_many2one_clickable = function(instance, local)
                     .first()
                     .then(function(param) {
                         if (param) {
-                            self.use_many2one_clickable = (param.value == 'true');
+                            self.use_many2one_clickable = ((param.value == 'True') or (param.value == 'true'));
```
pedrobaeza: Better to lower the string to also allow `TRue`, `TruE`, and so on

antespi:
@pedrobaeza changes done following your comments, thanks.

pedrobaeza:
:+1: 

Another good addition to have is an XML with the system parameter already set

PR #153, opened 2015-06-24
7.0 web percent

StefanRijnhart:
Ported from https://code.launchpad.net/~therp-nl/web-addons/7.0-add_percentage_widget/+merge/217602

### DIFF
--- a/web_percent/__init__.py
+++ b/web_percent/__init__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2014 TeMPO Consulting (<http://www.tempo-consulting.fr>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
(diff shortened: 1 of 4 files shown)


StefanRijnhart:
To answer @pedrobaeza's comments on the Launchpad merge proposal: the conversion mechamisn that you requested assume that the value of a percentage field is between 0.0 and 1.0. Now, exactly for lack of a widget that does these conversions, existing percentage fields are more likely to be a value between 0 and 100. That makes this widget backwards incompatible with these columns.

pedrobaeza:
I think the best name for the module would be _web_widget_percentage_ to indicate that it's a widget (we discussed this before).

What I said was to have the widget UI with a value between 0 to 100, and internally return a value between 0.0 to 1.0 to be consistent with the base functionality. Discount fields in sales and accounting don't follow this rule, but we can include a widget option to configure the range if you want.

ovnicraft:
@StefanRijnhart now the widget just add % at the field ? conversion to [0,1] and validation [0,100] does not support yet ?

StefanRijnhart:
A widget option as suggested by @Pedrobaeza is in fact a good idea. I'll think about that (sorry for losing track of this one).

pedrobaeza:
@StefanRijnhart, are you going to continue this PR?

pedrobaeza:
No answer, so closing. Feel free to reopen if needed.

PR #159, opened 2015-06-30
[FIX][8.0] web_shortcuts : Make sure that the menu still exists in odoo

AdilHoumadi:
[FIX] : make sure that the menu still exists in odoo
[IMP] : simplify the code

``` python
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/http.py", line 530, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/http.py", line 567, in dispatch
    result = self._call_function(**self.params)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/http.py", line 303, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/service/model.py", line 113, in wrapper
    return f(dbname, *args, **kwargs)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/http.py", line 300, in checked_call
    return self.endpoint(*a, **kw)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/http.py", line 796, in __call__
    return self.method(*args, **kw)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/http.py", line 396, in response_wrap
    response = f(*args, **kw)
  File "/usr/local/openerp/openerp-buildout/parts/addons/addons/web/controllers/main.py", line 949, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/usr/local/openerp/openerp-buildout/parts/addons/addons/web/controllers/main.py", line 941, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/api.py", line 241, in wrapper
  File "/usr/local/openerp/openerp-buildout/parts/addons/openerp/api.py", line 336, in old_api
    result = method(recs, *args, **kwargs)
  File "/usr/local/openerp/openerp-buildout/addons-web/web_shortcuts/model/web_shortcut.py", line 45, in get_user_shortcuts
    [x['menu_id'][0] for x in results if x['menu_id']]
TypeError: 'bool' object has no attribute '__getitem__'
```

### DIFF
--- a/web_shortcuts/model/web_shortcut.py
+++ b/web_shortcuts/model/web_shortcut.py
@@ -39,14 +39,17 @@ class web_shortcut(models.Model):
     @api.model
     def get_user_shortcuts(self, user_id):
         shortcuts = self.search([('user_id', '=', user_id)])
-        results = shortcuts.read(['menu_id'])
-        ir_ui_menu_obj = self.env['ir.ui.menu']
-        menus = ir_ui_menu_obj.search([('id', 'in', [x['menu_id'][0]
-                                                     for x in results])])
-        name_map = dict(menus.name_get())
-        # Make sure to return only shortcuts pointing to existing menu items.
-        filtered_results = filter(lambda result: result['menu_id'][0] in
-                                  name_map, results)
-        for result in filtered_results:
-            result.update(name=name_map[result['menu_id'][0]])
-        return filtered_results
+        res = []
+        for shortcut in shortcuts:
+            if shortcut.menu_id:
+                _name = shortcut.menu_id.name_get()
+                _name = _name[0][1] if len(_name) else ''
+                _id = shortcut.menu_id.id
+                res.append(
+                    {
+                        'id': shortcut.id,
+                        'name': _name,
+                        'menu_id': (_id, _name)
+                    }
+                )
+        return res
--- a/web_shortcuts/model/web_shortcut.py
+++ b/web_shortcuts/model/web_shortcut.py
@@ -40,16 +40,15 @@ class web_shortcut(models.Model):
     def get_user_shortcuts(self, user_id):
         shortcuts = self.search([('user_id', '=', user_id)])
         res = []
-        for shortcut in shortcuts:
-            if shortcut.menu_id:
-                _name = shortcut.menu_id.name_get()
-                _name = _name[0][1] if len(_name) else ''
-                _id = shortcut.menu_id.id
-                res.append(
-                    {
-                        'id': shortcut.id,
-                        'name': _name,
-                        'menu_id': (_id, _name)
-                    }
-                )
+        for shortcut in shortcuts.filtered('menu_id'):
+            _name = shortcut.menu_id.name_get()
+            _name = _name[0][1] if len(_name) else ''
+            _id = shortcut.menu_id.id
+            res.append(
+                {
+                    'id': shortcut.id,
+                    'name': _name,
+                    'menu_id': (_id, _name)
+                }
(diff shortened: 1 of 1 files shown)


web_shortcuts/model/web_shortcut.py
```diff
@@ -39,14 +39,17 @@ class web_shortcut(models.Model):
...
-                                                     for x in results])])
-        name_map = dict(menus.name_get())
-        # Make sure to return only shortcuts pointing to existing menu items.
-        filtered_results = filter(lambda result: result['menu_id'][0] in
-                                  name_map, results)
-        for result in filtered_results:
-            result.update(name=name_map[result['menu_id'][0]])
-        return filtered_results
+        res = []
+        for shortcut in shortcuts:
```
pedrobaeza, line 64: You can use `shortcuts.mapped('menu_id')` and it'll remove automatically False values

AdilHoumadi, line 64: Yeah sure, it's better.

AdilHoumadi, line 64: Thanks for the review, In the same vein, I finally used the filtered function because mapped function returns a set of ir.ui.menu instead of web.shortcuts objects

pedrobaeza, line 64: Yeah, but I thought that you only need menus (and it seems all the data you manipulate are from menu_id), not shortcut records.

AdilHoumadi, line 64: I tried to stay compatible with the old code :

``` python
filtered_results = [{
'menu_id': (543, u'Messaging/Users/Users'), 
'id': 26,  <--- come from the shortcut object
'name':  u'Messaging/Users/Users'
}]
```

If you agree we can manipulate ir.ui.menu instead of shortcuts

``` xml
<t t-name="Systray.Shortcuts.Item">
    <li>
        <a href="#" t-att-data-id="shortcut.menu_id[0]" t-att-data-shortcut-id="shortcut.id">
            <t t-esc="shortcut.name"/>
        </a>
    </li>
</t>
```

pedrobaeza, line 64: OK, I see it is used in several places, so please stay as you have put it. Thanks for the efforts

pedrobaeza:
:+1:

legalsylvain:
Hi, 
:question: 
I'm not sure to understand, but it seems that this part of code try to fix database bugs. In which (real) case an web.shorcut should not have a menu_id defined ?

If there is no use case, why not overload 'ir.ui.menu'.unlink function and delete obsolete shorcuts ? It will fix the problem and avoid to have a non consistent DB.

Thanks.

AdilHoumadi:
Hi,
Yeah indeed, we're fixing orphaned items. Another suggestion, I think is to change the column definition to be like that, what do you think ?

``` python
menu_id = fields.Many2one('ir.ui.menu', ondelete='cascade')
```

Thanks.

OSguard:
i like more the database change with 'cascade'

hbrunn:
@ah-taktik let's do both, then it works for already messed up databases and avoids the problem in the future

AdilHoumadi:
It's done !

adrienpeiffer:
:+1:

PR #162, opened 2015-07-02
[ADD] New module: web_readonly_bypass

JonathanNEMRY:
This Module provides a solution to the problem of the interaction between 'readonly' attribute and 'on_change' attribute when used together


web_readonly_bypass/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jonathan Nemry <jonathan.nemry@acsone.eu>
+* Laetitia Gangloff <laetitia.gangloff.eu>
```
adrienpeiffer: I think email address isn't correct

adrienpeiffer:
Can you remove `All Rights Reserved` from header ?

adrienpeiffer:
Little remarks on readme file and about header. Otherwise, :+1:

web_readonly_bypass/__openerp__.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Authors: Nemry Jonathan & Laetitia Gangloff
+#    Copyright (c) 2014 Acsone SA/NV (http://www.acsone.eu)
+#    All Rights Reserved
```
pedrobaeza: Remove this useless line

pedrobaeza:
Very interesting feature! :+1:  (code review)

web_readonly_bypass/views/readonly_bypass.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- vim:fdn=3:
+-->
+<openerp>
+    <data>
+        <template id="assets_backend" name="readonly_bypass" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/readonly_bypass/static/src/js/readonly_bypass.js"></script>
```
petrus-v: As you rename module  you miss to change the path:

Should be

``` xml
<script type="text/javascript" src="/web_readonly_bypass/static/src/js/readonly_bypass.js"></script>
```

web_readonly_bypass/static/src/js/readonly_bypass.js
```diff
@@ -0,0 +1,109 @@
+/*
+ * Allow to bypass readonly fi the value is changed
+ */
+
+openerp.readonly_bypass = function(instance) {
```
petrus-v: Same here, as you rename the module you forgot to rename the object name:

``` javascript
openerp.web_readonly_bypass = function(instance) {
```

otherwise the class is not loaded.

web_readonly_bypass/static/src/js/readonly_bypass.js
```diff
@@ -0,0 +1,109 @@
...
+     * @param boolean mode: True case of create, false case of write
+     * @param {Object} context->filter_out_readonly
+     * @param {Object} data field values to possibly be updated
+     * @param {Object} options Dictionary that can contain the following keys:
+     *   - readonly_fields: Values from readonly fields to merge into the data object
+     */
+    function ignore_readonly(data, options, mode, context){
+        if (options){
+            if (!('filter_out_readonly' in context && context['filter_out_readonly'] == true
+                    && 'readonly_fields' in options && options['readonly_fields'])) {
```
petrus-v: Be careful, if I'm not wrong, when evaluating condition with `&&` in javascript the right statement is not evaluated if the left is already false. By luck `options.readonly_fields` looks always defined.

I would prefer:

``` javascript
if ('readonly_fields' in options && options['readonly_fields'] &&
    !('filter_out_readonly' in context && context['filter_out_readonly'] == true ))
```

petrus-v:
Excellent :+1:

pedrobaeza:
DO NOT MERGE YET!

Thinking about this carefully, this module inverts the behaviour by default, so it applies to all existing codebase, and can lead to unexpected behaviours in modules that rely on the volatibility of onchange to change data, but they don't want to store it.

The module must act the contrary: use the normal behaviour by default, and include a context key `not_filter_out_readonly` or similar that allows that someone include this dependency and this value in the field context when he wants to change the behaviour.

sbidoul:
Although I don't have any concrete cases in mind, I agree with @pedrobaeza: it's risky to have a module that changes the standard behavior system wide, as this may have unexpected side effects.

@petrus-v since you are testing it these days, would you mind doing that change?

petrus-v:
@sbidoul 

I agree with @pedrobaeza! This behavior will be safer for me as well, I'll make change on #164 as well!

I wonder how to set the context correctly, I expected to set it on the field or the entire view but it looks more complicated as I expected to forward it properly!

Thanks to take me in consideration!

pedrobaeza:
I think the best approach is to define the context at field level, so you can fine-grain the control over each readonly field.

JonathanNEMRY:
@petrus-v 
an idea could be to say: `if we want this behavior then add it into the context action of the form`. And then the added context should simply be like that `{'readonly_by_pass': True}` or in a more complex way `{'readonly_by_pass': ['concerned_field', '...', ...]}`
By this way, it will be possible to extend the scope of this behavior, from a specific field to an entire form and the changes into the existing code are minimals

pedrobaeza:
Well, I prefer the duality of the key in the action window context or the key in the field context.

web_readonly_bypass/static/src/js/readonly_bypass.js
```diff
@@ -0,0 +1,109 @@
...
+     * @param boolean mode: True case of create, false case of write
+     * @param {Object} context->filter_out_readonly
+     * @param {Object} data field values to possibly be updated
+     * @param {Object} options Dictionary that can contain the following keys:
+     *   - readonly_fields: Values from readonly fields to merge into the data object
+     */
+    function ignore_readonly(data, options, mode, context){
+        if (options){
+            if ('readonly_fields' in options && options['readonly_fields'] &&
+                    !('filter_out_readonly' in context && context['filter_out_readonly'] == true )) {
```
petrus-v: @JonathanNEMRY , @pedrobaeza 

To be accurate, my problem was that, when I quickly tested to define context on the field like this on the view:

``` xml
<field name="my_readonly_field" context="{'filter_out_readonly': True}" />
```

or on the action windows. 

The javascript `context` parameter does'nt contain the `filter_out_readonly` properties :-/

It needs more  investigations ;-)

JonathanNEMRY:
Added a specific context into a field will be not propagated until the `dataset` or `datasetbuffer` action (create/write in our case). Context is propagated for relational fields if a context is present into the node: 
`build_context: function() {
        // only use the model's context if there is not context on the node
        var v_context = this.node.attrs.context;`
This method is not called for other fields so I agree with @pedrobaeza to have a dual context key `action or field` but it will be more difficult  and that's why I suggested to use the context action

pedrobaeza:
Ok, I don't know about the difficulty in JS side, so go ahead with the one you propose, @JonathanNEMRY

JonathanNEMRY:
So @petrus-v do you think it is OK for you to consider the special context key will be added into the action's context (`ir.action.act_window`)? (so it will be available into the `self.context` of `create/write` from the `dataset` classe )

petrus-v:
@JonathanNEMRY 

Yes I'm fine with that! I'll do the same. thanks for the tip!

petrus-v:
@JonathanNEMRY 

I've implemented a way to do it on PR #164 but I still get an issue as this.context on `BufferedDataSet` class does'nt contains window action context values. Do you have any idea/good practice to foward that context?

regards,

sbidoul:
Setting as WIP until 7.0 version is finished. Please review #164 first.

pedrobaeza:
Are this PR in line with the work done in 7.0?

JonathanNEMRY:
Hello @pedrobaeza ,
This PR is no more in line with the work done in 7.0
It should be more easier to port 7.0 in 8.0...what do you think? (any other ideas are welcome)

pedrobaeza:
Yeah, I think so

petrus-v:
@JonathanNEMRY 

In this case, not sure if cherry-pick commits would be easier than port from v7 ?!

Would you like I port it or you do it?

JonathanNEMRY:
You may continue the work if you feel you're ready to do it and as I had no time for the moment. But I am OK to do it later if you want. And you're right cherry-pick would be an easier way to have it aligned with the 7.0's modifications

petrus-v:
@JonathanNEMRY I've done some works there https://github.com/petrus-v/web/tree/8.0-web_readonly_bypass but I'm going to be off for a month, I could create a PR on acsone repo first, then let you follow this PR, or I create a new fresh PR on OCA/web repo, or you can directly merge my branch (becarefull I rebase on the top of oca/8.0)? What do you prefer ?

JonathanNEMRY:
@petrus-v  you can make a PR on https://github.com/acsone/web/tree/8.0-web_readonly_bypass-jne and then I will take care of the next steps! see you next month ;)

petrus-v:
@JonathanNEMRY 

Ok, thanks, I'll do that.

pedrobaeza:
Any update on this?

petrus-v:
@pedrobaeza, @sbidoul 

 I think we could replace `work in progress` tag by `needs review` tag as PR #164 was merged.

Obviously I give my :+1: as I've done the end of works, but I'm not sure it has values!

web_readonly_bypass/README.rst
```diff
@@ -0,0 +1,86 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
lmignon: Can you take a look to the latest version of the sample README.rst (https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst) and fix the following issues?
- add target to the image licence badge
- put a lot of '===' above the module title to make the title a real title
- add a link 'Try me on Runbot' in the usage section

web_readonly_bypass/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Read Only ByPass',
+    'version': '1.0',
```
lmignon: The version number should be 8.0.1.0.0 https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

web_readonly_bypass/static/src/js/readonly_bypass.js
```diff
@@ -0,0 +1,140 @@
+(function(){
+    var instance = openerp;
```
lmignon, line 3: Can you declare your JS as js-strict? https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#javascript

web_readonly_bypass/__openerp__.py
```diff
@@ -0,0 +1,37 @@
...
+    "author": "ACSONE SA/NV",
```
adrienpeiffer: Add OCA as author

JonathanNEMRY:
Hi,
I just add some improvements referenced to the last review and I've also tested the module with @petrus-v 's changes. It seems OK for me now. Just some more thumbs up before merging.

lmignon:
:+1:

web_readonly_bypass/README.rst
```diff
@@ -0,0 +1,79 @@
...
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jonathan Nemry <jonathan.nemry@acsone.eu>
+* Laetitia Gangloff <laetitia.gangloff@acsone.eu>
```
sbidoul, line 64: Add @petrus-v in contributors.

JonathanNEMRY:
@petrus-v is added to the contributors into README.rst: sorry for this omission and thanks again for your contribution!

petrus-v:
@JonathanNEMRY,

oh yes I missed it too, thanks! That was with pleasure ;), you did the hardest way

pedrobaeza:
:+1:

PR #164, opened 2015-07-02
7.0 web readonly bypass

petrus-v:
This is porting of ascone `web_readonly_bypass` module made by #162


web_readonly_bypass/__openerp__.py
```diff
@@ -0,0 +1,115 @@
...
+    "author": "ACSONE SA/NV, Anybox",
+    "maintainer": "ACSONE SA/NV, Odoo Community Association (OCA)",
+    "website": "http://www.acsone.eu",
+    'category': 'Technical Settings',
+    'depends': [
+        'base',
+        'web',
+    ],
+    'description': """
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
pedrobaeza: You can still put this into README.rst (although duplicated) for easy reading when browsing in GitHub

pedrobaeza:
Please check the 2 lint issues:

```
./web_readonly_bypass/__openerp__.py:80:80: E501 line too long (90 > 79 characters)
./web_readonly_bypass/__openerp__.py:81:80: E501 line too long (203 > 79 characters)
```

guewen:
:+1:

pedrobaeza:
Please take into account here also comments of https://github.com/OCA/web/pull/162

guewen:
@pedrobaeza I just read #162 so I removed by thumb up until it is settled.

petrus-v:
I still get an issue on fields on o2m fields (like sale order line) I would like to fix it before merge.

JonathanNEMRY:
Hello @petrus-v 
I just tested your PR on runbot OCA and debug too the JS. Seems that works very well and context is propagated until the Dataset.context. What I've done for my test:
- Debug mode Odoo
- Go into quotation
- Edit form view and make `payement_term` field readonly
- Edit Quotation's action and set context with `{'key': 'value'}`
- Reload the page and click on action.
- Edit the quotation and change customer to modify the payment term
- Add a breakpoint on `instance.web.DataSet.include/write` of `web_readonly_bypass`
- print self.context and it contains as well the `{'key': 'value'}`

Is there something different into my test case than your?
I also saw that you had added a test into the module: very nice work!

petrus-v:
@JonathanNEMRY,

Thanks for tested, yes there is something different! I probably wasn't explicit...

Problem happens on a Sale order line field, (I moved description(name field) as readonly on runbot) when the sale order line loose the focus, context is not filled. 

Then when we save the context is set but option is not there any more! 

Can you reproduce ?

Sorry, I forgot to remove some ugly logs, I have to improve my pre-commit hook!

JonathanNEMRY:
Seems the context is not computed into the `instance.web.BufferedDataSet` so try to use 

```
context = instance.web.pyeval.eval('contexts', self.context.__eval_context);
ignore_readonly(data, options, false, context);
```

into the `write/create` of  `instance.web.BufferedDataSet`

then the onchange will trigger the first `write` into the `BufferedDataSet` and `sale.order.line.name` will pass from options to data and then when saving the entire `sale.order` form: modification will be reported into the data of the record: `sale.order.order_line`

I think that should work with this modification

petrus-v:
Thanks @JonathanNEMRY it works smoothly!

web_readonly_bypass/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
JonathanNEMRY: you can remove the header from the `init.py`

JonathanNEMRY:
Code review & functional tests: it works well :+1: 
Thanks!!

petrus-v:
@JonathanNEMRY 

Header is removed in `__init__.py`!

web_readonly_bypass/__openerp__.py
```diff
@@ -0,0 +1,123 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Read Only ByPass',
+    'version': '1.0',
+    "author": "ACSONE SA/NV, Anybox",
```
pedrobaeza: Include here for indexing purposes OCA

pedrobaeza:
Is there any possibility to exercise this module in a test?

petrus-v:
Hi @pedrobaeza,

I'm not sure to understand what you mean... 

Would you like that I:
- create a test module (`web_readonly_bypass_test`) to let functional users test/try quickly on runbot
- Or running javascript unitest which appear to be already done by Travis according::

```
2015-07-31 08:04:22,971 22906 INFO openerp_test openerp.modules.loading: module web_readonly_bypass: loading static/test/web_readonly_bypass.js
```

Actually I wonder if _loading_ means _running test_ (should create a bug to make sure javascript unitest are tested by Travis...)
- Or add javascript unit test that make sure when a user interact with a readonly field that change by an onchage method the value will be send to the write method when saving data?

pedrobaeza:
Sorry for not being clear. I mean last option: create a JS unit test (or Python, I'm not sure) to make sure that the module works as expected.

JonathanNEMRY:
I think it is OK with the current added tests (that were not present before this PR) as the behavior of this module is to add readonly fields into the data passed to the server. Are you agree @pedrobaeza ?

pedrobaeza:
Sorry, I didn't see them. Yeah, of course, they are more than enough.

:+1: Merging this excellent module!

JonathanNEMRY:
Thanks for merging @pedrobaeza  and thanks @petrus-v  for this work!!

petrus-v:
Thanks fore merging and reviews!

@JonathanNEMRY thanks for helps and the base module, it save me time!

PR #161, opened 2015-07-01
[ADD] support_branding

hbrunn:

PR #165, opened 2015-07-03
[7.0][ADD] support_branding

sebalix:
This PR is the 7.0 backport of the #161 from Therp replacing the closed one #126.

~~There is still one bug I can't resolve (I'm not a guru with the JS framework): close the CrashManager dialog once the "Send to X" button is pushed (https://github.com/OCA/web/compare/7.0...osiell:7.0-support_branding?expand=1#diff-6d32c65850abff6e99face5adef67c36R124).
Any help is welcomed!~~


support_branding/static/src/js/support_branding.js
```diff
@@ -0,0 +1,170 @@
...
+                            // hoping there is a properly configured email
+                            // client
+                            $body.val($description.val() + '\n' + $body.val())
+                            $button.unbind('click');
+                            $button.click();
+                        })
+                        .then(function()
+                        {
+                            //TODO close the dialog
+                            $form.parents('.ui-dialog').hide();
```
hbrunn: Upstream does that: https://github.com/odoo/odoo/blob/7.0/addons/web/static/src/js/chrome.js#L289

So I think in this case, you need to save the original `this` and call `$(self).dialog("close");`

sebalix: Thanks,
I tried with a `var self = this` at the beginning of:
- `show_error: function(error)`
- then `jQuery('.support-branding-submit-form').each(function() ...`

But I always obtain the following error message:
`Uncaught Error: cannot call methods on dialog prior to initialization; attempted to call method 'close'`

I keep looking.

hbrunn: ah, `this` actually is meant to refer to the dialog dom element. Then select it and call `dialog("close")` on it

sebalix: If you meant `$(this).dialog("close");`, it doesn't work neither...
Maybe it's related to the fact that the event is based on a button created manually outside of the Dialog's constructor mechanism.

hbrunn: no, it's whatever the dialog dom element is. _Maybe_ it's as simple as `$form.parents('.ui-dialog').dialog("close")` (I don't have a 7.0 instance at hand currently). Just look up in your browser what elements are created for the dialog in the dom, one of them is the right one to call this function on

sebalix: Yes, it was my first tentative in fact, but the error is still the same.

sebalix:
Fixed, the commit a684551 is a bit weird but works well.

hbrunn:
:+1:

pedrobaeza:
:+1: Working in runbot

PR #169, opened 2015-07-08
[ADD][6.1] New module: web_exportall

craiggowing:


pedrobaeza:
Is this still interesting for anyone?

pedrobaeza:
Closing as no answer. Any way, FYI, `web_export_view` (in v8) now exports all the recordset when you select on the upper part.

PR #170, opened 2015-07-08
[ADD][7.0] New module: web_exportall

craiggowing:

### DIFF
--- a/web_exportall/AUTHORS
+++ b/web_exportall/AUTHORS
@@ -1,0 +1,1 @@
+Craig Gowing <craig.gowing@credativ.co.uk>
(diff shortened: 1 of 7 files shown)


laetitia-gangloff:
:+1: (functional only)
It is really great, thank you !

pedrobaeza:
Please apply same comments than in 8.0: https://github.com/OCA/web/pull/171

OSguard:
can we finish this module?

pedrobaeza:
Is this still interesting for anyone? AFAIK, there are comments unattended.

pedrobaeza:
Closing as no answer. Any way, FYI, `web_export_view` (in v8) now exports all the recordset when you select on the upper part.

PR #171, opened 2015-07-08
[ADD][8.0] New module: web_export_streamed

craiggowing:


pedrobaeza:
I think the functionality of this module should be integrated on https://github.com/OCA/web/tree/8.0/web_export_view

jjscarafia:
I think that when you use The select all checkbox, no matter if you are
seeing only 80, it exports all records (from other pages also). If you
deselect one, then it will export 79.
Is the default behaviour as far as I concern. I may be wrong of course...

web_exportall/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+    'category': 'Web',
+    'author': 'credativ Ltd,Odoo Community Association (OCA)',
+    'website': 'http://credativ.co.uk',
+    'license': 'AGPL-3',
+    "depends": [
+        'web'
+    ],
+    'init_xml': [],
+    'update_xml': [],
+    'demo_xml': [],
```
bwrsandman: These keys are deprecated and replaced with `"data"` and `"demo"`

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,201 @@
...
+                import_data = Model.export_data(cr,
+                                                uid,
+                                                ids,
+                                                field_names,
+                                                raw_data,
+                                                context=context
+                                                ).get('datas', [])
+                for data in import_data:
+                    yield data
+    return
```
bwrsandman: redundant

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,201 @@
...
+            for ids in all_ids:
+                import_data = Model.export_data(cr,
+                                                uid,
+                                                ids,
+                                                field_names,
+                                                raw_data,
+                                                context=context
+                                                ).get('datas', [])
+                for data in import_data:
+                    yield data
```
bwrsandman: It's nicer to format this as

``` python
from openerp import http, api

#...
def itter_data(Model, dbname, uid, all_ids, field_names, raw_data, context):
    """ Reads data in chunks of ids from all_ids and
        yields the rows one by one """
    with sql_db.db_connect(dbname).cursor() as cr, api.Environment.manage():
        for ids in all_ids:
            import_data = Model.export_data(
                cr, uid, ids, field_names, raw_data, context=context
            ).get('datas', [])
            for data in import_data:
               yield data
```

web_exportall/controllers/main.py
craiggowing: Could you explain why this is redundant? This is different from the existing function in that it returns an iterator rather than a list of all records.

If for instance you are downloading 1M+ records the existing code will certainly crash the server as it will consume all memory (at least in version 6.1 and 7.0, it may have changed in 8.0 but not as far as I am aware). This change which streams the data will use limited memory.

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,201 @@
...
+    """ Splits all_ids into more manageable chunks of
+        size step_size and yields them """
+    offset = 0
+    while True:
+        id_set = all_ids[offset:offset+step_size]
+        offset += step_size
+        if len(id_set):
+            yield id_set
+        else:
+            return
```
bwrsandman: Couldn't you put the `len` check into the while condition to simplify this?
Also seems like you can use a [`xrange` call instead of while](http://stackoverflow.com/a/312464)

craiggowing:
> I think that when you use The select all checkbox, no matter if you are
> seeing only 80, it exports all records (from other pages also). If you
> deselect one, then it will export 79.
> Is the default behaviour as far as I concern. I may be wrong of course...

I wasn't aware of that one, if that is the case then the extra drop down could be removed and the same logic could be used to see if we are exporting all or not.

I think the same use case still applies in 8.0, even if it is exporting all as standard, if you have 1M+ records you will find the server quickly running out of memory since it does not stream the response. This module adds the streaming functionality.

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,201 @@
...
+                                     request.uid,
+                                     domain,
+                                     context=context))
+        import_data = itter_data(Model,
+                                 request.cr.dbname,
+                                 request.uid,
+                                 ids,
+                                 field_names,
+                                 self.raw_data,
+                                 context)
```
bwrsandman: You can save a lot of lines by making your line break right after `(` and it keeps the code from looking lopsided.

``` python
        import_data = itter_data(
            Model, request.cr.dbname, request.uid, ids, field_names,
            self.raw_data, context
        )
```

web_exportall/__openerp__.py
craiggowing: Thanks for the hint, this was copied over from a 6.1 version so I must have forgot to remove them. They are removed in the latest commit.

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,201 @@
...
+                                                field_names,
+                                                raw_data,
+                                                context=context
+                                                ).get('datas', [])
+                for data in import_data:
+                    yield data
+    return
+
+
+class ExportAllBase(ExportFormat):
```
bwrsandman: You can probably simplify the inheriting classes even more: from_data could be defined here with abstract functions overwitten in the inheriting classes for things such as `write_row`, `write_header`

The StringIO logic can definitely be factored.

web_exportall/controllers/main.py
bwrsandman: A return at the end of a function or a generator is redundant. Unless I am missing something, the implicit last call in a python function is always `return None`.

craiggowing: I see what you mean. I guess it is a bad habit I got into!

Thanks for all the feedback, I'll look at refactoring the code based on your comments.

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,201 @@
...
+    with cr:
+        with openerp.api.Environment.manage():
+            for ids in all_ids:
+                import_data = Model.export_data(cr,
+                                                uid,
+                                                ids,
+                                                field_names,
+                                                raw_data,
+                                                context=context
+                                                ).get('datas', [])
```
bwrsandman: > If for instance you are downloading 1M+ records the existing code will certainly crash the server as it will consume all memory (at least in version 6.1 and 7.0, it may have changed in 8.0 but not as far as I am aware). This change which streams the data will use limited memory.

Unless `export_data` is a generator (I didn't check), the memory is already allocated here and `import_data` refers to it. As long as the return from this function doesn't copy this data, the memory is already used up until the generator of this function is done iterating or otherwise deallocated.

eLBati:
I agree with @pedrobaeza

craiggowing:
@pedrobaeza - I don't think this could be integrated with web_export_view since this module is exporting in XLS format and it is not possible to stream a response in this format AFAIK.

The web_exportall module combats the memory usage issue with XLS on the ORM side by reading a batch of IDs at a time to feed the input data into the XLS library, however the XLS file needs to be complete before it can be created and returned in the response.

Somebody please correct me if I am wrong about being able to stream an XLS file, this would be hugely useful if it was possible! (I'm not holding my breath, but the MS docs do refer to the encapsulated data as a "stream"... maybe there is a way!)

pedrobaeza:
Well, you have 2 features in this module:
- Export all records, that can be included as a feature in the other module.
- Export to CSV. This can be also another option in the module.
- Stream content, that also can be included only for the supported format/s.

dreispt:
@jjscarafia 

> I think that when you use The select all checkbox, no matter if you are seeing only 80, it exports all records (from other pages also). If you deselect one, then it will export 79. Is the default behaviour as far as I concern. I may be wrong of course...

I checked `web_export_view` and it only exports the current page.

I also agree with @pedrobaeza : 

This overlaps the `web_export_view` function, so no point in having both in the repo. It this provides superior features, it should be renamed and replace `web_export_view`.

pedrobaeza:
They are not superior, but complementary: `web_export_view` allows to export to XLS, and this one no. My proposal is https://github.com/OCA/web/pull/171#issuecomment-119710676

bwrsandman:
@pedrobaeza You're mistaken: https://github.com/OCA/web/pull/171/files#diff-79d6a21f53707ac3fc0a46719e389b2fR161 is XLS

pedrobaeza:
Ah, from the author I deduced XLS was not covered because the streaming problem

bwrsandman:
`web_export_view` is lacking in pdf export that modules like [`web_printscreen_zb`](https://github.com/zbeanz/openerp-7-web-addons.git) have

craiggowing:
A few updates with some refactoring based on your comments, also since 8.0 does export all or export selected already based on the selected check boxes I removed the view changes as they are redundant now.

dreispt:
@craiggowing What are the differences between this and [web_export_view](https://github.com/OCA/web/tree/8.0/web_export_view), and why should we need both modules?

craiggowing:
@dreispt - The main purpose of web_exportall is to allow the standard Odoo export to export many rows in a streamed response from the server. The main issue currently if you ever need to export millions of rows is that either the server will run out of memory, or the request will timeout waiting for the server to generate the file and make the response. This module will keep a very low memory footprint when exporting many rows and the streamed response will prevent timeout errors from occurring.

The web_export_view module has a completely different use case in that it exports what is seen on the current view, ie only currently viewed rows and columns if I understand it correctly.

dreispt:
@craiggowing OK, that sounds good.

pedrobaeza:
Call the module _web_export_streamed_ or something similar then

craiggowing:
@pedrobaeza - Thanks for the suggestion, I've renamed the module to web_export_streamed. It was historically called web_exportall since 6.1 when it did add the ability to export all rows, but I suppose in 8.0 the name is slightly outdated!

web_exportall/AUTHORS
```diff
@@ -0,0 +1 @@
+Craig Gowing <craig.gowing@credativ.co.uk>
```
pedrobaeza: Authors file is not used on OCA modules. Please add it to the README file and the manifest file.

web_exportall/README.rst
```diff
@@ -0,0 +1,22 @@
+Description
```
pedrobaeza: Please use the OCA's template for README: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

web_exportall/__openerp__.py
```diff
@@ -0,0 +1,35 @@
+# -*- encoding: utf-8 -*-
```
pedrobaeza: s/encoding/coding

web_exportall/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Web Export All',
+    'application': False,
+    'version': '1.1',
```
pedrobaeza: Use full version numbering scheme: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

web_exportall/controllers/main.py
```diff
@@ -0,0 +1,186 @@
...
+            self.from_data(columns_headers, import_data),
+            headers=[('Content-Disposition',
+                     'attachment; filename="%s"'
+                      % self.filename(model)),
+                     ('Content-Type', self.content_type)],
+            cookies={'fileToken': token}
+        )
+
+
+class CSVExportAll(ExportAllBase, http.Controller):
```
pedrobaeza: Please use CamelCase for class names

pedrobaeza:
I have tried the module on runbot and I can't find the dropdown to sellect "All records". Where is it?

craiggowing:
More updates pushed based on the OCA formats.
On the 8.0 version there is no longer any change to the UI for the CSV exporter. To use the streaming functionality you would just need to "select all" from the list view. I've also updated the readme based on this since the extra UI elements are only required pre 8.0

pedrobaeza:
Is this still interesting for anyone? AFAIK, there are comments unattended.

pedrobaeza:
Closing as no answer.

PR #173, opened 2015-07-10
[ADD] web_search_datetime_completion

hbrunn:
# More completion options for datetime fields

This module was written to extend the offers shown to the user when starting to fill in a date. In standard odoo, the only offer is the date at 00:00:00 hours, while it can be convenient for the user to have other options, too. When searching for the end of a period, `23:59:59` would come to mind, and there also might be domain specific times that make sense.
# Configuration

If you do nothing, only the time `23:59:59` is added to the completion dropdown.

In your code, you can add an options dictionary with a key `completion_options` which contains a list if dictionaries detailing the offsets that should be added to the parsed date, like::

<field name="my_datetime_field" options="{'completion_options': [{'hours': 12, 'minutes': 30, 'seconds': 0}]}" />

which would offer to search for the date at half past twelve. Note that when you set `completion_options`, you have to list all options you want to see, as the standard options will be replaced with your list.
# Usage

Start filling in your date in a search field and choose one of the additional options offered.
# Screenshot

web_search_datetime_completion_options


web_search_datetime_completion/README.rst
```diff
@@ -0,0 +1,57 @@
...
+<field name="my_datetime_field" options="{'completion_options': [{'hours': 12, 'minutes': 30, 'seconds': 0}]}" />
+
+which would offer to search for the date at half past twelve. Note that when you set `completion_options`, you have to list all options you want to see, as the standard options will be replaced with your list.
+
+Usage
+=====
+
+Start filling in your date in a search field and choose one of the additional options offered.
+
+For further information, please visit:
```
pedrobaeza, line 24: You can add here the new cool button "Try on runbot":

```
.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
   :alt: Try me on Runbot
   :target: https://runbot.odoo-community.org/runbot/162/8.0
```

nhomar:
:+1: :star2:

pedrobaeza:
Trying on runbot, I get this JS error: `ReferenceError: _t is not defined`

It's located in web_search_datetime_completion.js, line 60: `_t("Search %(field)s at: %(value)s")), {`

web_search_datetime_completion/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "More completion options for datetime fields",
+    "version": "1.0",
+    "author": "Therp BV",
```
OSguard: OCA as author is missing

OSguard:
get the same error

hbrunn:
@pedrobaeza @OSguard thanks for the reviews, fixed now

OSguard:
i try to test it on runbot:

http://3139456-173-1d4682.runbot1.odoo-community.org/web?db=3139456-173-1d4682-all&debug=#page=0&limit=80&view_type=list&model=sale.order&action=323

while adding: `<field name="date_order"  />` to search view, but i only get two dates search with "at" and not with "from" and "to". What did i wrong?

hbrunn:
@OSguard nothing. The trick is that you get this two offers instead of one. Everything else is up to the developer using the module. In the module I used for the screenshot, I said date_from should only offer midnight, while date_to should only offer a second before midnight

OSguard:
Ah, now i get it "Date From" ist the label of your field. What i suggest, was that if i have a field "date_order" in the search view, i get not only a suggested with `=` also with `<=` or/and `>=`
Would it be possible to add this as option in this module?

hbrunn:
would be a nice feature I think, but I'm not going to implement it any time soon

OSguard:
:+1: for this one
should we add this as possible improvement to the readme or create a github issue?

hbrunn:
@OSguard thanks, an issue seems fine to me

jholze:
:+1:  nice feature.

PR #174, opened 2015-07-10
[ADD] support 'open' on many2many_tags

hbrunn:
[ADD] support 'no_open_edit' on many2one
[FIX] typos

---

`no_open_edit` _boolean_ (Default: value of `no_open` which is `False` if not set)

  Causes a many2one not to offer to click through in edit mode, but well in read mode

`open` _boolean_ (Default: `False`)

  Makes many2many_tags buttons that open the linked resource

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -1,17 +1,13 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
 ==================================
 Add new options for many2one field
 ==================================
 
-
-Description
------------
-
 This modules modifies "many2one" and "many2manytags" form widgets so as to add some new display
 control options.
-
-**New: support many2manytags widget !**
-
-**New: support global option management with ir.config_parameter !**
 
 Options provided includes possibility to remove "Create..." and/or "Create and
 Edit..." entries from many2one drop down. You can also change default number of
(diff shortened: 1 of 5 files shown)


web_m2x_options/README.rst
```diff
@@ -84,6 +92,6 @@ Your XML form view definition could contain::
 Note
 ----
 
-Double check that you have no inherited view that remote ``options`` you set on a field ! 
-If nothing work, add a debugger in the first ligne of ``get_search_result method`` and enable debug mode in OpenERP. When you write something in a many2one field, javascript debugger should pause. If not verify your installation.
+Double check that you have no inherited view that remove ``options`` you set on a field ! 
+If nothing works, add a debugger in the first line of ``get_search_result method`` and enable debug mode in OpenERP. When you write something in a many2one field, javascript debugger should pause. If not verify your installation.
```
pedrobaeza: s/OpenERP/Odoo

pedrobaeza: Please take the occasion to update the template to current OCA's: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

pedrobaeza:
Thanks for bringing these options. One of my wishlist was to allow to open many2many tags, so it seems that you read my secret wishlist!

Let me suggest some suggestions that I send you as always to try to improve the feature:
- Instead of making the tags rectangle clickable, I think it's better to put the text as a clickable link, so we will get a consistent behaviour/aspect with other clickable elements (many2one...).
- In edit mode, it would be great to add an icon like the one on many2one fields to allow to open the many2many in a popup window.
- Include this feature as a configurable option via parameter to have this behaviour by default in all many2many tags.

web_m2x_options/README.rst
hbrunn: @pedrobaeza thanks, done

hbrunn:
thanks for your hints @pedrobaeza, but I won't have time to implement them any time soon, so let's merge this as it is for now

pedrobaeza:
Please include them at least as Known issues / Roadmap then

hbrunn:
@pedrobaeza done

pedrobaeza:
Thanks :+1:

web_m2x_options/README.rst
```diff
@@ -1,35 +1,35 @@
...
-**New: support global option management with ir.config_parameter !**
-
 Options provided includes possibility to remove "Create..." and/or "Create and
 Edit..." entries from many2one drop down. You can also change default number of
 proposition appearing in the drop-down. Or prevent the dialog box poping in
 case of validation error.
 
 If not specified, the module will avoid proposing any of the create options
-if the current user have no permission rights to create the related object.
+if the current has no permission rights to create the related object.
```
petrus-v: I guess you removed `user` by mistake?

petrus-v:
README typo and pylint complaints otherwise I give my :+1: Thanks for this adds!

hbrunn:
@petrus-v thanks, fixed

PR #181, opened 2015-07-17
[PORT][8.0] port web_popup_large

legalsylvain:
- port web_popup_large in v8;
- use now css file instead of js file;

Thanks for your review.

Kind regards.


sbidoul:
This looks similar to https://github.com/OCA/web/tree/8.0/web_dialog_size.

legalsylvain:
Yes. I didn't know this new V8 module. But IFAIK, the popup is opened in the normal size by default. 
Are you agree if I merge the two modules into your web_dialog_size module, setting by default the popup in large mode ?

Kind regards.

pedrobaeza:
I already have it! Let me propose it when I come back from the wedding.

legalsylvain:
OK ! Have a nice wedding @pedrobaeza 
Please in your PR, delete the **unported**/web_popup_large folder to avoid other mistakes.

Kind regards.

sbidoul:
I'm not sure having it unconditionally at 95% width is good. Most dialogs work fine with the default size, and many become ugly when enlarged on a wide screen. So I find it good to let the user enlarge it with a button when needed.

hbrunn:
how about an ir.config_parameter that decides this? (and if you do that, please cache it globally)

sbidoul:
@hbrunn it makes sense.

pedrobaeza:
Please see https://github.com/OCA/web/pull/190 for the PR. I close this one

PR #184, opened 2015-07-22
Port web_translate_dialog to 8.0

alexis-via:


dreispt:
Can you extract the description to a README.rst?
I tested this and my translation popup shows no lines to translate. Any clue?

alexis-via:
@dreispt That's very strange ; it works well here at my end. You mean, when you open the translation popup on a product, you don't have anything ? When I do that, I have a pop-name like this one:
http://people.via.ecp.fr/~alexis/web_translate_dialog.png

Did you reload the interface after the installation of the module ?

I'll take care of README.rst when I have time.

dreispt:
@alexis-via I see: this is for the translatable fields, not for the technical translations. Could you please make that clearer when extracting the README from the manifest?

web_translate_dialog/README.rst
```diff
@@ -0,0 +1,35 @@
+Web Translate Dialog
+====================
+
+This module replaces the standard translation view by an easy-to-use pop-up view where you can translate all the fields of the object in all the installed languages (for the old users of OpenERP, it may remind them the translation pop-up of OpenERP 6.1).
```
pedrobaeza: s/OpenERP (first one)/Odoo

pedrobaeza:
Tried on runbot and it's working fine. Please add in the "Known issues / Roadmap" a point about only open the affected field when clicking the flag icon instead of opening all the translatable fields, or if you want to improve it directly :wink: 

:+1:

dreispt:
:+1:

alexis-via:
@pedrobaeza In fact, I prefer the current behavior that shows all the translatable fields ; so I don't want to put in the "known issues" that it should open only the affected field :)

pedrobaeza:
But that confuses the user to which field he/she should translate. Would you block a modification by my part changing this?

For now, I'm merging this one.

alexis-via:
@pedrobaeza No, I wouldn't ! It's just my opinion, I'm not the author of this module.

pedrobaeza:
OK, I will let it as configurable via system parameter for satisfying all tastes :wink:

PR #186, opened 2015-07-23
[FIX] Now that we actually destroy editors, we need another check for…

### LINKED ISSUE #176: [8.0] web_ckeditor4 javascript error in Webkit based navigators when saving
module: web_ckeditor4
version: 8.0

**Steps to reproduce**
- Open a record containing a text_html widget in a Webkit based navigator (Chrome, Safari)
- Edit, leave the value in the text_html empty
- Save

**Current behaviour**
It raises the following error:
TypeError: undefined is not an object (evaluating 'self.editor.name').

**Expected behaviour**
The record should be saved without errors

 /cc @StefanRijnhart

StefanRijnhart:
… their currentness

Fixes #176

### DIFF
--- a/web_ckeditor4/static/src/js/web_ckeditor4.js
+++ b/web_ckeditor4/static/src/js/web_ckeditor4.js
@@ -38,7 +38,7 @@ openerp.web_ckeditor4 = function(instance)
                             self_arguments=arguments;
                         setTimeout(function()
                         {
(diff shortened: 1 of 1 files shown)


hbrunn:
:+1:

lefeverd:
Tested and approved!

StefanRijnhart:
@lefeverd Thanks for confirming!

PR #190, opened 2015-07-30
[IMP] web_dialog_size: Select default dialog size by config

pedrobaeza:
Following discussion on https://github.com/OCA/web/pull/181, I put this PR that switches the behaviour by default of the popup windows.


legalsylvain:
Hi @pedrobaeza : 
1/ could you delete unported/web_popup_large folder ?
2/ What do you think about the @sbidoul and @hbrunn remark's, on the other PR, about an ir.config_parameter ?
kind regards.
otherwise, :+1:. 

+++

pedrobaeza:
@legalsylvain, I have removed deprecated module (thanks for remembering me), and added an entry in the README file in "Known issues / Roadmap" section for the parameter, but just right now I can't afford the development of that feature.

pedrobaeza:
@rafaelbn, can you check this PR?

rafaelbn:
@pedrobaeza buttons to expand and shrink in dialog box are to close in the search bar

Here a picture

pedrobaeza:
That's not due to this module, but the _web_advanced_search_x2x_. Try to uninstall that module and try again. Anyhow, we should file an issue for that case to be solved.

rafaelbn:
OK, I opened #194 
:+1:

sbidoul:
For the reasons I mentioned in https://github.com/OCA/web/pull/181#issuecomment-122318521, I would not have this PR merged as is. It makes most dialog worse and improves only a few.

For this reason :-1: to merge this PR as is.

The only solution that would suit all needs it to have a cached system parameter that sets the default width in % or px, as suggested by @hbrunn in https://github.com/OCA/web/pull/181#issuecomment-122333286.

rafaelbn:
In https://github.com/OCA/web/pull/181: (closed by @pedrobaeza to finish functionality here)

@hbrunn 

> how about an ir.config_parameter that decides this? (and if you do that, please cache it globally)

@sbidoul 

> @hbrunn it makes sense

@pedrobaeza 

> Please see #190 for the PR. I close this one

Please @pedrobaeza / @legalsylvain could you add

> The only solution that would suit all needs it to have a cached system parameter that sets the default width in % or px, as suggested by @hbrunn in #181 (comment).

Then we could finish this need.

Thanks

pedrobaeza:
@hbrunn, can you point me to an example on how to cache globally a system parameter?

hbrunn:
@pedrobaeza I don't have an example at hand, what I would do it so write to `instance.$modulename.my_cache_object`, this should survive until the user reloads the web client.

hbrunn:
@pedrobaeza I made a PR on your branch to add the parameter to address @sbidoul's concerns.

Being on it, I took the liberty to reorganize the code a bit and to use fontawesome icons instead of text for the buttons, that looks nicer in my opinion:
icons

hbrunn:
PS: here the PR https://github.com/pedrobaeza/web/pull/1

pedrobaeza:
@hbrunn, thank you very much for taking care of this. As always, you're the JS master!

:+1: to the hbrunn's addition. @sbidoul, please validate it for merging this.

PR #195, opened 2015-08-05
IMP web_last_viewed_records - use full width of the header

ddico:
This change will allow the use of the full width when displaying the last viewed records.

Currently: 
screen shot 2015-08-05 at 1 37 06 pm

With this change:
screen shot 2015-08-05 at 1 38 03 pm


pedrobaeza:
:+1: Seems right

anajuaristi:
:+1: I like it!!! It's like having pages to navigate. Something that users are missing very often. From my side RTM

hbrunn:
:+1:

PR #196, opened 2015-08-05
IMP web_environment_ribbon - Minor improvement of transparency to all…

ddico:
…ow top menu visibility

Users can not see the top menu. Changing the transparency seems to be an easy fix

Currently:
screen shot 2015-08-05 at 1 50 47 pm

With this change:
screen shot 2015-08-05 at 1 50 23 pm

### DIFF
--- a/web_environment_ribbon/static/src/css/ribbon.css
+++ b/web_environment_ribbon/static/src/css/ribbon.css
@@ -16,7 +16,7 @@
     z-index: 10000;
     position: fixed;
     box-shadow: 0 0 3px rgba(0,0,0,.3);
-    background: #e43;
+    background: rgba(255, 0, 0, .6);;
 }
 
 .test-ribbon b {


pedrobaeza:
Thanks, good workaround. Although you can't still click over the ribbon, but at least now you see the content behind and can click on the left part.

:+1:

ddico:
Probably the best location for the ribbon would be over the logo (without going over the top menu)?

lmignon:
:+1: nice improvement

pedrobaeza:
Well, for now it's enough, but maybe it can be an improvement (without overlaying the left menus).

ddico:
The z-index of the banner is currently 10000, while the stack order for menu is 1001… so changing the order of the banner to a lower value (<1001) would do just what you suggested… I tried it and IMO it doesn’t look good… considering this is for testing and that you are able to see and use the menus I can agree with you that for now is good enough.

Thank you

pedrobaeza:
Merging then

PR #200, opened 2015-08-11
[8.0] web_action_conditionable

ovnicraft:
This module implements condition for actions in tree view.
do allows: edit=“false” create=“false” delete=“false”
This module allows too: delete=“state==‘draft’”, simple expression.


pedrobaeza:
Add a README according OCA's template (https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst) explaining the usage of the module.

web_action_conditionable/__openerp__.py
```diff
@@ -0,0 +1,14 @@
...
+    "name": 'web_action_conditionable',
+    "version": "0.1",
+    "depends": [
+        'base',
+        'web',
+    ],
+    'data': ['views/view.xml'],
+    "author": "Cristian Salamea,Odoo Community Association (OCA)",
+    "installable": True,
+    "active": False,
```
dreispt: The `active`key is deprecated.

dreispt:
Needs the README with the module description.

ovnicraft:
@pedrobaeza @dreispt done.

web_action_conditionable/README.rst
```diff
@@ -0,0 +1,77 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Web Action Conditionable
+========================
+
+This module was written to extend the functionality of actions in tree views.
+Odoo by default support delete="false" create="false", with this module you can:
+delete="state=='draft'".
```
dreispt: Use backticks around the code samples.

web_action_conditionable/README.rst
```diff
@@ -0,0 +1,77 @@
...
+Installation
+============
+
+To install this module, just follow basic steps to install an odoo module.
+
+
+Configuration
+=============
+
+No needed.
```
dreispt: s/No/Not

dreispt:
LGTM (no test) :+1:

ovnicraft:
@dreispt ok looks all is green :+1:

web_action_conditionable/static/src/js/views.js
```diff
@@ -0,0 +1,23 @@
+/*global openerp, _, $ */
+
+openerp.web_action_conditionable = function (instance) {
+  instance.web.View.include({
+    is_action_enabled: function(action) {
+      var attrs = this.fields_view.arch.attrs;
+      if (action in attrs) {
+	try {
+	  data = JSON.parse(attrs[action]);
```
hbrunn: don't do that yourself, call super here

web_action_conditionable/static/src/js/views.js
```diff
@@ -0,0 +1,23 @@
...
+	  return data;
+	} catch(error) {
```
hbrunn, line 10: does it make sense to inspect error and only do this if we had a json parse error?

web_action_conditionable/static/src/js/views.js
ovnicraft: why not ? call super gives any better point here ?

web_action_conditionable/static/src/js/views.js
ovnicraft, line 10: @hbrunn attrs comes with "true" when developer puts explicit and this module wants to support simple expression to decide.
parse  `state=='draft'` gives syntax error

web_action_conditionable/static/src/js/views.js
hbrunn: you don't know what happens in the future, and also not what other modules exist that might also do whatever with this function. By calling super and only acting on failures, you give other code the chance to do whatever that code is good for

web_action_conditionable/static/src/js/views.js
hbrunn, line 10: great, then you can do your parsing when super raises a syntax error

ovnicraft, line 10: will try your suggestion

ovnicraft:
@hbrunn sorry for delay, its fixed.

hbrunn:
@ovnicraft no worries :+1:

PR #201, opened 2015-08-11
[ADD] New module <web_group_expand>

oihane:


anajuaristi:
:+1: Functionally tested. RTM from my side. Very very usefull functionality where you can expand full group by level tree in one click

oihane:
@JayVora-SerpentCS has made other module that seems to have the same functionality, so anyone checking this PR please take a look to https://github.com/OCA/web/issues/204 , we have to decide which one is better :wink:

web_group_expand/static/src/js/web_group_expand.js
```diff
@@ -0,0 +1,43 @@
+openerp.web_group_expand = function(openerp) {
```
lmignon, line 2: use strict; is recommended for all javascript files (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#javascript)
https://github.com/odoo/odoo/blob/8.0/addons/web/static/src/js/openerpframework.js#L35

lmignon:
:+1: Works as expected and JS is easy to read even for non expert.

web_group_expand/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+    "category": "Web",
+    "version": "1.0",
+    "author": "OpenERP SA, "
+               "AvanzOSC, "
+               "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+               "Odoo Community Association (OCA)",
+    "depends": [
+        "web"
+    ],
+    "js": [
```
LeartS: As far as I know this has been deprecated since version 8 and does nothing. Replaced by extending `assets_backend`

web_group_expand/static/src/xml/expand_buttons.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- vim:fdl=1: -->
```
LeartS: No editor specific code in source!

web_group_expand/views/templates.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- vim:fdn=3:
```
LeartS: No editor specific code in source!

web_group_expand/README.rst
```diff
@@ -0,0 +1,58 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Group Expand Buttons
```
LeartS, line 5: Could you add an upper "decoration" too? See https://github.com/OCA/maintainer-tools/pull/109

web_group_expand/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+        "static/src/js/web_group_expand.js",
+    ],
+    "qweb": [
+        "static/src/xml/expand_buttons.xml",
+    ],
+    "css": [
```
LeartS: the css key does nothing since 8, same as javascript, infact your styling is not applied (just tried on runbot)

web_group_expand/views/templates.xml
```diff
@@ -0,0 +1,12 @@
...
+-->
+<openerp>
+    <data>
+        <template id="assets_backend" name="web_groupby_expand assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/web_group_expand/static/src/js/web_group_expand.js"></script>
```
LeartS: You should also link your css here.

LeartS:
If you can, please integrate the toggle button of this PR: https://github.com/OCA/web/pull/204

oihane:
@LeartS I have taken a new look to how do both modules work and they do not exactly work the same way, here you can expand by level, so if you group by in two levels you will have to push expand button twice. That's the main reason why there are both buttons.

oihane:
Changes done, continues with two buttons, but everything else done.

web_group_expand/static/src/js/web_group_expand.js
pedrobaeza, line 2: @oihane, please add that literal inside the function (http://www.w3schools.com/js/js_strict.asp)

pedrobaeza:
I have tried both proposed modules (SerpentCS one and this), and I prefer this approach, because it allows to click-through level by level, what can be very useful to avoid unfolding what you don't want. But this one needs to adjust the icon more centered on the button. You can do it altering the css classes.

pedrobaeza:
If you increase also a bit the size of the icons, it would be better

anajuaristi:
:+1:

pedrobaeza:
Can you squash your commits in one?

CarlosMolinaP:
:+1:

oihane:
@pedrobaeza squash done

pedrobaeza:
:+1:

PR #207, opened 2015-08-18
[FIX] description;

legalsylvain:
cosmetic changes in readme file;

### DIFF
--- a/web_dashboard_tile/README.rst
+++ b/web_dashboard_tile/README.rst
@@ -17,9 +17,9 @@ Usage
 =====
 
 * Dashboad sample, displaying Sale Orders to invoice:
-.. image:: web_dashboard_tile/static/src/img/screenshot_dashboard.png
+.. image:: ./static/src/img/screenshot_dashboard.png
 * Tree view displayed when user click on the tile:
-.. image:: web_dashboard_tile/static/src/img/screenshot_action_click.png
+.. image:: ./static/src/img/screenshot_action_click.png
 
 Kown issues/limits
 ==================


yvaucher:
:+1:

pedrobaeza:
But this change makes that the images are not shown on module form view:

seleccion_007

yvaucher:
so are visible either on Odoo or on github :/ that's sad.

And it is currently not visible on apps.odoo.com (I have little hope it could be shown there with this patch but it could not work with both.)

https://www.odoo.com/apps/modules/8.0/web_dashboard_tile/

lmignon:
@yvaucher https://www.odoo.com/apps/modules/8.0/web_dashboard_tile/static/src/img/screenshot_dashboard.png > Not found

legalsylvain:
@pedrobaeza. Agree. But better to show on github and apps than on form view. (and it must be fixed in Odoo)

pedrobaeza:
Are you sure this is going to be shown on apps.odoo.com?

yvaucher:
I think as @lmignon pointed it out, that it won't.

lmignon:
It won't be shown if the generated url looks like https://www.odoo.com/apps/modules/8.0/web_dashboard_tile/static/src/img/screenshot_dashboard.png .

legalsylvain:
@pedrobaeza  with this patch, the path  is now a classical relative path, and github can understand it. 
if Odoo apps or Odoo doesn't interprate it correctly, so we have to propose a fix in Odoo. Don't you think ?

CC : @gurneyalex.

PR #215, opened 2015-09-04
[IMP] check if we have an lxml with the upstream

hbrunn:
patch applied


pedrobaeza:
:+1:

lmignon:
:+1: can we add a comment somewhere to keep a ref to the issue on l ml?

OSguard:
i guess this can be merged?

PR #217, opened 2015-09-07
7.0 add extended binary field

ntodorova:
New version of binary field, including creation of ir_attachment object.


ntodorova:
@hbrunn can you please give me some tips to pass the coveralls test? : ))

extended_binary_field/__openerp__.py
```diff
@@ -0,0 +1,69 @@
...
+    ],
+    'css': [
+    ],
+    'demo': [
+    ],
+    'test': [
+    ],
+    'active': False,
+    'installable': True,
+    'auto_install': True,
```
legalsylvain: Why is this module auto_installable ?

pedrobaeza:
Is anyone interested in this one yet?

ntodorova:
I would like to bring it to the end if someone can help me with the coveralls. : )

hbrunn:
what kind of help do you need? what we want in the end is 100% test coverage, the good reads are https://www.odoo.com/documentation/8.0/reference/testing.html and https://docs.python.org/2/library/unittest.html

ntodorova:
But I think that the branch coveralls were decrease before my commit, or? And  on the branch in general, or I am wrong?

hbrunn:
yes, you're wrong. hopefully at least. the script is meant to measure how much test coverage a pr adds, and given you add code without adding tests, the pr decreases coverage

ntodorova:
ok, than you!

PR #218, opened 2015-09-07
7.0 add web text widget

ntodorova:
Add new 'text_limited' widget for TextField. You can change default values by context varibles 'limit_lines' and 'limit_chars'.


ntodorova:
@rami-wafaie would you please review this?

dreispt:
Could this instead extend the text widget to add it those options?

ntodorova:
Hi @dreispt, sorry for the delay. I changed the module according to your comment. :)
Can you review, please :)

web_text_widget/__openerp__.py
```diff
@@ -0,0 +1,61 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Web text limited widget',
+    'version': '0.1.0',
```
dreispt: 7.0.1.0.0

web_text_widget/__openerp__.py
```diff
@@ -0,0 +1,61 @@
...
+    'author': 'initOS GmbH & Co. KG',
+    'category': '',
```
dreispt: Maybe we should have something here...

web_text_widget/__openerp__.py
```diff
@@ -0,0 +1,61 @@
...
+    'description': """
+
+* Add new functionality for TextField, but data are limited to
+10 lines or 500 characters (by default).
+You can change default values by context varibles 'limit_lines' and
+'limit_chars'.
+If data contains more characters or lines, it will be cut.
+Example of usage:
+<field name="some_text_field"
+       widget="text_limited"
```
dreispt: This is outdated, right?

web_text_widget/__openerp__.py
```diff
@@ -0,0 +1,61 @@
...
+#
+##############################################################################
+{
+    'name': 'Web text limited widget',
+    'version': '0.1.0',
+    'author': 'initOS GmbH & Co. KG',
+    'category': '',
+    'description': """
+
+* Add new functionality for TextField, but data are limited to
```
dreispt: Probably it applies only to Text fields and not to Char fields.
It would be nice - and at least have it in a roadmap - to also support `limit_chars`on Char fields.

web_text_widget/__openerp__.py
```diff
@@ -0,0 +1,61 @@
...
+10 lines or 500 characters (by default).
+You can change default values by context varibles 'limit_lines' and
+'limit_chars'.
```
dreispt: HTML `maxlength`. Why not follow that and use `maxlength` and `maxlines`?

web_text_widget/static/src/js/text_limited.js
```diff
@@ -0,0 +1,56 @@
...
+var QWeb = instance.web.qweb;
+var _t = instance.web._t;
+
+
+instance.web.form.FieldText.include({
+    template: 'FieldText',
+    LIMIT_LINES_CONTEXT_KEY: 'limit_lines',
+    LIMIT_LINES_DEFAULT: 10,
+    LIMIT_CHARS_CONTEXT_KEY: 'limit_chars',
+    LIMIT_CHARS_DEFAULT: 500,
```
dreispt: Extending the standard widget, these default should be 0, meaning for the code below that no limit is to be applied.

ntodorova:
Hi @dreispt, sorry for the delay. I changed the module according to your new comment. :)
Can you review again, please :)

dreispt:
:+1: (no test)

rami-wafaie:
:+1:

rami-wafaie:
:+1:

legalsylvain:
Hi, 

Thanks for your contribution. LGTM. Just a question : why did you use context to set options. Why not define directly new params. (like here : https://github.com/OCA/web/blob/9.0/web_widget_x2many_2d_matrix/README.rst)
In your sample : 

```
field name="some_text_field" maxlines="8" maxlength"400"
```

Kind regards.

ntodorova:
There is no concrete reason for my choice, but I don`t think that it makes big difference. Isn't it?
Of course I can change it if you think it is necessary :)))

legalsylvain:
Thanks for the explanation.

legalsylvain:
Just a question : as it's now not a new widget, what about to rename the module by : 
web_widget_fieldtext_options
like other modules :
(https://github.com/OCA/web/tree/9.0/web_widget_datepicker_options)

ntodorova:
Hi @legalsylvain, sorry for the delay. I changed the name as you suggested.

legalsylvain:
thanks !

Issue #193, opened 2015-08-03
help_online not working in Chrome


amonang:
Tried the help_online module today but it did not work on Chrome (Latest - Version 44.0.2403.125 m). Opened up the same site on Firefox and it worked. Is this a known issue? Are there workarounds?

amonang:
Odoo version is 8.

pedrobaeza:
I have tried it on runbot and the module is working fine in both browsers, so it should be something on your side. Please try to reproduce the problem in runbot.odoo-community.org.

I close this for now.

amonang:
tested the module on runbot although the module installed the button generated is not clickable on Chrome. the tool tip is also not showing ("Create Help page for...")

nhomar:
2015-08-03 23:05 GMT-05:00 amonang notifications@github.com:

> tested the module on runbot although the module installed the button
> generated is not clickable on Chrome. the tool tip is also not showing
> ("Create Help page for...")

If you share a little vide, trying it in runbot (not locally, runbot) it
will be more easy to help you @amonang

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

dreispt:
I confirm the issue. It works with Firefox and Edge.
But tried with Chrome (v 44), and the help button does nothing.

pedrobaeza:
Uhm, it seems then a regression introduced in a new version of Chrome, because in my version (v38) it works.

amonang:
Was trying this again and I hope this info will help.
I did these steps and made the button clickable.
1. On chrome, Right click then inspect element on the button.
2. Drill through until the <a> tag
3. Right click then edit as html
4. Place a letter in between the <a>tags
5. Hover on botton shows the tool tip.
6. Click on button shows the add page dialog box

Although this method adds another icon on the button.

Unfortunately ran out of time today to play around with the js scripts for something more permanent.

amonang:
With the tags I mean the < a > tag

amonang:
solved this. but I don't know how to send solutions via github.
in the css file help_online.css change display from block to inline-block

``` diff
.openerp .oe_view_manager .oe_view_manager_switch .oe_list_button_help_online:after {
  font-size: 28px;
  content: "?";
  text-align: center;
  margin: 3px auto 4px;
  position: relative;
-display: block;
[...]

PR #219, opened 2015-09-08
[FIX] Issue #193. Display help button as inline-block

lmignon:
Tested with FF 40.0.3 and Chromium 44.0.2403.89 Ubuntu 15.04 (64-bit)


pedrobaeza:
:+1:

Cedric-Pigeon:
:+1:

lmignon:
This one can be merged...

PR #221, opened 2015-09-09
[ADD] web_download_disable_iframe

hbrunn:
Made for a customer with a very old firefox ESR where the original strangely enough goes wrong sometimes, and works other times


web_download_disable_iframe/README.rst
```diff
@@ -0,0 +1,52 @@
...
+window, the way it's done for iOS browsers in the standard code.
+
+Usage
+=====
+
+* go to ...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+    :alt: Try me on Runbot
+    :target: https://runbot.odoo-community.org/runbot/162/8.0
+
```
moylop260, line 18: Can you add: how to usage?

hbrunn, line 18: there's not really anything to write about how to use it. You click a download link, that's it

moylop260, line 18: Sorry, I tried in runbot but I don't see any change.

pedrobaeza, line 18: Remember that this button is only valid when the PR is merged, not now. For reviewing the change, you should go to the runbot on the CIs part of the PR.

moylop260, line 18: Yes, I reviewed in PR runbot build.

pedrobaeza, line 18: The button is on the README (just in the line you're commenting). That's why I thought you were trying the link in the button.

moylop260, line 18: Do I need test it with a particular browser?

moylop260, line 18: My test was: https://youtu.be/zoN6KwOWZAQ
What is wrong?
I downloaded a file with runbot-build of 8.0 stable version and downloaded a file with runbot-build of this PR.

hbrunn, line 18: You won't ever see any difference. That's the whole trick of this module. It simple causes old browsers (in my case a firefox esr from an elderly centos) to actually download the file properly

moylop260, line 18: @hbrunn 
Good I got it
Thank you
Can you add a link to a similar video like as mine in README?
- Without this change
- With this change.

hbrunn, line 18: no, too much work involved for no benefit as far as I can see

moylop260, line 18: images?

OSguard:
:+1: test and review

hbrunn:
upstream code was changed and supports ff esr out of the box, so this module is not necessary any more

PR #237, opened 2015-10-09
[WIP] 8.0 web list html widget

damdam-s:


web_list_html_widget/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "HTML widget for list/tree views",
+    "version": "0.1",
```
yvaucher: Put full version here: `8.0.0.1`

web_list_html_widget/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+    "depends": ["web"],
+    'author': "initOS GmbH & Co. KG, Odoo Community Association (OCA), "
+              "Camptocamp",
+    "category": "",
```
yvaucher: No category to provide?

web_list_html_widget/README.rst
```diff
@@ -0,0 +1,45 @@
+Widget Html rendering in list views
```
yvaucher: Missing the AGPL logo on top

web_list_html_widget/README.rst
```diff
@@ -0,0 +1,45 @@
...
+===================================
+
+This module allows to use a ``widget="html"`` for fields in list(tree) views.
+
+*WARNING* The content of the field will be rendered without HTML escaping and without any other security measure.
```
gurneyalex, line 11: WAT?

web_list_html_widget/README.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Features
+========
+
+* Add widget 'html' in a Html field on a tree view
+
+
```
yvaucher, line 21: Missing the try on runbot somewhere here, have a look at the template:

https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

web_list_html_widget/__openerp__.py
pedrobaeza: Please bump it to 8.0.1.0.0 as this is the first 8.0 version

web_list_html_widget/README.rst
damdam-s, line 11: Actually, I didn't succeeded in breaking Odoo security with this widget. Maybe someone can confirm/invalidate this ?
FYI, I made some tests from this site : https://www.owasp.org/index.php/Testing_for_Cross_site_scripting

damdam-s:
up ? anyone ?

web_list_html_widget/i18n/web_list_html_widget.pot
```diff
@@ -0,0 +1,16 @@
+# Translation of OpenERP Server.
```
pedrobaeza, line 1: Remove this file, as it's not needed anymore

web_list_html_widget/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza, line 1: Please shorten headers as in https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

pedrobaeza:
To be consistent with other names, please call it _web_widget_html_treeview_

damdam-s:
@pedrobaeza ok for the two first comments.
For the last, as it's a migration from v8 do we really need (have to) to change the module name ?
No problem if I have to change it but more opinions on this are welcome

web_list_html_widget/README.rst
dreispt, line 11: HTML fields support a `sanitize` attribute. You could recommend (or enforce?) to use only HTML fields with the `sanitize=True` attribute set.

yvaucher:
I agree with @pedrobaeza 
It would be more consistent with other module names:

```
web_widget_color 
web_widget_datepicker_options   
web_widget_float_formula    
web_widget_many2many_tags_multi_selection   
web_widget_text_markdown    
web_widget_x2many_2d_matrix
```

Plus it has no model definition that could make a migration from one to the other really tricky.

pedrobaeza:
There are still comments unattended and the module renaming. Are you going to finish this?

jcdrubay:
It seems like this is a duplicate with #902

PR #238, opened 2015-10-09
[web_tree_image]: Add support for image resize

LeartS:
Add support for server-side resizing of the images before they are sent to the web client tree view.

You may want to do this for:
- Saving data: no need to have the server send a 3200x1600 1,5MB image if you just need it to display a thumb of it in a 32 pixels height row of a tree view.
- Uniform images: no more rows with very different image widths! The web resizers can resize all images to any dimension (max 500x500) while keeping aspect ratio by centering the image and filling it with transparent background.

### DIFF
--- a/web_tree_image/README.rst
+++ b/web_tree_image/README.rst
@@ -1,9 +1,13 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+=====================================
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Interesting feature! :+1: (code review)

yvaucher:
:+1:

oscarolar:
:+1:

dreispt:
:+1:

PR #243, opened 2015-10-20
[FIX] count needactions from the same model only once

hbrunn:
and a small improvement to reload needactions when messages are read


pedrobaeza:
:+1:

lfreeke:
:+1: (test)

PR #244, opened 2015-10-20
[ADD] web_listview_show_advanced_search

jbaudoux:
This module allows to always show by default the advanced search on list view (like it is with graph view). No need to click on the arrow to show it.


pedrobaeza:
Thanks for the contribution! Why not doing the same in the kanban view?

jbaudoux:
@pedrobaeza I prefer having the same functionality for kanban in a separate module.

pedrobaeza:
In my opinion, both options should be together, but let's see what others say.

rafaelbn:
@jbaudoux why in separate module ?

jbaudoux:
Because not everybody want this feature on both listview and kanban.
From my experience, users are used to go on the listview to view and search data and they like to see the pre-defined filters and group by. Kanban is also used as a dashboard or control panel, for instance the "sales teams" in sales or "all operations" in warehouse, where you don't want the advanced search open. And for the kanban for customers, opportunities, products, from my experience, most users prefer to have the listview by default.
Having separate modules offer the flexibility of choosing where you want the feature to apply.

pedrobaeza:
Another option: have both on the same module, but selectable with configuration checks. This way, you reuse code, but letting the user to decide.

dreispt:
@pedrobaeza IMO it's easier to implement as two modules, unless @jbaudoux feels that one module is the way to go and renames the module to fit that.

jbaudoux:
Would be nice to get that PR merged and let other contributors improve it further if they want to...

luc-demeyer:
The list view is the main use case for this functionality, but I think it would be better to make the expansion configurable per view.
A suggestion that that is easy to implement is to make the expansion depend on the action context :
e.g. {'show_advanced_search': 1}
Replace the check in the JavaScript code (if e === 'list')  by 
(e === 'list' && this.dataset.context['show_advanced_search'])
and it's configurable.

As a future extension (but a bit more work), we could add a config parameter to determine the default behaviour per type of view (e.g. default advanced search tree) but that's imho not a priority. My preference goes to a default hidden advanced search and per list view configurable advanced search.

petrus-v:
Difficult question, I think it's to much related with the context of the view, so it depends!

What I would maybe prefer is like in odoo 9, if user expend search view somewhere it keep it expanded! (I know expand view in odoo9 is completely different of odoo8)

My 2 cts!

angelmoya:
It would be great if the advanced search were expanded on list views related with:
- field many2many on **Add an item**
- field many2one on **Search more...**

I also think search view is more related with list view than kanban view.

I think this module works great, and it could be merged yet. :+1:

hbrunn:
:+1: merging as all comments here are actually feature requests. My personal one would also be to have this steered by a context key for all kinds of views (`advanced_search_$viewtype_open` or the like)

PR #248, opened 2015-10-29
[ADD] web_developer_mode_optional

alejandrosantana:
Added web_developer_mode_optional module to allow access to developer mode, based on user groups.
Now this allows to show 'developer mode' link only to those users who are in a certain group.

Consider this as a clean substitute for PR https://github.com/OCA/web/pull/128


web_developer_mode_optional/static/src/js/chrome.js
```diff
@@ -0,0 +1,43 @@
+// License, author and contributors information in:
+// __openerp__.py file at the root folder of this module.
+
+/*---------------------------------------------------------
+ * OpenERP Web chrome
+ *---------------------------------------------------------*/
+openerp.web_developer_mode_optional = function () {
+    var instance = openerp;
```
hbrunn, line 8: your function is called with the correct reference to `instance`, use that

web_developer_mode_optional/static/src/js/chrome.js
```diff
@@ -0,0 +1,43 @@
...
+    openerp.web.chrome = {};
```
hbrunn, line 9: then you also don't need this, which is potentially harmful

web_developer_mode_optional/static/src/js/chrome.js
```diff
@@ -0,0 +1,43 @@
...
+    var QWeb = instance.web.qweb;
+    var _t = instance.web._t;
+
+    instance.web.UserMenuDevel =  instance.web.UserMenu.include({
+        on_menu_about: function() {
```
hbrunn, line 14: it's a deadly sin to overwrite a function without calling `super`. Refactor in a way such that you don't break modularity

hbrunn, line 14: probably you're better off doing your stuff in an override of the qweb view, and only set you flag in js

PR #252, opened 2015-11-05
[ADD] web_favicon

### LINKED ISSUE #199: Favicon feature
It would be nice to be have a feature allowing to set the favicon for an Odoo instance or, if possible, depending on the active Company.

Can experts comment if this is feasible? Given some guidelines I could try a PR.

hbrunn:
closes #199


web_favicon/controllers/web_favicon.py
```diff
@@ -0,0 +1,43 @@
...
+from openerp import http
+from openerp.tools.misc import file_open
+
+
+class WebFavicon(http.Controller):
+
+    @http.route('/web_favicon/favicon', type='http', auth="none")
+    def icon(self):
+        request = http.request
+        company = request.env['res.company'].search([], limit=1)
```
dreispt: This could be in the `else` part of the following `if`

web_favicon/controllers/web_favicon.py
```diff
@@ -0,0 +1,43 @@
...
+        if 'uid' in request.env.context:
+            user = request.env['res.users'].browse(request.env.context['uid'])
+            company = user.sudo(user.id).company_id
+        favicon = company.favicon_backend
+        favicon_mimetype = company.favicon_backend_mimetype
+        if not favicon:
+            favicon = file_open('web/static/src/img/favicon.ico')
+            favicon_mimetype = 'image/x-icon'
+        else:
+            favicon = StringIO.StringIO(base64.b64decode(favicon))
```
dreispt, line 27: In this case, the mimetype could be none. Is that OK?

web_favicon/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+from openerp import models, fields
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    favicon_backend = fields.Binary()
+    favicon_backend_mimetype = fields.Char(
+        help='Set the mimetype of your file. Usually, that\'s one of\n'
+        'image/x-icon, image/gif, image/png')
```
dreispt: Would if make sense to be a Selection field?

web_favicon/controllers/web_favicon.py
```diff
@@ -0,0 +1,43 @@
...
+##############################################################################
+import StringIO
+import base64
+from openerp import http
+from openerp.tools.misc import file_open
+
+
+class WebFavicon(http.Controller):
+
+    @http.route('/web_favicon/favicon', type='http', auth="none")
```
dreispt, line 13: Not that I'm knowledgeable in the subject, but shouldn't the route be just `/favicon` ?

web_favicon/models/res_company.py
```diff
@@ -0,0 +1,29 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp import models, fields
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    favicon_backend = fields.Binary()
```
dreispt, line 12: Why call it backend?

web_favicon/tests/test_web_favicon.py
```diff
@@ -0,0 +1,66 @@
...
+        return FakeResponse(data, headers)
+
+
+class FakeResponse(object):
+    def __init__(self, data, headers):
+        self.data = data
+        self.headers = dict(headers)
+
+
+
```
dreispt: PEP8 might complain about these extra lines

web_favicon/views/templates.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <template id="layout" inherit_id="web.layout">
+            <xpath expr="//link[@rel='shortcut icon']" position="replace">
+                <link rel="shortcut icon" href="/web_favicon/favicon" />
```
dreispt: Prefer to use `rel="icon"`([source](https://mathiasbynens.be/notes/rel-shortcut-icon))

web_favicon/views/templates.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
```
dreispt, line 3: Nitpick: maybe use the `<odoo>` tag and drop the `<data>` tag :-)

dreispt:
Thank you Holger, this is a great feature.
There  are some comments to take care of and make mr.Linter happy, but other than that LGTM.

web_favicon/README.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Note that most browsers cache favicons basically forever, so if you want your
+icon to show up, you'll most probably have to delete you browser cache.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+    :alt: Try me on Runbot
+    :target: https://runbot.odoo-community.org/runbot/{repo_id}/8.0
```
eLBati: Please replace {repo_id}

eLBati:
Thanks @hbrunn 
Tested on runbot
:+1: after the other remarks are solved

web_favicon/controllers/web_favicon.py
hbrunn, line 27: The mimetype is what the user fills in, and the view forces her to fill in something. Without mimetype, the browser has to guess which probably some obscure browsers do wrong. But if the mimetype is null here, it's a user error.
Of course we could guess the mimetype here, but I want to avoid the extra dependency

web_favicon/models/res_company.py
hbrunn, line 12: because the website favicon needs to be done in another module, and then I think it would be better to put it on the website model

web_favicon/controllers/web_favicon.py
hbrunn, line 13: i wouldn't know of any conventions here. Having everything live under $modulename seems a good choice in my opinion

web_favicon/models/res_company.py
hbrunn: we don't know which formats will be used tomorrow, so I don't think so

web_favicon/views/templates.xml
hbrunn, line 3: we're in 8.0

hbrunn:
@dreispt @eLBati thanks for your reviews, I took most of your points

web_favicon/models/res_company.py
dreispt: IMO the list you already provide is good enough, it it would be easy to extend it in case something more is needed.

web_favicon/controllers/web_favicon.py
dreispt, line 13: ok

web_favicon/models/res_company.py
dreispt, line 12: ok

dreispt:
It's :+1: after Travis is green.

hbrunn:
@dreispt thanks, please remove the needs fixing label then when appropriate

eLBati:
:+1:

web_favicon/models/res_company.py
pedrobaeza: I also prefer to provide a close list of possible values to guide the user what to choose/try. If in the future another format appears, then we can extend this list easily.

web_favicon/README.rst
```diff
@@ -0,0 +1,63 @@
...
+
+This module was written to allow you to customize your Odoo instance's shortcut
+icon (aka favicon). This is useful for branding purposes, but also for
+integrators who have many different Odoo instances running and need to see at a
+glance which browser tab does what.
+
+Configuration
+=============
+
+Upload your favicon (16x16, 32x32 or 64x64 pixels, 16 colors) on the company
```
pedrobaeza: A bigger resolution than 64x64 can be also used (and it's advisable for Android or iOS webapps for example). Also 16 colors is an older specification. Now the range color can be wider. Some reference: http://stackoverflow.com/questions/2268204/favicon-dimensions

pedrobaeza:
I was making something similar for website icon, but being static, so this one is very interesting, but your approach is a bit limited, because it doesn't take into account new specifications for declaring webapps icon for iOS and Android (which can be very useful, specially with responsive UI). Look at my module that already declared this: https://github.com/serviciosbaeza/serviciosbaeza-odoo-addons/blob/8.0/website_favicon/views/website_favicon.xml. You can also catch the nice icon I made: https://github.com/serviciosbaeza/serviciosbaeza-odoo-addons/blob/8.0/website_favicon/static/description/icon.png

I'm thinking of having multiple fields, one for each size, and the only required is the higher resolution one. Then, to have compute fields that catch the current resolution one if provided, and if not, a resized image to the proper size from the higher resolution image. All of this used in the links in the XML.

What do you think? Can you do it? May I propose the change to your branch?

hbrunn:
@pedrobaeza propose it, I'll be happy to merge it

dreispt:
@pedrobaeza AFAICT the icon is cached and the "high" resolution is not that high. Maybe a single field would be just fine.

pedrobaeza:
@dreispt, you're thinking only in traditional browsers, but there are more applications.

hbrunn:
@pedrobaeza is it a good idea to block this until we have the maximal version? Can't we merge now and add the fancy stuff whenever you find time to do it?

dreispt:
@hbrunn Put it in the Roadmap section of the README, as an idea for possible contributors..

hbrunn:
@dreispt done

dreispt:
I think this is good to merge. @pedrobaeza ?

pedrobaeza:
No, please let me work on the points I have told.

dreispt:
Is there a problem on that going in a separate PR?

pedrobaeza:
Yes, the need for a migration script.

hbrunn:
@pedrobaeza please either add the feature you want, or remove the work in progress tag. It's been two months now

hbrunn:
I think this can be merged now, it doesn't seem @pedrobaeza will add the feature he want any time soon

dreispt:
@pedrobaeza can we merge?

pedrobaeza:
I'm preparing a quick PR with some of the changes we told (not all), and I'll propose it to @hbrunn's branch.

pedrobaeza:
You have in https://github.com/hbrunn/web/pull/3 my little improvements. When I have time, I'll work in the rest of the roadmap.

hbrunn:
@pedrobaeza @dreispt merged

pedrobaeza:
Travis is fine, so I merge. Thanks to all!

PR #257, opened 2015-11-16
[FIX] web_m2x_options check Many2one and Many2many options

antespi:
This fix these issues:
- Do not overwrite options defined in view
- Read `no_create`, `no_create_edit` and `no_quick_create` options, instead of  `create`
- Replace tabs by spaces for cleaner alignment.

For testing propuse in runbot, you can do this:
- Set  `web_m2x_options.create` system parameter as False
- Go to any view and edit it setting any Many2one field adding options attribute as `options="{'no_create': True}"`

In this case, addon will remove 'Create...' (quick create) option, but view is removing both (quick create and create edit). With this fix addon is taking care of this.

I've done this in `res.partner.title` field in runbot

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -67,11 +67,11 @@ openerp.web_m2x_options = function (instance) {
             var self = this;
             // add options limit used to change number of selections record
             // returned.
-	    if (_.isUndefined(this.view))
-                return this._super.apply(this, arguments);
-            if (!_.isUndefined(this.view.ir_options['web_m2x_options.limit'])) {
-		    this.limit = parseInt(this.view.ir_options['web_m2x_options.limit']);
-	    }
+            if (_.isUndefined(this.view))
+                    return this._super.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


pedrobaeza:
In this case, is cleaner to put 2 commits: one with the fix itself, and another with the tab cleaning

antespi:
@pedrobaeza, but is the fix correct in your opinion?

pedrobaeza:
I don't know, because I haven't distinguished it in my first sight (that's why I told you to make that). Anyway, my knowledge in JS is still limited.

rafaelbn:
Tested in a customer devel environment. :+1:

JavierIniesta:
It shows an error message when I try to select a Partner Tag.

selection_006

antespi:
@JavierIniesta JS error fixed, thanks!

max3903:
Tested on runbot: http://3119835-258-e4e0b7.runbot2.odoo-community.org/
:+1:

JavierIniesta:
Tested in local environment :+1:

yvaucher:
tested on runbot :+1: 

And merging

PR #259, opened 2015-11-17
Web sheet full width migrate 90

njeudy:
- Migrate to V9
- Add README.rst with OCA infos
- Test for community and entreprise version.

### DIFF
--- a/web_sheet_full_width/README.rst
+++ b/web_sheet_full_width/README.rst
@@ -9,33 +9,12 @@ web_sheet_full_width
 This module was written to extend the functionality of web to support full width
 for form view
 
-Installation
-============
-
-To install this module, you need to:
-
-* Just click install button.
-
-Configuration
-=============
-
-To configure this module, you need to:
-
-* N/A
-
 Usage
 =====
-
-To use this module, you need to:
-
-* Go to ...
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/162/9.0
-
-.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
-.. branch is "8.0" for example
 
 Known issues / Roadmap
 ======================
@@ -66,7 +45,7 @@ Contributors
 ------------
 
 * Holger Brunn <hbrunn@therp.nl>
-* Nicolas JEUDY <nicolas@sudokeys.com>
+* Nicolas JEUDY - Sudokeys (https://github.com/njeudy)
 * Stephane (SOLIBRE) <stephane@omerp.net>
 
 Maintainer
--- a/web_sheet_full_width/README.rst
+++ b/web_sheet_full_width/README.rst
@@ -1,0 +1,85 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+web_sheet_full_width
+====================
+
+This module was written to extend the functionality of web to support full width
+for form view
+
+Installation
+============
+
+To install this module, you need to:
+
+* Just click install button.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* N/A
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to ...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/9.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
+
+Known issues / Roadmap
+======================
+
+* N/A
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/web/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+web/issues/new?body=module:%20
+web_sheet_full_width%0Aversion:%20
+9.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Holger Brunn <hbrunn@therp.nl>
+* Nicolas JEUDY <nicolas@sudokeys.com>
+* Stephane (SOLIBRE) <stephane@omerp.net>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/web_sheet_full_width/README.rst
+++ b/web_sheet_full_width/README.rst
@@ -9,6 +9,8 @@ web_sheet_full_width
 This module was written to extend the functionality of web to support full width
 for form view
 
+Note: Due to licence incompatibility you have not legal right to install this module on enterprise version.
+
 Usage
 =====
 
--- a/web_sheet_full_width/__openerp__.py
+++ b/web_sheet_full_width/__openerp__.py
@@ -20,8 +20,8 @@
 ##############################################################################
 {
     "name": "Show sheets with full width",
-    "version": "8.0.1.0.0",
-    "author": "Therp BV,Odoo Community Association (OCA)",
+    "version": "9.0.1.0.0",
+    "author": "Therp BV,Sudokeys,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "summary": "Use the whole available screen width when displaying sheets",
     "description": """
@@ -43,9 +43,6 @@ Icon courtesy of http://www.picol.org/ (size_width.svg)
         "view/qweb.xml",
     ],
     "auto_install": False,
-    'installable': False,
+    'installable': True,
     "application": False,
-    "external_dependencies": {
-        'python': [],
-    },
 }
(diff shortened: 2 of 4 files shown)


web_sheet_full_width/README.rst
```diff
@@ -0,0 +1,36 @@
+Show sheet over full width
```
pedrobaeza: Please use latest OCA's template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

web_sheet_full_width/README.rst
```diff
@@ -0,0 +1,36 @@
...
+==============================
```
pedrobaeza: This should have the same length of the title

pedrobaeza:
And please squash your commits in one, as there are little changes.

pedrobaeza:
Tested on runbot, it works OK.

web_sheet_full_width/__openerp__.py
```diff
@@ -43,7 +43,7 @@
         "view/qweb.xml",
     ],
     "auto_install": False,
-    'installable': False,
+    'installable': True,
     "application": False,
     "external_dependencies": {
```
elicoidal: Please remove empty keys

njeudy:
OK thanks for the review. I updated all I think.

lmignon:
@pedrobaeza @njeudy We should advertise that the module can only be installed on the community edition (the module web in the enterprise edition is a non lgpl fork of the module web in the community edition not an extend)

rvalyi:
+1 @lmignon you are absolutely right

pedrobaeza:
I'm very worried about this thing, because it means that everyone with the enterprise edition cannot install any of the OCA/web modules, isn't it? Cannot we make an enterprise version of the module with a compatible license?

rvalyi:
@pedrobaeza if so I suggest we sell them on the app-store at the profit of the OCA. My 2 cts.

rvalyi:
@pedrobaeza from what we hear, it's not like so many on-premise users choose the Enterprise proprietary version; eventually that offer will simply die just like the certification program or the SAP indexed prices, may be not worth harming OCA licenses for it. And as for Odoo saas users they cannot install community modules anyway so...

pedrobaeza:
Hehe, I knew your answer when I was writing my comment :stuck_out_tongue_winking_eye:, but this is very serious and it's a clear example of incompatibility: as Odoo has decided to release a separate enterprise web module instead of mounting over the community web module, this makes in practical terms incompatible with a lot of OCA modules, and for me this means that Odoo has dug its own grave.

rvalyi:
@pedrobaeza it's not like we never warned them...

lmignon:
@pedrobaeza I don't know how it will be possible to migrate OCA/geospatial to 9.0...

pedrobaeza:
Yeah, I have also read about that, but it can be migrated for Community version, isn't it?

lmignon:
@pedrobaeza yes but not run on enterprise at all.

pedrobaeza:
This is really worrying... The rest of the @OCA/board should be aware of this.

rvalyi:
This only my opinion here. But I really don't believe the current Enterprise offer where partners can install some proprietary modules on-premise is going anywhere. We might be worried or not (I' m not), but I don't believe that deserves changing our OCA licensing. Anyway this outcome was entirely predictable for years now...

njeudy:
@lmignon @rvalyi @pedrobaeza: this module is only css. can you explain me with it could not be installed on entreprise version ? I juste add some css rules to odoo et doesn't change any part of the code...

njeudy:
We have to change licence to LGPL for web modules ?

rvalyi:
@njeudy in this specific case where it's only css, then may be you may re-license it to some more permissive license to make things clear (I wasn't aware it was only css).
But still this is a serious issue for modules extending the web module in general.

max3903:
@njeudy If you want to allow Enterprise users to use your module, change the license of your module to LGPL. Otherwise, use AGPL.

rvalyi:
@max3903 @njeudy @pedrobaeza
in fact I'm not sure LGPL can derivate from something proprietary.
Does somebody have any clue about that?

pedrobaeza:
Seeing the post about licenses: https://odoo-community.org/blog/oca-news-1/post/oca-odoo-meeting-on-licenses-21, we cannot put an AGPL module over the propietary, but we can put a LGPL. The question is if the authors want to relicense the module to LGPL.

rvalyi:
@pedrobaeza it's written by Odoo SA indeed. Now is there any other source about LGPL telling that? It's just that in the past SA has been so approximative with licensing (see proprietary modules sold with AGPL v8...) that I don't really take everything they say for granted...

pedrobaeza:
Yeah, you're right about the uniqueness of the source. Anyway, I can't imagine an scenario where putting as LGPL can cause any trouble. Do you?

rvalyi:
@pedrobaeza I'm sure LGPL can be extended by proprietary, I'm just not sure about the reverse. That's probably possible, but clearing doubts would be great.

njeudy:
but I ported a module developed by THERP .. so I need to ask him for LGPL ?

pedrobaeza:
Yeah. As now, we have the security that it's incompatible with enterprise edition. Converting to LGPL, @rvalyi expresses also his doubts, but there are chances that it is compatible. You have to ask permission to the authors of the module, not the company.

lmignon:
@njeudy Do you know if your module works on entreprise? If it doesn't work, you just need to advertise that the module doesn't work on enterprise and can't be installed on enterprise. No licence change required in this case.

rvalyi:
@lmignon the Enterprise does not work. How can a community module work on it?
ok that one was free bashing.
-------> []

njeudy:
@lmignon yes it works on entreprise :)

pedrobaeza:
Let's query @hbrunn what he thinks about this and if he wants to switch to LGPL.

hbrunn:
I believe in viral licenses for the time being, so no. But wasn't part of the evil license discussion that OCA modules grant an exception for lgpl core modules?

guewen:
@hbrunn the exception states that an OCA AGPLv3 add-on can be installed aside an Enterprise proprietary add-on but not depend on it.

hbrunn:
thanks for the clarification

jgrandguillaume:
Hi,

I confirm what has been stated here :https://odoo-community.org/blog/oca-news-1/post/oca-odoo-meeting-on-licenses-21
We made our best that time to protect the interest of our contributors.

So, in brief:
-  an OCA AGPLv3 add-on can be installed aside an Enterprise proprietary add-on but not depend on it
- of course he contrary is true too (AGPL cannot extend Enterprise proprietary)
- to extend the Enterprise proprietary module with an OCA one, you need LGPL (or MIT eventually)
- to change the license, you need the agreement of all the authors. The OCA will not use its copyright granted through the CLA for this. An exception may be made once, but the assembly would have to vote.

Regards,

Joël

@lmignon For the Geospatial addons, we're working on it. It'll be available on community version.

pedrobaeza:
OK, so let's add for now the sentence "This module can only be installed on the community edition due to license incompatibilities" and we can continue.

elicoidal:
+1

dreispt:
> the module web in the enterprise edition is a non lgpl fork of the module web in the community edition not an extend

This is a mess - someone didn't do their homework (again). They only had to create a `web_enterprise`overriding stuff from the community `web` and all would be good!

As is causes for a lot of OCA modules to be incompatible with Enterprise.
Not only all OCA/web depend on `web`, but also modules in other repos extend on `web`, such as some timesheet modules.

Changing OCA modules licenses won't fix the root of problem. Odoo must fix doing a proper implementation of the enterprise features, that stays true to what they promised to the community.

elicoidal:
I propose that the OCA board prepare an official request to Odoo, sa.
I fear though that they wont commit such an extensive change for v9 and target might be v10 only.
An additional decision is to be made about the attitude of the OCA for v9.

hbrunn:
how about OCA providing an LGPL module that depends on `web` (enterprise version) and is called `web_enterprise`? This way, we have a marker what is targetted to the community version and what to the enterprise version. And if Odoo SA agrees with the rename, we simply remove this module again.

hbrunn:
PS: Shouldn't we move this to the mailinglist instead of hijacking the PR?

lmignon:
@hbrunn :+1:  for the mailinglist

elicoidal:
+1 for the module
+1 for bringing more light to this discussion

web_sheet_full_width/README.rst
```diff
@@ -0,0 +1,85 @@
...
+for form view
+
+Installation
+============
+
+To install this module, you need to:
+
+* Just click install button.
+
+Configuration
```
lmignon: You can remove this section since  it's useless

web_sheet_full_width/README.rst
```diff
@@ -0,0 +1,85 @@
...
+To configure this module, you need to:
+
+* N/A
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to ...
```
lmignon: These 2 previous sentences can be removed

web_sheet_full_width/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/9.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
```
lmignon: These commented lines can be removed.

OSguard:
I think for OCA it is impossible to maintain module that extend enterprise stuff. The reason is the enterprise license itself. We as partner has access to the enterprise code, but not allow to modified or distribute outside the company. So for modules depends on enterprise we can not use travis, runbot and so on to make sure that our OCA module works with the enterprise version.
As it is not possible to have our toolset work with enterprise stuff, we can not assure the OCA quality. So then we should simply don't do it.

rvalyi:
+1 OSguard
Not to say that if the OCA supports these proprietary modules, then we will have them around for more time. I would say the sooner they are discontinued or kicked out of the market, the better for us.

pedrobaeza:
@rvalyi, OCA cannot and shouldn't make boycott of any of the Odoo ways of financing while they still keeps a little commitment with the community. We only must guarantee the rights of our contributors and check for possible illegal actions.

rvalyi:
@pedrobaeza if financing means promising to VC's they'll close the thing and kill the community eco-system subtlety years after years in exchange of their money, then I believe the sooner we get out of this scheme, the saner for everybody. Sorry I hijacked the thread again, I wasn't sure this point would surface again.

jgrandguillaume:
+1 for the web_enterprise module
+1 for bringing this discussion on ML

For runbot and travis, we can ask Odoo sa to grant the OCA bot user an access to the enterprise repo to allow us to make the build on travis and runbot.

njeudy:
Hello,

What do I need to do so ? wait for OCA decision ?
For now All modules depending on `web` in OCA repos can't be migrate to 9.0 ?

If I have entreprise version it's not legal to use this module ?

sorry for my question but I'm not "licensing system master" I don't know if I should continue migrating web module to V9 :)

njeudy:
@lmignon will apply change soon, thanks

lmignon:
@njeudy 

> If I have entreprise version it's not legal to use this module ?

IMO yes it's not legal to use this module in enterprise

pedrobaeza:
@njeudy, can you please attend the comments and I'll merge?

njeudy:
@pedrobaeza : it's done sorry for the delay :)

njeudy:
@pedrobaeza do you know why this coverage is not ok ?

pedrobaeza:
It was probably a problem in the CIs in that moment. I have finally merged amending the README with latest agreement. Thanks for all your effort!

PR #261, opened 2015-11-17
[i18n] Adding translation files to web modules

PCatinean:


pedrobaeza:
Why are you adding empty translations? And why not using Transifex for this task?

PCatinean:
The main point was to have .po files present as runbot checks give warnings when they are missing

I can only assume now that this is not the proper procedure :)

pedrobaeza:
No, I don't think so :wink:

PR #262, opened 2015-11-19
Update web_m2x_options for v9

niboo-sam:
Update the module web_m2x_options to match the new javascript API.
Tell me what do you think about it.

### DIFF
--- a/web_m2x_options/__openerp__.py
+++ b/web_m2x_options/__openerp__.py
@@ -2,13 +2,12 @@
 
 {
     "name": 'web_m2x_options',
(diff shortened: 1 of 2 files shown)


sbidoul:
@samuellefever why did you close this one?

isabellerichard:
Hi,
I checked this PR, it works well for the part "hide Create if you have no creation access", but it doesn't work if you add `no_create` inside options of a field. Options `no_create` and `no_create_edit` are ignored.

isabellerichard:
Thanks @zakiuu, your fix works well :-)

PR #263, opened 2015-11-20
[FIX] disjunctions in advanced search

### LINKED ISSUE #222: [8.0][BUG] web_advanced_search_x2x - Wrong queries in filters
Hi @hbrunn ,

This happends with any query but I will show here only one.

I would like to filter projects which User (uid) is "manager" OR "member" or "follower".

In Odoo core we cannot make this search because the follower are res.partner and not res.user then we need to select Follower -> Is in selection and then in res.parter search User -> the user.

When Odoo creates the filter for this you can only select the res.partner and this is the query:

```
 ['|',
 '|', ['message_follower_ids', '=', 748], ['members', '=', 52],
 ['user_id', '=', 52],
 ['state', '!=', 'close'], ['state', '!=', 'cancelled']]
```

When I create the query using advance search it make this wrong:

```
 [['message_follower_ids.id', 'in', [748]], ['members', '=', 52], ['user_id', '=', 52],
 ['state', '!=', 'close'], ['state', '!=', 'cancelled']]
```

But it must create this:

```
 ['|',
 '|', ['message_follower_ids.user_ids', '=', 748], ['members', '=', 52],
 ['user_id', '=', 52],
 ['state', '!=', 'close'], ['state', '!=', 'cancelled']]
```

Regards

hbrunn:
fixes #222

### DIFF
--- a/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
+++ b/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
@@ -283,6 +283,42 @@ openerp.web_advanced_search_x2x = function(instance)
         },
     });
 
+    normalize_domain = function(domain)
+    {
+        // this is the js version of https://github.com/odoo/odoo/blob/8.0/openerp/osv/expression.py#L203
+        if(!domain.length)
+        {
+            return [(1, '=', 1)]
+        }
+        result = []
+        expected = 1
+        op_arity = {'!': 1, '&': 2, '|': 2};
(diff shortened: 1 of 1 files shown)


rafaelbn:
@pedrobaeza @antespi we need a review here. I'm rebuilding runbot for testing.

rafaelbn:
Tested un runbot :+1:  this is ready to be merged

cc @pedrobaeza @antespi @Tecnativa

pedrobaeza:
OK, merging.

PR #268, opened 2015-12-02
[9.0] web_dialog_size: Do make the dialog window moveable

wpichler:
Please accept this small addition to make the dialog moveable.
Very often the dialog does hide some things in the background you may need - so i think this is a great thing here.


pedrobaeza:
Interesting change, but please made it against 8.0 branch, because the module is not yet tested on v9 (and it's marked as not installable). If you have tried the module in v9 and it's working, please make the corresponding PR activating it.

wpichler:
I did check it against v8.0 - and have done a new push request

PR #271, opened 2015-12-09
[ADD] web_search_autocomplete_prefetch

hbrunn:
# 
# Prefetch autocomplete offers

When searching, the autocomplete options can be a bit frustrating because you
will be offered choices that won't yield a result. This addon searches for the
term in the background and only offers an option if this search has a result.
# Known issues / Roadmap
- some searches (especially via function fields) can be very heavy on the
  server.
  
  To disable prefetching on a per field basis, set the option
  `web_search_autocomplete_prefetch.disable`::
    options="{'web_search_autocomplete_prefetch.disable': true}"
  on your field in the search view.


NL66278:
:+1:  Did look at the code and seems generally OK to me, but I have limited knowledge of the Javascript interface to Odoo frontend. But installed and tested the module and looks like a big improvement for searching in lists.

web_search_autocomplete_prefetch/README.rst
```diff
@@ -0,0 +1,64 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: https

hparfr:
Very interesting and promising.

Tested on runbot in Messagin Todo.
Each time I enter a letter, 6 requests are be triggered => it's a huge impact on perf (client side and server side)

NL66278:
On performance: there should be a delay in autocomplete to check wether the user is still typing. Only when the user stops typing (let us say 0.3 seconds nothing typed) autocomplete should be triggered. This is a general autocomplete problem, nothing todo with the enhancement in this PR.

StefanRijnhart:
This looks like a great usability improvement!

But I think @hparfr could be right. Because prefetching is done on the client side, if you have six default search fields you get six search_count calls from client to server. Would it be possible to aggregate these into a single call? If the performance could be opitimized, it would be interesting to test this globally enabled by default instead of enabled on a field by field basis.

hbrunn:
@StefanRijnhart we could wrap all of this in one call, but the load on the server stays the same, so I'd rather keep it simple. Implementing @NL66278's suggestion about a delay for searching is on my list, with this, the server bleeds a lot less

hbrunn:
PS: this is global, after installing the module, you can _disable_ it for fields which take too much time

ivantodorovich:
:+1: this is awesome

lfreeke:
:+1:  Test

PR #273, opened 2015-12-16
Use clickable icon for image also in form view

ghost:
Add the widget also with name "image_preview" and add a corresponding form view widget that will also show a clickable icon for preview.
The name change to "image_preview" allows the widget to be used in a tree view with inline form view, such that the form view will still show the clickable icon for preview, instead of the default behavior of showing the full size image for the form view "image" widget.

### DIFF
--- a/web_tree_image/static/src/css/web_tree_image.css
+++ b/web_tree_image/static/src/css/web_tree_image.css
@@ -1,0 +1,3 @@
+.o_web_tree_image_clickable {
+    cursor: pointer;
+}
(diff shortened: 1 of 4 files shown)


ecino:
This is a good improvement :+1:

pedrobaeza:
Please add the documentation for the new features on the README file.

dreispt:
I'm not sure how this is used. I tried on runbot to click on a partner image and it did nothing. How can I see this working?

ecino:
@pedrobaeza What this PR does is adding a widget "image_preview" that behaves like "image" except that it doesn't display the full image when entering editing mode (icon or thumbnail instead).

@dreispt To test this, you must have a tree view with a field that has attribute widget="image_preview". I think the module does not add any view so the runbot does not display it.

PR #279, opened 2015-12-23
Web mail send using composition wizard instead of 'mailto:'

codingforfun:
Replace default email widget to open the internal composition wizard to send email instead of providing just a 'mailto:' link. 

Messages sent get appended to the chatter of the parent object they are sent from 

This still has some open issues. You are welcome to propose changes.

Issues:
- The default composition wizard from 'mail' seams to be unable to send mail to just normal email addresses, I only found way to add partner IDs but not to just add plain text email addresses. This is a problem for a general replacement. My current solution is to open another dialog to force to create a partner contact in this case. Previous solution was to just fall back to 'mailto:' in this case.


web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,85 @@
...
+                                additional_context: {
+                                    default_partner_ids: [partner.id],
+                                    default_composition_mode: 'comment',
+                                    // write to active model:
+                                    default_model: fm.dataset._model.name,
+                                    default_res_id: fm.datarecord.id,
+                                },
+                                on_close: function(){
+                                    /* Todo: refresh the chatter widget here
+                                    */
```
OSguard: @hbrunn do you know how we can trigger the function from a other widget. In this case the chatter?

hbrunn: you need to go up in the widget hierarchy (`getParent()` from the top off my head) until you find a form widget, and then recursively search the form's children for the chatter widget. Then call whatever function updates it.

hbrunn:
@codingforfun this can be very useful I think. Would you agree merging your PR to the social project, I feel it's more at home there

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun: thx

codingforfun:
@hbrunn 
I do not have any preference to which project to merge this. Initially I thought to place it here because it's basically a widget. 
I'm open to this if other people don't vote against it / other people vote for it.

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun: @hbrunn 
Is it right to assume that the widget hierarchy is different from the DOM tree? Or is it possible somehow to use jQuery to search the chatter in the DOM and then get the widget from the DOM element like this: `chatter = self.$el.parents('.oe_formview').find('.oe_chatter')` ?
If that is possible, how do I get from the chatter DOM-Node to its JS Widget?

hbrunn: It's correct to assume the widget hierarchy is totally different than the DOM. And there's no link from DOM elements to the widgets that produced them, so you really have to search through the widget hierarchy.

Is there a problem with the approach I suggested?

codingforfun: No. Not basically. I just had to work this out. I'm not yet so familiar with hacking JS in Odoo. Actually I found the solution. Looks scary but works.

hbrunn: js always looks like it's going to explode the next second, that's just something to get used to ;-)

codingforfun: I pushed this, if you like to review. Thx for your help!

Am 27.01.2016 um 16:08 schrieb Holger Brunn:

web_widget_mail_send_odoo/README.rst
```diff
@@ -0,0 +1,57 @@
...
+
+=====================================
+Web Widget - Email internal mail link
+=====================================
+
+This module was written to provide a new widget for Email fields to replace the
+standard behaviour of displaying a 'mailto:' link with JS to open the internal
+Email wizard from Odoo for message composition.
+
+The message you sent gets appended to the chatter as if you where using internal
```
hbrunn: s/sent/send, '...using _the_ internal...'

web_widget_mail_send_odoo/README.rst
```diff
@@ -0,0 +1,57 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
hbrunn: https

web_widget_mail_send_odoo/README.rst
```diff
@@ -0,0 +1,57 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================================
+Web Widget - Email internal mail link
```
hbrunn: reformulate to a sentence what this _does_, as in 'Use internal mail wizard for email links' or the like

web_widget_mail_send_odoo/__openerp__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# © 2015 initOS GmbH (<http://www.initos.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Web Widget - Email internal mail link',
```
hbrunn: if you change the heading in the readme, change it here too

web_widget_mail_send_odoo/__openerp__.py
```diff
@@ -0,0 +1,26 @@
...
+    'version': '8.0.1.0.0',
+    'category': 'Social Network',
+    'website': 'https://odoo-community.org',
+    'author': 'initOS GmbH, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'auto_install': False,
+    'depends': [
+        'web',
```
hbrunn: `mail` depends on `web` already, so this can go

web_widget_mail_send_odoo/assets.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+
+        <template id="assets_backend" name="custom bar assets" inherit_id="web.assets_backend">
```
hbrunn: remove the name attribute or give it a more module specific name

web_widget_mail_send_odoo/static/src/css/web_widget_mail_send_odoo.css
```diff
@@ -0,0 +1 @@
+.openerp .oe_form .oe_form_field_email_intern input,
```
hbrunn: people who don't understand German will be a bit puzzled by _intern_, consider replacing this by _internal_ at all occurences

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
+/* © 2015 initOS GmbH (<http://www.initos.com>)
+*  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+*/
+
+(function() {
```
hbrunn: This runs as soon as the javascript code is loaded, which can lead to timing problems. Use the standard pattern to define a function in the openerp namespace with the addon's name, getting `instance` as parameter: https://github.com/OCA/web/blob/8.0/support_branding/static/src/js/support_branding.js#L16

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+var instance = openerp;
+var _t = instance.web._t,
+   _lt = instance.web._lt;
+var QWeb = instance.web.qweb;
+
+instance.web.form.FieldEmailIntern = instance.web.form.FieldChar.extend({
+    template: 'FieldEmailIntern',
+    initialize_content: function() {
+        this._super();
+        var $button = this.$el.find('a');
```
hbrunn: a shorthand for this is `this.$('a')`

web_widget_mail_send_odoo/README.rst
codingforfun: I thought this is past: 'I sent you the message 2 hrs ago' vs. present: 'I send you a message in just a minute'. I think it's correct, but I'm not a native speaker.  
The message gets appended to the chatter AFTER it has been sent.

hbrunn: ah, right. My bad, sorry

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+                                additional_context: {
+                                    default_partner_ids: [partner.id],
+                                    default_composition_mode: 'comment',
+                                    // write to active model:
+                                    default_model: fm.dataset._model.name,
+                                    default_res_id: fm.datarecord.id,
+                                },
+                                on_close: function(){
+                                    // refresh the chatter widget here
+                                    $.each(self.view.getChildren(),
```
hbrunn: usually, the chatter sits right in the form, then this works. But if you place it somewhere else, this won't find anything. Consider searching recursively

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+                                        function(index, value){
+                                            if(value.widget=='mail_thread'){
```
hbrunn: I consider `instanceof instance.mail.RecordThread` slightly safer. Or do ducktyping: `if(value.root && value.root.thread && value.root.thread.message_fetch)`

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+                                                value.root.thread.message_fetch()
+                                            }
+                                        }
+                                    );
+                                },
+                            }
+                        );
+                    } else {
+                        console.log(self.build_context());
```
hbrunn: remove

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+                        var pop = new instance.web.form.FormOpenPopup(self);
+                        var context = new instance.web.CompoundContext(self.build_context(), {
+                            default_name: parsed_email[0],
+                            default_email: parsed_email[1],
+                        });
+                        pop.show_element(
```
hbrunn: how about

``` javascript
self.do_action({
    type: 'ir.actions.act_window',
    res_model: 'res.partner',
    target: 'new',
    context: {
        //defaults you need
    },
},
{
    on_close: yourfunction,
});
```

possibly preselecting the simple partner form. This integrates better with custom code I think

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+                        pop.on('create_completed', self, function (id) {
+                            self.on_clicked()
+                        });
+                    }
+                })
+        }
+    },
+});
+
+instance.web.form.widgets.add('email', 'instance.web.form.FieldEmailIntern')
```
hbrunn: consider adding a widget `email_mailto` to force the standard way if ever need for this arises

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -0,0 +1,91 @@
...
+        }
+    },
+    on_clicked: function() {
+        var self = this;
+        if (!self.get('value') || !self.is_syntax_valid()) {
+            self.do_warn(_t("E-mail Error"), _t("Can't send email to invalid e-mail address"));
+        } else {
+            // find partner id for email
+            var res_partner = new openerp.Model('res.partner');
+            var parsed_email = instance.mail.ChatterUtils.parse_email(self.get('value'));
```
hbrunn: if the email is invalid, `parsed_email[1]` will be `false`. This should generate an error.

Or better: Override `is_syntax_valid` to check this, then users can't fill in invalid email addresses and at this point, you're guaranteed to have a mail address

hbrunn:
@codingforfun you're fast :-) By reformatting the code, github considers all my comments already taken, but that's not true. Will you address the hidden ones that came later too?

codingforfun:
Sorry I didn't reloaded the page and somehow thought you had finished. 
I'll look at the hidden comments.

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun: I thought about this, but from what I've seen so far in the Odoo code concluded it won't happen since fields can only placed inside views. 
Do you have a real world example or reasoning how this could happen?

hbrunn: Fields can only be placed inside forms, that's true. But you could place the chatter widget wherever within the form, on some tab in a notebook for example. But that's indeed theoretical, I've never seen it in the wild

hbrunn:
Another thing that comes to mind: A tree view and kanban widget doing the same thing would be nice, but I won't block the PR for this

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun: Good point. As I said I'm not yet an JS expert. I'm basically copy&pasting myself through the existing code ;-)
I did a short try but did something wrong. As I'm under time constraints now I will leave it as is for now, but have a look later maybe.

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun: Good point. I'll try to fix this.

codingforfun:
Translation is also missing

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun: I tried your idea, but it makes it more complicate to find out if the partners has been created with success or not. You know a possible way to get the return state of the dialog from within the on_close handler. Before this was easier.

hbrunn: then leave as it is, you'll get some partner form with your approach anyways

codingforfun: Actually I'm pretty sure that it makes the same. Just looks more ugly than using an action. Basically it should open the default view for the model so it should work with custom code the same way as an action. 
I found this pattern on several places in code of other modules. 
To force the simple form I changed the context. This is hard coded into res.partner's fields_view_get() method in the core.

codingforfun:
I think i finished the work on all the comments.

hbrunn:
:+1:

OSguard:
:+1:

bud-e:
Hi...   @codingforfun @hbrunn @OSguard  @pedrobaeza  , 

Is the module moved to some different project | repo ??  If so then provide me the link to that...!!
Because I was unable to find the module in `9.0` & `10.0` branch of `web` project | repo.
Or Migration is still remaining for `9.0` & `10.0` ????

--
Thanks

PR #280, opened 2015-12-27
web_sheet_full_width working in 9.0

javieralmancevo:
I updated the module web_sheet_full_witdth to work with Odoo v9 Community.


pedrobaeza:
This is already proposed here: https://github.com/OCA/web/pull/259. Please check first for not duplicating efforts.

PR #285, opened 2016-01-06
[INIT][web_sequence_fix]

andrius-preimantas:


pedrobaeza:
Thanks for the contribution!

Can you please rename the module to _web_listview_sequence_fix_ to better match the use?

web_sequence_fix/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- encoding: utf-8 -*-
```
pedrobaeza: Use coding instead of encoding and use short headers like https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py#L1-L3

web_sequence_fix/__openerp__.py
```diff
@@ -0,0 +1,32 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Web Sequence Fix',
+    'version': '0.1',
```
pedrobaeza: Use full version numbering: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

andrius-preimantas:
cc @pedrobaeza , done

OSguard:
This is a cool fix, i test it on runbot on form view of sale.order. 

What i found out is, that you can sort the sequence field reverse order and have editable "bottom". Then i expect that it will stay on bottom to add a new line, but it jumps to the top after save. Maybe it is possible to fix this as well.

andrius-preimantas:
@OSguard , yeah, apparently with handle widget the last item receives default from _defaults dict. As my fix uses "default_sequence" key in context approach, my guess is that handle ignores/recreates context at some point. I'll sure investigate that and let you know.

Thanks for testing

Andrius

andrius-preimantas:
@OSguard , apparently my previous approach was not effective when sequence field was readonly (handle widget is treated the same as readonly). It's a limitation of Odoo as we might know that setting values for readonly fields does not necessary saves it.

So I've changed my approach and resequence fields every time list view gets a new record. If you could test it and let me know if you thing it works as expected I'd change README to represent my new approach.

Cheers

pedrobaeza:
You can use web_readonly_bypass module: https://github.com/OCA/web/tree/8.0/web_readonly_bypass

pedrobaeza:
Any news on this?

PR #289, opened 2016-01-11
Web switch company warning imp detection

hparfr:
Display a warning when you logout/login in another tab or window.

### DIFF
--- a/web_switch_company_warning/static/src/js/switch_company_warning.js
+++ b/web_switch_company_warning/static/src/js/switch_company_warning.js
@@ -25,7 +25,7 @@ openerp.web_switch_company_warning = function (instance) {
               if (msg.data.type !== 'newCtx')
                 return;
 
-                if(msg.data.newCtx != self.session.company_id) {
+                if(msg.data.newCtx != signature(self.session)) {
                     self.$el.show();
                 } else {
                     self.$el.hide();
@@ -33,7 +33,11 @@ openerp.web_switch_company_warning = function (instance) {
             });
 
             w.port.start();
-            w.port.postMessage(this.session.company_id);
+            w.port.postMessage(signature(this.session));
+
+            function signature(session) {
+                return [session.db, session.uid, session.company_id].join();
+            }
         }
     });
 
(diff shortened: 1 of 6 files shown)


web_switch_company_warning/static/src/js/switch_company_warning.js
```diff
@@ -33,7 +33,11 @@ openerp.web_switch_company_warning = function (instance) {
             });
 
             w.port.start();
-            w.port.postMessage(this.session.company_id);
+            w.port.postMessage(signature(this.session));
+
+            function signature(session) {
+                return [session.db, session.uid, session.company_id].join();
```
petrus-v, line 39: Can you move this function outside of the init method to easily overwrite it on customer project module if needs?

petrus-v:
Hi,

I'm just wondering is this for user convenience or technical limitation, as far as I know there are no limitation to login with 2 different user or on 2 different database?

regards

web_switch_company_warning/static/src/js/switch_company_warning.js
hparfr, line 39: Ok

petrus-v:
Thanks for this contribution!

:+1:  (code review only!)

hparfr:
> I'm just wondering is this for user convenience or technical limitation, as far as I know there are no limitation to login with 2 different user or on 2 different database?

It's a technical limitation.
As far as I know, you have only one session per host:port

I just tested it now and didn't succeed to maintain 2 sessions on differents db for the same host:port.

PS: I did commit with a wrong email, OCA CLAbot didn't recognize my account.

Thanks for the review petrus-v

hparfr:
I did squash the commit with the wrong email.
Should be ok from the CLA perspective.

bealdav:
tested on Runbot, OK, :+1: 
thanks

OSguard:
:+1:

StefanRijnhart:
CLA checked and confirmed. Tested on runbot and this change works nicely. Nice feature BTW (I had not noticed the module before). :+1:

PR #292, opened 2016-01-15
[ADD] Add web_fetchmail_no_followers module

cmsalmeida:
This module don't add mail author on followers


web_mail_no_followers/README.rst
```diff
@@ -0,0 +1,4 @@
+Description
```
hbrunn, line 1: use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

hbrunn:
please add tests

web_mail_no_followers/model/mail_thread.py
```diff
@@ -0,0 +1,61 @@
...
+
+    thread_id = message_post_super(self, cr, uid, thread_id, body=body,
+                                   subject=subject, type=type,
+                                   subtype=subtype, parent_id=parent_id,
+                                   attachments=attachments, context=nosub_ctx,
+                                   content_subtype=content_subtype, **kwargs)
+    return thread_id
+
+
+class MailNoFollowers(models.Model):
```
hbrunn, line 54: `mail.thread` is an abstract model, do the same

web_mail_no_followers/model/mail_thread.py
```diff
@@ -0,0 +1,61 @@
...
+##############################################################################
+
+from openerp.addons.mail.mail_thread import mail_thread
+from openerp import models, api
+
+
+message_post_super = mail_thread.message_post
+
+
+@api.cr_uid_ids_context
```
hbrunn, line 31: why not v8 api?

hbrunn:
I think this PR should be moved to crm or server-tools with a matching name change: `fetchmail_no_followers`

pedrobaeza:
Or maybe social

PR #293, opened 2016-01-17
[ADD] Web log followers change

cmsalmeida:
This module create a message every time a follower is added or removed

### DIFF
--- a/web_log_followers_change/README.rst
+++ b/web_log_followers_change/README.rst
@@ -1,0 +1,4 @@
+Description
+-----------
+Create a message every time a follower
+is added or removed on any object.
--- a/web_log_followers_change/__init__.py
+++ b/web_log_followers_change/__init__.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2010-2013 OpenERP s.a. (<http://openerp.com>).
+#    Copyright (C) 2013 ThinkOpen Solutions Brasil (<http://www.tkobr.com>).
+#    Author Carlos Almeida <carlos.almeida at tkobr.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import model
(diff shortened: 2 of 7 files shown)


web_log_followers_change/README.rst
```diff
@@ -0,0 +1,4 @@
+Description
```
elicoidal, line 1: Hi @cmsalmeida 
Thanks for your contribution to the OCA. 
Please follow the guidelines from:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
Please create a proper README file following: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

web_log_followers_change/__init__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
elicoidal, line 1: Please use simple headers in all py files following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

web_log_followers_change/__openerp__.py
```diff
@@ -0,0 +1,34 @@
...
+##############################################################################
+
+{
+    'name': 'Log every followers change',
+    'version': '8.0.0.1.0',
+    'author': "ThinkOpen Solutions Brasil, Odoo Community Association (OCA)",
+    'category': 'Web',
+    'website': 'http://www.tkobr.com',
+    'license': 'AGPL-3',
+    'images': [],
```
elicoidal, line 31: please remove empty keys (x2)

web_log_followers_change/model/mail_followers.py
```diff
@@ -0,0 +1,71 @@
...
+    return res
+
+
+class LogChangeMessageFollowers(osv.Model):
+    ''' I define this class so this code will be executed only if module
+        is installed in database. Otherwise it will be always executed.
+    '''
+    _name = 'mail.thread'
+
+    def __init__(self, *arg, **kwargs):
```
hbrunn, line 69: This has quite some potential side effects and doesn't play well with other modules doing the same thing. Look at https://github.com/hbrunn/social/blob/8.0-mail_follower_custom_notification/mail_follower_custom_notification/models/mail_thread.py for a safer approach

hbrunn:
@cmsalmeida thanks for this module, good idea. Please also add tests, and move your PR to https://github.com/OCA/social

cmsalmeida:
:+1:  thanks a lot

web_log_followers_change/model/mail_followers.py
cmsalmeida, line 69: :+1:

cmsalmeida, line 69: I'm not getting how that works, this seems the only way, if I do like that I got error stating that original method is not defined... can you point me to some documentation or guidelines on how to do it?
Thanks in advance

hbrunn, line 69: The documentation is odoo's source code I'm afraid. Starting points are https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L556 and https://github.com/OCA/OCB/blob/8.0/openerp/modules/registry.py#L170.

I'm toying with a mixing that helps with this, you _should_ be able to just grab that file https://github.com/hbrunn/server-tools/blob/8.0-base_encrypted_field/base_encrypted_field/models/base_patch_models_mixin.py and follow the instructions in the comment. Very much alpha, so ymmv

PR #295, opened 2016-01-18
[IMP] massive performance boost for big matrices

hbrunn:
On Firefox, big matrices can get very slow, that's because `get_totals`, which is called on every change in a field, reads from that dataset again, and this in turn always copys the whole dataset around. By accessing the cached rows directly (the widget doesn't allow adding/removing rows anyways), we're faster several orders of magnitude.

I took the occasion to also update `README.rst` to the newest version of the OCA template and to add an example on how to use the code. The example is necessary it seems, as I receive a lot of requests about this. Unfortunately, everyone I helped with this up to now couldn't be bothered to PR back whatever would have helped them in the first place. Actually, even a thank you is expected too much usually, so I eventually stopped replying on private mails on my modules in general.

### DIFF
--- a/web_widget_x2many_2d_matrix/README.rst
+++ b/web_widget_x2many_2d_matrix/README.rst
@@ -90,7 +90,6 @@ Known issues / Roadmap
 
 * it would be worth trying to instantiate the proper field widget and let it render the input
 
-===========
 Bug Tracker
 ===========
 
@@ -99,7 +98,6 @@ In case of trouble, please check there if your issue has already been reported.
 If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
 `here <https://github.com/OCA/web/issues/new?body=module:%20web_widget_x2many_2d_matrix%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
 
-=======
 Credits
 =======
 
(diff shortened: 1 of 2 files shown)


web_widget_x2many_2d_matrix/README.rst
```diff
@@ -51,12 +55,42 @@ show_row_totals
...
+
+<field name="task_ids" widget="x2many_2d_matrix" field_x_axis="project_id" field_y_axis="user_id" field_value="planned_hours" />
+
 Known issues / Roadmap
 ======================
 
 * it would be worth trying to instantiate the proper field widget and let it render the input
 
-
+===========
```
pedrobaeza: These sections should the line on top. It's only for the first title to highlight it

pedrobaeza:
Thanks for the improvement :+1: 

Indeed, when you receive several mails per day of people that wants your help, but with no feedback or thankfulness, it gets annoying. I stopped also and derive to forums/GitHub if they want to do it. At least, there are public records that can help others with the same problem.

elicoidal:
Hehe Like I use to say I am happy to help with community service but do not offer free private consulting

guewen:
:+1:

rafaelbn:
cc @antespi

dreispt:
@hbrunn As a rule of thumb, asking privately for help is hiring a professional and should be paid.
As Pedro said, questions and answers must be public to be considered community work.
My advice is to redirect people to open an Issue or ask on a public place.

PR #300, opened 2016-01-26
[ADD] create new addon to save updates in qweb views as diff

kaerdsar:
Create new addon for saving qweb views changes as diff instead of the hole view.

### DIFF
--- a/setup/web_x2many_delete_all/odoo_addons/__init__.py
+++ b/setup/web_x2many_delete_all/odoo_addons/__init__.py
@@ -1,0 +1,1 @@
+__import__('pkg_resources').declare_namespace(__name__)
--- a/setup/web_x2many_delete_all/odoo_addons/web_x2many_delete_all
+++ b/setup/web_x2many_delete_all/odoo_addons/web_x2many_delete_all
@@ -1,0 +1,1 @@
+../../../web_x2many_delete_all
--- a/setup/web_x2many_delete_all/setup.py
+++ b/setup/web_x2many_delete_all/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)


web_qweb_diff/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+    'version': '1.0',
+    'author': "bloopark systems GmbH & Co. KG ,",
+    'website': "http://www.bloopark.de",
+    'depends': ['web_editor'],
+    'data': [
+        'views/ir_ui_view.xml',
+        'security/ir.model.access.csv'
+    ],
+    'external_dependencies': {
+        'python': ['diff-match-patch']
```
hbrunn: add this to the requirements in https://github.com/OCA/web/blob/8.0/.travis.yml to make it work on travis

web_qweb_diff/README.rst
```diff
@@ -0,0 +1,32 @@
+Web Qweb Diff
```
hbrunn: use OCA's readme template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

kaerdsar:
@hbrunn I updated .travis.yml and the README file. It looks like the tests for reports are failing now, probably because of my module. Any idea?

hbrunn:
@kaerdsar I don't see how your code could have caused this, but it's early in the morning here, so I just forced a rebuild and we'll see how this comes out. Sometimes, travis is a bit brittle.

kaerdsar:
@hbrunn I already sent my CLA. Please, let me know, what should be the next step?

hbrunn:
@kaerdsar wait until the CLA is received, meanwhile fix travis

kaerdsar:
@hbrunn I already received an email confirming the reception of my CLA, so I'm in I think. I also entered in the PSC for Web addons.

I still no success with travis. I'll keep finding which is the problem. Thanks in advance.

hbrunn:
@kaerdsar believing you, I removed the cla label. The travis test looks like something goes wrong with the api mangling. Try calling super old style, and in any case, do the test locally and see what happens there. You can just copy the commandline from https://travis-ci.org/OCA/web/jobs/105832495#L223

kaerdsar:
@hbrunn Travis tests passed. I can't check what is the error in Runbot now.

hbrunn:
@kaerdsar looks good! :+1:

kaerdsar:
@hbrunn any updates about revisions?

hbrunn:
what do you mean?

kaerdsar:
I don't know what is the process to check the pull request. Do I have to do something more to get some code revisions?

hbrunn:
you mean reviews? not really, now at least one other person has to take the time to review, and if she also gives a thumbs up, we're ready to merge. Just mine is not good enough

kaerdsar:
Understood! Thanks for your time.

Another question .... I have a module to reset changes in website pages that depends of this module. Where should I add that module? In OCA/website?

hbrunn:
website sounds good

jholze:
@kaerdsar  i think you should push the last changes as well and we should also push the versioning addon or ?

dreispt:
What action is needed here?

jholze:
@kaerdsar i think this PR should be combined with the new Workspace handling we did for websites or what do you think ?

kaerdsar:
@dreispt as @jholze says, we have a new approach for handling workspaces in website, using a new version of the module in this PR. I'm not completely sure that the diff handling could be useful as a separated module. Then I think we may close this PR and I can create a new one in the OCA/website repository for the website workspace solution. What do you think?

dreispt:
@kaerdsar Correct me if I'm wrong (I didn't dig into the code), but it seems that this proposes a generic view related feature: you edit the XML and this module stores the change as an inherited view. If so, it would be more generic than website.

dreispt:
Also, I'm clueless about "Workspaces for Website". What is that?

kaerdsar:
@dreispt 
- This PR proposes a module where if you edit an XML then it will be saved a diff, which is something like a diff when you add new code in a repository with control version system. It is not stored a new inherited view.  Actually the idea about create inherited views sounds good and I think I can work on it too.
- The website workspace approach is about handle changes in website pages. Imagine you create some changes in a website page and don't want to publish them immediately. Then you create a workspace with those changes, this workspace can be published, can be reseted and also you can undo specific changes. This solution uses the diff objects to track the changes in the design process for the new website.

I think the module in this PR needs more discussion about its goal, or we can take your idea: store inherited views instead of diff objects for changes in XML views. I'm not completely sure, it would be nice get some feedbacks.

dreispt:
The "workspace" idea is something badly lacking in the Odoo CMS. The name is not intuitive, I think you mean view versions and publication control.
Technically, this doesn't require a diff feature: Odoo has out of the box supports for several versions of a view - the one with the lower priority is the one to be presented. Maybe a small extension on the Views model could do the trick and add versions, publication state, or even defining publication periods.

The XML diff to automatically create inherited views would be a great feature, but I'm afraid that it can be hard to implement. And probably it's best uses would be for backend view customization.

I suggest you to open a Blueprint Issue explain the goal and the design for the implementation.
You can get early feedback validating the idea and design, before jumping into the final coding.

pedrobaeza:
Is the "Needs fixing" label correct? Is there anything more to do here?

hbrunn:
the needs fixing came from me and I forgot to remove it when thumping up

pedrobaeza:
OK, thanks for clarifying

@kaerdsar can you rebase to get latest changes on the branch and if Travis goes green, I'll merge.

kaerdsar:
@pedrobaeza @hbrunn @dreispt I closed this PR because I won't keep this approach here, it has been used in one project I'm working on, but I think is to specific and maybe not useful as a generic module. Later I will work in a new PR trying to implement the suggestion from @dreispt about create inherited views from XML diff, but probably it takes some time. Thanks for the comments. If you have another good idea or there is another work in progress about it, just let me know.

pedrobaeza:
OK, no further progress for now.

PR #302, opened 2016-02-01
[FIX] undefined links if field is not set

schout-it:
fix to don't show links if field is not set.
previous 'undefined' would show if the relation was False, and an error would we thrown on click


pedrobaeza:
Can you please sign the CLA? https://odoo-community.org/page/cla

schout-it:
signed cla send

pedrobaeza:
:+1:

petrus-v:
:+1:

PR #304, opened 2016-02-02
[FIX][web_widget_color] Do not force mandatory color on JS level

andrius-preimantas:
With current code there's no way to leave color field empty because jscolor fills it automatically. This change fixes it. If you want field to be mandatory just set required="True" on xml level as usual.


andrius-preimantas:
cc @pedrobaeza , @OSguard 

Anything else is needed to process this PR?

Cheers

pedrobaeza:
It seems correct to me :+1:

OSguard:
:+1:

PR #305, opened 2016-02-03
[IMP] 7.0 add m2x options for view

azoellner:
Providing m2x widget options (`create`, `create_edit`, `m2o_dialog`, `limit`, `search_more`) at the view level via the `context` field of the `ir.actions.act_window`.

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -99,6 +99,33 @@ To add these parameters go to Configuration -> Technical -> Parameters -> System
 - web_m2x_options.search_more: True
 
 
+ir.actions.act_window context options
+-------------------------------------
+
+The following options can be specified in the ``context`` field of an
+``ir.actions.act_window`` to provide the same functionality at the form level:
+
+``create`` *boolean* (Default: depends if user have create rights)
+
+  Whether to display the "Create..." entry in dropdown panels.
+
+``create_edit`` *boolean* (Default: depends if user have create rights)
+
+  Whether to display the "Create and Edit..." entry in dropdown panels.
+
+``m2o_dialog`` *boolean* (Default: depends if user have create rights)
+
+  Whether to display the many2one dialog in case of validation error.
+
+``limit`` *int* (Default: openerp default value is ``7``)
+
+  Number of displayed records in drop-down panels.
+
+``search_more`` *boolean*
+
+  Whether to always display the "Search More..." entry in dropdown panels.
+
+
 Example
 -------
 
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -16,7 +16,7 @@ add some new display control options.
 The options provided include the possibility to remove "Create..." and/or
 "Create and Edit..." entries from many2one and many2many drop down lists. You
 can also change the default number of entries appearing in the drop-down, or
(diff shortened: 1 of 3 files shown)


OSguard:
azoellner is covered by the company CLA from initOS

hbrunn:
@OSguard then I suggest to check with @OCA/board to have this handle included, otherwise those messages will keep coming

hbrunn:
@azoellner I think if would be a lot more elegant to do your changes in https://github.com/OCA/web/pull/305/files#diff-d93aa186928a77371c865d1748af9b38R31, letting the appropriate context option win against the config parameter. Then the rest of the diff would only be you refactorings, which are badly needed i think

hbrunn:
Porting this to 8.0 at once would be highly appreciated, but it's not blocking

pedrobaeza:
Do you think these changes should be under the same module? They are for action windows, not for m2x fields.

hbrunn:
@pedrobaeza to my understanding, this code allows to set the options via the context, I think it's fine to have it in the same module

azoellner:
>  letting the appropriate context option win against the config parameter

For me, the most reasonable priorities seem to be (from lowest to highest)
1. default, i.e. the implicit global (Odoo database) level
2. `ir_config_parameter`, i.e. the explicit global level
3. `context` of the action/view, i.e. the form view level
4. `options` of the field widget, i.e. the individual field level

as they get more and more specific. I think this reflects in the code, as the checks successively override the option value that is ultimately used.

hbrunn:
@azoellner agreed, and that's exactly what will happen if you follow my suggestion, just less messy as it is now

azoellner:
@hbrunn Now I understand your suggestion. Agreed, moving the code into that function will avoid the ugly code duplication.

hbrunn:
@azoellner setting to needs fixing until you find the time to move the code

pedrobaeza:
@azoellner, are you going to finish this?

azoellner:
Sorry it took me so long. Hopefully my code now looks a bit better. Refactored the option-related common code from both the 'many2one' and the 'many2manytags' widget into common functions, and then refactored the multiple checks for determining the option values to a single common location.

pedrobaeza:
There's still 2 missing pylint problems:

```
web_m2x_options/__openerp__.py:3: [C8102(manifest-required-key), ] Missing required key "license" in manifest file
web_m2x_options/__openerp__.py:3: [C8106(manifest-version-format), ] Wrong Version Format "0.1" in manifest file. Regex to match: "(7.0)\.\d+\.\d+\.\d+"
```

azoellner:
> There's still 2 missing pylint problems

I didn't want to touch those, because I'm not sure about them. A reasonable `version` seems to be `7.0.0.1`, or maybe `7.0.0.2`. But `license` is not clear to me, since I'm not the original author. What do you suggest?

pedrobaeza:
You have to put 'AGPL-3' as license, and the version according OCA guidelines is 7.0.1.0.0 (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers)

pedrobaeza (approved):
Waiting @hbrunn's approval

hbrunn (approved):

PR #307, opened 2016-02-10
[8.0][FIX][web_m2x_options] Limit name_search to 80 like Odoo core does

antespi:
Limit name_search JSON-RPC to 80 results in order to not overload backend, like Odoo core does here: https://github.com/odoo/odoo/blob/8.0/addons/web/static/src/js/view_form.js#L3395

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -155,9 +155,12 @@ openerp.web_m2x_options = function (instance) {
                     values.push({
                         label: _t("Search More..."),
                         action: function () {
+                            // limit = 80 for improving performance, similar
+                            // to Odoo implementation here:
+                            // https://github.com/odoo/odoo/commit/8c3cdce539d87775b59b3f2d5ceb433f995821bf
                             dataset.name_search(
                                 search_val, self.build_domain(),
-                                'ilike', false).done(function (data) {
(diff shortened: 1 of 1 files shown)


antespi:
Please @yvaucher, @ecino, @davidZIVI and @njeudy review this little change. I don't know if this `false` was there for a prupose

pedrobaeza:
:+1:

JavierIniesta:
:+1:

pedrobaeza:
I have retired my thumb up because the concerns raised on IRC. Are you sure this doesn't limit the possible results?

antespi:
@pedrobaeza I'm checking Marmotte concerns...

Garamotte:
:+1:

antespi:
In this runbot http://3141343-307-78af88.runbot2.odoo-community.org/web?db=3141343-307-78af88-all I have changed Opportunities form, adding a domain `[('is_company', '=', True)]` to partner_id field. Then you can check that writing 'a', more than 8 results appears. Then you can click 'Search more...' so only companies are shown and user can define whatever filter he wants.

antespi:
@pedrobaeza I've change Odoo reference link to commit permanent link for better reference

pedrobaeza:
Great! :+1:

PR #311, opened 2016-02-17
TreeView -- Inhebit execution of actions for row click.

codingforfun:
In some cases it's usefull to forbit opening FormView or execute `tree_but_open` action for click onto a row in a Tree- or ListView. 
This module provides a new configuration attribute `open` similiar to `create`, `edit` and `delete`.


hbrunn:
if you have an action defined without form view, the result is the same. What's your use case that you can't simply do this?

codingforfun:
Unfortunately this doesn't apply in general. It does not work if the list view is embedded into a form view. Also I think it's more obvious and to read and easier to define.

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Web Tree No Open
```
elicoidal, line 6: Hi @codingforfun thanks for your contribution.
Please add a line before the document title.

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+================
+
+This module extends the functionality of tree views.
```
elicoidal: What about?
This module extends the functionality of tree views and adds a new attribute to deactivate the onclick behaviour in them.
The new configuration attribute 'open' is to be used in the same way as existing ones 'edit', 'create' and 'delete'.

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+The new attribute makes it possible to deactivate the onclick behaviour in tree
+views.
+::
+
+   <tree open="false" >
+
+Installation
+============
+
+To install this module, just follow basic steps to install an odoo module.
```
elicoidal: s/odoo/Odoo

ivantodorovich:
Have you considered also making a PR to odoo?

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+
+
+Configuration
+=============
+
+Not needed.
```
elicoidal: No configuration needed.

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+
+To use this module, you need to:
+
+* go to ...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/web/8.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
elicoidal: remove

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+.. branch is "8.0" for example
+
+For further information, please visit:
```
elicoidal: remove

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
```
elicoidal: remove if empty

web_tree_no_open/README.rst
elicoidal, line 6: Not clickable tree view?

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,81 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: https

web_tree_no_open/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+
+{
+    'name': 'web_tree_no_open',
+    'summary': 'Disable actions for click onto Tree- or ListView rows.',
+    'version': '8.0.0.1.0',
```
elicoidal: 8.0.1.0.0

web_tree_no_open/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+    'author': 'initOS GmbH, Odoo Community Association (OCA)',
+    'installable': True,
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'auto_install': False,
+    'depends': [
+        'base',
```
elicoidal: not sure base is necessary

codingforfun:
@hbrunn 

> What's your use case that you can't simply do this?

My use case is that I use a delegate model as a kind of proxy to unify relationships between some objects and some other objects. To list the relationship of an Object to it's connected other objects I list entities of the delegate model. There is a button to open the target of the delegate model. Therefor I do not want the default form view of the delegate model to get opened when it is clicked, because it is of no value to the user. 

My first try was to solve the delegation problem via `tree_but_open` action, but unfortunately this doesn't work on TreeViews embedded into FormViews and is not straight forward to extend.

codingforfun:
@ivantodorovich 

> Have you considered also making a PR to odoo?

You mean as a core patch? No not yet.

codingforfun:
@hbrunn 
Did I answered your question or is there something missing?

hbrunn:
@codingforfun nothing missing, but I'm still in doubt if your approach is the best solution. For your specific problem, I'd create a custom widget that does whatever your button does in `row_clicked`, this should do for a more consistent user experience.
And for the general case, I'll leave it to other reviewers if they think we need this module, I'll abstain.

codingforfun:
@hbrunn 
I do not insist on my solution in detail. If there is a better way it will be fine, however I think the need to create list views that do not react on click is there. If so it should be consistent for stand alone views and views embedded into form views. I think providing another attribute similar to 'edit' ... etc. is a natural approach to provide this. 

You are right. I think sooner or later for my special case I'll take the time and build a custom solution for my use case, but since I had this extension already there I would like to share. 

For my special case the most general solution would be to generalize the list view behaviour to also enable the `tree_but_open` mechanism to work for list views embedded into form views. I already had a look into this, but this looks a bit more complicated.

codingforfun:
Are there more opinions here about this?

codingforfun:
@hbrunn 
You are right in that this probably doesn't make much sense for a list view with associated form view. 
This only makes sense for list views embedded into form view via one2many. 
You think it would be a better PR if I restrict the change to `instance.web.form.One2ManyListView` and `instance.web.form.Many2ManyListView`? 

This is also related to https://github.com/OCA/web/pull/326. 
The same could be achieved by using https://github.com/OCA/web/pull/326 and providing a noop tree_but_open action, although this wouldn't be very intuitive to program.

OSguard:
@codingforfun is this opsolete to https://github.com/OCA/web/pull/326 ?

codingforfun:
> is this opsolete to #326 ?

not really. You can achieve similar effect with #326 by using a noop server action, but in my opinion this is a hacky approach. Doing so has following disadvantages over compared to the approach here: 
- Fair amount of extra code to write
- Not very obvious approach for a clearly defined simple requirement
- Extra RPC calls to get the noop action from the server. 
- The solution here integrates with the other tree options. E.g. you could make it dependent on context or other criteria.

OSguard:
:+1:

tremlin:
Code review :+1:

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Configuration
+=============
+
+No configuration needed.
+
+Usage
+=====
+
+To use this module, you need to:
```
pedrobaeza: Remove this empty paragraph

web_tree_no_open/README.rst
```diff
@@ -0,0 +1,67 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
pedrobaeza, line 3: Put target link

odooCroatia:
:+1:

remihb (approved):

codingforfun:
Someone had the same idea here: 
https://apps.odoo.com/apps/modules/10.0/web_tree_no_open/
https://github.com/aydi-nebil/web/tree/12.0/web_tree_no_open
so obviously there is a need for this

codingforfun:
In fact. Actually comparing the code for v10, I think it’s just a copy of mine.

PR #313, opened 2016-02-24
[ADD] Module web_calendar_year 8.0

astirpe:
Module Year calendar view.
Extends the default calendar widget with a year view.


pedrobaeza:
Hi, Andrea, thanks for the contribution!

I'm reviewing it.

web_calendar_year/README.rst
```diff
@@ -0,0 +1,3 @@
+Year calendar view
```
pedrobaeza: Please use full OCA's README template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

web_calendar_year/__init__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza, line 1: Use short headers: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

web_calendar_year/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+
+{
+    'name': 'Web Calendar Year',
+    'images': ['static/description/main_screenshot.png'],
+    'summary': """Year view on Odoo calendar""",
+    'author': 'ONESTEiN BV,Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'website': 'http://www.onestein.eu',
+    'category': 'Extra Tools',
+    'version': '1.0',
```
pedrobaeza: 8.0.1.0.0

web_calendar_year/__openerp__.py
```diff
@@ -0,0 +1,38 @@
...
+##############################################################################
+
+{
+    'name': 'Web Calendar Year',
+    'images': ['static/description/main_screenshot.png'],
+    'summary': """Year view on Odoo calendar""",
+    'author': 'ONESTEiN BV,Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'website': 'http://www.onestein.eu',
+    'category': 'Extra Tools',
```
pedrobaeza: Use category 'Web'

web_calendar_year/doc/index.rst
```diff
@@ -0,0 +1,9 @@
+This module adds a year view to all Odoo calendar views for all users.
```
pedrobaeza: Remove this file and company references

pedrobaeza:
2 more things:
- Remove Onenstein logo from the image.
- A more descriptive icon than the generic calendar one would be better.

web_calendar_year/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# © 2015 ONESTEiN BV (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Web Calendar Year',
+    'images': ['static/description/main_screenshot.png'],
+    'summary': """Year view on Odoo calendar""",
```
hbrunn: no need for a multiline string here

web_calendar_year/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+    'category': 'Web',
+    'version': '8.0.1.0.0',
+    'depends': ['web_calendar'],
+    'data': [
+        'views/web_calendar_year.xml',
+    ],
+    'demo': [],
+    'installable': True,
+    'auto_install': False,
+    'application': False,
```
hbrunn: remove empty keys and keys with default value

web_calendar_year/static/src/js/web_calendar_year.js
```diff
@@ -0,0 +1,39 @@
+openerp.web_calendar_year = function(instance) {
+
+    instance.web_calendar.CalendarView = instance.web_calendar.CalendarView.extend({
```
hbrunn: dont do this, this changes the reference of the object. Use `instance.web_calendar.CalendarView.include`

web_calendar_year/static/src/js/web_calendar_year.js
```diff
@@ -0,0 +1,39 @@
...
+            options.header.right = "year,month,agendaWeek,agendaDay";
+			options.eventAfterAllRender = function() {
+				console.log("eventAfterAllRender");
+				var tableBody = $('.oe_view_manager_body');
+				tableBody.find('.fake-thead').remove();
+				tableBody.scroll();
+			}
+            return options;
+        },
+		init_calendar: function() {
```
hbrunn: unindent

web_calendar_year/static/src/js/web_calendar_year.js
```diff
@@ -0,0 +1,39 @@
+openerp.web_calendar_year = function(instance) {
+
+    instance.web_calendar.CalendarView = instance.web_calendar.CalendarView.extend({
+        get_fc_init_options: function() {
+            var options = this._super();
+            options.buttonText.year = instance.web._t("Year");
+            options.header.right = "year,month,agendaWeek,agendaDay";
+			options.eventAfterAllRender = function() {
```
hbrunn: unindent

web_calendar_year/static/src/js/web_calendar_year.js
```diff
@@ -0,0 +1,39 @@
...
+				console.log("eventAfterAllRender");
```
hbrunn: remove debug print

web_calendar_year/views/web_calendar_year.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <template id="assets_backend" name="web_calendar_year assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/web_calendar_year/static/src/js/fullcalendar.js"></script>
```
hbrunn: libraries belong in `/static/lib`

hbrunn:
:+1:

pedrobaeza:
We are introducing here code from Adam Shaw (the JS lib), so we need him to sign the CLA also.

hbrunn:
@pedrobaeza is there any OCA-decision about that? I don't think it's feasible to have library authors sign our CLA, this effectively would mean we can't use existing libraries, that would be horrible

hbrunn:
PS: As the libraries have their own copyright, I think it's clear that the OCA can't have any claims on them

astirpe:
That js library (FullCalendar) is a modified copy of the js library belonging to the standard Odoo Module _web_calendar_

hbrunn:
oh, I was not aware of that. Isn't there a possibility to use Odoo's version and monkey patch whatever you need? That would be a lot cleaner, because this way, you never know which version of the library you have in hand currently.

astirpe:
@hbrunn indeed would be a lot cleaner! We will check asap for a better solution.

In the meanwhile, I would ask an example, in case someone is aware of any similar solution applied on a module (I mean import/refer js libs defined in a base module + extending or monkey-patching them). That would be very appreciated!

hbrunn:
have a look at https://github.com/OCA/web/blob/8.0/web_ckeditor4/static/src/js/web_ckeditor4.js#L25 - this is not exactly your case, but should give you the idea

astirpe:
@hbrunn thank you very much! Your example is very interesting. I'm still trying to figure out how to apply a similar solution for that FullCalendar library, I will be back on it as soon as possible ...

hbrunn:
if you post the diff of what you needed to change, we might be able to help you

astirpe:
Here is the [diff](https://github.com/astirpe/fullcalendar/commit/caafb6b053ab1061e1201aa4761e74883a8c907f) with the modifications applied to the original fullcalendar library.

As you can notice, the original code is entirely wrapped inside a kind of anonymous function:

```
(function($, undefined) {

    ...

})(jQuery);
```

an this is the main difficulty: extending or patching the original code contained in an anonymous function.

hbrunn:
luckily, this code puts most of its references into `$.fullCalendar`. So where you say

``` javascript
fcViews.year = YearView;
function YearView(element, calendar)
...
```

you should be also able to say

``` javascript
jQuery.fullCalendar.views.year = function(element, calendar)
...
```

then I mostly see `console.logs` which we don't want so badly anyways and a little adaption of `BasicView` of which I'm not sure it it's necessary.

pedrobaeza:
@astirpe, are you going to work in the cleaner implementation suggested by @hbrunn ?

astirpe:
@pedrobaeza  not for the time being, sorry. Feel free to close the PR (if that was the intention)

hbrunn:
let's set it to work in progress then, this way, it stays visible for whomever it concerns, but won't be shown in lists filtering for needs review. I also remove the CLA label as this is done too to my knowledge

astirpe:
I will not continue the development of this module, so I'm closing this old PR.
The original module will remain available [here](https://github.com/onesteinbv/addons-onestein/tree/8.0/web_calendar_year).

PR #316, opened 2016-02-29
[9.0] web_access_rule_buttons: Skip check on TransientModels without ID

guewen:
Fix:

If we call check_access_rule() without id, it will try to run a
SELECT without ID which will crash, so we just blindly allow the
operations

```
  File "/opt/odoo/src/openerp/http.py", line 643, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/src/openerp/http.py", line 680, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/src/openerp/http.py", line 316, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/src/openerp/service/model.py", line 118, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/src/openerp/http.py", line 309, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/src/openerp/http.py", line 959, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/src/openerp/http.py", line 509, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/src/addons/web/controllers/main.py", line 893, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/src/addons/web/controllers/main.py", line 885, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/opt/odoo/src/openerp/api.py", line 250, in wrapper
  File "/opt/odoo/src/openerp/api.py", line 381, in old_api
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/external-src/web/web_access_rule_buttons/models.py", line 24, in check_access_rule_all
    result[operation] = True
  File "/opt/odoo/src/openerp/api.py", line 248, in wrapper
    return new_api(self, *args, **kwargs)
  File "/opt/odoo/src/openerp/api.py", line 574, in new_api
    result = method(self._model, cr, uid, self.ids, *args, **old_kwargs)
  File "/opt/odoo/src/openerp/models.py", line 3554, in check_access_rule
    WHERE id IN %%s""" % self._table, (tuple(ids),))
  File "/opt/odoo/src/openerp/sql_db.py", line 139, in wrapper
    return f(self, *args, **kwargs)
  File "/opt/odoo/src/openerp/sql_db.py", line 218, in execute
    res = self._obj.execute(query, params)
ProgrammingError: syntax error at or near ")"
LINE 3:                           WHERE id IN ()
```

### DIFF
--- a/web_access_rule_buttons/models.py
+++ b/web_access_rule_buttons/models.py
@@ -19,6 +19,12 @@ def check_access_rule_all(self, operations=None):
         operations = ['read', 'create', 'write', 'unlink']
     result = {}
     for operation in operations:
+        if self.is_transient() and not self.ids:
+            # If we call check_access_rule() without id, it will try to run a
+            # SELECT without ID which will crash, so we just blindly allow the
+            # operations
+            result[operation] = True
+            continue
         try:
             self.check_access_rule(operation)
         except exceptions.AccessError:


yajo:
Is there anything preventing [this line](https://github.com/OCA/web/pull/316/files#diff-c06f72ad065d968be7ad95df0c3c3d76R38) from being executed when the module is not installed?

pedrobaeza:
Shouldn't that be monkey-patched on `_register_hook` method?

yajo:
That's what I was thinking. Otherwise it will be patched even without installing this.

yajo:
And what's worse, that's already in the stable branch

pedrobaeza:
I only link in my addons_path modules that I'm using, so this doesn't affect me, but a usual practice is to add the entire repo to the addons_path. Luckily, 9.0 isn't still a wide use branch.

guewen:
You are right, I'll make a second PR for that. This is not dramatic however because it adds a new method that will be used nowhere.

yajo:
Yes, but modules should not have effect when not installed. :+1: To this one then.

guewen:
> Shouldn't that be monkey-patched on _register_hook method?

On what model should I put the `_register_hook`?

web_access_rule_buttons/models.py
```diff
@@ -15,10 +15,17 @@ def check_access_rule_all(self, operations=None):
    :param operation: a list of ``read``, ``create``, ``write``, ``unlink``
    :return: {operation: access} (access is a boolean)
     """
+    self.ensure_one()
```
pedrobaeza: Do you need this? Why not simply check `self.ids` and this continues being multi-record?

pedrobaeza:
@guewen, in all models that call _register_hook, isn't it?

web_access_rule_buttons/models.py
guewen: Indeed :) fixed

pedrobaeza:
:+1:

guewen:
`_register_hook` patch is here: https://github.com/OCA/web/pull/318

web_access_rule_buttons/models.py
```diff
@@ -19,6 +19,12 @@ def check_access_rule_all(self, operations=None):
         operations = ['read', 'create', 'write', 'unlink']
     result = {}
     for operation in operations:
+        if self.is_transient() and not self.ids:
```
yajo, line 22: I'm pretty sure transient model records have an ID, the only thing is that they get deleted by a cron job. Are you sure `self.is_transient()` is not enough?

guewen, line 22: They do. But the wizard's form get created before they have one. This check is necessary.

yajo, line 22: OK

yajo:
:+1:

PR #318, opened 2016-03-01
Add method to BaseModel only on addon installation

guewen:


web_access_rule_buttons/models.py
```diff
@@ -28,4 +28,11 @@ def check_access_rule_all(self, operations=None):
...
 
 
-models.BaseModel.check_access_rule_all = check_access_rule_all
+# Could be any model, we just use a core model to have a 'register_hook'
+class IrModel(models.Model):
+    _inherit = 'ir.model'
+
+    def _register_hook(self, cr):
+        # Add method check_access_rule_all for all models
+        models.BaseModel.check_access_rule_all = check_access_rule_all
```
pedrobaeza, line 37: Shouldn't this be done to the corresponding instance:

```
self.check_access_rule_all = check_access_rule_all
```

guewen, line 37: @pedrobaeza this method is added on `BaseModel` because it can be called on any Model.

pedrobaeza, line 37: But this will make that in a multi-DB environment, the method is patched for all DBs, having the module installed or not.

guewen, line 37: That's right. But I don't think there is any solution to that. Another issue is that if you uninstall an addon, the code changed by `_register_hook` stays. This is a huge weakness of the way Odoo loads addons. This is by far not the only addon to suffer from that. Hopefully in v10, we'll be able to simply inherits the model 'base': https://github.com/odoo/odoo/pull/10288/files#diff-678bd201dfe1a0c69add630cf73e76baR58

But again, the worst case is that this method is added on the models, but will never be called because the js code that calls it is not loaded so that's not dramatic.

yajo, line 37: Even if you don't use the method, it does not feel good to alter all databases without installing the module. It can lead to hard to guess/debug situations.

Odoo provides `_patch_method()` for that. See this recent PR that does a similar thing: patches methods in modules based on a condition. Yours should be the same without the condition IMHO. https://github.com/OCA/server-tools/pull/352/files

Thanks for the link, indeed v10 will be an interesting one.

guewen, line 37: > Even if you don't use the method, it does not feel good to alter all databases without installing the module. It can lead to hard to guess/debug situations.

Never told otherwise! I agree, but Odoo is the culprit, I can't do anything about that.

> Odoo provides _patch_method() for that. See this recent PR that does a similar thing: patches methods in modules based on a condition. Yours should be the same without the condition IMHO. https://github.com/OCA/server-tools/pull/352/files

This is quite different. `_patch_method` monkey-patch an existing method on the current model. I add a method on `BaseModel`.

yajo, line 37: Hmm I see.

Further inspecting the code, I see you are doing a JSON-RPC call to this method, is it so? Then, what about just adding the `check_access_rule_all` method to `ir.model` and adding an additional `model` parameter to it? Then you call the method in `ir.model` and get the results for the model you ask for.

guewen, line 37: Yes it could be another solution which would not involves to add a method on `BaseModel` I didn't thought of, on the other hand, the API would be a bit confusing as it would take model + ids of the other model (as the method is a `@api.multi`). It would also no longer be coherent with `check_access_rule`.

This is not a change I would do on a stable module, but as this one is recent, chances that someone relies on this method to be on `BaseModel` are low... could be okay.

yajo, line 37: You can make it `@api.model` now, and about changing, we are fixing a bug here. Stability is desired, but IMHO lack of bugs is more, since submodules can be updated to new mehods.

hbrunn:
:+1:

lmignon:
:+1:

PR #324, opened 2016-03-11
[ADD] Module menu_collapsible 8.0

astirpe:
This module makes all menus collapsible for all users.


max3903:
@astirpe Please use OCA branding in your icon and screenshot. Thanks.

pedrobaeza:
Hi, @astirpe, thanks for the module.

Please call it _web_menu_collapsible_ for following the nomenclature about UI addons.

tafaRU:
Hi @astirpe, thanks for the module.
Please delete the files related to **menu_collapsible**.

bealdav:
tested, nice work

nhomar:
:+1: 

I just should add some print screens on the descriptor... It can help in the app store to be better used :-)

web_menu_collapsible/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================
+Collapsible menu
+================
+
+This module makes all menus collapsible for all users.
```
pedrobaeza, line 9: Please add: "Second level menus are collapsed by default".

web_menu_collapsible/README.rst
```diff
@@ -0,0 +1,59 @@
...
+Configuration
+=============
+
+No configuration is needed.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/8.0
+
+Bug Tracker
```
pedrobaeza, line 36: Add a section "Known issues / Roadmap" indicating: "There's no visual hint about the collapsed/uncollapsed state of a menu on second level menus"

web_menu_collapsible/README.rst
dreispt, line 9: By default? It means that it can be disabled?

dreispt:
I find curious that the top menu also collapses. Is this configurable?
And the left menu displays on top of the top menu.
IMO it should be the other way around: the top menu should display over the left menu.

web_menu_collapsible/README.rst
pedrobaeza, line 9: No, I mean before clicking in them :wink: 

Do you have a better sentence for it?

dreispt, line 9: I see. I'm afraid I don't. But it feels unexpected to me.

ivantodorovich:
I suggest adding an arrow, just like collapsable submenus. 
And maybe loose the animation to be consistent with already collapsable menues in odoo.

Also, since user will probably click them a lot now, I'd add this to the css classes:

``` css
  -webkit-user-select: none;  /* Chrome all / Safari all */
  -moz-user-select: none;     /* Firefox all */
  -ms-user-select: none;      /* IE 10+ */
  user-select: none;          /* Likely future */  
```

ivantodorovich:
@dreispt I believe you are testing something else. I've also noticed *-all database in runbot has something installed that does something like you're describing, but it's not this module!

ehahouimohssine:
:+1:

dreispt:
We already have two different  modules to hide menus: `web_hideleftmenu` and `web_menu_autohide`. 😕 
Do we need three different modules for such a similar feature?
What's the difference between each one of them? 
What prevents us from having a single, configurable, module?

tarteo:
@dreispt web_hideleftmenu hides the entire left menu and web_menu_autohide auto hides (on hover) and shows the top menu and left menu. This module enables the user to expand and collapse menu items in the left menu. All modules do something with either the left or right menu, but all do something else. I agree that all those modules should have a configurable option that turns the feature off or on, and that this configurable option should be an user preference. But I disagree with making one big 'god' module that contains all the menu features.

dreispt:
OK. I think I was tricked by the module description.
Now looking at it with the provided images I see what it does.

One suggestion: I miss a visual indication that the first level menus need to be clicked to open second level options. Is it possible to have an arrow or similar next to the first level options?

dreispt:
Looks perfect, thanks! :+1:

pankk:
Looking good! The only really minor thing I have is that the screenshots are not updated with the latest change (arrow icons) yet. Other than that :+1:

espo-tony:
This module gets the menu much more tidy. Good job! :+1:

andreschenkels:
Realy like this module :+1:

Issue #314, opened 2016-02-24
[8.0] Window action target `new` -- Readonly option


codingforfun:
Currently popups opened via `ir.actions.act_window` using option `target='new'` always open in edit mode. This make perfect sense for wizards, but sometimes we also need to show form view of objects linked somehow to another object. E.g. by triggering a button on a form view returning a window action. In this case the popup shouldn't open in edit mode. 

Currently there seam to be no standard way of doing this. 
As a workaround I've seen the solution of creating a custom form view with each field set to read only explicitly (module [mail_full_expand](https://github.com/OCA/social/blob/8.0/mail_full_expand/views/mail_full_expand.xml)). However this needs a lot of boiler plate code. A better solution would be if it would work with the models generic form view. 

After reviewing the code involved, I see two options to solve this pragmatically. 
1.) Add a new target option, e.g. `target='popup'` that behaves like `new` but opens in `view` mode instead of `edit` mode. This involves very few changes to JS code. 

2.) Add `readonly` option to `ir.actions.act_window`. This would be a more generic solution, but also would involve more code changes. 

I would really like to hear your opinions about that.

codingforfun:
@pedrobaeza @hbrunn 
I'm thinking about implementing this, because I need for a current project. What do you think about this? Which approach would you prefer? Thx.

codingforfun:
@OSguard 
ping

hbrunn:
if you have control over the actions, you can pass this in the `flags` dictionary of the action (https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/views.js#L374) which will be merged at some place with the view's `options` dictionary, so you should be able to simply write `'flags': {'initial_mode': 'view'}`: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_form.js#L116

But this doesn't work for `ir.action.act_window`s living in the database, in that case, you could use a server action that returns the the action with the flags you want.

So if you need any coding at all, I'd suggest to do a variation of (2) where you add a serialized field called `flags` that allows people to set any of those. This will also make it simple to add more stuff in the future.

codingforfun:
> so you should be able to simply write 'flags': {'initial_mode': 'view'}

@hbrunn 
Nice idea. Unfortunately this isn't working because it gets overridden here:
https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/views.js#L697

[...]

PR #326, opened 2016-03-16
New module 'web_one2many_list_action'

codingforfun:
Enable tree_but_open interface to be used for list views embedded into
form views for one2many relational fields.
My use case was to delegate opening a pop up view to another model. I.e. the line model is kind of a proxy where I do want to open the target model on click instead of the line model. 
But I think this can be useful for other circumstances too.


pedrobaeza:
Wow, this is amazing! I'll review it soon. Thanks for bringing this to OCA.

@oihane @anajuaristi, this is a clean alternative to https://github.com/odoomrp/odoomrp-utils/tree/8.0/sale_order_line_form_button, which has some problems I pointed out as you know.

codingforfun:
I hacked this very fast for my use case, so testing and input how this applies to the general case is very welcome.
One issue I already know is that actions with `target='new'` always open in edit mode (we discussed this here: https://github.com/OCA/web/issues/314) This is only loose related and not caused by my code, but you should be aware of if you want to do things similar to my use case and open popups of other models. I solved this by adding special flags into the computed action.

hbrunn:
:+1:

OSguard:
:+1: now can be merged

PR #329, opened 2016-03-17
[ADD] Added module web_draw

tarteo:


pedrobaeza:
This seems very similar to https://github.com/OCA/web/tree/8.0/web_widget_digitized_signature. Can you check?

tarteo:
With web_widget_digitized_signature you can select a signature (picture) from file system. With web_draw you can draw a signature (or any other drawing depends on context). So it's not similar.

pedrobaeza:
Why do you close it?

tarteo:
It's very similar.

pedrobaeza:
OK, thanks. Maybe if this one has some benefits or features, you can merge them in the other widget. The other also is for v8, and this one is for v9. You can make a mix of them and propose it for v9.

PR #331, opened 2016-03-22
[ADD] support color_field attribute

hbrunn:
This allows you to use some (probably function) character field as foreground color for rows in tree views. I don't think it's a perfect match to have this in this module, but we already have two being busy with colors in tree views, so I think adding it here is better than adding a third module.

The first commit is just a bit of cleanup because the module broke inheritance unnecessarily, the second contains the actual functionality and the last is some messing with the manifest. I figure this makes reviews simpler, I'll be happy to squash the commits once review is done.

### DIFF
--- a/web_tree_dynamic_colored_field/README.rst
+++ b/web_tree_dynamic_colored_field/README.rst
@@ -64,7 +64,7 @@ Usage
     ...
 
     With this example, the content of the field named `color` will be used to
-    populate the `color` CSS calue. Use a function field to return whichever
+    populate the `color` CSS value. Use a function field to return whichever
     color you want depending on the other record values. Note this this
     overrides the `colors` attribute, and that you need the tree to load your
     field in the first place by adding it as invisible field.
--- a/web_tree_dynamic_colored_field/README.rst
+++ b/web_tree_dynamic_colored_field/README.rst
@@ -4,14 +4,20 @@ Colorize field in tree views
 This module aims to add support for dynamically coloring fields in tree view
 according to data in the record.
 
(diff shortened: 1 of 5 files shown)


web_tree_dynamic_colored_field/__openerp__.py
```diff
@@ -19,9 +19,10 @@
 #
 ##############################################################################
 {
-    'name': 'web tree dynamic colored field',
+    'name': 'Colorize field in tree views',
+    'summary': 'Allows you to dynamically color fields on tree views',
     'category': 'Hidden',
-    'version': '8.0.1.0.0',
+    'version': '8.0.1.0.1',
```
pedrobaeza: As this includes a big new feature, I would put it 8.0.2.0.0

hbrunn: thanks, done

pedrobaeza:
:+1: (code review)

web_tree_dynamic_colored_field/README.rst
```diff
@@ -45,7 +51,23 @@ Usage
...
+        <tree string="View name" color_field="color">
+            ...
+            <field name="color" invisible="1" />
+            ...
+        </tree>
+    </field>
+    ...
 
+    With this example, the content of the field named `color` will be used to
+    populate the `color` CSS calue. Use a function field to return whichever
```
guewen: typo `s/calue/value/`

guewen:
:+1: code review

hbrunn:
@pedrobaeza @guewen can we merge this?

PR #332, opened 2016-03-23
[FIX] autodelete mails

hbrunn:
not autodeleting the mails is harmful if you send the mails to the same instance, for examlpe if you happen to have your issue tracker in the same database. Then the incoming mail will be ignored because odoo already knows the message id


pedrobaeza:
:+1:

yvaucher:
:+1:

guewen:
:+1:

PR #333, opened 2016-03-24
[RFR] in order to work properly, we need to use FieldOne2Many's caching logic

hbrunn:
while looking into https://github.com/OCA/commission/pull/78, I realized this widget is fundamentally flawed because it doesn't apply caching and directly writes into the database. This is different from standard one2many fields, so it should be adapted to use the same caching mechanism.

Mind you that this is basically a rewrite, so test thoroughly before merging.

### DIFF
--- a/web_widget_one2many_tags/static/src/js/web_widget_one2many_tags.js
+++ b/web_widget_one2many_tags/static/src/js/web_widget_one2many_tags.js
@@ -5,71 +5,137 @@
 openerp.web_widget_one2many_tags = function(instance)
 {
     openerp.web_widget_one2many_tags.FieldOne2ManyTags =
-    instance.web.form.FieldMany2ManyTags.extend({
+    instance.web.form.FieldOne2Many.extend(instance.web.form.ReinitializeFieldMixin, {
+        template: "FieldOne2ManyTags",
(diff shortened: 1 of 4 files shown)


guewen:
:+1: code review

pedrobaeza:
It seems to not work according https://github.com/OCA/commission/pull/78#issuecomment-202889021

jjscarafia:
Tested with https://github.com/OCA/commission/pull/78 and works great!
:+1:

pedrobaeza:
:+1:

web_widget_one2many_tags/static/src/js/web_widget_one2many_tags.js
```diff
@@ -4,97 +4,208 @@
...
+                        column.id + '__display',
+                        _(names)
+                            .map(function(name)
+                            {
+                                return name.display_name ||
+                                    instance.web._t('New record');
+                            })
+                            .join(', ')
+                    );
+                    console.log(record.get(column.id + '__display'));
```
codingforfun, line 204: Did you left a debug message?

hbrunn, line 204: thanks, what a beginner's mistake

PR #335, opened 2016-03-31
[9.0]Add web_duplicate_visibility

petrus-v:


web_duplicate_visibility/__openerp__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# © 2016 Pierre Verkest <pverkest@anybox.fr>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "web_duplicate_visibility",
```
legalsylvain: Could you set a more readable name. (without '_') ?

web_duplicate_visibility/demo/res_users_view.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<odoo>
+     <record id="view_users_form" model="ir.ui.view">
+        <field name="inherit_id" ref="base.view_users_form"/>
+        <field name="name">res.users.form</field>
```
legalsylvain: FYI, this line is useless. if this field is not here, odoo will generate automatically the name as 'model.name' + 'view.type'. here : res.users.form

web_duplicate_visibility/static/src/js/duplicate_visibility.js
```diff
@@ -0,0 +1,44 @@
...
+
+    var DuplicateVisibility = FormView.include({
+        /**
+         * Instantiate and render the sidebar if a sidebar is requested
+         * Sets this.sidebar
+         * @param {jQuery} [$node] a jQuery node where the sidebar should be inserted
+         * $node may be undefined, in which case the FormView inserts the sidebar in a
+         * div of its template
+         **/
+        render_sidebar: function($node) {
```
hbrunn, line 18: you replace super here, this breaks modularity. Implement your behavior by removing the button afterwards, or by patching `is_action_enabled`. But you need to call super.

legalsylvain, line 18: [Duplicate comment with @hbrunn]

Hi, I'm not sure if it's possible but I see that you're overwriting this function (1), instead of overloading and calling _super_. Do you think it's possible ?
something like (if I understood correctly your changes) : 
- call super ;
- remove_items 'Duplicate' if condition is not fulfilled ;

If it's not possible, could you just write a mention saying which line(s) you changed ?

(1) https://github.com/odoo/odoo/blob/9.0/addons/web/static/src/js/views/form_view.js#L204

web_duplicate_visibility/__openerp__.py
petrus-v: yes sure!

web_duplicate_visibility/demo/res_users_view.xml
petrus-v: Thanks, I didn't know!

web_duplicate_visibility/static/src/js/duplicate_visibility.js
petrus-v, line 18: Your are right guys! I think to both of those solutions, I don't know why I kept this one... maybe to avoid a redraw, my bad...

petrus-v:
@hbrunn, @legalsylvain 

Thank you very much for reviews! Last commit should fix the matter!

web_duplicate_visibility/README.rst
```diff
@@ -0,0 +1,83 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
+Web duplicate visibility
+========================
+
+This module allow to manage the visibility of duplicate button from the form
```
pedrobaeza: s/allow/allows

web_duplicate_visibility/README.rst
```diff
@@ -0,0 +1,83 @@
...
+view declaration.
+
+Usage
+=====
+
+While the default behavior of odoo is to display the duplicate button when user
+is allowed to create a new object. Your are now able to remove duplicate button
```
pedrobaeza: s/Your/You

web_duplicate_visibility/README.rst
```diff
@@ -0,0 +1,83 @@
...
+explicitly even if he is able to create new object::
```
pedrobaeza: s/he is/you are

pedrobaeza: Remove double `:`

danimaribeiro:
runbot :+1:

atchuthan:
runbot :+1:

hbrunn:
:+1:

PR #336, opened 2016-04-01
[FIX] preview image in README

bealdav:


web_dashboard_tile/README.rst
```diff
@@ -17,9 +17,9 @@ Usage
 =====
 
 * Dashboad sample, displaying Sale Orders to invoice:
-.. image:: web_dashboard_tile/static/src/img/screenshot_dashboard.png
+.. image:: /web_dashboard_tile/static/src/img/screenshot_dashboard.png
 * Tree view displayed when user click on the tile:
-.. image:: web_dashboard_tile/static/src/img/screenshot_action_click.png
+.. image:: /web_dashboard_tile/static/src/img/screenshot_action_click.png
 
 Kown issues/limits
```
pedrobaeza, line 24: Please take the occasion to put this correctly: s/Kown/Known

pedrobaeza:
With this change, are the images viewed on the apps?

bealdav:
@legalsylvain said me it's ok with both

pedrobaeza:
OK then :+1:

legalsylvain:
:+1:

petrus-v:
Sorry it doesn't appear to works on runbot... It wasn't before as well! so it's at least better on github
you get my :+1: 

But we should definitively find out a way to make it works in both cases

bealdav:
As far I know, it's here than rst is better markdown:
- rst don't need to prefix img tag with local folder (here module name)
- markdown needs local folder, and it's boring

It's seems rst converter on apps and on odoo is broken :(

PR #337, opened 2016-04-02
[MIG] web_searchbar_full_width : migrate to v9.0

bud-e:
Migrated `web_searchbar_full_width` in `v9.0`


web_searchbar_full_width/views/templates.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<openerp>
+<odoo>
     <data>
```
hbrunn: the data element can go away in v9

web_searchbar_full_width/static/src/xml/web_searchbar_full_width.xml
```diff
@@ -1,15 +1,8 @@
...
-            jQuery(this).attr('colspan', 4);
-        </t>
-        <t t-jquery="tr.oe_header_row_top td:nth-child(2)">
-            var $new_row = jQuery('<tr class="oe_header_row"/>');
-            jQuery(this)
-            .parents('tr')
-            .after($new_row);
-            $new_row.append(jQuery(this).attr('colspan', 4));
+    <t t-extend="ControlPanel">
+        <t t-jquery="div.oe-cp-search-view" t-operation="replace">
```
hbrunn: don't replace the whole element, just add the class(es) you need

web_searchbar_full_width/__openerp__.py
```diff
@@ -35,12 +36,7 @@
     "qweb": [
         'static/src/xml/web_searchbar_full_width.xml',
     ],
-    "test": [
-    ],
     "auto_install": False,
```
hbrunn: also remove this and `application`

web_searchbar_full_width/static/src/css/web_searchbar_full_width.css
```diff
@@ -1,20 +1,4 @@
-.openerp .oe_application .oe_breadcrumb_item:not(:last-child)
-{
-    max-width: inherit;
-}
```
hbrunn, line 4: this made the bread crumb full width, you probably need something similar as below for `div.oe-cp-title` in v9. As it is now, the breadcrumb is only half the width of the search bar

web_searchbar_full_width/static/src/xml/web_searchbar_full_width.xml
bud-e: @hbrunn  , As this is template inheritance... So, we have this [type of operations](https://www.odoo.com/documentation/9.0/reference/qweb.html#template-inheritance) , In that I've not found type that only adds class to element. So, I replace whole element.
And if I'm wrong then correct me...   Thanks... :)

hbrunn: use no operation at all and write the class attribute via jquery as in the code you removed:

``` xml
<t t-jquery="div.oe-cp-search-view">
    jQuery(this).removeClass('col-md-6');
    jQuery(this).addClass('col-md-12');
</t>
```

replacing nodes is never a good idea because you don't know which addons that possibly work on them are installed and what they do to the node.

bud-e: Ok. Thanks for guide, but can we go like this way... As in my latest commit i removed this whole `xml` file. And added `width=100%` style on both `breadcrumb & search-view`.
Revert me, if anything goes wrong... thanks..

hbrunn:
:+1:

pedrobaeza:
Thanks for the contribution.

:+1: 

Have you signed the CLA?

bud-e:
@pedrobaeza , yes I sent the mail with attached CLA to `cla@odoo-community.org`

pedrobaeza:
CLA checked, thanks.

danimaribeiro:
:+1: Tested on runbot

PR #339, opened 2016-04-06
[FIX][web_option_auto_color] Compatibility with MS Browsers.

adrienpeiffer:


hbrunn:
what is edge's problem with the original? If jquery operations don't work there, don't we need to fix this on a more fundamental level?

adrienpeiffer:
For information, this problem is also on IE. I thinks it's better and more clean to put `t-operation="attributes"`

image001

hbrunn:
okay, thanks. In this specific case, I think `position="attributes"` is a better choice anyways. But do I understand correctly that embedded javascript in QWeb never works in IE?

For this change: :+1:

pedrobaeza:
Following @hbrunn's expert approval, then my :+1:

jabibi:
I did not test it, but looking at the code seems correct :+1:

adrienpeiffer:
Thanks !

PR #340, opened 2016-04-08
[DEL] printf debugging

hbrunn:


pedrobaeza:
:+1: Fast-tracking as basic patch.

PR #346, opened 2016-04-18
9.0 web export view rv

ghost:


legalsylvain:
Hi, @rv-clearcorp, 

thanks for porting this module !

One Suggestion : What about to simplify UI, removing "Excel" button, and rename the button "Export Current View" into "Export To Excel". Users will gain a click ? 
CC  : @lepistone, @StefanRijnhart, @eLBati 

If not, could you fix the little padding / margin bug you can see on both screenshots ?

Otherwise LGTM, Thanks !

good-padding

bad padding

ghost:
Hi, @legalsylvain, 
CC : @lepistone, @StefanRijnhart, @eLBati

thanks for your observations, this already done:
- fixed the padding
- refactored the JS to use require

screenshot from 2016-04-20 12-36-18

regards.

likaiyuan:
image

eLBati:
Thanks @rv-clearcorp 
looking at the [diff with version 8](http://pastie.org/10835813) it seems some translations are missing.

Maybe this PR does not include all v8 changes? Please see https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#if-the-module-already-exists-in-the-90-branch-but-it-doesnt-contain-all-the-commit-history-or-youre-not-sure-if-it-have-it

web_export_view/view/web_export_view.xml
```diff
@@ -2,7 +2,7 @@
 
 <openerp>
     <data>
-        <template id="assets_backend" name="web_export_view assets" inherit_id="web.assets_backend">
+        <template id="assets_backend" name="web_export_view_assets" inherit_id="web.assets_backend">
```
elicoidal, line 5: why this change?

web_export_view/__openerp__.py
```diff
@@ -32,9 +16,6 @@
     'data': [
         'view/web_export_view.xml',
     ],
-    'qweb': [
-        'static/src/xml/web_export_view_template.xml',
-    ],
-    'installable': False,
+    'installable': True,
```
elicoidal, line 19: I think it is not necessary as this is the default value.

pedrobaeza, line 19: Better to let that value when migrating branches.

simahawk:
I'm testing this on OCB version

```
commit 57a1ad08b40b7d4621ba7496f5a90bf09b97794a
Merge: ad0ffd0 1d1debf
Author: OCA Git Bot <oca-git-bot@odoo-community.org>
Date:   Tue Jun 7 00:30:19 2016 +0200
```

and is not working: you get no records in the XLS. This is due to some changes in list view markup.

Here's the fix https://github.com/abstract-open-solutions/web/commit/8d36178a09be45c2d27131352423ce8596e26cbf

Also, looking for `'.o_list_view > tbody > tr[data-id]'` let you skip the check for "data-id" in the loop.

simahawk:
Ok, I found out why it was not working for me: I tested it on OCB + enterprise and they have slightly differences in the markup. I made a fix for it and now is working for both versions.

@rv-clearcorp would you mind to take a look (and integrate if working) my latest 2 commits here?

https://github.com/abstract-open-solutions/web/commits/web-export-view-v9

ghost:
Hi! @simahawk i will test and pull your changes :+1:

mcassuto:
@simahawk : hello I tested your version and I confirm, it works on v9 Enterprise.

However :-1:  because all the numbers arrive in the excel documnent as presented on the screen.
For instance, on the account_invoice tree view, the price 1320.56, is shown as "1 320.56 $"; I get the same thing in Excel instead of a cell with a number; which makes any computation impossible in Excel.

I believe most people who export to Excel, do it to manipulate numbers....

simahawk:
@mcassuto I don't get your point. The module just takes the raw text content of each cell in the HTML table, there's no formatting happening on client/server side AFAIK. So, if you get a currency symbol - or any other symbol that prevent numbers recognition - in cells' content then the problem is in the content of the HTML cell. If that's the case we can check if we can avoid it.
Am I missing something?

mcassuto:
@simahawk you got the point; the excel file comes with the currency symbol; it is not the raw text.
I even had one cell with the excel escape character  => '0.00

simahawk:
I'm sorry, I didn't look at the code since a while and indeed there are some conversions on client side based on css class. Unfortunately in the enterprise version they dropped classes like "oe_list_field_float" (there are no class at all, no one!) so the conversion is not happening and cell values are taken as bare text. 

I guess we can split this issue for enterprise into a new ticket tough, and maybe start lobbying Odoo SA for adding back those classes. Meanwhile we could override listing templates.
What do you think?

mcassuto:
@simahawk no offense ;-) I understand it may be a little awkward to block a feature because it does not work on EE. Also, doing that we will dig the gap between both versions...
Maybe we are off topic here. In fact, I am not the one to decide OCA's direction...

web_export_view/view/web_export_view.xml
StefanRijnhart: Well, it is an arbitrary change but it is non blocking for me. It does not impede inheritance because inheritance should refer to the template id, not the name.

StefanRijnhart:
So can @simahawk's changes be pulled now?

StefanRijnhart:
And then document in the README that the module is actually working on Enterprise, but with limitations.

pedrobaeza:
Any update on this?

rafaelbn:
This module is also being migrated to v10 here: https://github.com/OCA/web/pull/444

simahawk:
@rv-clearcorp can you pull my fixes and report the issue for enterprise version in the README?

pedrobaeza:
@simahawk, better that you open a new PR with your PR included and we close this one. There's also an interesting change to button in #444 that can be considered to be used here.

simahawk:
@pedrobaeza https://github.com/OCA/web/pull/461

pedrobaeza:
Closing this one then

PR #348, opened 2016-04-26
[ADD] web_ir_actions_act_window_none

hbrunn:
# 
# Noop action

This module was written to have an action that does nothing. A notorious case
where this is useful is buttons on popup windows, because they need to return
an action. Given the default here is `ir.actions.act_window_close`, you'll
have to jump through some hoops if you have a button that is not supposed to
close the popup window.
# Usage

To use this module, depend on it in your module and return::

```
{'type': 'ir.actions.act_window.none'}
```

The button will trigger a reload of the popup's form, in case you want to
suppress this, set key `reload` to `False` in the action you return.


hbrunn:
now let's hope I didn't do something stupid and there's a simpler way to keep wizard popups open on button clicks...

lasley:
Nice this issue is plaguing me as well, so I'm pretty sure you didn't do anything stupid. Demo would be nice, but I won't block for it mainly because I can't think of a quick strategy to do so while also keeping it obvious to the tester. 👍

pedrobaeza:
:+1: as well. Let's merge it after a while.

PR #349, opened 2016-04-28
[ADD] add module web_offline_warning

tremlin:
This PR adds a new module which replaces the generic browser error message when an RPC fails by a more user-friendly error message.

### DIFF
--- a/web_offline_warning/README.rst
+++ b/web_offline_warning/README.rst
@@ -1,0 +1,39 @@
+Offline warning message
+=======================
+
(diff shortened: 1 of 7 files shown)


petrus-v:
:+1: 

Thanks for this addition, code reviewed. Actually on runbot it's a bit annoying as far I understand an other module tried to get an information from server while opening the dialog (`web_dialog_size` ?) causing the same error, opening a new dialog etc... but I don't think it's the matter of this module but the other one...

Thanks to the throttling feature on chromium dev tools ;)

hparfr:
👍 Good idea

PR #351, opened 2016-05-04
[ADD] add input radio in a tree view inside a form

kaerdsar:


web_widget_radio_tree/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Web widget radio tree
+=====================
+
+This module allows to use input radio in a tree view inside a form, in order to ensure the user marks only one record.
```
hbrunn, line 9: judging from the code, you also provide a form widget, so this seems an unnecessary restriction to me

web_widget_radio_tree/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Example: You have a partner company form with many contacts. The contacts are shown in a tree and you want to specify only one as preferred.
+
+Usage
+=====
+
+In the view declaration, put widget='radio_tree' attribute in the field tag::
```
hbrunn: put the code-like stuff in double ``````

web_widget_radio_tree/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "Web Widget Radio Tree",
+    'summary': """Add radio buttons for records in tree.""",
```
hbrunn: no need for multiline string

web_widget_radio_tree/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
hbrunn: huse short headers: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

web_widget_radio_tree/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+{
+    'name': "Web Widget Radio Tree",
+    'summary': """Add radio buttons for records in tree.""",
+    'author': "bloopark systems GmbH & Co. KG",
```
hbrunn: add OCA: https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L10

web_widget_radio_tree/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+        'web',
+    ],
+    'data': [
+        'views/assets.xml',
+    ],
+    'qweb': [
+        'static/src/xml/widget.xml',
+    ],
+    'installable': True,
+    'auto_install': False,
```
hbrunn: you set those two to the default value - is there a good reason to do so or can you just remove them?

kaerdsar: 'installable' and 'auto_install'? I can remove them.

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,90 @@
+openerp.web_widget_radio_tree = function (instance) {
+
+    var QWeb = instance.web.qweb;
+
+    instance.web.list.columns.add('field.radio_tree', 'instance.web.list.RadioTreeColumn');
+
+    instance.web.list.RadioTreeColumn = instance.web.list.Column.extend({
+        _format: function (row_data, options) {
+            var name = options.model + '_' + this.id;
```
hbrunn: why not letting qweb taking care of the formatting and passing the whole options object to the template? And row_data too just for good measure.

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,90 @@
...
+            return _.template(
+                '<input type="radio" name="<%-name%>" <%-checked%> readonly="readonly"></input>', {
```
hbrunn: make this a qweb template for extensibility

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,90 @@
...
+
+    instance.web.form.RadioTree = instance.web.form.FieldBoolean.extend({
+        template: 'RadioTree',
+        start: function() {
+            var self = this;
+            this.$checkbox = $('input', this.$el);
+            var radio_name = this.__parentedParent.model + '_' + this.$checkbox[0].name;
+            this.$checkbox.attr('name', radio_name);
+            this.setupFocus(this.$checkbox);
+            this.$el.click(_.bind(function() {
```
hbrunn: use https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L404 and  https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L805

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,90 @@
...
+                self.clean_radio_in_records();
+                this.internal_set_value(true);
+            }, this));
+            var check_readonly = function() {
+                self.$checkbox.prop('disabled', self.get('effective_readonly'));
+                self.click_disabled_boolean();
+            };
+            this.on('change:effective_readonly', this, check_readonly);
+            check_readonly.call(this);
+            this._super.apply(this, arguments);
```
hbrunn: return `super()`'s result

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,90 @@
...
+        },
+        click_disabled_boolean: function(){
+            var $disabled = this.$el.find('input[type=radio]:disabled');
+            $disabled.each(function (){
+                $(this).next('div').remove();
+                $(this).closest('span').append($('<div class="boolean"></div>'));
+            });
+        },
+        clean_radio_in_records: function() {
```
hbrunn, line 40: this looks quite complicated. why doesn't an override of https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_form.js#L2039 work?

kaerdsar, line 40: I have to set value not only for the current record in the tree. When you click in a radio button all record should set the value to false and the current one to true. In this point is not clear how to access to the instances of the other records in the tree. Then I inspected the parent. If you have any other idea I would appreciate it.

web_widget_radio_tree/__openerp__.py
hbrunn: yes

web_widget_radio_tree/static/src/js/widget.js
hbrunn, line 40: ah, I see. Then I think you do the right thing in principle, but it could be a bit more elegant in my opinion:
- use https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L229 instead of the private field
- you use properties specific to `BufferedDataSet`, this way, your widget is not going to work with many2many fields. Better read all ids and call `write`, the implementation of the specific dataset will take care to do the right thing hopefully

kaerdsar, line 40: Thanks, I'm going to check your comments and try to improve the implementation.

kaerdsar, line 40: @hbrunn done! easier with dataset.write()

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -20,22 +20,16 @@ openerp.web_widget_radio_tree = function (instance) {
     instance.web.form.RadioTree = instance.web.form.FieldBoolean.extend({
         template: 'RadioTree',
         start: function() {
-            var self = this;
+            this._super.apply(this, arguments);
```
hbrunn: it's really crucial to return super's result, because this is usually a deferred object and it can mess up stuff seriously to just drop it: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L653

web_widget_radio_tree/static/src/js/widget.js
kaerdsar: @hbrunn done! updated to use qweb template.

kaerdsar:
@pedrobaeza could you please take a look to this PR?

web_widget_radio_tree/README.rst
pedrobaeza: Indicate that the field type is `Boolean`

pedrobaeza:
Tried in runbot and works flawlessly!

Minor remark, but :+1:

kaerdsar:
Thanks, documentation updated.

kaerdsar:
@hbrunn if you have a chance to re-check, please let me know if I should improve something else, thanks.

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,51 @@
...
+    instance.web.form.RadioTree = instance.web.form.FieldBoolean.extend({
+        template: 'RadioTree',
+        start: function() {
+            var _super = this._super.apply(this, arguments);
+            this.$checkbox = $('input', this.$el);
+            var radio_name = this.getParent().model + '_' + this.$checkbox[0].name;
+            this.$checkbox.attr('name', radio_name);
+
+            var self = this;
+            this.$el.click(_.bind(function() {
```
hbrunn: use the proxy function: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L805

web_widget_radio_tree/static/src/js/widget.js
```diff
@@ -0,0 +1,51 @@
...
+            var $disabled = this.$el.find('input[type=radio]:disabled');
+            $disabled.each(function (){
+                $(this).next('div').remove();
+                $(this).closest('span').append($('<div class="boolean"></div>'));
+            });
+        },
+        clean_radio_in_records: function() {
+            var parent = this.getParent();
+            var name = (this.$checkbox[0].name).split('_')[1];
+            _.each(parent.dataset.ids, function(id) {
```
hbrunn, line 43: can't you write all of them at once?

kaerdsar, line 43: The write method expects an id, not an array.

hbrunn, line 43: then use the `call` function to call write on a list of ids

kaerdsar, line 43: @hbrunn if I understand well you suggest to call directly the write method of the model (creating an RPC call). Now when we use the write of the BufferedDataSet there is not RPC call, only at the end when the user saves the form. Then maybe we call a write method once but do an RPC call each time the user click on a radio button in a record. Please, let me know if I'm missing something.

hbrunn, line 43: i was missing that this are memory writes. thanks for the explanation, then my comment is moot

hbrunn:
:+1:

antespi:
LGTM :+1:

lmignon:
:+1:  LGTM (Code review only) @kaerdsar Nice feature!

web_widget_radio_tree/static/src/js/widget.js
kaerdsar, line 43: Thanks for the review, it was very useful. I could improve the code.

PR #290, opened 2016-01-13
[MIG]Migrated web tree image in v9.

JayVora-SerpentCS:
Migrated web tree image in v9.

PR #355, opened 2016-05-14
web_tree_image ported to Version 9.0

docmfried:
Modifications to web_tree_image.js  to reflect the new Odoo 9.0 Javascript API.

### DIFF
--- a/web_tree_image/__openerp__.py
+++ b/web_tree_image/__openerp__.py
@@ -24,7 +24,7 @@
 ##############################################################################
 {
     'name': 'Show images in tree views',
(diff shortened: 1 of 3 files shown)


pedrobaeza:
There's already another migration proposed in https://github.com/OCA/web/pull/290. As this one is more recent, it has less priority than the other. Can you please check and review the other, and maybe improve it if there's something in your PR that the other doesn't have?

docmfried:
Thanks for pointing out. I checked and reviewed the mentioned migration proposed in #290 and all my changes are covered in the older one and no need to improve. I will close my PR and go for for PR #290.

PR #356, opened 2016-05-15
[MIG] web_dialog_size: Migrate to v9.0

bud-e:


eLBati:
Thanks :+1:

web_dialog_size/__openerp__.py
```diff
@@ -28,11 +28,13 @@
...
-    'author': "ACSONE SA/NV,Odoo Community Association (OCA)",
-    'website': "http://acsone.eu",
+    'author': "ACSONE SA/NV, "
+              "Odoo Community Association (OCA), "
+              "Siddharth Bhalgami",
 
+    'website': "http://acsone.eu",
     'category': 'web',
-    'version': '8.0.0.1.0',
+    'version': '9.0.0.1.0',
```
pedrobaeza: 9.0.1.0.0

pedrobaeza:
You have missed the latest 8.0 commits. In concrete, this one: https://github.com/OCA/web/commit/0578405b8b3406930c0edf132c1475de174a097d.

You need to follow this guide for the migration: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0, but in this case, as there's only one missing commit, you can only cherry-pick the commit. It's a very important one, because it allows to store the default window size.

lmignon:
:+1:  (Code review only)

pedrobaeza:
This still misses the latest commit, so please don't merge until then.

bud-e:
@pedrobaeza ,
I already cherry-pick this 2 commits https://github.com/OCA/web/commit/0578405b8b3406930c0edf132c1475de174a097d & https://github.com/OCA/web/commit/2238950e58366942db31f0a9c4c16581ee24061b from 8.0
So, I think this branch is updated. 

Thanks.

bud-e:
@pedrobaeza ,
I already cherry-pick this 2 commits https://github.com/OCA/web/commit/0578405b8b3406930c0edf132c1475de174a097d & https://github.com/OCA/web/commit/2238950e58366942db31f0a9c4c16581ee24061b from 8.0
So, I think this branch is updated. 

Thanks.

web_dialog_size/static/src/js/web_dialog_size.js
```diff
@@ -1,37 +1,41 @@
...
+Dialog.include({
+
+    init: function (parent, options) {
+        var self = this;
+        this._super.apply(this, arguments);
+        self.$modal.find('.dialog_button_extend').on('click', self.proxy('_extending'));
+        self.$modal.find('.dialog_button_restore').on('click', self.proxy('_restore'));
+
+        new Model('ir.config_parameter').query(['key', 'value']).
+        filter([['key', '=', 'web_dialog_size.default_maximize']]).all().then(function(default_maximize) {
```
sbidoul, line 16: Why not use the get_param method if ir.config_parameter as in v8? This should be more efficient than querying then filtering.

pedrobaeza:
I have tried the module and it works well, and make a quick test with get_param and it didn't work, so it seems the JS API has changed. We can merge this as is.

PR #357, opened 2016-05-15
New form view field web widget timepicker 

docmfried:
New form view field widget "timepicker" based on jquery timepicker plugin to ease user input.


web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+==============================
+Timepicker widget in form view
+==============================
+
+This module defines a timepicker widget, to be used with either char fields
+or (function) fields of type character. Use ``widget='timepicker'`` in your form view
```
dreispt: I find the char field choice strange: usually a Float field is used to enter time, and a widget `float_time` exists.
Why change this standard?

docmfried: The motivation orginated out of the UI experience comparing to the datetimepicker widget for date/datetime input.

timepicker_widget

The existing float_time widget has limited UI support to assist the end user. The implemented time handling and persistance could be changed from Char to Float within the javascript code. Currently the Field Widget is based on AbstractField, so we are open for any changes.

PS: BTW and out of topic - Compliments for your great Oddo development essentials. book - it got me started.

dreispt: I agree this widget is relevant. But from your explanation I believe that it would be best for it to work on float fields, and be an alternative to the standard float_time widget.

(Thanks, I'm glad my book helped you)

web_widget_timepicker/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
lasley: Would you mind using the short headers please? https://github.com/OCA/maintainer-tools/blob/master/template/module/controllers/__init__.py#L3

web_widget_timepicker/__openerp__.py
```diff
@@ -0,0 +1,51 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': '',
+    'version': '0.1',
```
lasley: Please use version numbers as described here - https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

This would be `9.0.1.0.0`

web_widget_timepicker/__openerp__.py
```diff
@@ -0,0 +1,51 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+
+{
+    'name': '',
```
lasley: Please add a name

web_widget_timepicker/static/src/js/jquery.timepicker.js
```diff
@@ -0,0 +1,1225 @@
+/*!
```
lasley: Might be better to use the CDN version of this instead of including directly. CDNJS has this library + assets avail - https://cdnjs.com/libraries/jquery-timepicker

web_widget_timepicker/static/src/js/timepicker_widget.js
```diff
@@ -0,0 +1,89 @@
+odoo.define('timepicker.form_widgets', function (require) {
```
lasley: I could be wrong here, but I believe it is SOP to namespace the JS modules the same as the actual module. `web_widget_timepicker.form_widgets` or similar

lasley:
Thanks for this submission, this is definitely something I can use. I agree with @dreispt regarding the use of a float fields as this is both more efficient, and allows for correct sorting.

docmfried:
Thanks for your feedback and the comments to improve the code quality. I will change the formal stuff as pointed out by @lasley. In addition I will change the internal handling from char to float as proposed by @dreispt.
As being a newcomer please allow me the question - what is the "correct" way to contribute the upcoming changes within this pull request. Is it sufficient to push the changes to my fork or is anything more needed? Thanks in advance.

lasley:
Hi @docmfried - Yup, just push the changes to the same branch & let us know when it's ready for review again

dreispt:
@docmfried Make sure you check the TravisCI log: there are a few lint issues to fix.

docmfried:
Hi @dreispt, @lasley ,
I committed some changes and improvements to reflect the formal errors and deviations from the OCA standards. Internally I changed the field persisance from char to float. The TravisCI check complains about Javascript usage of eval, which I use to convert the "custom" options from the form field record defintion to the actual timepicker widget object initialisation. Any recommendations or further remarks to improve the code?

lasley:
Hi @docmfried - Thanks for the changes, my issues have been addressed. Regarding the eval, we definitely can't merge with that as it is a security issue. 

I would recommend using `data-options` in the field instead, as it is the SOP for storing data in elements:

```
<field name="mytimefieldname" widget="timepicker" data-options="{'step': '15', 'disableTextInput': false}"/>
```

You can use jQuery to turn into an object:

```
var widgetOptions = $(this.node).data('options');
```

Just wrote the aboves on the fly, so may need a bit of tweaking, but should get you there. The jQuery `.data` operator automatically converts your data to an object for you. 

Alternatively, `JSON.parse(string)` works for string representations of objects

docmfried:
Hello @lasley, thanks for your advise. I changed from eval() to jquery data function and as a fallback to JSON.parse. As the user provided 'data-options' string may contain the wrong string format for the conversion (especially the "" and '' combinations) I added a cleanup function 'cleanup_str2json' to handle that situation. Any further comments or remarks to improve the current code?

lasley:
Looks good. Just noticed though that the external library is still included though. Would you mind switching out the timepicker JS and CSS to the CDN version so that we don't have to maintain it in this repo? https://cdnjs.com/libraries/jquery-timepicker

lasley:
Can just remove the files, then switch the URI in the assets file

docmfried:
Hi @lasley ,
I can see your point concerning the plugin assets, but I included the plugin source based on the following reasoning.
1. Self contained widget stable against external plugin changes
2. Independent of online connectivity during usage of application (keep the odoo installation offline and still be able to use the widget)
3. The existing plugins in the current odoo code base, like Datepicker, are also included in the existing addons.

If you prefer the external URI usage I will change the addon, but in my opinion the stability of the included timepicker sources outweights maintainibility by using the reference to external CDN.

hbrunn:
You should _never_ rely on external resources if this is not an explicit part of the plugin's purpose (like the google integration et al). Ther are a _lot_ of companies who use Odoo inhouse with a restrictive firewall, not to start about the privacy/security implications of pulling code from sources not under your control. That's perfectly fine for the next hipster website, but we build business applications here.

You should however, give credit in the appropriate section, which happens here.

web_widget_timepicker/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+        'static/src/xml/web_widget_timepicker.xml'
+    ],
+
+    # always loaded
+    'data': [
+        'views/web_widget_timepicker_assets.xml'
+    ],
+
+    # Installation options
+    "installable": True,
```
hbrunn: this one is not strictly necessary, as it's the default

pedrobaeza: Please don't omit it, or the script for the branch migration (in v10) might fail and create an invalid manifest. What I would remove is the comments about each of the keys.

web_widget_timepicker/static/src/js/web_widget_timepicker.js
```diff
@@ -0,0 +1,115 @@
...
+                disableTextInput: true,
+                forceRoundTime: true,
+                step: 15,
+                selectOnBlur: true,
+                timeFormat: 'H:i',
+                scrollDefault: 'now',
+            });
+        },
+        initialize_content: function() {
+            if(!this.get("effective_readonly")) {
```
hbrunn, line 29: why do you jump through all this hoops when the framework parses the `options` attribute (without `data-`) for you as json already? Compare with for example https://github.com/OCA/OCB/blob/9.0/addons/web/static/src/js/widgets/progress_bar.js#L31

docmfried, line 29: Hi @hbrunn,
thanks for your input. I agree totally with you that the external plugin source should be part of the repo to enable the application functionality without an internet connectivity.
The problem with the JSON string to object parsing based on the data-options attribute is the user dependent usage of " or ' within the field defintion. 
`data-options="{'step': '15', 'disableTextInput': false}"` could not be parsed with jquery data function.
but 
`data-options='{"step": "15", "disableTextInput": false}'` could be parsed with jquery data function.or JSON.parse

Thus the testing and the 2 phased handling of the user provided input during the initialize_content function.

hbrunn, line 29: I think you can stick to the standard `options` attribute, and as added bonus to my knowledge, the runtime also takes care of that. At least that worked in 8.0

hbrunn, line 29: here's an example of what I mean: https://github.com/OCA/web/tree/8.0/web_widget_datepicker_options

web_widget_timepicker/static/src/js/web_widget_timepicker.js
```diff
@@ -0,0 +1,115 @@
...
+                this.$input.val(show_value);
+            } else {
+                this.$(".oe_form_time_content").text(show_value);
+            }
+        },
+    });
+
+    core.form_widget_registry.add('timepicker', TimePicker);
+
+    return TimePicker;
```
hbrunn: make it a habit to return an object with memebers pointing to you references. That's for future extensibility, if you release this to the wild, everyone uses it and expects this one object, you'll break everything if for whatever reason you want to offer a second class. Not probable with this module, but a good pattern to follow.

web_widget_timepicker/static/src/js/web_widget_timepicker.js
docmfried, line 29: That was the addon as an example, which I initially tried to follow as an template, but could not get it working under version 9.0 as my odoo develop / target platform. The options were passed as `this.node.attrs.options` as string and not as an object. I debugged the whole thing under Firefox and the proposed code was the only working solution I found. But my Javascript skills are somewhat limited, so maybe I missed something in that context.

hbrunn, line 29: use `this.options`. I looked up the code for you: https://github.com/OCA/OCB/blob/9.0/addons/web/static/src/js/views/form_common.js#L660 - so this should work.

web_widget_timepicker/static/src/js/web_widget_timepicker.js
docmfried: Honestly I don't understand technically your comment. I checked the progressbar Widget as a comparsion, which also returns the Progressbar object. Please could you provide an example to check on?

hbrunn: it's a very theoretical comment, so I won't block you on ignoring it. This is what I mean:

Imagine my addon adds whatever functionality to the time picker. Then I say in my module:

``` javascript
[...]
var timepicker = require('web_widget_timepicker.form_widgets');
timepicker.include({
// hackhackhack
});
```

But later on, you add a second object you want to expose, say a column implementation (which I'll ask you to do actually, because otherwise, editable tree views show the default date widget in readonly mode). Now you'd have to return an object:

``` javascript
[...]
return {
    form_widget: TimePicker,
    tree_column: TheTreeColumn,
}
```

which makes my code break. So to play safe, just return your class wrapped into an object in the first place, then this problem won't arise.

hbrunn: btw: For small modules like yours, I think it's better to only use the module's name, your extra `.form_widgets` seems a bit too much. `web` does this because it's huge.

web_widget_timepicker/static/src/js/web_widget_timepicker.js
docmfried, line 29: Okay I undestand now based on the provided code example. I will try that. This also clearifies your statement about 

> returning object with memebers pointing to you references.

Thanks

web_widget_timepicker/static/src/js/web_widget_timepicker.js
docmfried: @hbrunn : That a lot. Your hint about "pyeval" to handle the options helped a great deal and made the code more elegant  (I hated this part myself). Also I changed the code according your advise about the return value - makes much sense - plus I renamed the module to the suggested name.
@pedrobaeza I removed the needless comments in the `__openerp__.py` file.

I will commit the changes once I am back online tonight.

hbrunn: but mind you that you don't need to do the pyeval yourself, you inherit from AbstractField, so you get this for free. Just use the `options` attribute instead of `data-options`

lasley: `data-options` was my bad advice, TIL for me as well. Thanks @hbrunn

web_widget_timepicker/__openerp__.py
hbrunn: will it? Then I need to adapt my reviewing, because I always ask people to remove keys with the default value.
Could you point me to the code that does the migration moving? Then I'll PR resilience against this key not existing.

web_widget_timepicker/README.rst
```diff
@@ -24,22 +23,21 @@ Usage
...
+See the available options at https://github.com/jonthornton/jquery-timepicker#timepicker-plugin-for-jquery.
 
 
-ToDo
-====
+Known issues / Roadmap
+======================
 
-Sanity check on options available in field defintion as override options for timepicker widget.
+* Absolutely no sanity check or validation on options.
```
hbrunn: I think this might confuse users, because it documents the development process, not the outcome

hbrunn:
:+1:

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+==============================
+Timepicker widget in form view
+==============================
+
+This module defines a timepicker widget, to be used with float fields
+or (function) fields. Use ``widget='timepicker'`` in your form view
+definition. It can be use as a replacement for the standard float_time widget.
```
hbrunn: s/use/used

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+If you use the widget with a field record, the input field has the following default
+timepicker options:
+
+* By default the possible selection is based on 15 minute interval (step: 15)
+* By default 24 hour mode with H:i format (timeFormat: 'H:i')
+* By default scroll selection starts at current time (scrollDefault: 'now')
```
hbrunn: remove 'By default ' because you introduce all of this as default values

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+The widget uses the jquery.timepicker plugin by Jon Thornton
+
+
+Usage
+=====
+
+This module defines a new widget type for form views input fileds.
```
hbrunn: remove this, you said that before

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+Set the attribute ``widget=timepicker`` in a ``field`` tag in a form view.
+
+You can pass custom options through the "timepicker" field in the options attribute:
```
hbrunn: s/through/to

lasley:
👍  here as well

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+    ...
```
hbrunn, line 45: remove the colons and double the colon after `attribute` above: http://docutils.sourceforge.net/docs/user/rst/quickstart.html#preformatting-code-samples

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+    <field name="mytimefieldname" `widget=timepicker`` options="{'step': '30', 'disableTextInput': false}"/>
+    ...
+
+See the available options at https://github.com/jonthornton/jquery-timepicker#timepicker-plugin-for-jquery.
```
hbrunn: make this a proper link: http://docutils.sourceforge.net/docs/user/rst/quickref.html#hyperlink-targets

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+Jon Thornton (https://cdnjs.com/libraries/jquery-timepicker)
+jquery.timepicker plugin - This software is made available under the open source MIT License. © 2014 Jon Thornton and contributors
+
+Odoo Community Association (OCA)
+
+
+Contributors
+------------
+
+* Michael Fried
```
hbrunn: add a spam-hardened mail address if you can

web_widget_timepicker/__openerp__.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# © 2016 Michael Fried @ Vividlab (<http://www.vividlab.de>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Web Timepicker Widget',
+    'version': '9.0.1.0.0',
+    'author': 'Michael Fried@Vividlab, Odoo Community Association (OCA)',
```
hbrunn: this is the first time I see you coming around, so maybe make this only your company? What you fill in here will show up in the app store, so for visibility, you'll probably want to choose one unique name for all the modules you contribute and not employee@company

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
```
pedrobaeza: Put target link

web_widget_timepicker/README.rst
```diff
@@ -0,0 +1,55 @@
...
+definition. It can be use as a replacement for the standard float_time widget.
+
+If you use the widget with a field record, the input field has the following default
+timepicker options:
+
+* By default the possible selection is based on 15 minute interval (step: 15)
+* By default 24 hour mode with H:i format (timeFormat: 'H:i')
+* By default scroll selection starts at current time (scrollDefault: 'now')
+
+The widget uses the jquery.timepicker plugin by Jon Thornton
```
pedrobaeza: Put this in the credits section

pedrobaeza:
A nicer icon would be welcome with the aspect of the widget.

web_widget_timepicker/README.rst
hbrunn: I didn't understand Pedro's comment in a PR before until I looked at the current README template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

hbrunn:
@pedrobaeza arent you usually the wizard that PRs great sodipodi files for that?

web_widget_timepicker/__openerp__.py
pedrobaeza: The problem would be with a manifest that doesn't include the final comma in the last key. Here you can see how I inject the `'installable': False` value in the manifest: https://github.com/OCA/maintainer-tools/blob/master/tools/migrate_branch.py#L167

pedrobaeza:
Yeah, I know, hehe, but as this time is more a screenshot than an artwork, I prefer if he can provide it.

docmfried:
Hello,
thanks for the postive feedback and the corrections on the non functional stuff to polish the PR. I am working on the needed changes. 
@pedrobaeza .I will add image(s) / screenshot(s) of the timepicker widget in action embedded in the README.rst file like . .. |picker| image:: ./images/picker.png and /or . |formview| image:: ./images/form_view.png as in the color widget addon,. Ok?
@hbrunn: Yes this is my first time contribution (I am feeling like a virgin :-). I will shed some thoughts about the unique name and the company branding topic. Thanks for the hint.

pedrobaeza:
@docmfried, thanks for the excellent contribution. I wish all the newcomers contribute with these superb modules.

docmfried:
Hello @pedrobaeza ,
thanks for the feedback. I am happy to contribute this module and maybe there will be more in the future ...
I added some screenshots to illustrate the widget usage and an custom icon for the module - don't if this is appropiate.
Hopefully the manifest is now correct. I checked it against the template contained in maintainer-tools. Please let me know if there is anythiny to fix or improve.

docmfried:
Hello @hbrunn,
Hello @pedrobaeza,
I am wondering if I missed a notification or whether you are just busy. If there is anything to improve in my last commit please let me now.

hbrunn:
the second one. :+1:

lasley:
👍

PR #362, opened 2016-05-24
Fix issue with bad position of Version div

qtheuret:
support_branding_issue
When support_branding module is installed and support_branding.release parameter is set, the scrollbar of the menu is moved and the 'Version' indication is not put below the 'Supported by' statement. You can see this on the attached image.


leemannd:
Thanks :+1:

pedrobaeza:
:+1:

lmignon:
:+1:  (tested on runbot)

PR #363, opened 2016-05-25
[IMP] Send mail internally for email links in kanban records

codingforfun:
Extend for email fields in kanban records (especially for res.partner)

### DIFF
--- a/web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
+++ b/web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
@@ -34,7 +34,6 @@ openerp.web_widget_mail_send_odoo = function(instance) {
             if (!self.get("effective_readonly") && 0 < self.$("input").size() && self.$("input").val()) {
                 try {
                     var parsed_email = instance.mail.ChatterUtils.parse_email(self.$("input").val());
-                    console.log(parsed_email)
                     if(!parsed_email[1])
                         return false;
                     return true;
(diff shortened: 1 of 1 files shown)


OSguard:
@hbrunn can we get a review from you?

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
 
     instance.web.form.widgets.add('email_mailto', 'instance.web.form.FieldEmail')
     instance.web.form.widgets.add('email', 'instance.web.form.FieldEmailIntern')
+
+    instance.web_kanban.FieldEmailInternKanban = instance.web_kanban.AbstractField.extend({
+        init: function(parent, field, $node) {
+            this.parent = parent;
```
hbrunn, line 109: you don't need this. the actual parent is set in https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L569. And what you want to keep here is the reference to `this.field_manager`, which another super does already: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_form.js#L1815

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
...
+            this._super.apply(this, arguments);
+            this.format_descriptor = _.extend({}, this.field, {'widget': this.$node.attr('widget')});
```
hbrunn, line 111: can this be anything else than `email`?

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
...
+                    if(!parsed_email[1])
+                        return false;
+                    return true;
+                } catch(e) {
+                    return false;
+                }
+            }
+            return true;
+        },
+        on_clicked: function() {
```
hbrunn, line 128: shouldn't you cancel the event so that the browser doesn't run its default action?

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
...
+            console.log("Kanban Email Widget ---- Click");
```
hbrunn, line 129: remove printf debugging

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
...
+                self.do_warn(_t("E-mail Error"), _t("Can't send email to invalid e-mail address"));
+            } else {
+                // find partner id for email
+                var res_partner = new openerp.Model('res.partner');
+                var parsed_email = instance.mail.ChatterUtils.parse_email(self.get('value'));
+                res_partner.query(['id'])
+                    .filter([['email', '=', parsed_email[1]]])
+                    .first().then(function(partner){
+                        if(partner){
+                            self.do_action(
```
hbrunn, line 141: return this. Then other code can hook into the deferred object that will be returned most of the time

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
...
+                                'mail.action_email_compose_message_wizard', {
+                                    additional_context: {
+                                        default_partner_ids: [partner.id],
+                                        default_composition_mode: 'comment',
+                                        // write to active model:
+                                        default_model: self.parent.view.dataset.model,
```
hbrunn, line 147: use `self.field_manager`

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
```diff
@@ -104,4 +103,67 @@ openerp.web_widget_mail_send_odoo = function(instance) {
...
+                                }
+                            );
+                        } else {
+                            self.do_warn(_t("E-mail Error"),
+                            _t("Can't send email:  No partner configured for email address."));
+                        }
+                    })
+            }
+        },
+        renderElement: function() {
```
hbrunn, line 159: better use a qweb template and attach your handler in the events dictionary as in https://github.com/OCA/server-tools/pull/480/files#diff-9de652377e95844f51139004242e0cfcR10 - then the framework takes care of this, you can remove the whole function and make the addon simpler to customize

web_widget_mail_send_odoo/static/src/js/web_widget_mail_send_odoo.js
codingforfun, line 141: Thx. I'll check your comments after my holidays.

PR #366, opened 2016-05-27
[FIX] web_dialog_size: do not expand by default and correctly honor default_maximize parameter

### LINKED ISSUE #365: [8.0] web_dialog_size default maximize has no effect
Apparently, the README of web_dialog_size is inconsistent.

It says both `By default, all dialog boxes are expanded.` and `By default, the module respects the caller's dialog_size option. If you want to override this and have all dialogs maximized by default, set the configuration parameter web_dialog_size.default_maximize to 1.`.

In my test, dialogs are expanded by default, but `web_dialog_size.default_maximize` seems to have no effect.

@pedrobaeza @hbrunn any opinion/advice on this?

sbidoul:
Fixes #365

### DIFF
--- a/web_dialog_size/README.rst
+++ b/web_dialog_size/README.rst
@@ -2,8 +2,7 @@ Expand Dialog
 =============
 
 A module that lets the user expand/restore the dialog box size through a button
-in the upper right corner (mimicking most windows managers). By default,
-all dialog boxes are expanded.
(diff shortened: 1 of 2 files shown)


hbrunn:
:+1:

lmignon:
:+1:

elicoidal:
LGTM

PR #368, opened 2016-06-02
web_company_logo_unclickable module

rami-wafaie:


ntodorova:
:+1:

web_company_logo_unclickable/static/src/xml/base.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<templates id="template" xml:space="preserve">
+    <t t-extend="WebClient">
+        <t t-jquery="a[class=oe_logo]" t-operation="replace">
```
hparfr, line 5: Why not just remove the _src_ attribute of **a** ?

hbrunn, line 5: your mean the `href` attribute, but yes, I think that's much cleaner

hparfr, line 5: Correct, I mean href instead of src. ;-)

rami-wafaie, line 5: I tried to, but I need to keep "Edit Company Data" the way it is, just remove the clickable link from the logo, so any ideas to make it better?

hbrunn, line 5: untested from the top off my head:

``` xml
<t t-jquery="a[class=oe_logo]">
jQuery(this).removeAttr('href');
</t>
```

PR #370, opened 2016-06-06
[ADD] adding module web_invalid_tab

kaerdsar:
Highlights a tab when fields inside are invalid.

### DIFF
--- a/web_invalid_tab/README.rst
+++ b/web_invalid_tab/README.rst
@@ -1,0 +1,45 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+Web invalid tab
+===============
+
+This module highlights a tab when fields inside are invalid. Is useful when you have a form with many tabs.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/web/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+web/issues/new?body=module:%20
+web_invalid_tab%0Aversion:%20
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Cesar Lage <kaerdsar@gmail.com>
+* Robert Rübner <rruebner@bloopark.de>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/web_invalid_tab/README.rst
+++ b/web_invalid_tab/README.rst
@@ -6,7 +6,7 @@
 Web invalid tab
 ===============
 
-This module highlights a tab when fields inside are invalid. Is useful when you have a form with many tabs.
+This module highlights a tab when fields inside are invalid. It's useful when you have a form with many tabs.
 
 Bug Tracker
 ===========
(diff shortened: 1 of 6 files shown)


bealdav:
Tested on runbot. It works, thanks
:+1:

web_invalid_tab/README.rst
```diff
@@ -0,0 +1,45 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============
+Web invalid tab
+===============
+
+This module highlights a tab when fields inside are invalid. Is useful when you have a form with many tabs.
```
antespi: s/Is useful/It's useful/

antespi:
Tested in runbot :+1: 
Thanks, this is a cool addon for improvement UX

web_invalid_tab/static/src/js/view_form.js
```diff
@@ -0,0 +1,62 @@
...
+            if (!this.disable_utility_classes) {
+                if (this.field_manager.get('display_invalid_fields')) {
+                    this.$el.toggleClass('oe_form_invalid', !this.is_valid());
+                    this._check_invalid_tab();
+                }
+            }
+        },
+        _check_invalid_tab: function() {
+            var tab = this.$el.closest(tab_selector);
+            if (tab && tab.attr('id')) {
```
hparfr: tab is always a jQuery object, you can then do simply:
`if (tab.attr('id'))`

web_invalid_tab/static/src/js/view_form.js
```diff
@@ -0,0 +1,62 @@
+openerp.web_invalid_tab = function(instance) {
+
+    var tab_selector = 'div[role="tabpanel"]';
+
+    function tab_link(tab) {
+        return $("a[href='#" + tab.attr('id') + "']");
+    }
+
+    function is_visible(tab, element) {
+        if ($(element).hasClass('oe_form_invisible') || element.style.display == 'none') {
```
hparfr, line 10: $(element).is( ":hidden" ) is more expressive

web_invalid_tab/static/src/js/view_form.js
```diff
@@ -0,0 +1,62 @@
...
+                this.$el.find('.oe_field_translate').toggle(this.field_manager.get('actual_mode') !== "create");
+            }
+            if (!this.disable_utility_classes) {
+                if (this.field_manager.get('display_invalid_fields')) {
+                    this.$el.toggleClass('oe_form_invalid', !this.is_valid());
+                    this._check_invalid_tab();
+                }
+            }
+        },
+        _check_invalid_tab: function() {
```
hparfr: if the name is __check__invalid_tab then we are expecting an idempotent function but we can find {remove|add}Class.

web_invalid_tab/static/src/js/view_form.js
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+        }
+    });
+
+    instance.web.FormView.include({
+        on_form_changed: function() {
+            this._super();
+            $(tab_selector).each(function(i, tab) {
+                var invalid = _.detect($(tab).find('.oe_form_invalid'), function(x) {
```
hparfr, line 52: you can may be achive a similar result wih better perfs using .css selectors

hparfr:
Usefull module. Thanks

web_invalid_tab/static/src/js/view_form.js
kaerdsar, line 10: @hparfr yes is more expressive, but if a required field is inside a tab which is not active then I will get that the field is hidden. Then I can not use ':hidden' or ':visible' because I need to know if the field is visible when the tab is active. That's the reason why I created the function is_visible. Do you have any other idea?

hparfr, line 10: I'm not sure to understand well what you want.

Let's say there is field well visible (and empty) in the current tab (tab 1).
I then switch to another tab (tab 2).
This field will not change, and will not be visible anymore. it's tab 2 which will have his property display set to none.
is:visible gives you this info without the need to traverse the DOM and query each layer

kaerdsar, line 10: Try this case: Open an Order form, which has 3 tabs. In the third tab there is a required field. When you open the form the first tab is active, you fill the fields that you see and try to save. I have to highlight the third tab even the user never touch it (activate it), because inside has a field that should be filled.

kaerdsar, line 10: @hparfr could you please check the last comment? I would like to know if I can improve the code in some way. Thanks

kaerdsar:
@hbrunn could you please take a look at this PR?

bealdav:
test are ok, somebody to merge ?

chafique-delli:
test are ok
:+1:

hbrunn:
:+1:

kaerdsar:
@pedrobaeza @bealdav should I update something more here? or is ready to merge?

lasley (approved):
👍  from me, code no test. 

Can someone check CLA status? If this is something I can also do, please point me in a direction instead of doing it for me 😉

bealdav:
@kaerdsar just sign CLA, thanks

legalsylvain:
CLA is OK.

PR #376, opened 2016-06-23
do not completely overide List.render and List.render_cell in web_tree_dynamic_colored_field module

suvit:
more correct add colorize function to qweb context


hbrunn:
I did basically the same thing as a side effect in https://github.com/OCA/web/pull/331

hbrunn:
Thanks for your input, but I'll close this one

PR #379, opened 2016-07-01
[ADD] web_widget_float_time_second: new module

agaldona:
*New widget(float_time_second) for floats with HH:MM:SS format.
*New configuration check(Float time (HH.MM.SS)) to add seconds to float_time widget.

There is a problem in tree views, the first time you load a view with float_time widget and you have selected the check in configuration it only loads time with seconds in the first line. In order to show time with seconds in other lines you have to select them.

### DIFF
--- a/web_widget_float_time_second/README.rst
+++ b/web_widget_float_time_second/README.rst
@@ -1,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+========================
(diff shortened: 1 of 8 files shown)


dreispt:
> There is a problem in tree views, the first time you load a view with float_time widget and you have selected the check in configuration it only loads time with seconds in the first line. In order to show time with seconds in other lines you have to select them.

This sounds like a bug.
Can't we instead propose a fix?

PR #386, opened 2016-07-20
[9.0][MIG] web_environment_ribbon

antespi:
- Migration to v9
- Compliance to OCA guidelines

### DIFF
--- a/web_environment_ribbon/data/ribbon_data.xml
+++ b/web_environment_ribbon/data/ribbon_data.xml
@@ -8,5 +8,17 @@
             <field name="value">TEST</field>
         </record>
 
+        <!-- Add ribbon color configuration parameter -->
+        <record id="set_ribbon_color" model="ir.config_parameter">
+            <field name="key">ribbon.color</field>
+            <field name="value">#f0f0f0</field>
+        </record>
+
+        <!-- Add ribbon background color configuration parameter -->
+        <record id="set_ribbon_background_color" model="ir.config_parameter">
+            <field name="key">ribbon.background.color</field>
+            <field name="value">rgba(255,0,0,.6)</field>
+        </record>
+
     </data>
 </openerp>
--- a/web_environment_ribbon/data/ribbon_data.xml
+++ b/web_environment_ribbon/data/ribbon_data.xml
@@ -1,24 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
-<openerp>
-    <data noupdate="1">
+<!-- Copyright 2015 Francesco OpenCode Apruzzese <cescoap@gmail.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo noupdate="1">
 
-        <!-- Add ribbon name default configuration parameter -->
-        <record id="default_ribbon_name" model="ir.config_parameter">
-            <field name="key">ribbon.name</field>
-            <field name="value">TEST</field>
-        </record>
+<!-- Add ribbon name default configuration parameter -->
+<record id="default_ribbon_name" model="ir.config_parameter">
+    <field name="key">ribbon.name</field>
+    <field name="value">TEST</field>
+</record>
 
-        <!-- Add ribbon color configuration parameter -->
-        <record id="set_ribbon_color" model="ir.config_parameter">
-            <field name="key">ribbon.color</field>
-            <field name="value">#f0f0f0</field>
-        </record>
+<!-- Add ribbon color configuration parameter -->
+<record id="set_ribbon_color" model="ir.config_parameter">
+    <field name="key">ribbon.color</field>
+    <field name="value">#f0f0f0</field>
+</record>
 
-        <!-- Add ribbon background color configuration parameter -->
-        <record id="set_ribbon_background_color" model="ir.config_parameter">
-            <field name="key">ribbon.background.color</field>
-            <field name="value">rgba(255,0,0,.6)</field>
-        </record>
+<!-- Add ribbon background color configuration parameter -->
+<record id="set_ribbon_background_color" model="ir.config_parameter">
+    <field name="key">ribbon.background.color</field>
+    <field name="value">rgba(255,0,0,.6)</field>
+</record>
 
-    </data>
-</openerp>
+</odoo>
--- a/web_environment_ribbon/static/src/css/ribbon.css
+++ b/web_environment_ribbon/static/src/css/ribbon.css
@@ -1,7 +1,5 @@
 .test-ribbon{
     width: 200px;
-    background: #e43;
-    position: absolute;
     top: 50px;
     left: -50px;
     text-align: center;
@@ -16,7 +14,7 @@
     z-index: 1000;
     position: fixed;
     box-shadow: 0 0 3px rgba(0,0,0,.3);
-    background: rgba(255, 0, 0, .6);;
+    background: rgba(255,0,0,.6);
 }
 
 .test-ribbon b {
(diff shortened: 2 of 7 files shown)


rafaelbn:
Hi :+1: thanks, just functionally tested

web_environment_ribbon/static/src/js/ribbon.js
```diff
@@ -1,35 +1,57 @@
...
+/* Copyright 2015 Sylvain Calador <sylvain.calador@akretion.com>
+   Copyright 2015 Javi Melendez <javi.melendez@algios.com>
+   Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
 
-    This program is free software: you can redistribute it and/or modify
-    it under the terms of the GNU Affero General Public License as
-    published by the Free Software Foundation, either version 3 of the
-    License, or (at your option) any later version.
+(function ($) {
```
lasley: This is the old JS API, please upgrade to v9

web_environment_ribbon/data/ribbon_data.xml
```diff
@@ -1,4 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2015 Francesco OpenCode Apruzzese <cescoap@gmail.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <openerp>
```
lasley: This could be simplified as `<odoo noupdate="1">` in v9

web_environment_ribbon/static/src/css/ribbon.css
```diff
@@ -16,7 +17,7 @@
     z-index: 1000;
     position: fixed;
     box-shadow: 0 0 3px rgba(0,0,0,.3);
-    background: rgba(255, 0, 0, .6);;
+    background: rgba(255,0,0,.6);
```
lasley, line 20: TBH if we were going to alter the diff to do anything with the commas, it would be better to add commas to the line above and leave the ones that were here. Is there a specific reason for removing them?

web_environment_ribbon/view/base_view.xml
```diff
@@ -1,21 +1,28 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2015 Francesco OpenCode Apruzzese <cescoap@gmail.com>
+     Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <openerp>
```
lasley: This can be simplified as `<odoo>` w/ no `<data>` el necessary

lasley:
Thanks for the submission! Comments inline

pedrobaeza:
cc @Tecnativa

web_environment_ribbon/static/src/css/ribbon.css
antespi, line 20: Following guidelines to migrate v8 to v9 (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0), I've respected commits merged in v8 branch. Please @jabibi answer this question.

web_environment_ribbon/static/src/js/ribbon.js
antespi: I've followed this guidelines (https://github.com/OCA/maintainer-tools/blob/ca0941c654a4c319686aff9cd82b66ab0fb9a06b/template/module/static/src/js/module_name.js#L4) to migrate to v9.
I didn't know any new JS API in v9, could you indicate where to find it?

web_environment_ribbon/data/ribbon_data.xml
antespi: OK, done

web_environment_ribbon/view/base_view.xml
antespi: Done

pedrobaeza:
Tried on runbot. It's working as expected.

:+1:

web_environment_ribbon/static/src/css/ribbon.css
pedrobaeza, line 20: @lasley, this change was made by 8.0 contributors, but it appears here because we have brought 8.0 commits that are not in 9.0 yet.

rafaelbn:
@lasley could you take a look here? your comments have been processed. Thanks!

web_environment_ribbon/static/src/js/ribbon.js
lasley: Hmm yeah seems that just hasn't been updated; the v9 API uses a new requires system, which makes essentially all v8 JS incompatible.

The v9 API is described in the Odoo docs, the important portion starting at the [Module Overview](https://www.odoo.com/documentation/9.0/reference/javascript.html#javascript-module-system-overview) & an example can be seen [in website_field_autocomplete](https://github.com/OCA/website/blob/9.0/website_field_autocomplete/static/src/js/field_autocomplete.js#L5)

web_environment_ribbon/static/src/css/ribbon.css
lasley, line 20: Ah thanks for the clarification, no problem here then 😉

web_environment_ribbon/static/src/js/ribbon.js
pedrobaeza: I have tried and this is working as expected (changing colors through parameters), so maybe it's not so incompatible at all?

lasley: This must be an edge case due to no real dependencies then. 

I don't guess I'll block over this, but I think we may be setting a bad precedence by allowing old JS API in. I see zero examples of this old API in Odoo v9 code; they are all using the new style.

I'm still learning the true injection point for an app that should fire JS when the DOM is ready, but from what I understand the code should look similar to:

``` javascript
odoo.define('web_environment_ribbon.ribbon', function(require){
  "use strict";

  var $ = require('$');
  var Model = require('web.Model');
  var core = require('web.core');

  var model = new Model('ir.config_parameter');

  function validStrColour(strToTest) { ... }

  core.bus.on('web_client_ready', null, function () { Code to run here })
});
```

pedrobaeza: @Yajo, can you take a look on this?

antespi:
@lasley Thanks so much for your detailed sample. Now new JS API used in this addon. Please, check it out!

antespi:
runbot is running as expected. Please review again, JS file has changed. Thanks

antespi:
I think this is ready for review, please mark it as needsreview. Thanks

web_environment_ribbon/static/src/js/ribbon.js
yajo: Sorry, I have to agree with @lasley.

[`compatibility.js`](https://github.com/odoo/odoo/blob/9.0/addons/web/static/src/js/compatibility.js) makes this work, the same way as old api still works in v9, but an api upgrade is very desirable. The example a bove, together with [the docs](https://www.odoo.com/documentation/9.0/reference/javascript.html#web-client), are good start points.

pedrobaeza: The change has be done to new API, so no need to worry about. Good to know about compatibility.js.

pedrobaeza:
:+1: Tried again on runbot. Waiting for @lasley approval to merge this.

antespi:
Squashed for being merged

lasley:
Hooray it works! 👍

PR #389, opened 2016-07-22
[8.0][ADD] web_widget_image_download

kutyel:
# Web Widget - Image Download

This module was written to extend the functionality of the widget image and allows to download it.
# Usage

1. Go to the section `Contacts`.
2. Click on a contact.
3. Edit the contact.
4. Click the `Download` button (between `Edit` and `Clear`).

@Yajo @Tecnativa


pedrobaeza:
You can put a better icon using a screenshot of the usual icon size (128x128) showing the download button. If you let the default icon, there's no need of keeping the SVG file, as it's already in the template.

pedrobaeza:
CLA ready

web_widget_image_download/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
yajo: Replace with `web`

web_widget_image_download/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
yajo: LGPL

web_widget_image_download/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+===========================
+Web Widget - Image Download
+===========================
+
+This module was written to extend the functionality of the widget image and allow the download of the contact avatar.
```
yajo: s/widget image/image widget/

... and not all images are contact avatars. :wink:

web_widget_image_download/static/description/icon.svg
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
yajo: If you use the default OCA icon, you don't need to include its SVG.

web_widget_image_download/static/src/css/web_widget_image_download.css
```diff
@@ -0,0 +1,6 @@
+/* Copyright 2016 Flavio Corpa <flavio.corpa@tecnativa.com>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+.oe_form_binary_file_download {
+  color: #eee !important;
```
yajo: If `!important` is not required, it's best to avoid it.

web_widget_image_download/static/src/css/web_widget_image_download.css
```diff
@@ -0,0 +1,6 @@
...
+}
```
yajo: Always end files with newline.

web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2016 Flavio Corpa <flavio.corpa@tecnativa.com>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+ */
```
yajo, line 4: You can save a couple of LOC removing those dangling comment endings, [like in the template](https://github.com/OCA/maintainer-tools/blob/master/template/module/static/src/js/module_name.js#L1-L2). Not a blocker though.

web_widget_image_download/static/src/css/web_widget_image_download.css
kutyel: @Yajo you are right, I didn't want to include it, but there was another class from odoo that override my widget custom css, so I had to do it 😕

web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -0,0 +1,19 @@
...
+openerp.web_widget_image_download = function(instance) {
+    'use strict';
+
+    instance.web.form.web_widget_image_download = instance.web.form.FieldBinaryImage.include({
+        render_value() {
```
yajo, line 9: What's the current browser compatibility for this?

web_widget_image_download/static/src/xml/web_widget_image_download.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2016 Flavio Corpa <flavio.corpa@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<template>
+    <t t-extend="FieldBinaryImage">
+        <t t-jquery=".oe_form_binary_file_edit" t-operation="after">
+            <a class="fa fa-download fa-1g col-md-4 oe_form_binary_file_download" title="Download" download="avatar.png"></a>
```
pedrobaeza: Call it `image.png`, not `avatar.png`

web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -0,0 +1,19 @@
...
+            this._super();
```
yajo, line 10: Shouldn't you return this?

web_widget_image_download/README.rst
kutyel: @Yajo you want me to change all the licenses to LGPL? 😨

web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -0,0 +1,19 @@
...
+
+            // remove `col-md-offset-5` from following trash icon (improves visibility)
+            this.$el.find('.oe_form_binary_file_clear').removeClass('col-md-offset-5');
```
yajo: View replacements should be done through QWeb if possible.

web_widget_image_download/static/src/xml/web_widget_image_download.xml
pedrobaeza: You should translate this text. @Yajo, can you explain to him the method to translate templates?

web_widget_image_download/README.rst
antespi: All OCA addons are AGPL-3, no LGPL
Why LGPL, can you explain it?

web_widget_image_download/static/src/xml/web_widget_image_download.xml
yajo: What's this `download` attribute?

web_widget_image_download/README.rst
yajo: If you ever want to create an LGPL website addon (such as a theme) and you want to depend on this module, you'd be violating the license.

pedrobaeza: LGPL should be for website (frontend) modules, not web (backend) modules

web_widget_image_download/static/src/css/web_widget_image_download.css
yajo: Can't you just use a more specific selector to get higher priority?

web_widget_image_download/README.rst
pedrobaeza: Talking with @Yajo, he is right about the possibility of using this widget through https://github.com/OCA/website/tree/8.0/website_backend_views, and in this case, having an LGPL widget that can be reused in any frontend is a better option.

web_widget_image_download/static/src/xml/web_widget_image_download.xml
kutyel: @Yajo its the [standard HTML5 download attribute](http://www.w3schools.com/tags/att_a_download.asp), required to download the image.

web_widget_image_download/static/src/js/web_widget_image_download.js
kutyel, line 9: Quite [good](http://kangax.github.io/compat-table/es6/#test-object_literal_extensions), is the new _shorthand method_ for object literals. 😄

kutyel:
Thank you all for the feedback @Tecnativa! :+1:

web_widget_image_download/static/src/js/web_widget_image_download.js
kutyel, line 10: @Yajo I don't think so, the parent widget doesn't and it does not use it either.

web_widget_image_download/static/src/js/web_widget_image_download.js
antespi: QWeb Engine in JS side doesn't support attributes position. In this case, we must use replace. And I don't like replaces because they could break inheritance. In this case I think that JQuery manipulation is the safest way.

rafaelbn:
Tested un runbot :+1:

antoniocanovas:
Tested un runbot 👍

web_widget_image_download/README.rst
legalsylvain: Hi, 
I understand that for compatibility reasons, it is better to have LGPL, but contributors has the right to share their work with OCA in AGPL.

pedrobaeza: @legalsylvain, he is a new guy in our team (Tecnativa), so we are deciding publicly which license to use.

legalsylvain: OK, I did'nt know.

:+1:

web_widget_image_download/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/web/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
```
kutyel, line 36: Should have removed this section, the image was mine.

web_widget_image_download/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Flavio Corpa <flavio.corpa@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Web Widget - Image Download",
+    "summary": "Allows to download the avatar for the contact",
```
kutyel, line 6: This should be:

```
This module was written to extend the functionality of the widget image and allows to download it.
```

web_widget_image_download/views/assets.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Flavio Corpa <flavio.corpa@tecnativa.com>
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
+
+<openerp>
```
kutyel, line 5: I bet this will be `<odoo>` in v9.0?

web_widget_image_download/README.rst
pedrobaeza, line 36: Yeah, but you can make another PR to amend it. There are 2 more things that I have thought about this weekend:
- If there's no image, is there the button to download?
- If the image is not png, the extension shouldn't be png (for example, jpg or gif). We should detect the type. Is there an easy way for that?

web_widget_image_download/views/assets.xml
pedrobaeza, line 5: Yeah, and you can remove `<data>` also in that version

web_widget_image_download/README.rst
kutyel, line 36: Yes, @Yajo told me this weekend that an improvement would be to use odoo's `t-attr-` tags to retrieve the file name (and format) from the server and then add it dynamically.

web_widget_image_download/__openerp__.py
yajo, line 6: Actually the "this module was written to ..." part is more for the README, and lately just removed (all are written modules after all). However something like "Allows to download any image from its widget" seems better.

yajo:
Congratulations @kutyel! :tada:

PR #390, opened 2016-07-25
[8.0] Add web_graph_sort

ghost:


web_graph_sort/README.rst
```diff
@@ -0,0 +1,57 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to ...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/web/8.0
```
pedrobaeza: This link is not correct. Please see README template (https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst)

web_graph_sort/README.rst
```diff
@@ -0,0 +1,57 @@
...
+==============
+
+This module allows to sort pivot tables.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to ...
```
pedrobaeza: Please complete this

web_graph_sort/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+{
+    'name': "Web Graph Sort",
+    'version': '8.0.1.0.0',
+    'author': (
+        "NDP Systèmes, "
+        "Savoir-faire Linux, "
+        "Odoo Community Association (OCA)"
+    ),
+    'license': 'AGPL-3',
+    'category': 'Technical',
```
pedrobaeza: `web` for being consistent with other web modules

ghost: sure

pedrobaeza:
I have tried the module and the problem comes when you have multiple columns. For one column it works perfectly, but for multiple ones, the sort is made in all columns at the same time, but in a weird mode. Is it possible to make it only in one column?

ghost:
@pedrobaeza what you are asking is more complex. I agree it would be better. For now, I don't have the time to implement it that way. I can add this in Known Issues.

For now, the sorting is always done on the total of the row.

pedrobaeza:
OK, you should clarify then on the instructions and add it in the Known issues, because I don't see clear the exact sorting. Is it possible to put only the arrow on that column?

Anyway, good start and very appreciated!

ghost:
@pedrobaeza fixed, the arrow will now appear only on the total column.

Thanks for the review!

pedrobaeza:
Thanks for the changes

:+1:

web_graph_sort/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+This module allows to sort pivot tables.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to any pivot table. Click on a column header. The table will be sorted by that column.
+
```
rafaelbn, line 21: Maybe a user doesn't know what is a pivot table. You could clarify a lot saying Go To Reporting -> ....

rafaelbn:
Thanks for this contribution :+1: tested in runbot. Just one comment to clarify end users.

ghost:
@rafaelbn thanks for the review, I updated the README file.

PR #391, opened 2016-07-27
[8.0][IMP][web_widget_image_download] save image with the correct type

kutyel:
This improvement fixes the following points:
- If there's no image, is there the button to download?

**No, it disappears**.
- What happens if the extension of the image is not `png` (for example, `jpg` or `gif`).

**It detects the type**.
- Will this widget break inheritance from other widgets?

**No, cause it uses `JQuery` to safely modify the main _widget_, without replacing it**.

@Tecnativa

---

**To whom it might merge:** please merge this PR with the _squash_ option, in order to preserve a clean history in the remote repo.

### DIFF
--- a/web_widget_image_download/static/src/js/web_widget_image_download.js
+++ b/web_widget_image_download/static/src/js/web_widget_image_download.js
@@ -2,15 +2,36 @@
  * Copyright 2016 Flavio Corpa <flavio.corpa@tecnativa.com>
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
  */
-openerp.web_widget_image_download = function(instance) {
+openerp.web_widget_image_download = function (instance) {
     'use strict';
 
     instance.web.form.web_widget_image_download = instance.web.form.FieldBinaryImage.include({
         render_value() {
             this._super();
 
-            // take image URI from preceding <img> tag
-            this.$el.find('.oe_form_binary_file_download').attr('href', $('img[name="image"]').attr('src'));
+            const widget = this.$el.find('.oe_form_binary_file_download');
+
+            this.imgSrc = this.$el.find('img[name="image"]').attr('src');
+
+            $.ajax({
+                type: 'HEAD',
+                url: this.imgSrc,
+                complete(xhr) {
+                    // retrieve image type from server ("Content-Type" header)
+                    widget.attr('download', xhr.getResponseHeader("Content-Type").replace('/', '.'));
+                }
+            });
+
+            // use jquery instead of `replace` with qweb (to avoid breaking inheritance)
+            if (this.has_custom_image()) {
+                this.$el.find('.oe_form_binary_file_clear').removeClass('col-md-offset-5');
+            }
+
+            widget.attr('href', this.imgSrc);
+        },
+        has_custom_image() {
+            // check if the image of the widget is different from the default placeholder
+            return this.imgSrc && !this.imgSrc.includes('/placeholder.png');
         }
     });
 }
--- a/web_widget_image_download/static/src/js/web_widget_image_download.js
+++ b/web_widget_image_download/static/src/js/web_widget_image_download.js
@@ -16,7 +16,7 @@ openerp.web_widget_image_download = function (instance) {
             $.ajax({
                 type: 'HEAD',
                 url: this.imgSrc,
-                complete(xhr) {
+                complete: function (xhr) {
                     // retrieve image type from server ("Content-Type" header)
                     $widget.attr('download', xhr.getResponseHeader("Content-Type").replace('/', '.'));
                 }
--- a/web_widget_image_download/static/src/js/web_widget_image_download.js
+++ b/web_widget_image_download/static/src/js/web_widget_image_download.js
@@ -6,10 +6,10 @@ openerp.web_widget_image_download = function (instance) {
     'use strict';
 
     instance.web.form.web_widget_image_download = instance.web.form.FieldBinaryImage.include({
-        render_value() {
+        render_value: function () {
             this._super();
 
-            const widget = this.$el.find('.oe_form_binary_file_download');
+            var $widget = this.$el.find('.oe_form_binary_file_download');
 
             this.imgSrc = this.$el.find('img[name="image"]').attr('src');
 
@@ -18,7 +18,7 @@ openerp.web_widget_image_download = function (instance) {
                 url: this.imgSrc,
                 complete(xhr) {
                     // retrieve image type from server ("Content-Type" header)
-                    widget.attr('download', xhr.getResponseHeader("Content-Type").replace('/', '.'));
+                    $widget.attr('download', xhr.getResponseHeader("Content-Type").replace('/', '.'));
                 }
             });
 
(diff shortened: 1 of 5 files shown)


web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -2,15 +2,32 @@
...
         render_value() {
             this._super();
 
-            // take image URI from preceding <img> tag
-            this.$el.find('.oe_form_binary_file_download').attr('href', $('img[name="image"]').attr('src'));
+            const widget = this.$el.find('.oe_form_binary_file_download');
+
+            this.imgSrc = this.$el.find('img[name="image"]').attr('src');
+
+            $.get(this.imgSrc, (response, status, xhr) => {
```
yajo: Have you tried downloading only the HTTP headers, given that's all you need? http://stackoverflow.com/a/30891843/1468388

kutyel: I think there is no much of a difference, one is using `$.ajax` and the other `$.get` right away...

lasley: This would be a significant performance improvement, particularly when dealing with images. 

The major difference is GET vs HEAD. You are technically also using `.ajax`, just with a type of `GET`.

web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -2,15 +2,36 @@
...
+openerp.web_widget_image_download = function (instance) {
     'use strict';
 
     instance.web.form.web_widget_image_download = instance.web.form.FieldBinaryImage.include({
         render_value() {
             this._super();
 
-            // take image URI from preceding <img> tag
-            this.$el.find('.oe_form_binary_file_download').attr('href', $('img[name="image"]').attr('src'));
+            const widget = this.$el.find('.oe_form_binary_file_download');
```
lasley: It is best practice to prefix jQuery vars with `$` as it helps in terms of readability. In this instance `const $widget =`

web_widget_image_download/static/src/js/web_widget_image_download.js
```diff
@@ -2,15 +2,36 @@
...
+
+            this.imgSrc = this.$el.find('img[name="image"]').attr('src');
+
+            $.ajax({
+                type: 'HEAD',
+                url: this.imgSrc,
+                complete(xhr) {
```
lasley: `complete: function(xhr) {`

lasley:
Thanks for the submission @kutyel - few comments inline

One note, please try not to squash your original commits with ones in reference to pull request comments. I was reviewing your code, then it disappeared and I had to review a whole new set of code 😦

kutyel:
I am sorry @lasley in our company @Tecnativa it is customary to do so in order to preserve a clean history, right @pedrobaeza? :wink:

web_widget_image_download/static/src/js/web_widget_image_download.js
kutyel: Is that not kind of personal, or it is part of some odoo general guidelines? I like the recommendation, though :wink:

web_widget_image_download/static/src/js/web_widget_image_download.js
kutyel: As I already explained [here](https://github.com/OCA/web/pull/389/files#r71903339), that is perfectly valid _es2015_ JavaScript.

lasley: Apologies, this was due to the rebase. Comment got lost.

lasley:
I agree with a clean history & fully support squashing. This did destroy the conversation, require me to re-review, and require you to have to point out a comment that was clearly in the revision history before the rebase.

Maybe I am interpreting the contribution guidelines incorrectly, so please someone correct me if I am wrong, but they state to "make a single commit per logical change set". I would consider a PR Review to be an independent logical change set from the original submission. If we do not interpret as such we cannot determine what exactly was changed between the original and the review, thus a full code review is required all over again.

EDIT: 👍  code+functional btw

web_widget_image_download/static/src/js/web_widget_image_download.js
lasley: I guess it somewhat is because JS/jQuery people never really wrote down a style guide that I have been able to find; but it is incredibly common in JS/jQuery code (post ECMA5) to do this. I had some links with a bunch of arguments for and against, but can't seem to find at the moment. Not a blocker though.

pedrobaeza: I think you have a point. We can adapt the flow making 2 things:
- Commiting each time in a PR after a change done because of a review.
- Squashing on merge, for avoiding a dirty history when the PR arrives the repository.

lasley: This sounds like a solid plan to tackle the issue from both ends. 

An interesting way to do this would be a `git merge --squash`. Looks like Github supports this from the UI too - https://help.github.com/articles/about-pull-request-merge-squashing/

pedrobaeza: Indeed we can do it from GitHub UI... if the merger do it. Let's try to "educate" people then for doing so when needed. @lasley, I think you also deserve to be have rights to merge in some of the repositories you are working around. Please tell me which repositories you are interested and I'll propose you for being in those PSCs.

lasley: I agree with education, particularly because then it would be a question of exactly what the commit said - which should really be up to the committer IMO. I'll try to think of a good way to word this for the contribution docs.

Oh wow, I'm honored! I'll get back to you on which repos via email in the next week or so 😄

kutyel: @lasley that option from GitHub is just awesome! No more rebase/squash anymore! :tada:

yajo:
I tried on runbot.

I can see the HEAD request, but not the button, when I load the preferences window:

captura de pantalla de 2016-07-28 16-10-57

kutyel:
Is that the admin panel? I should give it a try 😨

yajo:
User preferences, from top right menu

web_widget_image_download/static/src/js/web_widget_image_download.js
yajo: Please make sure PhantomJS supports ES6 features before adding them.

web_widget_image_download/static/src/js/web_widget_image_download.js
yajo: Please make sure PhantomJS supports ES6 features before adding them.

web_widget_image_download/static/src/js/web_widget_image_download.js
lasley: This is a good point; IIRC Phantom barely even supports ES5. There is https://www.npmjs.com/package/karma-es6-shim but eff if I know how to load it into Odoo's test runner.

kutyel: That's right! Adding the `es6-shim` to odoo would allow everyone to use the latest JavaScript features without worrying about browser compatibility (as it is a pollyfill library).

yajo:
:+1:

lasley:
Changes look good 👍

pedrobaeza:
Thanks :+1: 

Waiting the merge of the patch!

PR #306, opened 2016-02-07
[8.0] Add web timeline view

lmignon:
Here it's a PR that demonstrate an alternative to the gant view to display temporal data. The new view is based on http://visjs.org/timeline_examples.html. The PR extends project to showcase the usage of this new addon by adding a timeline view on tasks.
image
Feel free to finalize this work if you need it...
IMO this kind of view can be a good candidate to porvide a replacement to the gant view in V9 freely available.

PR #393, opened 2016-08-01
[9.0] [MIG] web_timeline

pedrobaeza:
# 
# Timeline view

Define a new view displaying events in an interactive visualization chart.

The widget is based on the external library 
http://visjs.org/timeline_examples.html
# Configuration

You need to define a view with the tag <timeline> as base element. These are
the possible attributes for the tag:
- date_start (required): it defines the name of the field of type date that
  contains the start of the event.
- date_end (optional): it defines the name of the field of type date that
  contains the end of the event.
- date_delay (optional): it defines the name of the field of type date that
  contains the end of the event.
- default_group_by (required): it defines the name of the field that will be
  taken as default group by when accessing the view or when no other group by
  is selected.
- event_open_popup (optional): when set to true, it allows to edit the events
  in a popup. If not (default value), the record is edited changing to form
  view.
- colors (optional): it allows to set certain specific colors if the expressed
  condition (JS syntax) is met.

You also need to declare the view in an action window of the involved model.

Example:

```
<odoo>
    <record id="view_task_timeline" model="ir.ui.view">
        <field name="model">project.task</field>
        <field name="type">timeline</field>
            <timeline date_start="date_start"
                      date_stop="date_end"
                      string="Tasks"
                      default_group_by="user_id"
                      event_open_popup="true"
                      colors="#ec7063:user_id == false;#2ecb71:kanban_state=='done';">
            </timeline>
        </field>
    </record>

    <record id="project.action_view_task" model="ir.actions.act_window">
        <field name="view_mode">kanban,tree,form,calendar,gantt,timeline,graph</field>
    </record>
</odoo>
```
# Usage

For accessing the timeline view, you have to click on the button with the clock
icon in the view switcher. The first time you access to it, the timeline window
is zoomed to fit all the current elements, the same as when you perform a
search, filter or group by operation.

You can use the mouse scroll to zoom it or out in the timeline, and make click
on any free area and drag for panning the view in that direction.

The records of your model will be shown as rectangles whose widths are the
duration of the event according our definition. You can select them clicking
on this rectangle. You can also use Ctrl or Shift keys for adding discrete
or range selections. Selected records are hightlighted with a different color
(but the difference will be more noticeable depending on the background color.
Once selected, you can drag and move the selected records across the timeline.

When a record is selected, a red cross button appears on the upper left corner
that allows to remove that record. This doesn't work for multiple records
although they were selected.

Records are grouped in different blocks depending on the group by criteria
selected (if none is specified, then the default group by is applied).
Dragging a record from one block to another change the corresponding field to
the value that represents the block. You can also click on the group name to
edit the involved record directly.

Making double click on the selected record, you would be able to edit it.
Making double click on a free area, you will create a record taking the group
and the start date where you click as initial data.
# Known issues / Roadmap
- Implement support for vis.js timeline range item addition (with Ctrl key
  pressed).
- Implement a more efficient way of refreshing timeline after a record update.


pedrobaeza:
cc @Tecnativa

pedrobaeza:
You can try the module in this PR with a practical example in projects: https://github.com/OCA/project/pull/187

web_timeline/README.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Usage
+=====
+
+For accessing the timeline view, you have to click on the button with the clock
+icon in the view switcher. The first time you access to it, the timeline window
+is zoomed to fit all the current elements, the same as when you perform a
+search, filter or group by operation.
+
+You can use the mouse scroll to zoom it or out in the timeline, and make click
```
yajo: s/it/in

yajo: you don't "make click on", you just "click on"

web_timeline/README.rst
```diff
@@ -0,0 +1,140 @@
...
+on any free area and drag for panning the view in that direction.
+
+The records of your model will be shown as rectangles whose widths are the
+duration of the event according our definition. You can select them clicking
+on this rectangle. You can also use Ctrl or Shift keys for adding discrete
+or range selections. Selected records are hightlighted with a different color
+(but the difference will be more noticeable depending on the background color.
```
yajo: close parenthesis

web_timeline/README.rst
```diff
@@ -0,0 +1,140 @@
...
+although they were selected.
+
+Records are grouped in different blocks depending on the group by criteria
+selected (if none is specified, then the default group by is applied).
+Dragging a record from one block to another change the corresponding field to
+the value that represents the block. You can also click on the group name to
+edit the involved record directly.
+
+Making double click on the selected record, you would be able to edit it.
+Making double click on a free area, you will create a record taking the group
```
yajo: "Double-clicking" (x2)

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
+/*---------------------------------------------------------
+ * Odoo web_timeline
+ * Copyright 2015 ACSONE SA/NV
+ * Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+ *---------------------------------------------------------*/
```
yajo: I just hate those lines that give that big emphasis to the least important piece of the file. If you remove them, you'll make my eyes happier :smile:. Not required though.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+            this.current_window = null;
+        },
+
+        get_perm: function(name){
+            var self = this;
+            var promise = self.permissions[name];
+            if(!promise) {
+                var defer = $.Deferred();
+                new Model(this.dataset.model)
+                .call("check_access_rights", [name, false])
```
yajo, line 63: Isn't this redundant? You can assume any call without permissions will fail by itself, not?

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+            var attrs = fv.arch.attrs;
+            var self = this;
+            this.fields_view = fv;
+            this.parse_colors();
+            this.$timeline = this.$el.find(".oe_timeline_widget");
+            this.$el.find(".oe_timeline_button_today").click(self.on_today_clicked);
+            this.$el.find(".oe_timeline_button_scale_day").click(self.on_scale_day_clicked);
+            this.$el.find(".oe_timeline_button_scale_week").click(self.on_scale_week_clicked);
+            this.$el.find(".oe_timeline_button_scale_month").click(self.on_scale_month_clicked);
+            this.$el.find(".oe_timeline_button_scale_year").click(self.on_scale_year_clicked);
```
yajo: This is probably not doing what you expect. It does not matter if you call `on_scale_year_clicked` from `this` or from `self`. The main confusion about the `this` keyword in JS is that, unlike in Python and other languages, JS sets it at calling time, not at definition or bind time. jQuery says that, inside the event execution context, `this = event.target`.

You need a proxy that gets that and applies the context you want, so you should use `$.proxy(this.on_scale_year_clicked, this)` (and same for all the above).

BTW [`self` is already in scope](http://www.w3schools.com/jsref/prop_win_self.asp), so it is not a good name for this kind of stuff (but you don't need it anyway if you use proxies).

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+
+            this.$el.addClass(attrs['class']);
+
+            this.name = fv.name || attrs.string;
+            this.view_id = fv.view_id;
+
+            //this.start = py.evaluate(attrs.start || 'None', {});
+            this.mode = attrs.mode;                 // one of month, week or day
+            this.date_start = attrs.date_start;     // Field name of starting
+                                                    // date field
```
yajo: Not sure what these comments apply to. Could you reformat them please?

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+            var init = function () {
+                    self.init_timeline().then(function() {
+                        $(window).trigger('resize');
+                        self.trigger('timeline_view_loaded', fv);
+                        self.ready.resolve();
+                    });
+                };
+
+            var test = $.when(self.fields_get, self.get_perm('unlink'), self.get_perm('write'), self.get_perm('create'));
+            return $.when(test).then(init);
```
yajo: You don't need to put a `when` inside another.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+                });
+                return defer;
+            } else {
+              return promise;
+            }
+        },
+
+        set_default_options: function(options) {
+            this._super(options);
+            _.defaults(this.options, {
```
Yajo: If using underscore, then you should `require('_')` at top of file.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+            });
+            var edit_check = new Model(this.dataset.model)
+                .call("check_access_rights", ["write", false])
+                .then(function (write_right) {
+                    self.write_right = write_right;
+
+                });
+            var init = function () {
+                    self.init_timeline().then(function() {
+                        $(window).trigger('resize');
```
Yajo: you have to `var $ = require('$')` at top.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+            dialog.on('create_completed', self, function(id) {
+                self.dataset.ids = this.dataset.ids.concat([id]);
+                self.dataset.trigger("dataset_changed", id);
+                self.dataset.read_ids([id], self.fields).done(function(records) {
+                    var new_event = self.event_data_transform(records[0]);
+                    var items = self.timeline.itemsData;
+                    items.add(new_event);
+                    self.timeline.setItems(items);
+                });
+            });
```
yajo: Again, you can define a `on_create_completed` method in the object and call it here with a proxy, to let it be inherited/extended.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+                    context: this.dataset.get_context(),
+                    title: title,
+                    view_id: +this.open_popup_action,
+                }).open();
+                dialog.on('write_completed', self, function() {
+                    self.dataset.trigger("dataset_changed", id);
+                    self.current_window = self.timeline.getWindow();
+                    self.reload();
+                    self.timeline.setWindow(self.current_window);
+                });
```
yajo: Again, you can define a `on_write_completed` method in the object and call it here with a proxy, to let it be inherited/extended.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -0,0 +1,514 @@
...
+        },
+
+        on_scale_year_clicked: function(){
+            this.scale_current_window(24 * 365);
+        },
+    });
+
+    core.view_registry.add('timeline', TimelineView);
+    return TimelineView;
+});
```
Yajo: eol

web_timeline/static/src/css/web_timeline.css
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.timeline-navigation-zoom-out .ui-icon{
+  background: none !important;
+}
+.timeline-navigation-move-left .ui-icon{
+  background: none !important;
+}
+.timeline-navigation-move-right .ui-icon{
+  background: none !important;
```
yajo: Please avoid using `!important` if not absolutely required. It hardens the theming task later. Use a more specific CSS selector if you need to override some core behavior if possible please.

web_timeline/static/src/css/web_timeline.css
```diff
@@ -0,0 +1,42 @@
...
+
+/* gray background in weekends, white text color */
+.vis.timeline .timeaxis .grid.saturday,
+.vis.timeline .timeaxis .grid.sunday {
+  background: gray;
+}
+/* .vis.timeline .timeaxis .text.saturday,
+.vis.timeline .timeaxis .text.sunday {
+  color: white;
+} */
```
yajo: Remove these kind of comments, please. If code is not required, then just drop it.

web_timeline/static/src/css/web_timeline.css
```diff
@@ -0,0 +1,42 @@
...
+.timeline-navigation-move-right .ui-icon{
+  background: none !important;
+}
+/*.vis.timeline .timeaxis .grid.odd {
+  background: #f5f5f5;
+} */
+
+/* gray background in weekends, white text color */
+.vis.timeline .timeaxis .grid.saturday,
+.vis.timeline .timeaxis .grid.sunday {
```
yajo: Given we are on v9 now, this file would be much simplified if you use Less or Sass. Not a requirement though.

web_timeline/static/lib/vis/vis.css
```diff
@@ -0,0 +1,810 @@
+.vis .overlay {
```
yajo, line 1: I know it's common to include the libraries in the source code, but [given this one is npm-installable](http://visjs.org/index.html#download_install), it would be much nicer IMHO [if you add a `package.json` file that declares the dependencies](https://docs.npmjs.com/files/package.json#dependencies) and add to installation instructions that you should execute `npm install` after downloading the module to make it work. OTOH, you could even execute `cd static/src/js && npm list vis || npm install vis` manually in the `pre_init` script, and provide a useful exception if you have no permissions to write it and it was not found.

Of course, package would be now under `static/src/js/node_modules`.

web_timeline/static/src/js/web_timeline.js
pedrobaeza: I haven't changed this. It's from 8.0 code. What do you want me to do then?

web_timeline/static/lib/vis/vis.css
pedrobaeza, line 1: I'm not sure about this one, because the deployment with npm is harder than including the library, and you wouldn't find version problems, so I prefer this way.

lmignon, line 1: I agree with @pedrobaeza. In Odoo the 'implicit' convention is to include the libraries in the source code. Moreover, npm requires that you have acces to internet from your server and it's not always the case.

web_timeline/static/src/js/web_timeline.js
yajo: Replace these `self.on_*` for `$.proxy(this.on_*, this)`

web_timeline/static/lib/vis/vis.css
yajo, line 1: With npm you can install a concrete version, just like with `pip`. It would make the module much thinner, but nevermind. It's not a blocker.

lmignon, line 1: I know the benefits of npm (versioning and dependencies management) but my concern is to not introduce an additional skill requirement to be able to install this addon in odoo...

web_timeline/README.rst
```diff
@@ -0,0 +1,140 @@
...
+that allows to remove that record. This doesn't work for multiple records
+although they were selected.
+
+Records are grouped in different blocks depending on the group by criteria
+selected (if none is specified, then the default group by is applied).
+Dragging a record from one block to another change the corresponding field to
+the value that represents the block. You can also click on the group name to
+edit the involved record directly.
+
+Making double click on the selected record, you would be able to edit it.
```
elicoidal: Double-click on the record to edit it.
Double-click in open area to create a new record with the group and start date linked to the area you clicked in.

web_timeline/static/lib/vis/vis.css
```diff
@@ -0,0 +1,810 @@
...
+  z-index: 1;
+}
+.vis.timeline .customtime {
+  background-color: #6E94FF;
+  width: 2px;
+  cursor: move;
+  z-index: 1;
+}
+.vis.timeline.root {
+  /*
```
elicoidal, line 348: Can you remove dead code?

pedrobaeza, line 348: It belongs to the library, so we shouldn't remove it

web_timeline/static/src/js/web_timeline.js
pedrobaeza, line 63: It's needed for allowing certain parts of the "UI" (double click for creation, red cross for unlink...)

pedrobaeza:
Comments fixed, but I'm working on some other things:
- Image of the assigned person
- Other fields rendering (maybe Qweb template)

web_timeline/static/lib/vis/vis.css
yajo, line 348: OK

web_timeline/static/src/js/web_timeline.js
yajo, line 63: OK

yajo:
This is a great module! Hoping to see it finished :wink:

funbaker:
@pedrobaeza Do you plan to implement the Critical Path Method?

pedrobaeza:
Not for now, so you can work on that if you want.

rafaelbn:
@pedrobaeza this PR is working fine, maybe is better to merge and the rest add it in known issues?

yajo:
Is this still WIP?

rafaelbn:
@Yajo still pending https://github.com/OCA/web/pull/393#issuecomment-236938066 (nice to have)

lasley:
Maybe we should just add https://github.com/OCA/web/pull/393#issuecomment-236938066 to Roadmap & review/merge the existing?

pedrobaeza:
Yes, I think this is enough for a first version indeed, although I would like to add the other thing.

lmignon (changes requested):
The history from https://github.com/OCA/web/pull/521 is not preserved! 
see https://github.com/OCA/web/pull/306#issuecomment-292890328

pedrobaeza:
@lmignon @adrienpeiffer I have fixed commit history to reflect correctly the dates. Please give your final blessing for merging this.

pedrobaeza:
Travis is failing due to web_widget_float_formula

lmignon (approved):

pedrobaeza:
Merging this finally! Thanks to all.

PR #397, opened 2016-08-05
[9.0][ADD] web_stock_picking: Mobile & Barcode Picking Interface

lasley:
I originally created https://github.com/OCA/stock-logistics-workflow/pull/245 for this, but it is obvious to me this belongs here now that I have tweaked a bunch of things. 

This is still WIP in the sense that I need to add test coverage & refactor some things, but everything should work on a functional side. 

I did have a most stupid issue injecting a button into the Stock Picking KanBan view & [left my code there w/ a comment](https://github.com/laslabs/web/blob/15a5cb6dd811438dcb6f941fe26145af04c7967f/web_stock_picking/views/stock_picking_type.xml#L17) just in case someone knows what I'm doing wrong. Other than that, we're stuck with the generic `Mobile/Barcode` link in the sidebar 😦 
# 
# Web Stock Picking

This module provides a website frontend for stock pickings.

It is designed to be particularly useful with a mobile device and barcode
scanner.
# Installation
- Install Module
# Configuration
- You must set a Barcode Nomenclature on any Stock Picking Type that will
  utilize the barcode integration
# Usage
- Go to Inventory => Mobile/Barcode
- Choose a picking, or refine search
- Utilize picking form in similar fashion as backend form

Picking form also contains a barcode listener interface, and will increment/decrement
quantities based on scanned barcodes. It currently supports the following nomenclature
types:
- Product
- Lot
# Known issues / Roadmap
- Properly handle permissions in view
- Support for lot scanning is at the unit level only. The lots themselves are not bound
  to the picking in any way
- Implement proper errors instead of alerts - requires keybindings for enter/cancel
- Implement commented out view actions
- It would be nice to also allow for a BarcodeScanner device, instead of just the
  element events
- Button injection in stock.picking.type Kanban fails silently (have also tried jQuery
  template injection)


legalsylvain:
Hi,
huge work ! Did you think to pull request against this repository ?
https://github.com/OCA/stock-logistics-barcode
regards.

lasley:
@legalsylvain - Hah I did, then I moved it here & in hindsight I probably should have left it. You're thinking this would be best in the barcode logistics repo?

pedrobaeza:
I would bet for the one said by @legalsylvain or OCA/website, but not here indeed.

lasley:
Sweet I'll switch over to logistics-barcode. I don't think this will work in website - it is only dependent on web.

pedrobaeza:
Ah, then the description should be improved, because it says: "This module provides a website frontend for stock pickings.", so I understood that it was a website view.

lasley:
Good call - that was leftover from before I figured out how to redesign the web interface. This was going to be dependent on website until I learned all the stuff about web in the app_drawer PR 😄

PR #400, opened 2016-08-10
[9.0] [PORT] web_ir_actions_act_window_message

lisandrogallo:
I'm not sure if it is the correct approach, but I could fix the bug of the dialog box not showing the message in v9.

### DIFF
--- a/web_ir_actions_act_window_message/__openerp__.py
+++ b/web_ir_actions_act_window_message/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 {
     "name": "Client side message boxes",
-    "version": "8.0.1.0.0",
+    "version": "9.0.1.0.0",
     "author": "Therp BV, "
               "Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


hbrunn:
What you should to is to port this to the v9 javascript api. All the requirejs-like stuff et al.

Further, you seem to be working on an old version of this addon, because the current v8 looks like this: https://github.com/OCA/web/blob/8.0/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js (and it can do a lot more)

To get the latest version, follow the procedure as outlined in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#if-the-module-already-exists-in-the-90-branch-but-it-doesnt-contain-all-the-commit-history-or-youre-not-sure-if-it-have-it

lisandrogallo:
Thanks for the feedback @hbrunn. I will check the info you provided me and see if I can correctly port the module.

PR #401, opened 2016-08-10
[9.0] [PORT] web_ir_actions_act_window_message

lisandrogallo:
Merged all commit history from 8.0 branch for this module and added a minor fix. Now it seems to be working properly in v9.
Related to PR #400.


web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
```diff
@@ -24,18 +24,31 @@ openerp.web_ir_actions_act_window_message = function(instance)
     instance.web.ActionManager.include({
```
lasley: Hi @lisogallo - Thanks for the submission!

This JavaScript is using the deprecated 8.0 API. While it does work now due to some compatibility imports, this will not be the case going forward. 

There is a new Requires system that must be used in order to upgrade to the 9.0 API. Please take a look at the [Javascript module system overview](https://www.odoo.com/documentation/9.0/reference/javascript.html#web-client) for a full description.

For an high level abstract example, you can also take a look at this [PR upgrading the template module](https://github.com/OCA/maintainer-tools/pull/208/files).

lisandrogallo:
Hi @lasley, 

I'll see if I can adapt the JS to v9 API based on the links you provided me. Thanks for your feedback!

web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
```diff
@@ -19,32 +19,90 @@
...
+                size: 'medium',
+                buttons: buttons,
+                title: action.title,
+                $content: $('<div>', {
+                    text: action.message,
+                }),
+            }).open();
+
+          },
+            ir_actions_act_window_message_get_buttons: function(action, close_func)
```
hbrunn: watch your indentation

PR #396, opened 2016-08-05
[9.0][ADD] Community Web v2 - Backend Menus

lasley:
This PR is scoped towards the goal of moving the side menu to a mobile-compliant top menu, similar to that seen in Enterprise. It should also provide a mobile compatible Apps drawer in lieu of the Apps screen available in Enterprise.

High Level Prelim Feature Breakdown:
- Provide a [SlideoutJS](https://mango.github.io/slideout/) slider for the "Apps Drawer"
- Implement a method of displaying module icons in the "Apps Drawer"
- Remove the existing left sidebar ("Secondary Menu"), and provide a mobile-compliant Bootstrap menu bar instead
- Move the User and Systray into the Secondary Menu

This will be paired with a few other modules in order to provide a fully mobile compliant interface to Odoo backend. It will not actually be dependent on any of those modules, however, and can be developed in parallel.

To Do:
- Move the user & systray to Secondary Menu
- Move logo to Apps Drawer?
- Display App Icons in place of textual representations
- Shrink Secondary Menu
- Make it all look better

Screenshots of current implementation:

image

image

PR #402, opened 2016-08-10
[9.0][ADD] JS DataModel Allowing Public User Sessions

lasley:
This PR is from our excitement in #396. Please someone tag as WIP.

The intent here is to break the backend sandbox with no API breakage, providing seamless use of backend Widgets in Website views.

Following is an excerpt of relevant talking points between @Yajo and I regarding implementation tactics. Current tactic in this PR is the Controller overrides that seem to have been disproven.

Our biggest struggle with breaking the backend sandbox will be the same that is noted in the website_backend_views ReadMe - lack of support for anonymous user sessions. Anon sessions are obviously a critical component in any website, but the web.DataModel and underlying JsonRPC interface do not support it.

Public user sessions are not allowed in the underlying DataModel controllers, thus yielding a Session Expired error when attempting to access resources.

We explored the possibility of creating a password for the public user, logging in with `web.Session` JS, then working off of that. This was theoretically disproven due to use session overlap in situations such as `website_sale` shopping cart. We have not tested this idea beyond theory.

The next idea was to override the DataModel controllers to allow public sessions, which is the code currently in this PR - dubbed the hammer-style approach. 

Regarding this approach, however - At first sight, if you log in, go to backend, open 2 browser tabs, log out in 2nd, and try to search something in 1st, Odoo dies badly, instead of showing this like it usually does:
image

@Yajo checked that ultimately, [initiating a session calls res.users.check](https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/res/res_users.py#L498-L511). Maybe if you overload that and when there is no password, if user is public (check that by group), then succeed (else, super()), then you could simply authenticate with require("web.session").session_authenticate(db, login, False) and get a working session with public user.

I will be attempting the above strategy next.

cc @dreispt @antespi @pedrobaeza @t3ddftw


model_public_session/README.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. code:: xml
+
+    <input type="text"
+           class="js_website_autocomplete"
+           data-query-field="name"
+           data-display-field="name"
+           data-value-field="id"
+           data-limit="10"
+           data-domain='[["website_published", "=", true]]'
```
dreispt: Shouldn't this domain should be implicitly enforced?

model_public_session/demo/field_autocomplete_demo.xml
```diff
@@ -0,0 +1,36 @@
...
+                    <div class="row">
+                        <label for="url">URL</label>
+                        <input type="text"
+                               name="url"
+                               class="js_website_autocomplete"
+                               data-query-field="name"
+                               data-display-field="name"
+                               data-value-field="url"
+                               data-limit="10"
+                               data-model="website.menu"
```
dreispt: res.partner might be a better example

model_public_session/README.rst
```diff
@@ -0,0 +1,92 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+============================
+Website Field - AutoComplete
```
dreispt: Should the technical name be `website_autocomplete` of similar?

model_public_session/controllers/main.py
```diff
@@ -0,0 +1,43 @@
...
+class Website(Website):
+
+    @http.route(
+        '/website/field_autocomplete/<string:model>',
+        type='http',
+        auth='public',
+        methods=['GET'],
+        website=True,
+    )
+    def _get_field_autocomplete(self, model, **kwargs):
```
dreispt: When is this controller method used?
I checked the js code for the autocomplete, I see it use `search_read` and can't find any reference to an `/website/field_autocomplete/` URL.

model_public_session/README.rst
yajo: I think he somehow copypasted https://github.com/OCA/website/pull/214

lasley: Whoops nothing to see here, folks!

web_session_allow_public/models/res_users.py
```diff
@@ -0,0 +1,22 @@
...
+class ResUsers(models.Model):
+
+    _inherit = 'res.users'
+
+    def check(self, db, uid, passwd):
+        """ Catch AccessDenied error and allow public session to proceed """
+        if not passwd:
+            cr = self.pool.cursor()
+            try:
+                is_public = self._has_group(cr, uid, 'base.group_public')
```
yajo: Can't you use new api?

web_session_allow_public/demo/web_session_allow_public_demo.xml
```diff
@@ -0,0 +1,14 @@
...
+
+<!--
+    Copyright 2016 LasLabs Inc.
+    @license AGPL-3 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="assets_frontend" name="Public Session Demo" inherit_id="website.assets_frontend">
+        <xpath expr="." position="inside">
+            <script src="/web_session_allow_public/static/src/js/web_session_allow_public.demo.js" />
```
yajo: Thinking about the objective of this module, it would make sense to include `web/static/src/js/framework/data_model.js` in frontend.

web_session_allow_public/static/src/js/web_session_allow_public.demo.js
```diff
@@ -0,0 +1,20 @@
...
+  var base = require('web_editor.base');
+  
+  base.ready().done(function() {
+    new Model('website.menu')
+      .call('search_read', [])
+      .then(function(result) {
+        console.log('Model call was a success!');
+        console.log(result);
+      });
+  });
```
yajo: You can transform this in a tour and run it automatically. Check templates in maintainer-tools scaffolding.

web_session_allow_public/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
```
yajo: Just remembering https://github.com/OCA/web/pull/396#issuecomment-238164559

web_session_allow_public/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "web",
+        "website",  # Required to test sandbox break
+    ],
+    "demo": [
+        "demo/web_session_allow_public_demo.xml",
```
yajo: Did you know this is an awesome idea for including demo assets only in db instances? I'm thinking on adding it to official template.

yajo:
Just enter runbot and open JS console: 

captura de pantalla de 2016-08-12 09-25-43

web_session_allow_public/models/res_users.py
lasley: Admittedly I didn't try it. I'm pretty sure I am hitting the non-api method here though because I'm not within an api method & don't have an env yet. The core code has three methods, api.v7 api.v8 and no api.

web_session_allow_public/README.rst
lasley: Good call, I need to start using Bob templates or something so I stop getting these copy paste errors. Will run a find/replace

web_session_allow_public/__openerp__.py
lasley: Hahah then I don't feel so bad for how long I sat staring at my IDE trying to figure out how to do this!

The template repo is my go-to doc, I'm all for adding every bit of advice there as possible 👍

web_session_allow_public/static/src/js/web_session_allow_public.demo.js
lasley: It's already running automatically. What would be the functional difference between this and a tour?

web_session_allow_public/demo/web_session_allow_public_demo.xml
lasley: Yup you're right, we definitely want that. I'm going to import in the real assets instead of demo though.

web_session_allow_public/models/res_users.py
lasley: Yup this was just ambiguous naming of the input `cr` as `db` instead. This very much worked as an `api.model`

web_session_allow_public/static/src/js/web_session_allow_public.demo.js
yajo: You are currently running this code in every rendered page. With tours, you can run them when testing the module, and get a Travis :x: if something fails. Besides, it's the standard way of launching JS website tests in Odoo. Check the [template for JS tour](https://github.com/OCA/maintainer-tools/blob/master/template/module/static/src/js/module_name.tour.js) and [Python launcher test case](https://github.com/OCA/maintainer-tools/blob/master/template/module/tests/test_something.py#L36-L43).

lasley: Ahhh cool this makes sense, thanks for the explanation. Will make this a tour 👍

lasley:
I have narrowed this down to the [IrHttp model](https://github.com/odoo/odoo/blob/05645343c326bee607af750623d1257a1c5d9ba3/openerp/addons/base/ir/ir_http.py#L77), but we're back to the whole needing to login using the public user issue. I'm going to give it a shot, but I'm thinking we're going to end up with the infinite session again.

This is what the `request.session` looks like, before it is kicked back with the `SessionExpiredException`:

``` python
(Pdb) request.session
<OpenERPSession {'uid': None, 'context': {'lang': 'en_US'}, 'geoip': {}, 'login': None, 'password': None, 'db': u'test'}>
```

A thought that is occurring - maybe we force inject something into the context when using JsonRPC over JS. This way we would hopefully isolate this override to the JS widgets

lasley:
Hmmm turns out it was actually that easy. The only side effect seems to be a slightly altered error message when you are operating in 2 tabs, logout in one, and search in the other:

image

Seems negligible though because the error still hints at bad auth, it just doesn't say Expired.

Next orders of business:
- [ ] Inject [DB info and web client creation](https://github.com/odoo/odoo/blob/9.0/addons/web/views/webclient_templates.xml#L306) into `website` template
- [ ] Inject `web.backend_assets` into `website` template.
- [ ] Identify out how to move this JS session query test into a tour & run it as a public user

simahawk:
@Yajo @lasley I have similar issue I described [here](http://stackoverflow.com/questions/40630972/anonymous-user-access-to-rpc-api-security-hole). Would you mind to give me your insights?  Will be much appreciated... probably I can use this PR to solve my issue too.

yajo:
Indeed, you issues is mostly what this solves: allow RPC calls from public users.

simahawk:
@Yajo ok, I'll try it and provide feedback then :)

lasley:
I should probably pick this PR up again now that the web_responsive is merged. I'll try to do that sometime this week, but this does seem to work from what I remember.

yajo:
@lasley I don't remember if I told you this in last Odoo Experience, but @ged-odoo said that they're decoupling the widgets from views and from web client, meaning all of them should become plugable anywhere. We may finally use backend widgets in the frontend for v11 :blush:

Of course, we'd need something like this for that purpose.

However this has security implications. And you know what? That's fun :laughing: because you are supposed to use good permissions in your addons, but once you enable crawlers to ask your database random things, then you realize maybe everybody did not do everything right.

Maybe we could have a default security measure here: adding a new boolean field to ir.model that enables exposure through public XMLRPC API. An ir.config_parameter can be added to disable this check too, for those of us that trust in permissions (nicer if found in a settings page).

simahawk:
@lasley @Yajo yep, they are rewriting all the form/widgets story and it will be much much better :)
I agree, we could use a setting for exposing models as you can do with fields in website form editor (enterprise).

yajo:
[Actually the base for that is CE](https://github.com/odoo/odoo/blob/10.0/addons/website_form/controllers/main.py#L24). But yes, that's the idea. However some can consider such extra security step is unnecessary if your record rules are OK, and some others just may need this public XMLRPC access to combining their DB with a mobile app or whatever, and want all exposed by default, so I'd add that option to force or not this check.

The fun part is that you could add that only by using actual rules in database, instead of hardcoding a controller as they do there :grin:

lasley:
Hmmm wouldn't the public user require an explicit read access rule in order to be able to access the data? It seems like this pattern is already handled for us in the default permissions handling, unless I'm totally overlooking something here.

simahawk:
That was my fear and I double check that removing the portal permission (in my case for res.partner.category) was enough for locking access to the model.
It should be fine, we were just wondering about adding a more explicit way to enable this feature to protect who is not protecting him/herself :)

lasley:
The magic of this is happening in the [`ir.http`](https://github.com/OCA/web/pull/402/files#diff-da9d303b88b4fa1c2154543ea0c6571fR15) model, and not actually in a model for searching. Basically I am just explicitly logging in a new public user session during the initial user authentication, which just kind of makes everything else work. AFAIK it's only happening once per page load as well, even though many models can be queried afterwards.

I'm not quite sure how I would inject permissions onto the models being searched too - maybe something in the base model (<=v10). But if I did that, then the question is how to identify that the public user session in question is via RPC or Odoo.

I'm up for ideas though, I can see the use case here. I'll keep pondering as well.

web_session_allow_public/tests/ir_http.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from openerp import models
+from openerp.http import SessionExpiredException
+
+
+class IrHttp(models.Model):
```
yajo: Shouldn't this file be a test? :confused:

yajo:
I guess that this kind of manual model discrimination should be made in the controller itself, but yes, after all is reinventing a wheel that's been there since the beginning. All you need is to have good permissions and you're fine. (BTW I'm happy seeing @simahawk using permissions :stuck_out_tongue_winking_eye:)

So maybe you should simply forget about it @lasley :laughing:

lasley:
Security in a nutshell right there, if it's not implemented correctly you're screwed regardless. I'll forget the feature then 😄 

As I'm looking at this, it's basically done outside of the tests. I'll polish those in the next few days so we can go for a merge.

lasley:
Turns out that a JS test is impossible for this, because of the mockRPC injection that is happening inside of Phantom. I went ahead and fell back to Python, which should hopefully go 🍏  and subsequently be ready for review.

lasley:
On second thought, this may belong in `server-tools`. I was able to remove all but `base` dependency because all of this is happening in `ir_http`

Maybe call it `ir_http_public`? Or maybe `base_rpc_public` to be a bit more descriptive to the humans?

web_session_allow_public/tests/test_ir_http.py
```diff
@@ -0,0 +1,25 @@
...
+MODULE_PATH = 'openerp.addons.base.ir.ir_http'
+
+
+class TestIrHttp(HttpCase):
+
+    @mock.patch('%s.request' % MODULE_PATH)
+    def test_public_session(self, request):
+        """ It should allow HTTP authentication on public user """
+        request.session.uid = False
+        with mock.patch.object(
```
yajo, line 21: You are the `mock` guy indeed :laughing: 

But, did you know you can actually make RPC calls in tests? [you have some helpers inside `HttpCase`](https://github.com/odoo/odoo/blob/9.0/openerp/tests/common.py#L246-L249) and [an example here](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/odoo/addons/base/tests/test_xmlrpc.py).

yajo (changes requested):
About moving to `server-tools`, seems a good idea. I'd call this simply `base_rpc_public` as you said.

web_session_allow_public/tests/test_ir_http.py
lasley: > You are the mock guy indeed 

I've said it once and will say it again, I ❤️  mock. I'd probably marry it if there weren't laws against that 😛 

> But, did you know you can actually make RPC calls in tests? 

Ah yes, I spent about two hours trying to make this work yesterday. It seems that the internal XML proxy is authenticating from a lower level than the external.

Internal RPC calls directly the [`res.users.check`, which explicitly fails on no passwd](https://github.com/odoo/odoo/blob/457a705bf0927e281d5269014f955d3e2a828e11/odoo/addons/base/res/res_users.py#L496). I tried to override this in [one of my earlier attempts](https://github.com/OCA/web/pull/402/commits/1f8162292bf012c3671b3b3fd27553604a76152c), and [you reported bad results](https://github.com/OCA/web/pull/402#issuecomment-239379531). 

We are now overriding in [the ir.http session authentication](https://github.com/OCA/web/pull/402/commits/643d128cd53c9cc60c71072f7f01a23d2db7beec), which it seems the internal XML-RPC completely bypasses for some reason.

My frustration ala code for further elaboration. The last line illustrates why a mock was required:

```
(Pdb) user = self.env.ref('base.public_user')

(Pdb) self.xmlrpc_object.execute('test', user, '', 'res.users', 'search_read')
2016-11-22 17:16:41,331 27267 INFO test werkzeug: 127.0.0.1 - - [22/Nov/2016 17:16:41] "POST /xmlrpc/2/object HTTP/1.1" 200 -
*** Fault: <Fault 3: 'Access denied'>

(Pdb) self.env['res.users'].check('test', user, False)
*** AccessDenied: Access denied

(Pdb) openerp.service.security.check('test', user, False)
*** AccessDenied: Access denied

(Pdb) self.env['res.users'].browse(user).check_credentials(False)
(Pdb) 2016-11-22 02:03:43,309 7962 INFO test werkzeug: 10.211.55.1 - - [22/Nov/2016 02:03:43] "POST /longpolling/poll HTTP/1.1" 200 -

(Pdb) super(IrHttp, self)._auth_method_user()
*** SessionExpiredException: Session expired

self._auth_method_public()
*** ValueError: URI connections not allowed
```

lasley: On second thought I'm wondering if the JS Model is using JsonRpc vs XmlRpc, which could be creating the difference...

yajo: But the problem with current tests is that you are checking that the method raises the exception, but not checking if you actually can call XML/JSONRPC anonymously and still get access to public records. I mean, it's better than no tests, but still a little bit incomplete IMHO.

pedrobaeza:
I think this has no sense now that v11 has arrived with a clear separation of widgets and DataModel. Please close it if you think the same.

lasley:
I think we may still be missing this piece. I haven't had time to play yet, but I can't find any examples of a backend widget/view being used in the website side of things under an anonomyous session. Do we know if this is possible?

_Edit: added caveat of anonymous session_

pedrobaeza:
It was one of the goals when designing new JS views, but I haven't seen either an example on core.

simahawk:
I confirm the issue is still there: for anon users any ajax call is locked out. I just restored the fix on a project migrated from v9 to v11.

PR #404, opened 2016-08-11
Model based quick create config option.

codingforfun:
Add option to disable quick create for a complete odoo instance based on the target model of the relation. E.g. for product.product. 

Similiar to https://github.com/OCA/server-tools/tree/9.0/base_optional_quick_create but disable quick create completely for that model instead of falling back to full create form. 

Same effect as if you would add `no_create` to all m2x fields in all fields with that target model in their relation.

### DIFF
--- a/web_m2x_options/models/model.py
+++ b/web_m2x_options/models/model.py
@@ -1,0 +1,9 @@
+# -*- coding: utf-8 -*-
+
+from openerp import models, fields
+
+
+class IrModel(models.Model):
+    _inherit = 'ir.model'
+
+    disable_quick_create = fields.Boolean('Disable quick create')
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -100,8 +100,19 @@ openerp.web_m2x_options = function (instance) {
 
             var create_rights;
             if (!(self.options && (self.options.no_create || self.options.no_create_edit))) {
-                create_rights = new instance.web.Model(this.field.relation).call(
-                    "check_access_rights", ["create", false]);
+                // check quick create options
+                var target_model = this.field.relation
+                create_rights = new instance.web.Model('ir.model').
+                    query(['disable_quick_create']).
+                    filter([['model', '=', target_model]]).
+                    first().
+                    then(function(result){
+                        if(result.disable_quick_create)
+                            return $.when(false);
+                        else
+                            return new instance.web.Model(target_model).call(
+                                "check_access_rights", ["create", false]);
+                    });
             }
 
             $.when(search_result, create_rights).then(function (data, can_create) {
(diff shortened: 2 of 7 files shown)


web_m2x_options/models/model.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+
+from openerp import models, fields
+
+
+class IrModel(models.Model):
+    _inherit = 'ir.model'
+
+    disable_quick_create = fields.Boolean('Disable quick create')
```
sbidoul, line 9: Would you mind naming this field `avoid_quick_create` so users of `base_optional_quick_create` can easily migrate to your approach (which is more elegant)

web_m2x_options/static/src/js/form.js
```diff
@@ -100,8 +100,19 @@ openerp.web_m2x_options = function (instance) {
 
             var create_rights;
             if (!(self.options && (self.options.no_create || self.options.no_create_edit))) {
-                create_rights = new instance.web.Model(this.field.relation).call(
-                    "check_access_rights", ["create", false]);
+                // check quick create options
+                var target_model = this.field.relation
+                create_rights = new instance.web.Model('ir.model').
```
sbidoul, line 105: Do you have an idea if this additional query would have a performance impact? Would some kind of caching be beneficial?

web_m2x_options/models/model.py
codingforfun, line 9: I don't mind how the field is called, but `base_optional_quick_create` is doing something different so there might be people who need both modules. Of course setting both options at once wouldn't make much sense for the same model, but having both options for different models might be a valid use case. 
This is why I called it `disable_quick_create` to enforce the **disable** vs. **avoid** character of this approach.

web_m2x_options/static/src/js/form.js
codingforfun, line 105: It's an extra RPC call of course, but I don't think it will be noticeable. This is only in the form view and I think the number of m2x fields per view isn't that big that this should be a concern here. 
I don't have an idea how to cache here. Any suggestions?

OSguard:
:+1:

pedrobaeza:
:+1:

codingforfun:
Any further remarks here? Does anyone feel like merging this?

web_m2x_options/static/src/js/form.js
lasley: I think this is a good idea to put on the Roadmap if you don't mind @codingforfun - I'll merge after that

lasley:
@codingforfun - One comment & will merge after that. Thanks for the excellent contribution 👍

web_m2x_options/static/src/js/form.js
codingforfun: What Roadmap? You mean in the module description README.rst?

codingforfun:
Thanks

PR #405, opened 2016-08-12
[8.0][IMP] web_dashboard_tile: context_today and a few other improvements.

ivantodorovich:
- ~~Removed unused field `active`~~
- Removed unused code and cleaned it up
- Moved helper field computation into a separated method, to avoid evaluating calculations while editing the tile in form view. (before, if domain conained an error it throwed an exception and it couldn't be edited to fix it. Now it can be edited, even if there are errors)
- Implemented `context_today` and context variables.
- safeguarded `eval()` by using `safe_eval`
- FIXED a bug that sometimes throwed this error "_FIX: Non-db action dictionaries should provide either multiple view modes or a single view mode and an optional view id._"
- "Add Tile" is now inside of the "Add to Dashboard" menu.


ivantodorovich:
This PR is ready for review.

I'm working on further improvements of this module, but I intend to do them in a separate PR once this get merged.

ivantodorovich:
@pedrobaeza would you please add 'Needs Review' label?

pedrobaeza:
Done

ivantodorovich:
Ty :heart:

web_dashboard_tile/__init__.py
```diff
@@ -1,26 +1,7 @@
...
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2010-2013 OpenERP s.a. (<http://openerp.com>).
+# © 2014 initOS GmbH & Co. KG (<http://www.initos.com>).
+# © 2015-Today GRAP
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from . import models
+from . import models
```
lasley: Add a newline

web_dashboard_tile/__openerp__.py
```diff
@@ -1,38 +1,27 @@
...
         'mail',
         'web_widget_color',
     ],
-    'author': "initOS GmbH & Co. KG,GRAP,Odoo Community Association (OCA)",
+    'author': 'initOS GmbH & Co. KG, '
+              'GRAP, '
+              'Odoo Community Association (OCA)',
     "category": "web",
     'license': 'AGPL-3',
+    'contributors': [
```
lasley, line 20: Is this manifest key actually used in Odoo? If not, please remove and use the Contributors section in ReadMe

web_dashboard_tile/models/tile_tile.py
```diff
@@ -1,155 +1,127 @@
...
+    def _get_eval_context(self):
+        def _context_today():
+            return fields.Date.from_string(fields.Date.context_today(self))
+        context = self.env.context.copy()
+        context.update({
+            'time': time,
+            'datetime': datetime,
+            'relativedelta': relativedelta,
+            'context_today': _context_today,
+            'current_date': time.strftime('%Y-%m-%d'),
```
lasley: fields.Date.today()

web_dashboard_tile/models/tile_tile.py
```diff
@@ -1,155 +1,127 @@
...
-    font_color = fields.Char(default='#FFFFFF')
-    sequence = fields.Integer(default=0, required=True)
+    helper = fields.Char(compute='_compute_function_helper')
+
+    @api.one
+    def _compute_data(self):
+        self.count = 0
+        self.computed_value = 0
+        # Compute count item
+        model = self.env[self.model_id.model]
```
lasley: Why the reassignment through `env`? `self.model_id` is already the usable model isn't it?

web_dashboard_tile/__openerp__.py
ivantodorovich, line 20: I don't really know, to be honest. 
[I've seen this in other modules, so I copied from them](https://github.com/search?p=3&q=org%3AOCA+language%3Apython+contributors&ref=searchresults&type=Code&utf8=%E2%9C%93)

web_dashboard_tile/models/tile_tile.py
```diff
@@ -1,155 +1,127 @@
...
+        eval_context = self._get_eval_context()
+        self.count = model.search_count(eval(self.domain, eval_context))
+        # Compute datas for field_id depending of field_function
+        if self.field_function and self.field_id and self.count != 0:
+            records = model.search(eval(self.domain, eval_context))
+            vals = [x[self.field_id.name] for x in records]
+            if self.field_function == 'min':
```
lasley: Something more Pythonic that would accomplish the same w/o the overhead of all these if statements:

```
function_map = {
    'min': lambda x: min(x),
    'avg': lambda x: sum(x) / len(x),
    ...
}
try:
    self.computed_value = function_map[self.field_function](vals)
except KeyError:
    pass
```

web_dashboard_tile/models/tile_tile.py
```diff
@@ -1,155 +1,127 @@
...
+            }
+            self.helper = _(helpers.get(self.field_function, '')) % desc
+
+    # Constraints and onchanges
+    @api.one
+    @api.constrains('model_id', 'field_id')
+    def _check_model_id_field_id(self):
+        if self.field_id and self.field_id.model_id.id != self.model_id.id:
+            raise ValidationError(
+                _("Please select a field of the selected model."))
```
lasley: s/of the/from the

web_dashboard_tile/models/tile_tile.py
ivantodorovich: To be honest, IDK.
That line is from the original code. I didn't really changed it.

web_dashboard_tile/models/tile_tile.py
```diff
@@ -1,155 +1,127 @@
...
+
+    @api.one
+    @api.constrains('field_id', 'field_function')
+    def _check_field_id_field_function(self):
+        if self.field_id and not self.field_function or\
```
lasley: Admittedly somewhat confused by this if statement, but I think this hits it and makes slightly more clear:

```
validations = self.field_id, self.field_function
if any(validations) and not all(validations):
```

web_dashboard_tile/models/tile_tile.py
```diff
@@ -166,7 +138,7 @@ def open_link(self):
         }
         if self.action_id:
             res.update(self.action_id.read(
-                ['view_type', 'view_mode', 'view_id', 'type'])[0])
+                ['view_type', 'view_mode', 'type'])[0])
             # FIXME: restore original Domain + Filter would be better
```
lasley: Would you mind moving this comment to the roadmap of readme please?

web_dashboard_tile/static/src/css/tile.css
```diff
@@ -1,44 +1,52 @@
...
     font-style: italic;
 }
+
+.openerp .oe_searchview_drawer .oe_searchview_dashboard .oe_dashboard_tile_form {
+    display: none;
+}
+
+.openerp .oe_searchview_drawer .oe_opened .oe_dashboard_tile_form {
+    display: block;
+}
```
lasley: Please add newline

lasley:
Thanks for the submission @ivantodorovich - comments inline

web_dashboard_tile/models/tile_tile.py
ivantodorovich: Just tried your way and it doesn't work. It needs to go trhough `env`

web_dashboard_tile/models/tile_tile.py
ivantodorovich: I didn't changed these lines, that's from the original code.
Although I agree with you, I'm working on another PR that will rewrite this function from scratch, so maybe leave it like this is for now?

web_dashboard_tile/models/tile_tile.py
ivantodorovich: _Please select a field from the selected model._?

lasley: Yes please, sounds a bit better IMO

web_dashboard_tile/models/tile_tile.py
lasley: No problem, I won't block over this

web_dashboard_tile/models/tile_tile.py
ivantodorovich: I believe it is already in the original readme:

> combine domain of menue and filter so can not restore origin filter;

lasley: Woot, let's just delete the comment then please

web_dashboard_tile/static/src/css/tile.css
ivantodorovich: Done!

web_dashboard_tile/models/tile_tile.py
lasley: Hmmm that makes no sense. Oh well, I won't block over existing code weirdness

web_dashboard_tile/models/tile_tile.py
ivantodorovich: Done. I've added a new line to the readme too, because I fear without that comment the issue wasn't clear.

ivantodorovich:
All done!

lasley:
👍 

Once we get our second thumb, would you also squash your commits please @ivantodorovich

ivantodorovich:
Yes of course :)

legalsylvain:
Hi, 

~~1/ Please do not remove active field because it is not useless. compute active field is based on access right. Use case :~~

~~\- you set a tile based on pos.order.~~
~~\- some user are pos user, others not.~~

~~With your PR, users that are not pos user will have an error.~~
[FIXED in https://github.com/OCA/web/pull/405/commits/5dfe52840452953c71b4e4fd4236de173ad3213d]

2/ Thanks a lot for all the others improvments ! 

kind regards.

ivantodorovich:
You are right @legalsylvain. 
I've corrected it in the last commit.

chermed:
@ivantodorovich I want to suggest adding other things

To support relative datetime in the domain, why not import time and datutil.relatvedelta packages and passing them in the localcontext in the `safe_eval` function ?

ivantodorovich:
@chermed I kinda copied from [here](https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/ir/ir_actions.py#L931):

```
    def _get_eval_context(self, cr, uid, action, context=None):
        """ Prepare the context used when evaluating python code, like the
        condition or code server actions.
        :param action: the current server action
        :type action: browse record
        :returns: dict -- evaluation context given to (safe_)eval """
        obj_pool = self.pool[action.model_id.model]
        env = openerp.api.Environment(cr, uid, context)
        model = env[action.model_id.model]
        obj = None
        if context.get('active_model') == action.model_id.model and context.get('active_id'):
            obj = model.browse(context['active_id'])
        return {
            # python libs
            'time': time,
            'datetime': datetime,
            'dateutil': dateutil,
            # NOTE: only `timezone` function. Do not provide the whole `pytz` module as users
            #       will have access to `pytz.os` and `pytz.sys` to do nasty things...
            'timezone': pytz.timezone,
            # orm
            'env': env,
            'model': model,
            'workflow': workflow,
            # Exceptions
            'Warning': openerp.exceptions.Warning,
            # record
            # TODO: When porting to master move badly named obj and object to
            # deprecated and define record (active_id) and records (active_ids)
            'object': obj,
            'obj': obj,
            # Deprecated use env or model instead
            'self': obj_pool,
            'pool': self.pool,
            'cr': cr,
            'uid': uid,
            'context': context,
            'user': env.user,
        }
```

```
            eval_context = self._get_eval_context(cr, uid, action, context=context)
            condition = action.condition
            if condition is False:
                # Void (aka False) conditions are considered as True
                condition = True
            if hasattr(self, 'run_action_%s_multi' % action.state):
                run_context = eval_context['context']
                expr = eval(str(condition), eval_context)
```

ivantodorovich:
@legalsylvain if you think it's ok now, would you give it a thumbs up?

I'd like to see it merged :)

---

I'll squash commits once it get the seconds thumbs up

legalsylvain:
Hi, LGTM now for me. Thanks ! 
Could you just add in the module description, a line with a sample of how to use context_today. (for end users that doesn't know the syntax ?)
:+1:

ivantodorovich:
@legalsylvain I've pushed an update to the readme. Let me know what you think.

I didn't want to give it much relevance since these are the same variables available when you create filters. The user is expected to know that if he wants to use them. It's not a concept added by this module.

legalsylvain:
:+1: Thanks !

ivantodorovich:
Hum, I wanted to squash commits before merging :/

legalsylvain:
Sorry. I just created a revert by github, but not sure that it will delete unsquached previous commits. @pedrobaeza , any idea?

pedrobaeza:
I'm afraid it's not possible without push forcing the branch, having the risk of leading to broken git pull of someone. We will have to live without these commits (it's not the most populated PR anyway).

PR #406, opened 2016-08-18
[8.0][FIX] web_group_expand  -   Fix #249

### LINKED ISSUE #249: [8.0] web_group_expand - expand buttons shown multiple times
Checked on runbot, when you go to Sales > Products > Products by Categories, and you enter into a product and then go back, the expand buttons are shown in multiple cases. This has been spotted only here, but I think that is a problem with this kind of view.

image

image

ivantodorovich:
Fixes https://github.com/OCA/web/issues/249


pedrobaeza:
@oihane, please check.

oihane:
:+1: LGTM

pedrobaeza:
:+1:

PR #395, opened 2016-08-05
[FIX][web_widget_image_download] Make button look good everywhere (I hope).

yajo:
Fix https://github.com/OCA/web/pull/391#issuecomment-235906320.

This is some sort of backport from #392 

@Tecnativa

PR #407, opened 2016-08-23
[9.0][ADD] web_widget_darkroom: Add o2m DarkroomJS widget

lasley:
# Odoo DarkroomJS Widget

This module provides a [DarkroomJS](https://github.com/MattKetmo/darkroomjs) web widget for use with images fields.

This widget will allow you to perform the following actions on images:
- Zoom
- Rotate
- Crop
- Step back in history client-side (before save)
# Usage

- Install web_widget_darkroom
- Add the to any One2many image relation by using the `darkroom` widget. Options can be passed through to Darkroom using the `options` key::
  
  <field name="image_id" widget="darkroom"
                         options="{'minWidth': 100}" />

The Odoo DarkroomJS widget passes options directly through to Darkroom, which are copied from the source below::

```
  // Default options
  defaults: {
    // Canvas properties (dimension, ratio, color)
    minWidth: null,
    minHeight: null,
    maxWidth: null,
    maxHeight: null,
    ratio: null,
    backgroundColor: '#fff',

    // Plugins options
    plugins: {},

    // Post-initialisation callback
    initialize: function() { /* noop */ }
  },
```
# Known Issues/Roadmap
- Plugins are not able to be added without inheriting, then redefining the widget in the registry due to JS inheritance.
  *\* This is not scalable because there would need to be an explicit dependency chain in order to avoid registry overwrite.


lasley:
WIP because I would like to add some tests. This has been in use for a while though & works great

pedrobaeza:
Isn't better to allow to use Darkroom in any image widget? Adding a button like in this module: https://github.com/OCA/web/pull/395

lasley:
Hmmm I do like the thought of implicit usage. 

The one issue I am already running into though is that there are a lot of buttons as part of this widget. I'm going to be fixing the plugin framework too, so there's a lot of places this can go. I'm planning a M2M/O2M widget as well.

I could maybe add a button to pop out a modal w/ this widget embedded in it though. Best of worlds type deal maybe?

So many buttons:
image

pedrobaeza:
I prefer that you add a button in the regular image widget that represents an "edit" action, and clicking on it, a modal screen allows to represent the full Darkroom UI, including Save/Discard buttons. For tree views, you can include a server action that allows to be called within a button with the same effect as clicking the edit button.

lasley:
I see where you're coming from @pedrobaeza . My project requires an inline widget for some streamlining of some workflows, however I think I can easily work the button modal operation into scope as well - it will increase the versatility by leaps and bounds.

obulkin:
I've added the modal functionality, fixed a few bugs, and generally cleaned up the WIP code, removing a number of unnecessary Darkroom files. The module should be ready for review now, @lasley.

web_widget_darkroom/README.rst
```diff
@@ -0,0 +1,89 @@
...
+  * maxWidth
+  * maxHeight
+  * ratio (aspect ratio)
+  * backgroundColor
+
+* Open a form view that contains an image in edit mode and hover over the 
+  image widget. You should see a Darkoom button that can be clicked to open 
+  the image in a Darkroom modal, where it can be edited and the changes can be 
+  saved.
+
```
simahawk, line 47: can we have some screenshots here?

simahawk (approved):
except missing screenshot: code LGTM

web_widget_darkroom/README.rst
obulkin: Sure thing. I've added some screenshots to the README.

simahawk: :+1:

moylop260 (approved):

moylop260:
Could you check if the new beta messages errors of lint are valid to fix them?
https://travis-ci.org/OCA/web/jobs/211978937

obulkin:
@moylop260 - I've made a number of changes based on the linter results. Here's my "beta feedback" about the ESLint rules 😄 :

- Currently the [`multiline-ternary`](http://eslint.org/docs/rules/multiline-ternary) rule is set up to enforce multiline ternaries rather than to prevent them. I'm guessing this is the opposite of what you intended, so the value needs to be changed to `"never"`.
- The [`no-mixed-operators`](http://eslint.org/docs/rules/no-mixed-operators) rule is conflicting with the [`no-extra-parens`](http://eslint.org/docs/rules/no-extra-parens) rule. The latter has a setting that should prevent this, but I would suggest dropping the operator rule entirely, since it's really inflexible and even demands parens in cases like `var a = 5 + 10 / 2;`, where the order of operations is sufficient to make the result obvious.
- The [`no-undefined`](http://eslint.org/docs/rules/no-undefined) rule conflicts with [`no-void`](http://eslint.org/docs/rules/no-void). Maybe drop the latter?

pedrobaeza:
@moylop260 can you check these comments about the linter?

moylop260:
@obulkin
Thanks for comment about beta eslint.
I will change a few checks for these cases.
Thanks

:+1:

lasley:
LGTM, thanks @obulkin

PR #410, opened 2016-08-25
[IMP] make css applyable by ui on 'web_sheet_full_width_selective' module

bealdav:
Now width css class can be applied by functionnal user on any model.

img1


web_sheet_full_width_selective/__openerp__.py
```diff
@@ -1,37 +1,19 @@
...
 
 {
     'name': 'Show selected sheets with full width',
-    'version': '8.0.0.1.0',
+    'version': '8.0.2.0.0',
     'license': 'AGPL-3',
-    'author': 'Noviat, Odoo Community Association (OCA)',
-    'category': 'Hidden',
+    'author': 'Noviat, Akretion, Odoo Community Association (OCA)',
+    'category': 'Web',
```
pedrobaeza: This is in lower as the standard: `web`

bealdav:
ping @luc-demeyer

web_sheet_full_width_selective/models/model.py
```diff
@@ -0,0 +1,59 @@
+# coding: utf-8
```
legalsylvain, line 1: could you split this file in two file. (one file per model), and name the file ir_ui_view.py and ir_model.py ?

mourad-ehm:
:+1:  code review

web_sheet_full_width_selective/models/model.py
bealdav: Sylvain,
Whatever the rules define in guidelines, what is the gain to put 9 lines in another file whereas the whole file is only 61 lines ?
I have a better developer experience (#productivity) when I have a reasonable number of lines in my files than switch between tabs (search, replace, etc...).
Too many rules kill the rules. Python != Java. Don't you think ?

pedrobaeza: @bealdav It's a question of homogeneity in the files layout, no matter the number of lines. Please split the file and name each file with the model name.

legalsylvain: Hi ! 
well, another reason is to know more easely, which models are overloaded, when you discover an OCA module. if there is ir_ui_view in ir_model.py file, it is more difficult to read.
regards.

rvalyi: Hello, I won't go against the OCA rules, but let me tell you I agree with @bealdav and I would like this topic be debated again. With this rule, some modules like in our BR localization will soon have 50 files instead of 10 making them totally impossible to grasp... Just like Java projects...

rvalyi: and having 50 class files in a folder was doable with Java because you had the one click IDE cross file navigation when clicking on a method, but that's something we dont' have for Odoo Python dev...

pedrobaeza: I see more impossible to have 50 classes in a file, or several files with a weird grouping classification. The exception are models like sale.order.line or other one2many models that can go in the same file as sale.order.

We can't debate again and again each question. Please follow the already established rules.

pedrobaeza: A lot of Python IDEs do that, like Pycharm

lasley: > The exception are models like sale.order.line or other one2many models that can go in the same file as sale.order.

@pedrobaeza - That exception doesn't seem to be in our guidelines. I interpreted it as a hard 1:1 rule, which I fully agree with.

@rvalyi Komodo does this as well

pedrobaeza: @lasley Indeed it isn't, but it's in our mind. Here is the PR: https://github.com/OCA/maintainer-tools/pull/217

bealdav: > It's a question of homogeneity in the files layout, no matter the number of lines. Please split the file and name each file with the model name

@pedrobaeza then you consider you can follow the same practices with the same efficiency on all points, for this very small module and a very large modules as brazilian locale or account or connector. Homogeneity can't be full between large and small modules. They have different architect constraints.
I'm ok to say **"Developers must split python files to have an adhoc readibility. A good balance must be found with class or code with nearest concept in the same file and avoid long files**"

No metric here, just avoid real excess, trust in developer, give us enjoy develop for oca

Any text editor (often free) have a navigator class plugin (vim, pycharm, komodo, sublime). If they are there it's to be used ?

You talk about one2many relation exception. But they other cases where foreign is narrow between models: think about inheritS. You may have 'res.partner' with inherits model 'res.partner.skilled' or 'res.partner.carrier'. Have them in the same file is really benefit for readibility.

We shouldn't define all exceptions strictly. In which case it can be in the same file or not. Only general approach should be promote. We works an open source project all day. We must keep our minds opened. **Objective is readibility and productivity** of the developer (so many work to do tomorrow).
Rules must help OCA to grow and stay up to date with each odoo version, not bring us to restricted our contributions to oca

I'm not convinced we should follow strict rule on all points. Point on this matter could be simplify.

I know Pedro, you're finally a pragmatic developer, and community leader which have a long term vision. Keep important rules (python mostly). 

I have ever convinced you on an important point previously because you're pragmatically opened.

my 2 cents.

Thanks

legalsylvain: Well. Lot of noise ! Please @bealdav, just split this file (or let me do it !) ;-)
1/ It's not the place to discuss to change this rule
2/ having a ir_ui_view class in a file named model.py is confusing. Not really "productive" as you say

LGTM otherwise. thanks.
Regards.

pedrobaeza: In fact, in the name of the pragmatism, we can continue as this case is very trivial, but extending this practice across all your contributions will lead to bloated files like the Odoo ones, and new contributors that examine other modules to copy practices will be very confused with the heterogeneity.

Anyway, the final word is in the reviewers or the PSC, so in l10n-brazil you can agree another criteria skipping one of the rules/guidelines, but please in general repos like this one try to keep the rule.

bealdav:
@legalsylvain model is a generic filename that englobe all classes (2, 61 lines) of models.Model, no reason to rename.

web_sheet_full_width_selective/models/model.py
bealdav: @legalsylvain model is a generic filename that englobe all classes (2, 61 lines) of models.Model, no reason to rename. Thanks to have review all other points of the module.

See file name which indicate what are the models as you say, don't give so much info on module content.

With my text editor I see at glance what are the models of my file.
sub

With class browser you have the whole info, may be, Sylvain, it's time to move outside gedit ?
sub2

bealdav: > but extending this practice across all your contributions will lead to bloated files like the Odoo ones

@pedrobaeza, you're right, I don't want to have bloated files as Odoo. Then if your are OK, I can make a PR to define a way to express guidelines more comfortable for everyone on this matter.

pedrobaeza: I already opened a PR: https://github.com/OCA/maintainer-tools/pull/217

bealdav:
I try to write test, but my fields_view_get() call don't fall down in make_fields_view_get(), my hook

Any idea ?
EDIT: @pedrobaeza I would like say: module works fine, but I can't find a way to simulate the complete fields_view_get in my test like if it was a click on a menu. 

``` python
CSS_CLASS = 'oe_form_sheet_full_screen'


class TestAddCssClass(common.TransactionCase):

    def test_apply_css_class(self):
        '''' Test if css class is applied to sheet tag '''
        contact_form_view = self.env.ref('base.view_partner_form')
        contact_form_view.write({'form_width': CSS_CLASS})
        # action = self.env.ref('base.action_partner_form')
        contact = self.env['res.partner'].search([])[0]
        res = contact.fields_view_get(view_id=contact_form_view.id)
        # res = action.fields_view_get(view_id=contact_form_view.id)
        doc = etree.XML(res['arch'])
        node = doc.xpath('//sheet')
        class_in_tag = []
        if node:
            for current_node in node:
                class_in_tag = [
                    x[1] or '' for x in current_node.items()
                    if x[0] == 'class'] or ['']
        self.assertTrue(CSS_CLASS in class_in_tag[0])
```

web_sheet_full_width_selective/models/model.py
```diff
@@ -0,0 +1,72 @@
...
+        string='Form Width', selection='_get_form_width',
+        help="Allow to set the form view to the max width "
+             "to have a better usability on data entry")
+
+
+class ModelExtended(models.Model):
+    _inherit = 'ir.model'
+
+    def _css_class_to_apply(self, node, css_class):
+        ''' Complete class if exist '''
```
pedrobaeza: Docstring is put with doble quota

pedrobaeza:
Take a look at how at how base_field_validator does: https://github.com/OCA/server-tools/blob/8.0/base_field_validator/models/ir_model.py#L34-L83

web_sheet_full_width_selective/models/model.py
legalsylvain: Hi, 
I still really don't see the interest to break this rule, specially in that case (only two models). With your norm, we have to open the file to know which model is overloaded (or use rgrep, but the change of snake case to camel case for model is not finished I think). With the current OCA norm, it is not necessary.
The current https://github.com/OCA/maintainer-tools/pull/217 doesn't seem to allow this kind of design you are promoting.
I do not only discover code on my PC, but on github too (after odoo-code-search research) In that case, we have to open each model.py file to know what is inside. I'm not sure that github allows rgreping repository online...

That being said, that's not a "casus belli" for me, and I prefer to see your improvment in this OCA that not your improvment at all.

So, abstain.
Note : please, the choice of editor OCA developers usually use is really out of the scope of this PR...Otherwise, let's go talking about vim and emacs ! ;-)

Regards.

bealdav: Hi Sylvain, 
If subject is to avoid one click to see what are the tables.models, then promote to remove 'models' folder, and then you directly jump into the file with the whole code. 
In small modules (80 % of my contributions) they only have 1-2 files in model folder.
Ok, I go out now ;-)

PS: you can split my file in part if you want, no offense for me

bealdav:
@pedrobaeza @legalsylvain 
Simple question. With these changes, there is possibility to set width to different range according to css settings. Then is the module name 'web_sheet_full_width_selective' is the right name ? or should I create another module like 'web_sheet_size_selection' ?

pedrobaeza:
Yeah, the functional coverage has changed, but I think we can stay with this name, and take note for changing the name for v10 (as v9 already has the module migrated)

PR #411, opened 2016-08-27
[9.0] web_hide_db_manager link - Port to v9

JordiBForgeFlow:
This PR migrates the module web_hide_db_manager link from 8.0 to 9.0
cc @alejandrosantana


web_hide_db_manager_link/README.rst
```diff
@@ -19,6 +19,17 @@ To install this module, you need to:
  * Search by module name "Hide link to database manager in login screen" or by module technical name "*web_hide_db_manager_link*" 
  * Click install button
 
+Usage
```
elicoidal, line 22: Can you update to the latest README template?

elicoidal, line 22: (check the beginning of the file)

elicoidal:
:+1:

web_hide_db_manager_link/views/webclient_templates.xml
```diff
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
+<!-- Copyright Anubía, soluciones en la nube,SL (http://www.anubia.es)
+     Alejandro Santana <alejandrosantana@anubia.es>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <openerp>
```
lasley: `<odoo>`

web_hide_db_manager_link/views/webclient_templates.xml
```diff
@@ -1,10 +1,12 @@
...
     <data>
```
lasley: This can be removed

lasley:
👍

carlosdauden:
:+1:

web_hide_db_manager_link/README.rst
```diff
@@ -1,37 +1,57 @@
...
+.. image:: https://odoo-community.org/website/image/ir
+.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/9.0
+
 Known issues
 ============
 
-As Odoo v8 templates only live in the database, this module must be installed in each database you want this to be in effect.
+* As Odoo v8 templates only live in the database, this module must be
```
sbidoul, line 44: @jbeficent is this still the case in v9?

I'm surprised that such a module (that is by nature database independent) needs to be installed in databases.

JordiBForgeFlow: @sbidoul Yes, still the case in v9.

The login screen is generated from a template, and need to fetch data from some database. It will therefore be created using the definition set in the catched DB. When you switch to another DB, if that module is not installed the link will be visible again.

The database manager (web/database/manager) is, in contrast, a static HTML page and, as such, does not depend on the DB to render it's content.

sbidoul: Ah ok, thanks for clearing up my confusion.

PR #417, opened 2016-09-08
[IMP] help_popup module

bealdav:
- [x] Improve report esthetic
- [x] Better way to complete doc: dedicate view on action window
- [x] Allow to display / print the doc for a complete base menu (Sales, Purchase, etc) 
- [x] Provide a mechanism to update partially advanced_help and advanced_help_model fields: any other modules should can add or update some parts of these fields (data of several module in one fields)
- [x] Provide a mechanism to delete properly help data at uninstall of the module (several modules have modified help fields)
- [ ] add test unit

popup

---

`When you click on 'Sales help' above, you open the report below (all documentation of the menu sales).`
sales_doc

---

`Here is a dedicate view on action window to modify end users help fields`

img3


help_popup/model.py
```diff
@@ -1,35 +1,177 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
-
-from openerp import models, fields
+# © 2015 David BEAL @ Akretion <david.beal@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from BeautifulSoup import BeautifulSoup as BSHTML
```
dreispt: This is a new dependency, right? Isn't LXML suitable for the job?

dreispt:
I'm not trying to push anything on you, and only as a reminder it would be desirable that help_popup and help_online eventually converge into a single module.

bealdav:
@dreispt why not, have you a plan in mind about functionalities that should stay or be extended and with which dependencies ?
cc @lmignon @Cedric-Pigeon

rafaelbn:
Hi, I have been reviewing this module with @bealdav and it looks very simple and useful. About @dreispt comment https://github.com/OCA/web/pull/417#issuecomment-246958576, I agree that it's better not to depend in website as help_online do: https://github.com/OCA/web/blob/9.0/help_online/__openerp__.py#L30

This module is very interesting as potentially could solve the problem of Odoo with a full functional documentation made by end-users. In this moment the doc which end user write down is stored in Odoo database so if I do the doc I must export somehow to import for my other customers.

I will wait to have travis and runbot green to final review.

help_popup/tests/test_help.py
```diff
@@ -0,0 +1,44 @@
...
+    #     partner_act.with_context(install_mode=True).write(
+    #         {'advanced_help': 'toto'})
+    #     # import pdb; pdb.set_trace()
+    #     # vals = {'advanced_help': 'toto'}
+    #     # partner_act._update_help_field(vals, 'advanced_help', 'help_popup')
+    #     self.assertEqual(
+    #         partner_act.advanced_help,
+    #         u'\nnew_content<help_base>toto</help_base>',
+    #         "Partially updating of 'ir.model'.advanced_help is broken")
+    #     # partner_act.remove_obsolete_help('base')
```
petrus-v: Could you remove or fix comment code?

help_popup/models/action_window.py
```diff
@@ -0,0 +1,179 @@
...
+import inspect
+
+from openerp import _, api, models, fields
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from bs4 import BeautifulSoup as BSHTML
+except ImportError:
+    _logger.debug(
```
petrus-v, line 15: I guess this could be at least a warning error, if we want to install this module I guess this import is required

bealdav: this logger is for not installed module (oca guidelines)
There is raise in standard odoo installation module with 'external_dependencies' data in manifest.
No way to install if no dependency satisfied.

gurneyalex:
please look into https://travis-ci.org/OCA/web/jobs/164951115#L544

gurneyalex (changes requested):
please fix the tests: https://travis-ci.org/OCA/web/jobs/164951115#L544

PR #416, opened 2016-09-07
[9.0] Migrate web_m2x_options to 9.0

zakiuu:

PR #418, opened 2016-09-12
[9.0] Migrate web_m2x_options to 9.0

zakiuu:


sbidoul:
@zakiuu have you handled this https://github.com/OCA/web/pull/416#discussion_r78050624 from the other PR?

web_m2x_options/static/src/js/form.js
```diff
@@ -394,5 +392,25 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
                 return values;
             })
         },
+
+       render_value: function(){
+           var self = this;
+           return jQuery.when(this._super.apply(this, arguments))
```
lmignon: At top of your file add 

``` javascript
var $ = require("$");
```

And replace `JQuery.when` with `$.when`

web_m2x_options/static/src/js/form.js
```diff
@@ -394,5 +392,25 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
...
+           .then(function(){
+               if(self.options.open){
+                   self.$el.find('.badge')
+                   .css('cursor', 'pointer')
+                   .click(function(e){
```
lmignon: you should call a method declared on `FieldMany2ManyTags` so it will be possible to override it.  (??? `on_badge_click` ???)

web_m2x_options/static/src/js/form.js
```diff
@@ -394,5 +392,25 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
...
+                       var id = parseInt(jQuery(this).attr('data-id'));
```
lmignon: s/this/e.target

web_m2x_options/static/src/js/form.js
```diff
@@ -1,20 +1,57 @@
...
                    'web_m2x_options.search_more',
                    'web_m2x_options.m2o_dialog',];
 
-    instance.web.form.FieldMany2One = instance.web.form.FieldMany2One.extend({
+    var M2ODialog = Dialog.extend({
+        template: "M2ODialog",
+        init: function(parent) {
+            this.name = parent.string;
+            this._super(parent, {
+                title: _.str.sprintf(_t("Create a %s"), parent.string),
```
lmignon, line 32: PlZ add a `var _ = require('_');` since underscore is used in this JS file

lmignon (changes requested):
I don't know what's going wrong but the module seems to not work
image

web_m2x_options/static/src/js/form.js
```diff
@@ -150,15 +179,21 @@ openerp.web_m2x_options = function (instance) {
...
 
-                if (values.length > self.limit || self.search_more) {
+                if (values.length > self.limit && (can_search_more || search_more_undef || search_more)) {
                     values = values.slice(0, self.limit);
                     values.push({
                         label: _t("Search More..."),
                         action: function () {
+                            // limit = 80 for improving performance, similar
+                            // to Odoo implementation here:
+                            // https://github.com/odoo/odoo/commit/8c3cdce539d87775b59b3f2d5ceb433f995821bf
```
sbidoul: upstream, it's 160 not 80.

BTW, the code is so different from the upstream `get_search_result` that I wonder if it has diverged elsewhere.

sbidoul (commented):
@zakiuu don't forget this one, not a big deal but let's have the same behaviour as upstream.

lmignon:
@zakiuu It seems that you have to report a bug to Odoo. Your implementation can only work with Odoo Enterprise. Indeed in the community version of `web` the  widget `FieldMany2One` that you extend at https://github.com/OCA/web/pull/418/files#diff-d93aa186928a77371c865d1748af9b38R55 is not exposed by the js module `web.form_relational`  https://github.com/odoo/odoo/blob/9.0/addons/web/static/src/js/views/form_relational_widgets.js#L1734

The return should be 

``` javascript
return {
    FieldMany2ManyTags: FieldMany2ManyTags,
    AbstractManyField: AbstractManyField,
    FieldMany2One: FieldMany2One,
};
```

as in the enterprise version. Therefore the module can't work in 9c.

sbidoul:
@lmignon can you do the PR to Odoo? Can you also check with @adrienpeiffer I think he has done a similar PR recently.

lmignon:
@sbidoul Yes I can.

zakiuu:
@lmignon Also the M2ODialog, that's why I had to rewrite the whole code part related to it

lmignon:
This PR depends of https://github.com/odoo/odoo/pull/13518

pedrobaeza:
I had the same issue with one2many field and what I do is:

```
var FieldOne2Many = core.form_widget_registry.get('one2many');
```

web_m2x_options/static/src/js/form.js
njeudy: should use upstream one instead ?

lmignon:
@pedrobaeza I've implemented your proposal to make the module compatible with Odoo 9.c without the merge of the fix proposed to odoo in odoo/odoo#13518

web_m2x_options/static/src/js/form.js
```diff
@@ -19,6 +19,11 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
                    'web_m2x_options.search_more',
                    'web_m2x_options.m2o_dialog',];
 
+    // In odoo 9.c FieldOne2Many is not exposed by form_relational
```
pedrobaeza: s/FieldOne2Many/FielMany2One (in two places)

lmignon:
@pedrobaeza @njeudy changes done

pedrobaeza:
Please squash commits by author and check Travis status.

lmignon:
@pedrobaeza Errors in travis are in `web_widget_float_formula` I've already rebased the branch on 9.0...

pedrobaeza:
OK, thanks, what about the squashing? Is that possible?

lmignon:
@pedrobaeza done

pedrobaeza:
@Yajo, can you help us about the fail in the JS test?

yajo:
@lmignon is right; the guilty addon is `web_widget_float_formula`.

This is most probably one of the expected consequences of https://github.com/odoo/odoo/pull/13743: the test was probably failing before, but nobody noticed because Travis was getting fake :white_check_mark: status.

Another possibility is that this PR introduces the failure, and that's really possible because last builds from stable branches are all green... Could you check if your addon introduces any incompatibility, please @lmignon?

pedrobaeza:
@Yajo, I don't think so, because it fails in other PRs too. I have restarted the build in CI to be sure if in main branch is also incorrect. Are you able to provide a fix to get green Travis again?

yajo:
Right now I'm stuck in another task, I hope I can soon.

lmignon:
@pedrobaeza @Yajo Therefore, this one can be  reviewed and merged..

yajo:
#460 should fix your problem.

yajo:
#460 merged, please rebase

lmignon:
thank you @Yajo Rebase done...

pedrobaeza:
Can you please include https://github.com/OCA/web/pull/434 and squash commits by author and logic?

lmignon:
@pedrobaeza sorry but we spent too much time on this PR. I've already squashed the commit for members of Acsone and now you request the inclusion of a PR made after this one. Changes of #434  should be proposed in another PR and can not block this one.

pedrobaeza:
OK, @lmignon, I'm not blocking it. I'm asking if you do the favor of including that interesting PR, but if you don't want, I can't force you. Anyway, I would ask again for a better squashing, as I still continue seeing a lot of commits from "OCA Transbot" or Holger.

lmignon:
@pedrobaeza The commits from Holger and OCA Transbot are those from 8.0 (IMO we must keep these commits). The cherry-pick of #434 introduces conflicts and I've no time to try to resolve these conflicts (and to do the appropriate tests once merged to be sure that everything still works). 
My main concern is that PR after PR I  wonder if the balance benefits/costs remains positive when contributing to OCA....

pedrobaeza:
@lmignon the idea after squashing commits is to keep a clean commit history in each branch, which is treated isolated from previous versions. Doing this, we reduce the git size, ease searches through history and so on. This is something not about OCA, but code version management in general. If you don't want to do it, I can do it, but the PR then will count as closed instead of merge (as I have to change commit history before pushing).

lmignon:
@pedrobaeza I strongly disagree to squash commits from others branches. Keeping the same commits is the easiest way to compare 2 branches and detect which commits are missing. (except OCA transbot)

adrienpeiffer (approved):
:+1:

pedrobaeza:
OK, I don't want to push too much. Thanks for the squash.

Merging...

jlzhou:
#482 Migrate to odoo 10.0, add new 'open' option to control external button next to drop down.

PR #420, opened 2016-09-14
[IMP] web_widget_x2many_2d_matrix: New option field_att_

pedrobaeza:
Adding this option to the widget, you can define a field in the x2many
model that is evaluated for enabling or not the field in the matrix.
With this, you have a way to control if some matrix combinations are
not allowed.

NOTE: This doesn't prevent to require to fill the full matrix with
all the combination records.

### DIFF
--- a/web_widget_x2many_2d_matrix/README.rst
+++ b/web_widget_x2many_2d_matrix/README.rst
@@ -1,5 +1,6 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-    :alt: License: AGPL-3
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
 
 ===========================
 2D matrix for x2many fields
@@ -8,7 +9,8 @@
 This module allows to show an x2many field with 3-tuples
 ($x_value, $y_value, $value) in a table
 
-          $x_value1   $x_value2 
+========= =========== ===========
+\          $x_value1   $x_value2
 ========= =========== ===========
 $y_value1 $value(1/1) $value(2/1)
 $y_value2 $value(1/2) $value(2/2)
@@ -23,7 +25,9 @@ result could look like this:
 .. image:: /web_widget_x2many_2d_matrix/static/description/screenshot.png
     :alt: Screenshot
 
-The beauty of this is that you have an arbitrary amount of columns with this widget, trying to get this in standard x2many lists involves some quite agly hacks.
+The beauty of this is that you have an arbitrary amount of columns with this
+widget, trying to get this in standard x2many lists involves some quite ugly
+hacks.
 
 Usage
 =====
@@ -54,11 +58,23 @@ show_row_totals
     If field_value is a numeric field, calculate row totals
 show_column_totals
     If field_value is a numeric field, calculate column totals
+field_att_<name>
+    Declare as many options prefixed with this string as you need for binding
+    a field value with an HTML node attribute (disabled, class, style...)
+    called as the `<name>` passed in the option.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/8.0
 
 Example
 =======
 
-You need a data structure already filled with values. Let's assume we want to use this widget in a wizard that lets the user fill in planned hours for one task per project per user. In this case, we can use ``project.task`` as our data model and point to it from our wizard. The crucial part is that we fill the field in the default function::
+You need a data structure already filled with values. Let's assume we want to
+use this widget in a wizard that lets the user fill in planned hours for one
+task per project per user. In this case, we can use ``project.task`` as our
+data model and point to it from our wizard. The crucial part is that we fill
+the field in the default function::
 
  class MyWizard(models.TransientModel):
     _name = 'my.wizard'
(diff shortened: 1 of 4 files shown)


pedrobaeza:
@hbrunn your review is welcome.

cc @Tecnativa

web_widget_x2many_2d_matrix/README.rst
```diff
@@ -50,15 +53,25 @@ field_label_y_axis
     Use another field to display in the table header
 field_value
     Show this field as value
+field_editability
+    Use this field to evaluate if a cell should be editable or not. It
```
hbrunn: finish the sentence, probably explaining that this is supposed to be truthy if editable, and not truthy otherwise. How about a serialized field that returns a dict with properties such as `readonly` and `invisible`?

web_widget_x2many_2d_matrix/static/src/xml/web_widget_x2many_2d_matrix.xml
```diff
@@ -16,7 +16,13 @@
                         <th><t t-esc="widget.get_y_axis_label(y)" /></th>
                         <td t-foreach="widget.get_x_axis_values()" t-as="x" t-att-class="'oe_list_field_cell' + (widget.is_numeric ? ' oe_number' : '')" t-att-data-x="x">
                             <span t-att-class="widget.get_xy_value_class()">
-                                <input class="edit" t-att-data-id="widget.get_xy_id(x, y)" t-att-value="widget.format_xy_value(widget.get_xy_value(x, y))" />
+                                <t t-set="editable" t-value="widget.get_xy_editability(x, y)"/>
```
hbrunn: you could also say `t-attr-disabled="not widget.get_xy_editability(x, y) or None"` (an attribute with value `None` won't be set) - this way, you don't need to duplicate the code for the input

web_widget_x2many_2d_matrix/README.rst
hbrunn: now reading your actual implementation, how about a dict that is simply merged into the attributes using `t-att="widget.get_xy_editability(x, y)"`? Then it might make sense to rename the field, but I envision a lot of cool crazy stuff people can do with this

pedrobaeza: I don't get you. Can you explain what you want to do?

web_widget_x2many_2d_matrix/static/src/xml/web_widget_x2many_2d_matrix.xml
pedrobaeza: I think I tried that, but this is a special attribute that should be present or not. It's not possible to put `None`. Let me try again...

hbrunn: that's the trick with `None`. Either that or `False` (don't remember which one) causes the attribute not to exist at all. So the  function should return either `'disabled'` or `None`

web_widget_x2many_2d_matrix/README.rst
hbrunn: imagine I can return

``` python
{
    'disabled': 'disabled', # that's what you need
    'class': 'some class', # we can add some fancyness per field
    'onwhatever': 'injected javascript', # might be helpful in some use cases
    'style': 'display: none', # or just hide it?
    'out of ideas': 'but theres probably more cool stuff to do with this',
}
```

and all keys become attributes of the input. Very flexible.

web_widget_x2many_2d_matrix/static/src/xml/web_widget_x2many_2d_matrix.xml
pedrobaeza: This is a JS expression, but you're right, doing this works: `t-att-disabled="widget.get_xy_editability(x, y) ? undefined : 'disabled'"`

pedrobaeza:
@hbrunn I have implemented your suggestion and I'm improving slowly my JS skills. I have to implement an extra field in my TransientModel for computing `not bool(product_id)`, but it works. I have squashed directly because I'm starting the migration of the module and I need to cherry-pick this directly.

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -38,12 +21,16 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
         field_y_axis: 'y',
         field_label_y_axis: 'y',
         field_value: 'value',
+        // Putting this field by default, we always ensure editability
+        field_editability: 'id',
```
hbrunn: this one is obsolete then, right?

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -261,6 +254,26 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
...
+                    this.by_x_axis[x][y], this.fields_att[att]);
+                // Discard empty values
+                if (val) {
+                    vals[att] = val;
+                }
+            }
+            return vals;
+        },
+
+        get_xy_editability: function(x, y)
```
hbrunn: obsolete too?

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -261,6 +254,26 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
             return this.by_x_axis[x][y]['id'];
         },
 
+        get_xy_att: function(x, y)
+        {
+            vals = {};
```
hbrunn: in js, you should declare variables with the var keyword. Otherwise, you'll pollute the global namespace which can lead to weird and hard to debug interactions

hbrunn:
this is great! A few recommendations, but otherwise :+1:

pedrobaeza:
@hbrunn fixed!

lasley (approved):
TIL this module, very neat!

Changes look good, code and functional 👍

elicoidal:
:+1:

pedrobaeza:
Can you please merge as it's approved?

pedrobaeza:
Thanks!

PR #424, opened 2016-09-16
[IMP] Make web_environment_ribbon more generic

Garamotte:
By creating the ribbon div on the fly in Javascript, this module doesn't depend anymore on the webclient structure. I made this change to make the module compatible with the enterprise web client, but this can also simplify migrations :)

I hope that proposing a change in the module to make it compatible with the enterprise web client is allowed, as long as there is no specific code for this web client.

The diff displayed in Github is a bit hard to read because of a full block deindent.
Prefer "git show -b 60f6adca59fe50515eb7eaf45d1e8c86cfd20213" to have an easily readable diff.

### DIFF
--- a/web_environment_ribbon/static/src/css/ribbon.css
+++ b/web_environment_ribbon/static/src/css/ribbon.css
@@ -3,7 +3,7 @@
 
 .test-ribbon{
     width: 200px;
-    top: 50px;
+    top: 25px;
     left: -50px;
     text-align: center;
     line-height: 50px;
@@ -14,10 +14,11 @@
     -moz-transform: rotate(-45deg);
     -o-transform: rotate(-45deg);
     transform: rotate(-45deg);
-    z-index: 1000;
+    z-index: 9999;
     position: fixed;
     box-shadow: 0 0 3px rgba(0,0,0,.3);
     background: rgba(255,0,0,.6);
+    pointer-events: none;
 }
 
 .test-ribbon b {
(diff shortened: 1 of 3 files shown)


web_environment_ribbon/static/src/css/ribbon.css
```diff
@@ -14,7 +14,7 @@
     -moz-transform: rotate(-45deg);
     -o-transform: rotate(-45deg);
     transform: rotate(-45deg);
-    z-index: 1000;
+    z-index: 9999;
```
pedrobaeza, line 17: Why do you change this z-index? This makes the menu being behind the ribbon

Garamotte: This was intended, because long texts are partially hidden behind the menu instead.

If the ribbon can completely hide small menus, maybe making it thicker (10 px less) can avoid the issue ?

pedrobaeza: The idea is to not put too much large texts. Anyway, you can offset the text a bit to let it uncovered, but make this thicker is not the solution, as it's still over the menu.

Garamotte: If your concern is about clicking on the menu through the ribbon, I just added the "pointer-events: none" CSS property, to allow that.

sbidoul (approved):
Works fine on runbot. (functional review only)

web_environment_ribbon/static/src/css/ribbon.css
pedrobaeza: OK, this makes the trick good, thanks.

pedrobaeza (approved):

PR #425, opened 2016-09-20
[MIG] web_widget_float_formula: v9 with i18n

obulkin:
- Update JS to use v9 module system
- Fix non-functioning logic for obtaining the value of an input element
- Eliminate redundant calls (e.g. there were multiple calls to eval)
- Modify formula cleanup to use localized decimal point and thousands separator
  characters
- Add JS unit tests


web_widget_float_formula/views/web_widget_float_formula.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright GRAP
+    Copyright 2016 LasLabs Inc.
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <data>
```
lasley: `<odoo>` tag includes `<openerp>` and `<data>` in one. Remove this

lasley (approved):
Excellent job @obulkin - one minor note

obulkin:
Note: When I run the Python test in this module locally, it produces the following trace:

```
ERROR: test_js (openerp.addons.web_widget_float_formula.tests.test_js.TestJS)
File "/media/sf_LasLabs/v8_to_v9_Upgrades/Web_OCA/web_widget_float_formula/tests/test_js.py", line 14, in test_js
login="admin",
File "/var/www/odoo/openerp/tests/common.py", line 430, in phantom_js
self.phantom_run(cmd, timeout)
File "/var/www/odoo/openerp/tests/common.py", line 381, in phantom_run
self._wait_remaining_requests()
File "/var/www/odoo/openerp/tests/common.py", line 392, in _wait_remaining_requests
thread.join(0.05)
File "/usr/lib/python2.7/threading.py", line 940, in join
raise RuntimeError("cannot join current thread")
RuntimeError: cannot join current thread
```

This even happens with a completely empty JS test file, so there's likely some sort of issue in one of those HttpCase methods.

web_widget_float_formula/README.rst
```diff
@@ -1,55 +1,88 @@
...
+If you spotted it first, help us smash it by providing detailed and welcomed 
+feedback.
 
 Credits
 =======
 
+Images
+------
+
+* Odoo Community Association:
```
pedrobaeza: There's no icon, so this is not needed. Maybe you can add an explanatory itself by yourself.

obulkin: The module does have an `icon.png`. It's just not showing up in the PR because I didn't change it in any way. Do you still want me to remove this section?

pedrobaeza: Ah, OK, sorry for the confusion. Anyway, the icon is custom, so this reference is not valid. Please remove it.

obulkin: Fair enough. I'll remove it. v8 doesn't explicitly give anyone credit for the image, so I suppose it'll stay that way.

pedrobaeza: It seems the icon was added by @legalsylvain in https://github.com/OCA/web/commit/d6483bbc9ff761837bffa0489183414346ddc044

legalsylvain: Yes, no need to credit ! It's copyleft.

obulkin: Cool. Sounds like we're good to go then. I removed the portion crediting OCA for the image and made the change that @lasley requested above.

pedrobaeza (approved):

PR #426, opened 2016-09-20
New module web_notify

lmignon:
This technical module allows you to send instant notification messages from the server to the user in live.


hbrunn:
I think this is great! Just add some tests, then it's perfect

hbrunn:
BTW: I know already now that I'm going to use this to implement odoo's version of `wall` - or are you planning to do that anyways?

lmignon:
@hbrunn  I don't know what is 'wall' therefore nothing is planned on this topic.
My primary  need is just to send simple notification message to the end-users without raising a UserError. Moreover  it becomes possible  to send feedback informations for tasks run asynchronously .
Do you have some samples on the way I can unit test  this module?

hbrunn:
`wall` is the utility that allows admins of unixoid operating systems to send short messages to their users. Stuff like: "I'm going to take the server down in 5, save your stuff!": http://www.computerhope.com/unix/wall.htm - no worries about that, I'll come up with a PR soon enough.
About testing: It's difficult of course to test the asynchronous part, but also not necessary imho. If you just call your functions and then assert that they pushed the appropriate channel/string combination into postgres'  queue, or listen to them with a second cursor, that's fine I think

lmignon:
@hbrunn Thank you for the explanation. I've added the tests and a little bit improved the logic.

hbrunn (approved):
thanks!

web_notify/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+    @api.multi
+    @api.depends('create_date')
+    def _compute_channel_names(self):
+        for record in self:
+            res_id = record.id
+            record.notify_info_channel_name = 'notify_info_%s' % res_id
+            record.notify_warning_channel_name = 'notify_warning_%s' % res_id
+
+    notify_info_channel_name = fields.Char(
+        compute='_compute_channel_names', store=True, required=True)
```
hbrunn: you could consider removing the required flag, as the constraints can't be set (this only works if you set a default; see runbot errors), and the value is calculated anyways afterwards

lmignon: Thank you for the tip. Done

web_notify/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+    def _compute_channel_names(self):
+        for record in self:
+            res_id = record.id
+            record.notify_info_channel_name = 'notify_info_%s' % res_id
+            record.notify_warning_channel_name = 'notify_warning_%s' % res_id
+
+    notify_info_channel_name = fields.Char(
+        compute='_compute_channel_names', store=True)
+    notify_warning_channel_name = fields.Char(
+        compute='_compute_channel_names', store=True)
```
sbidoul: Any particular reason to have store=True?

lmignon: just to avoid to compute the values on each access even if in this case the computation is very simple

sbidoul: I'd think the perf gain is negligible and not worth an additional database column.

lmignon: Ok I've removed `store=True`

sbidoul:
:+1:

sbidoul (approved):

adrienpeiffer:
:+1:

pedrobaeza:
Can you add an screenshot before merging and maybe a nice icon?

lmignon:
@pedrobaeza Sure it would be a plus to have a screenshot and a nice icon.I'll provide a screenshot but I'm not the right people to provide a nice icon...

lmignon:
@pedrobaeza screenshot added

guewen (approved):

PR #427, opened 2016-09-21
[9.0][MIG] web_widget_x2many_2d_matrix: Migration and enhancement

pedrobaeza:
Migration of the module to v9 with these features:
- New JS API
- Solved problem with data feed from onchanges.
- Changes in structure.

There's also this improvement:
- Add 2 new attributes for the clickable possibility in the axis: `x_axis_clickable` and `y_axis_clickable`

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
@@ -2,12 +2,16 @@
  * Copyright 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
 
-openerp.web_widget_x2many_2d_matrix = function(instance)
-{
-    instance.web.form.widgets.add(
-        'x2many_2d_matrix',
-        'instance.web_widget_x2many_2d_matrix.FieldX2Many2dMatrix');
-    instance.web_widget_x2many_2d_matrix.FieldX2Many2dMatrix = instance.web.form.FieldOne2Many.extend({
+odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
+    "use strict";
+
+    var core = require('web.core');
+    var formats = require('web.formats');
+    var FieldOne2Many = core.form_widget_registry.get('one2many');
+    var Model = require('web.Model');
+    var data = require('web.data');
+
+    var WidgetX2Many2dMatrix = FieldOne2Many.extend({
         template: 'FieldX2Many2dMatrix',
         widget_class: 'oe_form_field_x2many_2d_matrix',
 
@@ -46,7 +50,7 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
             this.field_editability = node.attrs.field_editability || this.field_editability;
             this.show_row_totals = node.attrs.show_row_totals || this.show_row_totals;
             this.show_column_totals = node.attrs.show_column_totals || this.show_column_totals;
-            return this._super.apply(this, arguments);
+            return this._super(field_manager, node);
         },
 
         // return a field's value, id in case it's a one2many field
@@ -67,10 +71,10 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
         },
 
         // setup our datastructure for simple access in the template
-        set_value: function()
+        set_value: function(value_)
         {
             var self = this,
-                result = this._super.apply(this, arguments);
+                result = this._super(value_);
 
             self.by_x_axis = {};
             self.by_y_axis = {};
@@ -150,7 +154,7 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
                             {
                                 return;
                             }
-                            var model = new instance.web.Model(self.fields[field].relation);
+                            var model = new Model(self.fields[field].relation);
                             deferrends.push(model.call(
                                 'name_get',
                                 [_.map(_.keys(rows), function(key) {return parseInt(key)})])
@@ -171,7 +175,7 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
                             self.compute_totals();
                             self.setup_many2one_axes();
                             self.$el.find('.edit').on(
-                                    'change', self.proxy(self.xy_value_change));
+                                'change', self.proxy(self.xy_value_change));
                             self.effective_readonly_change();
                         }
                         return jQuery.when.apply(jQuery, deferrends);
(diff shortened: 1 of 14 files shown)


pedrobaeza:
cc @Tecnativa

pedrobaeza:
@hbrunn you'll be interested in this one too.

hbrunn (approved):
looks good! But the onchange-story bothers me a bit. I must admit I never tried this on 8, do we have the same problem there? For my feeling, if a widget needs help here, the widget is wrong. Can't we make this behave exactly as a standard one2many in this respect?
did you also consider having the widget generate empty records as necessary to get the full Cartesian product of the values available? Then using the widget would be a lot simpler.
I also experimented a bit with injecting fields here instead of using custom inputs, this turned out to be quite a hassle in 8. I _think_ (didn't try yet) this should be a lot simpler in 9
[the last two are just ideas of course, not in any way blocking]

pedrobaeza:
About the field injection and the Cartesian product, I have exhausted the development time (and also my spirit for trying something more in JS), so that can go better to roadmap section. It's of course very interesting.

The problem with onchange seems something inherent to ORM. We have found the same problem on one2many fields and in onchange https://github.com/OCA/server-tools/pull/492#issuecomment-237594285, so we need to see if we can patch it at one2many general widget or at ORM.

pedrobaeza:
@hbrunn I'm facing a problem resolving name of many2one fields when using a default value. Do you figure out why it's not resolved in that case?

From onchange:

seleccion_002

From default:

seleccion_003

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -50,10 +53,18 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
             this.field_y_axis = node.attrs.field_y_axis || this.field_y_axis;
             this.field_label_x_axis = node.attrs.field_label_x_axis || this.field_x_axis;
             this.field_label_y_axis = node.attrs.field_label_y_axis || this.field_y_axis;
+            this.x_axis_clickable = node.attrs.x_axis_clickable != undefined ? this.parse_boolean(node.attrs.x_axis_clickable) : this.x_axis_clickable;
+            this.y_axis_clickable = node.attrs.y_axis_clickable != undefined ? this.parse_boolean(node.attrs.y_axis_clickable) : this.y_axis_clickable;
```
yajo: To feed your love for JS a bit more... `null != undefined` is false.
Maybe you will want to use `!==`.

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -50,10 +53,18 @@ openerp.web_widget_x2many_2d_matrix = function(instance)
...
-            this.show_column_totals = node.attrs.show_column_totals || this.show_column_totals;
-            return this._super.apply(this, arguments);
+            for (var property in node.attrs) {
+                if (property.startsWith("field_att_")) {
+                    this.fields_att[property.substring(10)] = node.attrs[property];
+                }
+            }
+            this.field_editability = node.attrs.field_editability || this.field_editability;
+            this.show_row_totals = node.attrs.show_row_totals != undefined ? this.parse_boolean(node.attrs.show_row_totals) : this.show_row_totals;
+            this.show_column_totals = node.attrs.show_column_totals != undefined ? this.parse_boolean(node.attrs.show_column_totals) : this.show_column_totals;
```
Yajo: Same about `!==`

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
hbrunn: how about `this.parse_boolean(node.attrs.x_axis_clickable || (this.x_axis_clickable ? '1': '0'))`?

pedrobaeza: Well, in this case I think it can't be null. It exists or not, but not assigned to null. Am I wrong?

pedrobaeza: @hbrunn what do we win with that expression? Take into account that `this.x_axis_clickable` has been declared a couple lines above, and its value is fixed to `true`

hbrunn:
@pedrobaeza I'd put a breakpoint at https://github.com/OCA/web/pull/427/files/00ab29fbe49960abdd66c1f77b71923b9fd6961c#diff-d25cee8c92486a0e4cbf5ba21a360bbaR156 to check if the value there is something funny in the case of a default value. Probably you need an extra check for some other data type

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
hbrunn: then `this.parse_boolean(node.attrs.x_axis_clickable || '1')` would be simpler. The proposal is mainly about using javascript's notion of truthyness as we'd do in Python

pedrobaeza: OK, got it. I though JS doesn't convert correctly to boolean any data type the same way Python does

yajo: It converts to booleans implicitly, but not exactly like Python in all cases.

hbrunn: I'd be cautious about 'correctly', there are a lot of weirdnesses like empty arrays not being falsy and similar things. But for checking if something is set `value || default` works perfectly fine, and then you don't need to go into the details of `undefined`, `null`, `false` etc, as all of those are falsy

pedrobaeza: But for this case I think it is. I'm making tests right now to check.

pedrobaeza: Tried in all cases and it's working correctly, so I let Holger's version.

yajo: Like in Python, `"0"` is true. Not sure if it applies here, otherwise ignore this :grin:

pedrobaeza: No, it doesn't apply here. We only want to have a default in case you don't specify this attribute in the XML. If the widget user is putting specifically "0" in the option, then it should be passed to `parseBoolean` function, which is what is done right now.

pedrobaeza:
@Yajo, can you make the final check and approve this?

yajo (approved):

pedrobaeza:
@hbrunn can we proceed with the merge?

pedrobaeza:
Thanks!

PR #430, opened 2016-09-29
MIG 9.0 support_branding

angelmoya:


support_branding/README.rst
```diff
@@ -52,6 +54,20 @@ add the following XML::
...
 the default is empty!
 
+Usage
+=====
+
+To use this module, you need to:
+
+#. On developer mode go to Settings > Technical > Paremeters > System Parameters
+#. Configure parameters related with support branding
+#. Support branding is showed on left menu and when raises an exception
```
hbrunn, line 64: something's missing here

support_branding/README.rst
```diff
@@ -69,6 +85,7 @@ Contributors
 
 * Holger Brunn <hbrunn@therp.nl>
 * Stefan Rijnhart <srijnhart@therp.nl>
+* Angel Moya <odoo@tecnativa.com>
```
hbrunn, line 88: after you rewrote the js part to v9 api, I'd consider it a serious contribution. As it is, I think this is not warranted

support_branding/static/src/js/support_branding.js
```diff
@@ -1,22 +1,11 @@
...
-   customers without an OPW, you can brand the OpenERP instance
-   accordingly using this module. Please enter the two variables
-   in the code below, 'support_name' and 'support_link'. They will
-   replace the unfriendly message about the OpenERP instance not
-   being supported.
-           
- */
-   
-openerp.support_branding = function(instance) {
+odoo.support_branding = function(instance) {
```
hbrunn: rewrite to v9 api

hbrunn (changes requested):
you miss commits from 8.0 - follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0 to have them in your PR

angelmoya:
@hbrunn thanks for yout comments I will rewrite

moylop260 (approved):

angelmoya:
@hbrunn I rewrite js, please check it

support_branding/static/src/js/support_branding.js
```diff
@@ -1,114 +1,90 @@
...
-        {
+odoo.define('web.support_branding', function(require) {
+    var CrashManager = require('web.CrashManager');
+    var core = require('web.core');
+    var _t = core._t;
+    CrashManager.extend({
+        init: function() {
             var self = this,
-                ir_config_parameter = new openerp.web.Model('ir.config_parameter');
+                ir_config_parameter = new Model('ir.config_parameter');
```
hbrunn, line 13: where do you require `Model`? I think you should require `web.Model`

hbrunn (changes requested):
on runbot, go to the app store, this will trigger an exception. But the exception popup doesn't have the send button, nor the text field to fill in stuff. So it seems you'll have to adapt something there too

angelmoya:
You need to set support email on parameters to get this form and send
button.
I will fix this js

support_branding/static/src/js/support_branding.js
angelmoya: It was fixed

hbrunn:
I just filled in an email address, but still no send button on the exception

angelmoya:
I test... send buttons is showed and open system app to send email... but you need to scroll to see that

angelmoya:
OK, tested on Runbot with module test-exceptions....

It's works nice with Undefined type error, not with other error types, I will check it

angelmoya:
Finally I tested with test-exceptions, and is working well.

It show email form on ERROR and not show on WARNING.

I can do also to show form on warning, but I think that email form is needed for ERROR and not for WARNING, because warning could be use to show information to user.

This is the xml that is extended to show email form:

https://github.com/OCA/OCB/blob/9.0/addons/web/static/src/xml/base.xml#L31

And this is the record for the warning:

https://github.com/OCA/OCB/blob/9.0/addons/web/static/src/xml/base.xml#L22

If it's needed to show email form on warning we only need to add to this record.

I think that is not needed for warning, what do you think @hbrunn ?

pedrobaeza:
@angelmoya Where's the updated code?

angelmoya:
I did not update code, it was right, I only review functionality

pedrobaeza:
OK, let's see what @hbrunn says

hbrunn:
I just made @angelmoya's homework and debugged through this, finding that this cannot ever work because of https://github.com/angelmoya/web/blob/d0865bbfa1d0fb79604837ee7db13b567840ecd4/support_branding/static/src/js/support_branding.js#L10 - `extend` returns a reference to the new object, but doesn't touch the original object. What you want here is `include`. Do you actually test the code you write?

And then there's still https://github.com/OCA/web/pull/430#discussion_r81072080 pending

hbrunn:
superseded by https://github.com/OCA/web/pull/469

PR #434, opened 2016-10-06
web_m2x_options functionality aligned with the specification

jan-sedivy:
According to the module documentation https://www.odoo.com/apps/modules/8.0/web_m2x_options/
that there should be new options available at the field level called "create" and "create_edit", but there are none such used in the get_search_result(). Instead the function looks for no_create, no_create_edit and no_quick_create not mentioned in the documentation.
On top of it, the condition
if (!no_quick_create && ((m2x_create_undef && can_create) || m2x_create)) {
seems a bit odd to me as looks like the the m2x_create (self.view.ir_options['web_m2x_options.create'] == "True") entirely overrides the can_create parameter returned probably from the access right checking. That sounds quite dangerous to me.

We had been using this code on version 8.0 for more than half a year with no problems observed.
The code base of version 9.0 is slightly different but as the module is anyway listed in the "Unported addons" section of the 9.0 description I believe branch 8.0 is the right place to commit our changes

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -184,15 +184,19 @@ openerp.web_m2x_options = function (instance) {
                 var raw_result = _(data.result).map(function (x) {
                     return x[1];
                 });
-                var no_quick_create = (
-                    self.options && (self.options.no_create ||
-                                     self.options.no_quick_create)
-                )
-                var m2x_create_undef = _.isUndefined(self.view.ir_options['web_m2x_options.create'])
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Have you sent the CLA?

jan-sedivy:
Yes, I had sent the CLA on October 6th.

pedrobaeza:
CLA checked

pedrobaeza (approved):

njeudy (approved):
Tested on V8.0

Ok for me

Issue #431, opened 2016-10-02
[9.0] web_widget_float_formula


fmdl:
Hello

This module don't works when I use in fload field (I have use this formula : =2*3), I have an error message :

Uncaught TypeError: Cannot read property 'toString' of undefined
http://localhost:8069/web/content/1080-ac38307/web.assets_backend.js:7187
Traceback:
TypeError: Cannot read property 'toString' of undefined
    at OdooClass._process_formula (http://localhost:8069/web/content/1080-ac38307/web.assets_backend.js:7187:683)
    at OdooClass._compute_result (http://localhost:8069/web/content/1080-ac38307/web.assets_backend.js:7191:87)
    at OdooClass.trigger (http://localhost:8069/web/content/512-863c2d1/web.assets_common.js:2618:180)
    at OdooClass.trigger (http://localhost:8069/web/content/512-863c2d1/web.assets_common.js:2620:669)
    at HTMLInputElement.blur (http://localhost:8069/web/content/1080-ac38307/web.assets_backend.js:3701:3209)
    at HTMLInputElement.dispatch (http://localhost:8069/web/content/512-863c2d1/web.assets_common.js:740:451)
    at HTMLInputElement.elemData.handle (http://localhost:8069/web/content/512-863c2d1/web.assets_common.js:713:173)

Regards

lasley:
@IguanaYachts - Thanks for the report, issue confirmed. 

For further information, the error is superficial and only occurs once per page load. The math still seems to be taking place correctly despite the error, and subsequent formulas do not yield an error.

Offending code. Solution is to check if `formula` is defined before regex. I will submit a PR in the next few days

```
            _process_formula: function(formula) {
                var clean_formula = formula.toString().replace(/^\s+|\s+$/g, '');
```

PR #437, opened 2016-10-10
[IMP] web_widget_float_formula: Focused bug fix

obulkin:
- Refactor to address bug reported in #431 in more focused fashion, rather
  than with broad try/catch


lasley (approved):
Better than my fix for sure. Thanks for investigating the root cause @obulkin

pedrobaeza (approved):

PR #440, opened 2016-10-12
[ADD] web_pivot_zoom

hbrunn:
# 
# Zoom in pivot views

This module was written to allow your users to zoom into cells in pivot views
by simply clicking a cell in question.
# Installation

Don't install this module manually, it won't do anything. A module that uses it
will pull it as dependency.
# Usage

# . depend on this module
# . in your measure elements in your pivot view, declare an options dictionary

   as in::
    <field name="field" type="measure" options="{'web_pivot_zoom': {}}" />

   The options dictionary must have a key `model` to inform the module which
   model should be shown, and a key `domain` for the domain to be used.
   The domain is subject to evaluation, so you can use the grouped values in
   there. Note that if a field is not grouped, it will evaluate as `None`, so
   the `=?` is your friend for expressions refering to the grouped fields.

   Note that not all field types support the `=?` operator. For many2many
   fields, you'd have to use a construction like
   `('many2many_field', many2many_field and '=' or '!=', many2many_field)`
   to simulate this to some degree.

   Consult the demo data for an example, there, a graph view is added to the
   users model, and clicking cells sends you to partners with the company in
   question.

Measures with a configuration enabling zooming will have a small link icon next to the column title.


web_pivot_zoom/static/src/js/web_pivot_zoom.js
```diff
@@ -0,0 +1,205 @@
+//-*- coding: utf-8 -*-
+//© 2016 Therp BV <http://therp.nl>
+//License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
thomaspaulb, line 3: Shouldn't you use [this syntax](https://github.com/OCA/maintainer-tools/blob/master/template/module/static/tests/js/module_name.js#L1)?

web_pivot_zoom/static/src/js/web_pivot_zoom.js
```diff
@@ -0,0 +1,205 @@
...
+            if(!model || !domain)
+            {
+                return;
+            }
+            return this._web_pivot_zoom_action(
+                model, domain, current_field, row_header, col_header
+            );
+        },
+        _web_pivot_zoom_action: function(
+            model, domain, current_field, row_header, col_header
```
thomaspaulb, line 130: Why are you passing the `col_header` here if you are not using it later?

web_pivot_zoom/static/src/js/web_pivot_zoom.js
```diff
@@ -0,0 +1,205 @@
...
+            {
+                group_values[measure.field] = measure_values[i];
+            });
+            _.each(this.pivot_options.invisible_measures, function(measure, i)
+            {
+                group_values[measure] = measure_values[
+                    i + self.pivot_options.measures.length
+                ];
+            });
+            var model = this.graph_view._web_pivot_zoom[current_field.field]
```
thomaspaulb: You are using this expression two times, also 4 lines below, can you give it a meaningful name and declare it once instead?

web_pivot_zoom/static/src/js/web_pivot_zoom.js
```diff
@@ -0,0 +1,205 @@
...
+                ).attr('data-id'),
+                row_id = $current_row.find('[data-id]').attr('data-id'),
+                current_field = this.pivot.measures[
+                    (e.currentTarget.cellIndex - 1) %
+                    this.pivot.measures.length
+                ],
+                row_header = this.pivot.get_header(row_id),
+                col_header = this.pivot.get_header(col_id),
+                measure_values = this.pivot.get_values(
+                    Math.min(col_id, row_id), Math.max(col_id, row_id)
```
thomaspaulb, line 70: Why don't you just use `col_id`, `row_id` here? Does it have to be in order of id's, and if yes, can you document why?

web_pivot_zoom/static/src/js/web_pivot_zoom.js
```diff
@@ -0,0 +1,205 @@
...
+                    ),
+                    col_value = group.attributes.value.slice(index),
+                    row = self.find_or_create_header(
+                        row_headers, row_value, group
+                    ),
+                    col = self.find_or_create_header(
+                        col_headers, col_value, group
+                    ),
+                    values = self.get_values(
+                        Math.min(col.id, row.id), Math.max(col.id, row.id)
```
thomaspaulb, line 208: You are repeating this trick again here - maybe clearer to define a common function that does this, and comments why it is so?

pedrobaeza:
As information, Odoo 10.0 does this in core.

web_pivot_zoom/static/src/js/web_pivot_zoom.js
hbrunn: this has nothing to do with database ids, those is how the internal datastructure works: https://github.com/OCA/OCB/blob/8.0/addons/web_graph/static/src/js/pivot_table.js#L88 - I can add a link to that, but I'd expect interested readers to read this piece of code themselves

web_pivot_zoom/static/src/js/web_pivot_zoom.js
hbrunn: I do that as soon as the the amount of lines to define the function exceeds the amount of lines the function replaces

web_pivot_zoom/static/src/js/web_pivot_zoom.js
hbrunn: this is a precaution for the future - we might need it later, and then it would be a pity to have to change the api. I usually try to make my code extensible by default

web_pivot_zoom/static/src/js/web_pivot_zoom.js
thomaspaulb: OK perhaps it's just a matter of taste, but improving code readibility and adding DRY trumps the number of lines IMO

web_pivot_zoom/static/src/js/web_pivot_zoom.js
thomaspaulb: As there are also no OCA guidelines on documenting code, it's your choice how you do it and my responsibility to study the code and supply you a PR with docstrings on every part. 

That said, other contributors and reviewers will be productive on your code much faster if the meaning of a piece of code is self-evident from the naming of the variables and functions, and if that is not possible, carry a small note of explanation. The same way you expect junior OCA contributors to polish their contribution to make things as easy as possible for the senior reviewer, the senior contributor could also strive to make life easier for junior contributors and reviewers - all in the interest of producing better code with less bugs.

hbrunn: you're right. the thing is that I seriously started a commit addressing that, but what I came up with won't help, because I either reproduced the whole code in natural language, or write too little, and then you need to code anyways.
Now it's just the link above.

PR #442, opened 2016-10-14
[LANG] Add Dutch tranlations for help_online

Yenthe666:
Added some Dutch for you guys.

### DIFF
--- a/help_online/i18n/nl.po
+++ b/help_online/i18n/nl.po
@@ -1,0 +1,169 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* help_online
+#
+msgid ""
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Are you sure the target branch is correct? You have also Transifex for this. We have here the risk to override not synchronized translations (which is not too harm though)

Yenthe666:
Hi @pedrobaeza,

Hmm yes it is probably better to do this in Transifex, I just wanted to translate it here quickly.
Perhaps I should just close this one and do it on Transifex?

Regards,
Yenthe

pedrobaeza:
Yes, please. You can also import PO files in Transifex I think.

PR #443, opened 2016-10-14
[MIG][10.0] web_responsive

lasley:
[MIG] web_responsive: Migrate to v10  …
- Bump version
- Change all openerp to odoo
- Fix qunit suite injection
- Remove jQuery require
- Change to new selectors:
  - `oe_leftbar` to `o_sub_menu`
  - `oe_logo` to `o_sub_menu_logo`
  - `oe_footer` to `o_sub_menu_footer`
  - `oe_secondary_menus_container` to `o_sub_menu_content`
- Add style to hide oe_footer instead of remove to not break `support_branding`
- Add note in ReadMe explaining override of `support_branding`
  - [x] WIP until #446 is merged in order to include all fixes here


web_responsive/static/src/js/web_responsive.js
```diff
@@ -4,7 +4,6 @@
 odoo.define('web_responsive', function(require) {
     'use strict';
 
-    var $ = require('$');
```
lasley: Both this and `require('jQuery');` fail. I assume this means it is no longer necessary

lasley:
This actually looks a lot better on v10 too, pleasant surprise

pedrobaeza:
I'm wondering if we should be able soon to backport master version of _web_enterprise_ module passed to community.

lasley:
I was thinking the same thing, and shared this thought with @Yajo & @antespi at the Odoo days. 

Jairo shared his experiences regarding his backports of the v10 portals to v9 and somewhat scared me away from the initiative. 

I've been thinking, however, that we may be able to bring in Odoo SA LESS without anything else. This might hopefully maintain a compatibility bridge & not require a bunch of stylistic changes for OCA conformity.

Just thoughts - I've been watching master for some hints as to what the new web will look like, but nothing as of yet.

pedrobaeza:
OK, let's keep an eye on the master activity and we can judge then.

About this PR, to mention that the drawer is hidden when you click on Discuss app. Please check.

lasley:
Nice catch, fixed

web_responsive/static/src/less/navbar.less
```diff
@@ -3,6 +3,7 @@
 
 .main-nav {
     margin-bottom: 0;
+    z-index: 9000;
```
pedrobaeza: With this we are hiding some elements behind (like the notification bar for adding in the browser)

yajo:
They said that in v11 the web client will be refactored. Maybe backporting the LESS would be enough, but we won't know until it hits master, so I guess this is the current path to follow, mostly given most job is already done.

lasley:
#446 changes integrated & a few style fixes applied to compensate. This is ready for review

legalsylvain:
Hi.
I tested, installing sale module (on runbot, and menu is now unavailable). http://3270342-443-603eaf.runbot1.odoo-community.org/
do you face the same problem ? 
regards.

(Ubuntu / firefox 49.0)

lasley:
Thanks @legalsylvain looks like we regressed back to https://github.com/OCA/web/pull/443#issuecomment-253995640 - will look into it 😄

lasley:
Woot the solution was as simple as adding `o_web_client` to the body. Gotta love it when it's easy!

pedrobaeza:
Trying the module I have found 2 problems:
- Some apps appears duplicated and without icon.
- Tooltips appears behind the drawer.

seleccion_004

lasley:
Thanks for the review @pedrobaeza - I was able to replicate the tooltip issue, but can't seem to replicate the duplication. What browser are you in? Are you seeing it on all sizes, or is it particular to medium / mobile?

yajo:
There's another problem still. Go to "Sales" in runbot and see that the message count badge is misaligned:

captura de pantalla de 2016-10-25 09-34-13

pedrobaeza:
I'm afraid this happens also in v9.

lasley:
Hah we missed so many bugs - I did warn that I suck at design before starting this 😆  

@pedrobaeza - have you been able to replicate the duplication and no icon issue? If so, what browsers and sizes?

yajo (changes requested):
Among the changes that were introduced in the code sprint, if I don't remember badly, the above part was scrolling too.

Right now, the above part is fixed while the bottom one scrolls, and it is not much useful in phones, where the screen is small enough already. Could you fix that too please?

captura el 2016-10-26 a las 10 24 22

pedrobaeza:
About the duplication, it happens when the system is not fully loaded, and the image is not fetched, but the alt of the image. We can dismiss that issue.

lasley:
Ah yeah I get some weird issues when I start getting click happy on initial load. Glad we can ignore that 😉 

Doesn't seem that we have scrolling for the Control Panel in any version, but I should probably be able to make it happen with a little CSS

lasley:
I have mixed feelings on what I just did, so it's got a few z-index bugs while deciding how it should be finalized. 

Scroll issue is fixed by changing the control panel margin-top to negative of the scroll position. This yields a pretty nice effect where the control panel scrolls with the page contents. 

It also hides the top nav bar once the page scrolls further than the height of the control panel, which is where I have the mixed feelings. On one hand, it's nice having the full screen space for views. On the other, that forces a scroll to top in order to nav.

On iPhone this isn't too big of a deal because you can click the status bar at the top of the screen to init a scroll to the top. Does Android have similar functionality? 

Any other thoughts?

yajo:
> On iPhone this isn't too big of a deal because you can click the status bar at the top of the screen to init a scroll to the top. Does Android have similar functionality? 

I guess it depends on the browser, but usually scroll speed in Android is much faster than on iPhones, so a quick drag gets you to the top in no time.

yajo (approved):

pedrobaeza (approved):

pedrobaeza:
Let's merge for now

pedrobaeza:
Please squash last commits, Dave.

lasley:
Sounds good, I need to fix some z-index issues with the new feature before merge. Will do that today

image

web_responsive/static/src/less/navbar.less
```diff
@@ -0,0 +1,46 @@
+/* Copyright 2016 LasLabs Inc.
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+.main-nav {
+
+    margin-bottom: 0;
+    min-height: @app-drawer-navbar-height;
+    z-index: 1;
```
lasley, line 8: This is what I added to fix overlap of elements onto navbar

lasley:
I only changed one line to fix the overlap issue mentioned above, so I went ahead and squashed anyways. The one change is noted inline on navbar.less

yajo:
OK, let's wait for 3rd approval to merge.

web_responsive/views/web.xml
```diff
@@ -0,0 +1,254 @@
...
+
+                <nav class="navbar navbar-default main-nav"
+                     role="navigation"
+                     groups="base.group_user,base.group_portal"
+                     >
+                    <div class="container-fluid">
+
+                        <div class="navbar-header">
+
+                            <a href="#"
```
yajo: Could you please remove this href? It's creating a history entry each time you click on the drawer, and I think it will work equally without it...

yajo (changes requested):

lasley:
Note to self; Need to raise z-index on `o_user_menu dropdown-menu`

image

lasley:
Another z-index note, notifications below app drawer should not appear in app drawer:

image

image

yajo (approved):

yajo:
Labeled as needs fixing until you approve yourself :laughing:

lasley (commented):
😆  I now approve myself 👍 

I rebased so it can be merged. The fixes were just a few minor z-index tweaks to allow a bit better layering

web_responsive/static/src/js/web_responsive.js
yajo: [So it seems...](https://github.com/odoo/odoo/commit/d56b35e26e4b578c5f6f434d336edd006d321860#diff-dfbc96e4758b063efc18cb19c9719fc9L41)

yajo (changes requested):
Something got broken with last update, on runbot click on the drawer and all the window gets greyed out.

lasley:
Crap I forgot Odoo 10 requires restarts for most changes to take place on web now - I've never run into so many disparities between my local and Runbot before. Oh well, back to the dev table 😛

lasley:
At the risk of sounding like a broken record, done! 😆

yajo (approved):

pedrobaeza (approved):

Issue #435, opened 2016-10-06
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
- [x] help_online - By @lmignon - #526 
- [ ] help_popup - By @daramousk - #624
- [x] support_branding - ~~By @nikiwaibel - #465~~ By @robertrottermann - #465 
- [x] web_access_rule_buttons - By @daramousk - #628 
- [ ] web_action_conditionable - By @daramousk - #635 
- [x] web_advanced_search_wildcard - By @tafaRU - #822 
- [x] web_advanced_search_x2x - By @rami-wafaie - #600
- [ ] web_ckeditor4 - By @daramousk - #820 
- [x] ~~web_color~~ - Replaced by _web_widget_color_
- [x] web_confirm_window_close - By @stefanrijnhart - #631 
- [ ] web_context_in_colors - By @spacemonkeythe - #771 
- [ ] web_dashboard_open_action - By @spacemonkeythe - #770 
- [ ] web_dashboard_tile
- [x] web_dialog_size - By @bud-e - #550 
- [ ] web_dom_model_classes
- [ ] web_duplicate_visibility - By @lmignon - #557 
- [x] ~~web_easy_switch_company~~ (Obsolete) By @legalsylvain #450
- [x] web_editor_background_color - By @Yajo - #665 
- [x] web_environment_ribbon - By @sylvain-garancher - #436 
- [x] web_export_view - By @jlzhou - #444
- [x] web_favicon - By @nikiwaibel - #459 
- [ ] web_graph_improved
- [ ] web_group_expand - By @JayVora-SerpentCS
- [x] web_hide_db_manager_link - by @JayVora-SerpentCS - #571
- [x] web_ir_actions_act_window_message - By @benwillig - #640 
- [x] web_ir_actions_act_window_page - By @StefanRijnhart - #729 
- [ ] web_last_viewed_records - By @JayVora-SerpentCS
- [x] ~~web_listview_custom_element_number~~ Included in core
- [x] web_m2x_options - By @chienandalu - #622
- [ ] web_menu_navbar_needaction - By @hbrunn - #1028 
- [x] web_notify - By @JayVora-SerpentCS - #491 
- [ ] web_option_auto_color
- [ ] web_polymorphic_many2one > web_widget_many2one_polymorphic - By @hbrunn - #884 
[...]

PR #444, opened 2016-10-14
[MIG 10.0] web_export_view

jlzhou:
Based on Rodney's pull request '9.0 web export view rv #346'.
Avoided to export the sequence column with handle widget.
Adapted to css class name of odoo 10.0.


jlzhou:
migrate web_export_view to odoo 10.0 #435

rafaelbn:
This module is also being migrated to v9 here: https://github.com/OCA/web/pull/346

jlzhou:
I've signed and sent the CLA out, new line was added to the end of 2 files. What's the next?

jlzhou:
@rafaelbn updated, please review.

jlzhou:
@lasley Please advise what else should be fixed? Thanks.

web_export_view/__init__.py
```diff
@@ -1,21 +1,8 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2016 Henry Zhou (http://www.maxodoo.com)
```
lasley: Are all these copyrights necessary for a file with one line of code? Surely not everyone on this list modified this file.

web_export_view/__manifest__.py
```diff
@@ -1,40 +1,28 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2016 Henry Zhou (http://www.maxodoo.com)
```
lasley, line 2: Missing `Domsense srl` copyright

web_export_view/controllers/__init__.py
```diff
@@ -1,21 +1,8 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2016 Henry Zhou (http://www.maxodoo.com)
```
lasley: Are all these copyrights necessary for a file with one line of code? Surely not everyone on this list modified this file.

web_export_view/controllers/controllers.py
```diff
@@ -1,32 +1,18 @@
...
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2016 Henry Zhou (http://www.maxodoo.com)
+# Copyright 2016 Rodney (http://clearcorp.cr/)
+# Copyright 2012 Agile Business Group
+# Copyright 2012 Therp BV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 try:
```
lasley: AFAIK Odoo requires Python 2.7.x, which will always have the `json` library. This guard is unnecessary

web_export_view/i18n/de.po
```diff
@@ -1,44 +0,0 @@
-# Translation of OpenERP Server.
```
lasley, line 1: Why deleting these files?

web_export_view/static/src/js/web_export_view.js
```diff
@@ -1,38 +1,24 @@
...
-//    but WITHOUT ANY WARRANTY; without even the implied warranty of
-//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-//    GNU Affero General Public License for more details.
-//
-//    You should have received a copy of the GNU Affero General Public License
-//    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-//
-//#############################################################################
-openerp.web_export_view = function (instance) {
+odoo.define('web_export_view.web_export_view', function (require) {
```
lasley: No need to duplicate the name IMO `odoo.define('web_export_view'` is fine

web_export_view/static/src/xml/web_export_view_template.xml
```diff
@@ -2,11 +2,8 @@
 
 <templates id="template" xml:space="preserve">
 <t t-name="AddExportViewMain">
-    <div class="oe_form_dropdown_section">
-        <button class="oe_dropdown_toggle oe_dropdown_arrow">Export Current View</button>
-        <ul class="oe_dropdown_menu">
-            <li class="oe_sidebar_export_view_xls"><span>Excel</span></li>
-        </ul>
+    <div class="o_cp_sidebar btn-group btn-group-sm">
+        <button class="btn btn-default export_treeview_xls" type="button"><i class="fa fa-file-excel-o"></i></button>
```
lasley: It would be nice if this were on multiple lines, the wrapping caused me to have to double take that it was all the same line.

Also, that `<i>` should have a `t-translation="off"` in order to avoid funky translations

web_export_view/views/web_export_view_view.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<openerp>
+<odoo>
     <data>
```
lasley: You can remove this data tag, it is implicit with the new odoo tag

lasley (changes requested):
Thanks @jlzhou - comments inline - code, no test

web_export_view/__manifest__.py
```diff
@@ -1,40 +1,28 @@
...
 
 {
-    'name': 'Export Current View',
-    'version': '8.0.1.2.0',
+    'name': "Web Export Current View",
+    'version': '10.0.1.0.0',
     'category': 'Web',
-    'author': "Agile Business Group,Odoo Community Association (OCA)",
-    'website': 'http://www.agilebg.com',
+    'author': 'Henry Zhou (MAXodoo), '
```
lasley: Why removing an author?

web_export_view/controllers/__init__.py
```diff
@@ -1,21 +1,8 @@
...
+# Copyright 2016 Rodney (http://clearcorp.cr/)
+# Copyright 2012 Agile Business Group
+# Copyright 2012 Therp BV
```
lasley: Where did Therp come from in this?

web_export_view/i18n/de.po
jlzhou: For improving user experience, I changed the UI to use button instead of menu item, so the translation of menu is unnecessary.

web_export_view/controllers/__init__.py
jlzhou: Since this pull request referred to Therp BV's pull request. https://github.com/OCA/web/pull/346

web_export_view/static/src/xml/web_export_view_template.xml
jlzhou: good advice, thanks!

web_export_view/static/src/xml/web_export_view_template.xml
```diff
@@ -2,11 +2,10 @@
 
 <templates id="template" xml:space="preserve">
 <t t-name="AddExportViewMain">
-    <div class="oe_form_dropdown_section">
-        <button class="oe_dropdown_toggle oe_dropdown_arrow">Export Current View</button>
-        <ul class="oe_dropdown_menu">
-            <li class="oe_sidebar_export_view_xls"><span>Excel</span></li>
-        </ul>
+    <div class="o_cp_sidebar btn-group btn-group-sm">
+        <button class="btn btn-default export_treeview_xls" type="button">
```
yajo: Add a `title` to this to give a hint on what it does, or add some text next to the icon please.

yajo (changes requested):

web_export_view/static/src/xml/web_export_view_template.xml
jlzhou: OK, that will be done.

jlzhou:
@lasley @Yajo Please review, thanks.

pedrobaeza:
@jlzhou, thanks for the migration you have done and the improvements. I want to ask you for another possible improvements:
- When you have groups, they are not exported to Excel. It would be desirable to have this option.
- One of the problems with this module is that you can't export data from a view with mode="tree". Changing the approach to have the button always visible (we should relocate it also to another place, as the current location is not visible for these views), and digging correctly in the DOM elements for this view (very similar to the normal tree view one) will do the trick. This will also help users to locate the feature, as it's hidden now by default and users don't think about selecting records. The behavior will be: nothing selected > you export all (including groups). Something or all selected: export the selection.

What do you think about this?

yajo (approved):

jlzhou:
@pedrobaeza Many thanks for your suggestions, plan to implement it in the next PR.

lasley (approved):

pedrobaeza:
@jlzhou are you going to work in the suggestions?

lasley:
At the minimum we should put them in the roadmap. Preserve good ideas IMO

augustosamame:
Hi @jlzhou . I pulled this branch from your repo and tried to install it in my v10 community test instance. The module installs just fine. However, when displaying a tree view I get the following error:

```
Uncaught Error: QWeb2: Template 'AddExportViewMain' not found
http://odoo_dev.devtechperu.com:8069/web/content/251-65914b7/web.assets_common.js:2908
Traceback:
Error: QWeb2: Template 'AddExportViewMain' not found
    at Object.exception (http://odoo_dev.devtechperu.com:8069/web/content/251-65914b7/web.assets_common.js:2908:7)
    at Engine._render (http://odoo_dev.devtechperu.com:8069/web/content/251-65914b7/web.assets_common.js:2952:103)
    at Engine.render (http://odoo_dev.devtechperu.com:8069/web/content/251-65914b7/web.assets_common.js:2948:151)
    at Class.redraw (http://odoo_dev.devtechperu.com:8069/web/content/435-7733302/web.assets_backend.js:4115:349)
    at Class.redraw (http://odoo_dev.devtechperu.com:8069/web/content/251-65914b7/web.assets_common.js:3012:371)
    at Class.add_items (http://odoo_dev.devtechperu.com:8069/web/content/435-7733302/web.assets_backend.js:1928:530)
    at http://odoo_dev.devtechperu.com:8069/web/content/435-7733302/web.assets_backend.js:1928:754
    at Function._.each._.forEach (http://odoo_dev.devtechperu.com:8069/web/content/251-65914b7/web.assets_common.js:12:558)
    at Class.add_toolbar (http://odoo_dev.devtechperu.com:8069/web/content/435-7733302/web.assets_backend.js:1928:588)
    at Class.render_sidebar (http://odoo_dev.devtechperu.com:8069/web/content/435-7733302/web.assets_backend.js:2021:480)
```

It doesn't make much sense to me since I see you are defining the AddExportViewMain template here:

```
static/src/xml/web_export_view_template.xml

<templates id="template" xml:space="preserve">
<t t-name="AddExportViewMain">
    <div class="o_cp_sidebar btn-group btn-group-sm">
        <button class="btn btn-default export_treeview_xls" type="button" title="Exporta datos seleccionados">
            <i t-translation="off" class="fa fa-file-excel-o"></i>
        </button>
    </div>
</t>
</templates>
```

But I'm an Odoo newbie so maybe I'm missing something obvious. I did restart the server after installing the module just in case.

jlzhou:
You should restart your odoo server after adding new addon module into addons loading path.

augustosamame:
Thank you Henry, but as indicated at the end of my question, I did restart
the server after module install.

jlzhou:
suggest you to restart odoo first after placing new addon, then install it.

web_export_view/__manifest__.py
```diff
@@ -1,40 +1,27 @@
...
+        'views/web_export_view_view.xml',
     ],
     'qweb': [
-        'static/src/xml/web_export_view_template.xml',
+        "static/src/xml/web_export_view_template.xml",
     ],
-    'installable': False,
-    'auto_install': False,
+    'installable': True,
+    "auto_install": False,
```
yajo: Why all these quote changes? Can't you leave them as they were and reduce the diff?

yajo (changes requested):
AFAICS there's only this little comment I add below (I think this should be squashed by merger, BTW) and the point of adding @pedrobaeza's suggestion to Known Issues section in README. Once done those little 2 things, ready to merge :blush:

yajo (approved):

alexis-via (approved):
Tests only (no code review). Works well... but there is a bug on float fields when the language of the user is French, see my comment below.

pedrobaeza:
Why changing the icon to a generic one? I see correct to let the Excel icon, as you are exporting to Excel.

alexis-via:
I just found a bug after my "approval": when I switch my user from English langugage to French language and I export a list view that has float fields (sale orders for example), when I open the generated XLS file with libreoffice, I don't see the decimal part of the numbers. For example, if I have a sale order with total amount 57.60, then in my spreadsheet the value will be 57. It's an important problem.

jlzhou:
To be consistent with exporting excel icon and hint in the pivot view.

yajo:
About floating point... maybe you can use the current context's language settings and ensure they are exported as such.

pedrobaeza:
I don't see such consistency with the current download icon. Please switch again to the Excel one.

jlzhou:
@pedrobaeza
Please see my attached screenshot of pivot view, it has a export button too.
pivot

pedrobaeza:
Well, I think Odoo is not correct putting that icon, although it makes more sense in that context where you have less options. I still think we have to put the Excel icon. What others think?

pedrobaeza (approved):

bistaray:
When exporting Customer Invoices, the "Total" and "Amount Due" fields are BLANK.  I am using the with Enterprise, installed from https://www.odoo.com/apps/modules/10.0/web_export_view/

Also, some fields show up in Excel with HTML formatting inside the field.

I don't know if this is related to Enterprise, but I am not getting good results from the module.

PR #445, opened 2016-10-15
[MIG][8.0] web_responsive

lasley:
[MIG] web_responsive: Migrate to v8  …
- Drop version
- Downgrade JS api
- Switch to openerp tags
- Switch selectors
  - `oe_leftbar` from `div` to `td`
  - Remove `id` selector from `nav` xpath

cc @hbrunn


lasley:
This module is not compatible with [`support_branding`](https://github.com/OCA/web/blob/8.0/support_branding/views/qweb.xml#L11) due to removal of [`.oe_footer`](https://github.com/OCA/web/pull/445/files#diff-b2e18ea72a902d3738259f602168f4d5R106). Not really sure of a good way to fix that either - I guess we could add branding into the App Drawer. Seems meh though IMO.

```
Error details:
Element '<xpath expr="//div[@class='oe_footer']">' cannot be located in parent view

Error context:
View `menu_secondary`
[view_id: 198, xml_id: support_branding.menu_secondary, model: n/a, parent_id: 167]" while parsing None:102, near
<data inherit_id="web.menu_secondary" name="App Drawer - Secondary Menu">

        <xpath expr="//a[@class='oe_logo']" position="replace"/>

        <xpath expr="//div[@class='oe_footer']" position="replace"/>

        <xpath expr="//div[@class='oe_secondary_menus_container']/t" position="replace">

            <t t-foreach="menu_data['children']" t-as="menu">
                <ul style="display: none" class="oe_secondary_menu nav navbar-nav" t-att-data-menu-parent="menu['id']">
                    <t t-call="web.menu_secondary_submenu"/>
                </ul>
            </t>

        </xpath>

    </data>
```

pedrobaeza:
Can we add a fake footer? Or maybe put it on the top part in the middle, only visible for largest displays.

We will have the same problem in v9, now that the module is being ported.

lasley:
I can hide it, but otherwise we have branding somewhere in the top bar. I specifically had to remove the logo in the top in order to save space & this is significantly more space than a logo.

I hid it & added a note about the override. We should have 🍏  tests now hopefully. I'll submit forward ports once it passes.

lasley:
I ended up having to add a few fixes for support of the `share` module, which was killed in v9.

There's a Lint fail that I'm not so sure about though - https://travis-ci.org/laslabs/web/jobs/167878323#L301

```
help_online/tests/test_export_help_wizard.py:25: [W7936(missing-manifest-dependency), ] Missing external dependency "anybox.testing.openerp" from manifest. More info: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#external-dependencies
```

web_responsive/README.rst
```diff
@@ -0,0 +1,91 @@
...
+The following keyboard shortcuts are implemented:
+
+* Toggle App Drawer - `ActionKey <https://en.wikipedia.org/wiki/Access_key#Access_in_different_browsers`_+``A``
+* Navigate Apps Drawer - Arrow Keys
+* Type to select App Links
+* ``esc`` to close App Drawer
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/154/9.0
```
yajo, line 38: 8.0

web_responsive/views/web.xml
```diff
@@ -0,0 +1,242 @@
...
+<openerp>
+<data>
+
+    <template id="webclient_bootstrap"
+              inherit_id="web.webclient_bootstrap"
+              name="App Drawer - Web Client"
+              >
+
+        <xpath expr="//td[@class='oe_leftbar']" position="attributes">
+            <field name="style">display: none;</field>
```
yajo: I guess I'd do this in Less/CSS instead.

web_responsive/views/web.xml
```diff
@@ -0,0 +1,242 @@
...
+                                 groups="base.group_user,base.group_portal"
+                                 >
+                                <t t-call="web.menu_secondary" />
+                            </div>
+                            <div class="nav navbar-nav navbar-right">
+                                <ul class="nav navbar-nav navbar-right oe_user_menu_placeholder"
+                                    style="display: none;"
+                                    />
+                                <ul class="nav navbar-nav navbar-right oe_systray"
+                                    style="display: none;"
```
yajo, line 92: You know, these inline styles always itch me. :laughing:

web_responsive/views/web.xml
```diff
@@ -0,0 +1,242 @@
...
+        <xpath expr="//a[@class='oe_logo']" position="replace" />
+
+        <xpath expr="//div[@class='oe_footer']" position="attributes">
+            <attribute name="style">display: none;</attribute>
+        </xpath>
+
+        <xpath expr="//div[@class='oe_secondary_menus_container']/t" position="replace">
+
+            <t t-foreach="menu_data['children']" t-as="menu">
+                <ul style="display: none" class="oe_secondary_menu nav navbar-nav" t-att-data-menu-parent="menu['id']">
```
yajo: Same as above.

yajo (changes requested):
Code is mostly fine, but if you just enter runbot as demo user, you get:

captura de pantalla de 2016-10-19 09-31-49

yajo:
BTW, I can't see the tests being run in travis. Is travis problem or this addon's?

web_responsive/README.rst
```diff
@@ -0,0 +1,91 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
```
pedrobaeza, line 3: For v8 it should be AGPL-3.

web_responsive/views/web.xml
lasley: This is copied from Odoo core, and is required for their Javascript unfortunately

web_responsive/views/web.xml
lasley: This is copied from Odoo core, and is required for their Javascript unfortunately

web_responsive/views/web.xml
lasley: I'll be pulling in changes we came up with on #446 - so we'll be good here when I do that 😉

lasley:
Seems it's a Travis issue. They should be failing at the moment - http://3269812-445-536557.runbot1.odoo-community.org/web/tests?module=web_responsive?db=3269812-445-536557-all

web_responsive/views/web.xml
yajo: Man, Odoo really needs some picky geek :astonished:

yajo:
Who could we ping about it?

lasley:
Hmmm honestly I'm not too sure, v8 isn't really my thing - I started Odoo dev in 9 for the most part. @hbrunn maybe?

I just took a look at the 8.0 branch on Travis & none of those are showing Phantom tests either. Deeper dive shows [they're not running on OCA Runbot either](http://pastebin.com/YYUpmpCY). 

I would say this likely means we're doing something wrong on the autoload of the tests. [Odoo v8 web test_js.py](https://github.com/odoo/odoo/blob/8.0/addons/web/tests/test_js.py) seems to be doing the same thing we are for the Python files though. [Their Runbot is running the tests](http://runbot9.odoo.com/runbot/static/build/176140-8-0-85b34f/logs/job_20_test_all.txt), so I'm utterly confused.

Edit: apology reviewers - I stupidly integrated some of #446 here before it was ready, which was creating major conflicts on the cherry pick. I ended up having to rebase and destroy the diff timeline 😦

yajo:
In OCA I think tests are run in Travis and runbot is kept just for functional tests, but in any case none of them runs any JS test.  :confused: 

BTW, bots are :x: still.

lasley:
Sorry v8 users, I'm abandoning this one officially. The current version works across all Odoo versions without many changes. This one requires a lot of changes and will be a long term maintenance burden IMO

pedrobaeza:
Yeah, it's better to concentrate on migrating to new versions instead of backporting infinitely to old versions.

PR #446, opened 2016-10-15
[FIX] web_responsive: Hide footer instead of remove

lasley:
- Add style to hide oe_footer instead of remove to not break `support_branding`
- Add note in ReadMe explaining override of `support_branding`
- Add `oe_main_menu_navbar` id to navbar

This is the v9 fix for `support_branding` in #445

### DIFF
--- a/web_responsive/views/web.xml
+++ b/web_responsive/views/web.xml
@@ -45,7 +45,10 @@
                                accesskey="A"
                                >
                                 <span class="sr-only">Toggle App Drawer</span>
-                                <i class="fa fa-th fa-lg app-drawer-icon-open" />
+                                <i class="fa fa-th fa-lg app-drawer-icon-open"
+                                   t-translation="off"
+                                   aria-hidden="true"
+                                   />
                             </a>
 
                             <button type="button"
@@ -62,7 +65,10 @@
                                     data-target="#odooMenuBarNav"
                                     >
                                 <span class="sr-only">Toggle Navigation</span>
-                                <i class="fa fa-bars fa-lg" />
+                                <i class="fa fa-bars fa-lg"
+                                   t-translation="off"
+                                   aria-hidden="true"
+                                   />
                             </button>
 
                         </div>
@@ -102,12 +108,13 @@
 
         <xpath expr="//a[@class='oe_logo']" position="replace" />
 
-        <xpath expr="//div[@class='oe_footer']" position="replace" />
-
         <xpath expr="//div[@class='oe_secondary_menus_container']/t" position="replace">
 
             <t t-foreach="menu_data['children']" t-as="menu">
-                <ul style="display: none" class="oe_secondary_menu nav navbar-nav" t-att-data-menu-parent="menu['id']">
+                <ul class="oe_secondary_menu nav navbar-nav"
+                    t-att-data-menu-parent="menu['id']"
+                    style="display: none;"
+                    >
                     <t t-call="web.menu_secondary_submenu" />
                 </ul>
             </t>
@@ -206,12 +213,26 @@
 
             <div class="panel-default app-drawer-app-panel" id="appDrawerAppMenu">
                 <div class="panel-heading" id="appDrawerAppPanelHead">
-                    <h4 class="app-drawer-panel-title">
-                        <a href="#" class="app-drawer-icon-close drawer-toggle">
-                            <i class="fa fa-lg fa-chevron-left" />
-                            Apps
+                    <div class="col-xs-6">
+                        <h4 class="app-drawer-panel-title pull-left">
+                            <a href="#" class="app-drawer-icon-close drawer-toggle">
+                                <i class="fa fa-lg fa-chevron-left"
+                                   t-translation="off"
+                                   aria-hidden="true"
+                                   />
+                                Apps
+                            </a>
+                        </h4>
+                    </div>
+                    <div class="col-xs-6">
+                        <a class="oe_logo pull-right" t-attf-href="/web?{{ keep_query() }}">
+                            <i class="fa fa-pencil-square-o oe_logo_edit"
+                               aria-hidden="true"
+                               t-translation="off"
+                               />
+                            <img src='/web/binary/company_logo'/>
                         </a>
-                    </h4>
+                    </div>
                 </div>
                 <div class="panel-body" id="appDrawerAppPanelBody">
                     <ul class="row list-unstyled oe_application_menu_placeholder"
--- a/web_responsive/README.rst
+++ b/web_responsive/README.rst
@@ -40,12 +40,18 @@ The following keyboard shortcuts are implemented:
 Known issues / Roadmap
 ======================
 
+Note: Data added to the footer ``support_branding`` is not shown while using
+this module.
+
 * Provide full menu search feature instead of just App search
 * Drag drawer from left to open in mobile
 * Figure out how to test focus on hidden elements for keyboard nav tests
 * If you resize the window, body gets a wrong ``overflow: auto`` css property
   and you need to refresh your view or open/close the app drawer to fix that.
 * Override LESS styling to allow for responsive widget layouts
+* Adding ``oe_main_menu_navbar`` ID to the top navigation bar triggers some
+  great styles, but also `JavaScript that causes issues on mobile
+  <https://github.com/OCA/web/pull/446#issuecomment-254827880>`_
 
 Bug Tracker
 ===========
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -85,7 +85,9 @@ odoo.define('web_responsive', function(require) {
             this.initDrawer();
             var $clickZones = $('.openerp_webclient_container, ' +
                                 'a.oe_menu_leaf, ' +
-                                'a.oe_menu_toggler'
+                                'a.oe_menu_toggler, ' +
+                                'a.oe_logo, ' +
+                                'i.oe_logo_edit'
                                 );
             $clickZones.click($.proxy(this.handleClickZones, this));
             core.bus.on('resize', this, this.handleWindowResize);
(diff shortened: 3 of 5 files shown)


pedrobaeza:
We should take the occasion to hide all replaced elements instead of just removing them, so we maximize the compatibility with other possible modules.

One thing: what about putting the company logo a bit more smaller in the upper right corner of the drawer?

lasley:
I originally hid everything, but was told by @Yajo to remove instead in the following comments:
- https://github.com/OCA/web/pull/396#r76193405
- https://github.com/OCA/web/pull/396#r76193550
- https://github.com/OCA/web/pull/396#r76193557
- https://github.com/OCA/web/pull/396#r76193727
- https://github.com/OCA/web/pull/396#r76193743

So I should revert all that? I agree with his logic TBH - there's a lot of JS happening & menus getting filled in that adds a lot of load to the page without replacing.

pedrobaeza:
When there's JS logic involved, I agree on the optimization principle, but we don't know if other modules will mess up with these removals. Is it so heavy to let them?

lasley:
Regarding the company logo - I can add in the title pane of the app drawer in v9, but it shrunk considerably in v10 and I don't think would be viable:

image

vs

image

I admittedly like the smaller UI of 10, so I hesitate to change. Is the missing logo something that will bug people you think?

lasley:
It's pretty heavy. If I leave them, the entire menus are duplicated. The interesting thing about the way the menus are generated too is that every menu is added, but hidden. This entire section would be duplicated - 4.5k lines of code on my local:

image

pedrobaeza:
There are 2 cons for not having logo:
- The company branding, that all company managers wants to have in their systems (well, everywhere)
- In multi-company environments, users look at the logo to see in which company they are.

pedrobaeza:
OK, let's see what we get with the replacings. Don't touch that part for now.

pedrobaeza:
Even smaller, in v10 the logo can be there as something testimonial (or include a system parameter to configure the height at request)

lasley:
It shouldn't be too difficult to add the logo in the bottom center of the App Drawer. Will that work?

pedrobaeza:
Well, if we put it with absolute positioning plus float feature, it would be enough, but it's useless if the logo is not shown when opening because you have to scroll down.

lasley:
Agreed. I just looked at Enterprise for some possible inspiration, but it seems they put it at the bottom and required scroll.

I can increase the height of the App Drawer title bar in 10 to support it at the top then. I can't really think of a better spot for it, and there is space there. I am planning on adding a Search Bar there too, but that can go top middle. Or maybe the logo top middle and search top right.

I don't think a param will work because the height is controlled in CSS, so I wouldn't be able to grab the param (at least I don't think so, let me know if I'm missing some Odoo magic).

pedrobaeza:
Isn't better to put the logo inside the top bar instead of overlapping the apps icons?

lasley:
Yeah that was the intent, and what it looked like on my local - This definitely wasn't what I was going for 😆

pedrobaeza:
hehe, we definitively need that docker that mimicks Runbot behavior on local :wink:

lasley:
Seriously - my dev environment situation sucks at the moment. This is the first I've bothered with the other Odoo versions, which turns out to be a real pain. 

Something will be figured out, sooner rather than later - that's for sure. My poor juniors will be so 😕  with this workflow.

lasley:
Rebased for Travis changes so my Runbot doesn't skip the builds anymore. The last commit should hopefully fix the logo issue too

lasley:
Note to self, looks like junk on mobile 

image

lasley:
Annnnnd now it doesn't! This is ready for review again I believe.

Note that this is still incompatible with the `support_branding` module as detailed in the ReadMe (build isn't broken, but `support_branding` assets are simply hidden). A glue module will need to be created in order for the `support_branding` stuff to be injected somewhere other than `web.secondary_menu`, which is the top bar in this interface.

pedrobaeza:
Thanks for the efforts, Dave. I have tried and I feel that the logo is a bit small, leaving a big margin on top and bottom. Or is this to be the same for v10?

About support_branding, we can remove the word incompatible, because as said it doesn't fail, but one of the features isn't simply shown. But the other is still there: the error dialog includes the possibility of sending the traceback to a support mail.

lasley:
Alright updated the ReadMe. Logo sizing is better now too

image

lasley:
Dang now that I see the ruler guides against the "Apps" link, I realize that the Apps link is not center. That's going to bug me now that I know about it, so will fix

pedrobaeza:
This looks much better, but I'm afraid there's still one thing to do: restore the "Edit company data" link. In v9, without it, you can't edit company data, unless you activate "Multi companies".

web_responsive/views/web.xml
```diff
@@ -102,7 +103,9 @@
 
         <xpath expr="//a[@class='oe_logo']" position="replace" />
 
-        <xpath expr="//div[@class='oe_footer']" position="replace" />
+        <xpath expr="//div[@class='oe_footer']" position="attributes">
+            <attribute name="style">display: none;</attribute>
```
yajo: Better add class `hidden` if possible.

web_responsive/views/web.xml
```diff
@@ -206,12 +209,19 @@
...
+                    <div class="col-xs-6">
+                        <h4 class="app-drawer-panel-title pull-left">
+                            <a href="#" class="app-drawer-icon-close drawer-toggle">
+                                <i class="fa fa-lg fa-chevron-left" />
+                                Apps
+                            </a>
+                        </h4>
+                    </div>
+                    <div class="col-xs-6">
+                        <a class="oe_logo pull-right" t-att-href="'/web/?debug' if debug else '/web'">
```
yajo: Can't you use `keep_query`?

web_responsive/views/web.xml
```diff
@@ -206,12 +209,19 @@
...
             <div class="panel-default app-drawer-app-panel" id="appDrawerAppMenu">
                 <div class="panel-heading" id="appDrawerAppPanelHead">
-                    <h4 class="app-drawer-panel-title">
-                        <a href="#" class="app-drawer-icon-close drawer-toggle">
-                            <i class="fa fa-lg fa-chevron-left" />
-                            Apps
+                    <div class="col-xs-6">
+                        <h4 class="app-drawer-panel-title pull-left">
+                            <a href="#" class="app-drawer-icon-close drawer-toggle">
+                                <i class="fa fa-lg fa-chevron-left" />
```
yajo: You'll probably get dirty translations with this. Better use a div for fontawesome icons or add `t-translation="off"` please.

yajo (changes requested):

web_responsive/views/web.xml
lasley: Is there a way to append into the existing class? The goal here is to allow compatibility with addons that may target the div. At the moment it's `oe_footer`, but I don't know if other modules inject classes into that or other things.

yajo: Then why not just add `.oe_footer {display:none}` to the Less file?

lasley: Because I didn't think of it? 😆  👍

lasley:
Alright we should be good. I added a bonus fix on the toggle button padding/border too 😄 

Note the padding had to be made a bit smaller due to [adding of `oe_main_menu_navbar` id](https://github.com/OCA/web/pull/446/commits/0ae026047a364f56c2f2326db9c82980ac3848e3#diff-b2e18ea72a902d3738259f602168f4d5R38) which triggered some new CSS. Looks more like the v10 branch now, which makes me happy.

pedrobaeza:
Very good! Just one minor thing: edit icon shouldn't be visible when you're not admin, as you don't have rights to edit company data.

lasley:
Fixed - brought in the [v10 style](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/web/static/src/less/webclient_layout.less#L54), just nested differently

pedrobaeza:
Uhm, entering as demo user, I don't see any app nor menu...

lasley:
Should demo user have access to `Settings` or `Apps` menus? I installed a few apps and it seems to work fine.

image

lasley:
Taking it a step further, I added `Settings` to `Demo User` Admin access rights & answered my own question. Demo user does not have access to the initial 2 app icons - but can see them once the change is made:

image

pedrobaeza:
But at least "Discuss" app should be available I think...

lasley:
Nope - discuss is an additional install it seems. This is a screencap from a fresh installed DB:

image

pedrobaeza (approved):
Thanks for checking.

oscarolar (approved):
Thanks!

pedrobaeza:
@Yajo, please check

yajo (changes requested):
Enter mobile mode as demo user and choose "Sales". When the submenu is hidden, you see this:

captura de pantalla de 2016-10-19 09-18-47

using firefox

lasley:
Well that's sad - I have to fully remove the oe_main_menu_navbar id in order to stop a bunch of the core JS from firing. This resolved the [ @Yajo comment](https://github.com/OCA/web/pull/446#pullrequestreview-4814058)

I don't see an easy way around this & there wasn't a particular reason for me doing it outside of simply wanting the previous ID to exist for other modules that may target it, so I removed & reverted the styling fixes associated with it.

pedrobaeza:
Waiting for @Yajo's confirmation to merge.

lasley:
Once he confirms, I would like to squash the commits logically as well in order to allow for easier cherry-picking into the forward/backports & to not conflict with our OCD for a clean diff 😄

yajo (approved):
It's sad that you had to remove that, but this now works, so it's OK.

Maybe you could add that to known issues if you want, but it's OK anyway.

Waiting until you rebase to merge.

lasley:
I agree that it's sad I had to remove it - I was pretty excited when I added it, particularly when I noticed the styling differences.

Adding to known issues is a good call, I can see that being trying again later. Done.

Rebase is done as well, turns out the logical commit was simply the one. Was worth the check though IMO - I believe we're good for merge once we get CI 🍏

PR #448, opened 2016-10-15
[IMP][10.0] travis: Upgrade travis config

lasley:
- Switch to new matrix
- Add Transifex
- Add travis_after_tests_success


pedrobaeza (approved):
Thanks, I was wondering why we don't have Codecov... I'll fast-track when CIs finish

PR #449, opened 2016-10-16
10.0 port web sheet full width

legalsylvain:
PORT web_sheet_full_width module. 
I made more simple the css file. Review welcome if it's enough. 

thanks for your review.

CC @njeudy ; @hbrunn.


pedrobaeza:
It would be great if you make a less file instead.

legalsylvain:
Hi @pedrobaeza, thanks for your review.

I'm not comfortable with less concept. Could you point OCA sample ? Or alternatively make a PR on mine ?

Is less an OCA norm ?

pedrobaeza:
This is something new from 9.0. Now less is automatically compiled and converted to css. Although not required, less is simpler to be read than traditional css, and can programatically add logic on css: https://en.wikipedia.org/wiki/Less_(stylesheet_language)

legalsylvain:
Hi, thanks ! 
Well on this file, there are only 5 lines.
https://github.com/legalsylvain/web/blob/27c226d5c508c7f467d1d686340743a9b22efaed/web_sheet_full_width/static/src/css/web_sheet_full_width.css
I don't know how to make it simplier. 

Maybe : 

```
@web-sheet-margin: 15px;
```

and then : 

```
.o_form_view .o_form_sheet_bg .o_form_sheet{
    margin: @web-sheet-margin;
}
```

What do you think ? 

kind regards.

lasley:
You could also nest the declarations:

```
@web-sheet-margin: 15px;

.o_form_view {
    .o_form_sheet_bg .o_form_sheet {
        margin: @web-sheet-margin;
        max-width: none;
    }
    .oe_chatter{
        max-width: none;
    }
}

```

pedrobaeza:
I think the question about less is the inheritability. @Yajo can inform us about that, as he is the expert in that.

yajo:
I'd do as @lasley said. It's simpler than plain CSS and makes the same. However, instead of this custom `@web-sheet-margin`, [I'd use `@padding-base-horizontal`](https://github.com/twbs/bootstrap/blob/v3.3.6/less/variables.less#L91). And you don't need to declare the variable, just use it; it gets declared later in the stack.

legalsylvain:
Thanks all for all this explanation !
I made the according changes.

Note : to apply less changes, I had to restart my odoo server, instead of tiping F5. Is it normal ? Is there an other way to refresh css ? Or did I miss something ?

regards.

yajo:
Maybe Ctrl+F5 would have done the trick.

yajo (approved):

pedrobaeza (approved):

Garamotte (approved):

legalsylvain:
@yajo no Ctrl F5 does not refresh. stopping server is necessary. I guess that compilating less is done once at startup.
(Debug mode does not change this behaviour).
@all thanks for your reviews !

yajo:
Then maybe it's something related to the precompilation of qweb templates that they added in v10, because in v9 a f5 usually works. Try running `odoo.py --dev` to have instant view updates.

BT-fgarbely:
Hello @legalsylvain 

I tried this module (v10) in my enterprise odoo v10 from Dec 2016.
I can see the sheets aligned on the left side, but not in full width.

web_sheet_full_width


If I would re-add "width: auto !important;", the sheets will be displayed in full width.
Am I doing something wrong?

Regards,

pedrobaeza:
In enterprise you don't need this module AFAIK. Everything is in full screen by default.

pedrobaeza:
Well, you need to install the corresponding responsive module, of course.

BT-fgarbely:
Which "corresponding responsive module" do you mean?

lukebranch:
@BT-fgarbely ,

I believe @pedrobaeza is referring to the web_enterprise module.

BT-fgarbely:
in runbot (v10, enterprise), the module web_enterprise is installed, but the sheets are not full screen.
Maybe there is another module which handles this functionality.

pedrobaeza:
OK, I see that forms still have fixed width. Well, the module has been tried over community edition, as there's no possibility to handle runbot for enterprise modules, so these things can happen. You can provide a patch, but you must avoid `!important` as this messes up CSS inheritance. Locate a more specific selector that modifies what you want, and you should handle also that this works in both versions.

lukebranch:
@BT-fgarbely ,

I see what you mean in regards to the sheets not being completely full width on all screens, perhaps there is room for adding enterprise compatibility with this module in the future.

PR #454, opened 2016-10-22
[ADD] web_widget_x2many_2d_matrix_demo

JordiBForgeFlow:
# 
# 2D matrix for x2many fields demo

This module prepares an example for the module 'web_widget_x2many_2d_matrix'.

It implements a simple wizards that allows users to enter the planned hours
for all the projects.
# Usage

Go to 'Project / Enter Planned hours'

### DIFF
--- a/web_widget_x2many_2d_matrix_demo/README.rst
+++ b/web_widget_x2many_2d_matrix_demo/README.rst
@@ -1,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+
+================================
+2D matrix for x2many fields demo
+================================
+
+This module prepares an example for the module 'web_widget_x2many_2d_matrix'.
+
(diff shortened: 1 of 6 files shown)


JordiBForgeFlow:
I cannot make the module 'web_widget_x2many_2d_matrix' work. I have created a demo of this widget in order to validate the behavior, establish a productive conversation on how to use it, and debug possible errors, 

At this moment, when I open the matrix nothing is shown:
image

As you can see, at least a task should appear.

Tested also in 8.0 with same results.

cc @pedrobaeza @hbrunn

pedrobaeza:
You don't need this module for a demo. See https://github.com/OCA/sale-workflow/tree/9.0/sale_order_variant_mgmt

JordiBForgeFlow:
Ok, thanks Pedro! I will follow up on the example that you provide. But I will keep this open for a while. I created it as advertised by @hbrunn in the module readme, and it will not work. I'd like to understand better why.

JordiBForgeFlow:
I could make it work, but forcing the x and y values to be chars. If I put them as non-chars, it won't work.

image

pedrobaeza:
In my module axis are many2one. Check how is done.

PR #456, opened 2016-10-28
[MIG] web_tree_many2one_clickable

antespi:
Addon migrated to v9

Changes:
- Migrated to new JS API
- Now all many2one are clickable by default when installing this addon.
- Read system parameter `web_tree_many2one_clickable.default` only once per view.
- Allow to disable just one field with `widget="many2one_unclickable"`

### DIFF
--- a/web_tree_many2one_clickable/__openerp__.py
+++ b/web_tree_many2one_clickable/__openerp__.py
@@ -1,38 +1,26 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright
-#      (C) 2013 Therp BV (<http://therp.nl>).
-#      (c) 2015 Serv. Tecnol. Avanzados (http://www.serviciosbaeza.com)
-#               Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2013 Therp BV (<http://therp.nl>).
+# Copyright 2015 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+# Copyright 2015 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
 {
     "name": "Clickable many2one fields for tree views",
-    "version": "8.0.1.0.0",
+    "summary": "Open the linked resource when clicking on their name",
+    "version": "9.0.1.0.0",
+    "category": "Hidden",
+    "website": "https://odoo-community.org/",
     "author": "Therp BV, "
-              "Serv. Tecnol. Avanzados - Pedro M. Baeza, "
+              "Tecnativa, "
               "Odoo Community Association (OCA)",
-    "category": "Dependency",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
     "depends": [
         'web',
     ],
     "data": [
+        'data/ir_config_parameter.xml',
         'views/asset.xml',
     ],
-    'installable': False,
 }
--- a/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
+++ b/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
@@ -1,115 +1,101 @@
-//-*- coding: utf-8 -*-
-//############################################################################
-//
-//   OpenERP, Open Source Management Solution
-//   This module copyright
-//     (C) 2013 Therp BV (<http://therp.nl>).
-//     (c) 2015 Serv. Tecnol. Avanzados (http://www.serviciosbaeza.com)
-//              Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
-//
-//   This program is free software: you can redistribute it and/or modify
-//   it under the terms of the GNU Affero General Public License as
-//   published by the Free Software Foundation, either version 3 of the
-//   License, or (at your option) any later version.
-//
-//   This program is distributed in the hope that it will be useful,
-//   but WITHOUT ANY WARRANTY; without even the implied warranty of
-//   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-//   GNU Affero General Public License for more details.
-//
-//   You should have received a copy of the GNU Affero General Public License
-//   along with this program.  If not, see <http://www.gnu.org/licenses/>.
-//
-//############################################################################
+/* Copyright 2013 Therp BV (<http://therp.nl>).
+ * Copyright 2015 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+ * Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
 
-openerp.web_tree_many2one_clickable = function(instance, local)
-{
-    instance.web.list.Column.include({
-        /*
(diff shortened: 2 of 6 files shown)


pedrobaeza (changes requested):
web_tree_many2one_clickable.default system parameter is not put by default.

The rest LGTM

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -1,104 +1,112 @@
...
+    return $.when();
+}
+
+ListView.Column.include({
+    init: function(id, tag, attrs) {
+        this._super(id, tag, attrs);
+        if (this.widget == 'many2one_clickable') {
+            this.use_many2one_clickable = true;
+        // } else if (this.widget == 'many2one_unclickable') {
+        //     this.use_many2one_clickable = false;
```
yajo: :hankey:

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -1,104 +1,112 @@
...
+                if (param) {
+                    use_many2one_clickable_default = (
+                        param.value.toLowerCase() == 'true');
+                }
+                ir_option_clickable_loaded.resolve();
+            });
+        return ir_option_clickable_loaded;
+    }
+    return $.when();
+}
```
yajo, line 34: This whole method is... weird.

I'd just do:

```
var clickable_default = $.Deferred();
new Model("ir.config_parameter")
.call("get_param", ["web_tree_many2one_clickable.default"])
.done(function(value){clickable_default.resolve(value.toLowerCase() === "true");})
.fail(function(){clickable_default.reject();})
```

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -1,104 +1,112 @@
...
+        this._super(id, tag, attrs);
+        if (this.widget == 'many2one_clickable') {
+            this.use_many2one_clickable = true;
+        // } else if (this.widget == 'many2one_unclickable') {
+        //     this.use_many2one_clickable = false;
+        } else if (this.type == 'many2one') {
+            if (use_many2one_clickable_default) {
+                this.use_many2one_clickable = true;
+            } else {
+                get_options();
```
yajo: Following comment above about this method, change this to:

```
clickable_default.done($.proxy(this, function(value){this.use_many2one_clickable=value}))
```

web_tree_many2one_clickable/__openerp__.py
```diff
@@ -1,38 +1,25 @@
...
+# Copyright 2015 Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+# Copyright 2015 Antonio Espinosa <antonio.espinosa@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
 {
     "name": "Clickable many2one fields for tree views",
-    "version": "8.0.1.0.0",
+    "summary": "Open the linked resource when clicking on their name",
+    "version": "9.0.1.0.0",
+    "category": "Dependency",
```
yajo: I think this should be "Hidden"

yajo (changes requested):

pedrobaeza:
Please rebase to get a green base.

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
antespi: Why is weird?

This method allows:
- Inheritance
- Only loads configuration once per view

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
antespi: Sorry because my ignorance, what does `$.proxy` method here?

web_tree_many2one_clickable/__openerp__.py
antespi: This was the original code.
@hbrunn what category would be the best in this case, `'category': 'Web'`?

hbrunn: I don't really have an opinion about this, but Hidden/Dependency is what I use a lot for this kind of modules

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
hbrunn: it takes care that `this` in the function points to the object from the outer `this`

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
yajo: If I'm not missing something, this is quite overengineered. Let me explain. You currently do this:
1. [Create a `false` default value](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R20). This is weird, because instead of having defaults, you should subscribe to the promise.
2. [Create an empty variable](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R21) that is expected to [become a promise later](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R25). This is weird because you should define it in the 1st step and simply reject or resolve it later.
3. [Query `ir.config_parameter` manually](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R27) and check for results, weird because you could simply call its `get_param` method, that already includes the possibility to have a default value.
4. [The returned promise](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R39) only tells you the parameter got loaded, which is weird because normally you are expected to get a result from the promise.
5. [There is a check to know what promise to return](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R24), but why do you return different promises instead of just resolving or rejecting the one you should already have declared?
6. Later in the stack, [you check if there was a positive default](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R48) and [then apply it](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R49), and [try to fetch the default if it is negative](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R51). This is wierd too, because either you could reach that line before the promise is resolved and you are not applying the value when it does, or you could get there after resolution, but the returned parameter was `false` and then you'd be calling `get_options` all the time. Yes, [it checks for the existence of the `ir_option_clickable_loaded` variable](https://github.com/OCA/web/pull/456/files#diff-f06f62892caad3befc7bded8af2e3c50R24) to avoid asking the server again, but that is again not required if you do my solution because when a promise is already resolved and you attach a handler later to it, it gets immediatly executed automatically.
7. You are not exposing this method nor the promise, so this is not really inheritable from another addon.

So yes, this is very weird as it is right now; when doing async programming there are these things to keep in mind.

Mi suggestion avoids the `use_many2one_clickable_default` + `ir_option_clickable_loaded` combination; it simply sets a promise called `clickable_default` that gets evaluated only once when the module is loaded and that is correctly resolved with a value. Besides, it is much shorter.

If you still want to make it inheritable, which is good, then simply expose the promise and other addons will be able to subscribe to it.

antespi:
@Yajo, @pedrobaeza all changes done. Now with a right promises implementation, JS do not call `get_param()` until a many2one column is going to be rendered

web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
```diff
@@ -45,11 +39,10 @@ ListView.Column.include({
...
         } else if (this.type == 'many2one') {
-            if (use_many2one_clickable_default) {
-                this.use_many2one_clickable = true;
-            } else {
-                get_options();
-            }
+            this.use_many2one_clickable = false;
+            clickable_default($.proxy(function(value){
+                this.use_many2one_clickable = value;
+            }, this));
```
yajo, line 45: OK, I get your point of just asking once and only when the widget gets initialized, and it's nice, but let me try to do this more straightforward.

This would go above:

```
var asked = false, default = $.Deferred();

function ask_once() {
  if (asked) {return default}
  asked = true;
  new Model("ir.config_parameter")
            .call("get_param", ["web_tree_many2one_clickable.default", false])
            .done(function(value){
                default.resolve(String(value).toLowerCase() === "true");
            })
            .fail(function(){
                default.reject();
             });
}
```

Then, here, inside the `init()`:

```
this.use_many2one_clickable = false;
ask_once();
default.done($.proxy(function(value){
                this.use_many2one_clickable = value;
            }, this));
```

It makes the same, but all is easier to understand.

yajo (changes requested):

yajo (approved):

pedrobaeza (approved):

PR #458, opened 2016-11-01
[ADD] web_widget_pattern

hbrunn:
Input patterns
==============

This module was written to allow you to define a regex on character fields a user's input is tested against.

Usage
=====

- depend on this module
- define a key ``pattern`` in the field's options dictionary::

       <field name="email" options="{'pattern': '[\S]+@[\S]+'}" />
- if you need pattern modifiers, use the key ``pattern_modifiers``


hbrunn:
to test this on runbot, try to fill in a malformed email address on a partner

NL66278:
I logged into runbot instance but looked very strange. So not able to test.

web_widget_pattern/demo/res_partner.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
NL66278, line 1: If this modifies standard partner view, should it not go under views instead of under demo?

hbrunn: it's just a demo, so no ;-)

hbrunn:
@NL66278 for testing: All the web modules are installed there, so the UI is somewhat distorted. Especially by the addon that hides the main menu. If you move to the left, the menu should pop up. There, you can select Organizer/Contacts to get a partner view

NL66278:
@hbrunn I do not know what the intention of the hide main menu module is, but the whole runbot instance has been rendered totally unusable. Anyway I do not see an organizer/contacts menu, so testing is impossible.
schermafdruk - 01-11-16 - 16 05 33

andreschenkels:
:+1: I added the CRM module and tested it on a email adres. Works!

andhit-r:
Tested functionally on runbot :+1:

pedrobaeza:
Please use new approval flow so that the future statistics of reviewers will be correct.

web_widget_pattern/README.rst
```diff
@@ -0,0 +1,62 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+#. depend on this module
+#. define a key ``pattern`` in the field's options dictionary::
+
+       <field name="email" options="{'pattern': '[\S]+@[\S]+'}" />
+#. if you need pattern modifiers, use the key ``pattern_modifiers``
```
pedrobaeza: What is a pattern modifier?

web_widget_pattern/README.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. depend on this module
+#. define a key ``pattern`` in the field's options dictionary::
+
+       <field name="email" options="{'pattern': '[\S]+@[\S]+'}" />
```
pedrobaeza, line 19: Specify that the pattern is expressed as a regular expression and a general link to an explanation to this would be appreciated, like https://en.wikipedia.org/wiki/Regular_expression

pedrobaeza (changes requested):

web_widget_pattern/README.rst
hbrunn: that's written a paragraph above, and a programmer who doesn't know what a regex is should go back to the library before starting to mess with an erp system in my opinion...

web_widget_pattern/README.rst
hbrunn: http://www.w3schools.com/jsref/jsref_obj_regexp.asp, section modifiers

web_widget_pattern/README.rst
pedrobaeza: Well, I know what is a regular expression, but I don't remember always the syntax, so that link is very handy for me.

web_widget_pattern/README.rst
pedrobaeza: This link would be very useful too

hbrunn: I added both, even though I disagree strongly with the mindset that stuff always should be made simple for programmers. Incompetence should fail early in my opinion, not only when it already caused a major security problem or privacy messup. (and to be sure: I don't mean you, but the people who produce all the horrible code out there)

pedrobaeza (approved):
Thank you very much!

PR #464, opened 2016-11-04
[10.0] web_tree_image: Porting to 10.0

danieltorres7:
- Changed model to installable.
- Tested and everything seems ok.

### DIFF
--- a/web_tree_image/__manifest__.py
+++ b/web_tree_image/__manifest__.py
@@ -27,7 +27,7 @@
 ##############################################################################
 {
     'name': 'Show images in tree views',
-    'version': '9.0.1.1.0',
+    'version': '10.0.1.1.0',
     'author': "Therp BV, "
               "MONK Software, "
               "Odoo Community Association (OCA), "
@@ -44,5 +44,5 @@
     'data': [
         'view/assets.xml',
     ],
-    'installable': False,
+    'installable': True,
 }
--- a/web_tree_image/static/src/js/web_tree_image.js
+++ b/web_tree_image/static/src/js/web_tree_image.js
@@ -19,7 +19,7 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-odoo.define('web.tree_image', function (require) {
+odoo.define('web.WebTreeImage', function (require) {
     "use strict";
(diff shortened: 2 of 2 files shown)


danieltorres7:
@leorochael, can you take a look?

pedrobaeza:
You should switch to new JS API

leorochael:
@pedrobaeza Can you give us some pointers? Where can we find docs or examples of the new JS API?

pedrobaeza:
I'm afraid there isn't. You need to see sample in OCA/web repository for 9.0 and odoo itself.

leorochael:
@pedrobaeza, if it's a v9 new JS API, isn't the changes landed in #290, enough?

pedrobaeza:
Yeah, indeed, sorry for the noise. I though it was a direct migration from 8.0. Then it's enough.

web_tree_image/__manifest__.py
```diff
@@ -27,7 +27,7 @@
 ##############################################################################
 {
     'name': 'Show images in tree views',
-    'version': '9.0.1.1.0',
```
leorochael, line 30: Please merge this commit with the previous one

leorochael (changes requested):

web_tree_image/static/src/js/web_tree_image.js
```diff
@@ -19,7 +19,7 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-odoo.define('web.tree_image', function (require) {
+odoo.define('web.WebTreeImage', function (require) {
```
leorochael, line 22: @pedrobaeza @StefanRijnhart, What do you think of this change?

It looks ok to me, though it doesn't seem strictly necessary to port to v10

leorochael (commented):
LGTM :+1:

web_tree_image/static/src/js/web_tree_image.js
pedrobaeza: This only impacts in possible module that inherits this, so no actual problem I think.

StefanRijnhart (approved):
All good, thank you!

pedrobaeza (approved):

PR #469, opened 2016-11-08
9.0 support branding

hbrunn:
fixed version of #430


hbrunn:
@pedrobaeza @moylop260 could you transfer your approval to this PR?

moylop260 (approved):

pedrobaeza (approved):
Is it possible to squash all the "OCA transbot" commits?

legalsylvain:
@pedrobaeza : Are you sure it's a good idea to squash OCA runbot commit ? Correct me if I'm wrong but I guess that all the commits come from V8.0 commits that have been cherry picked ? The commits exist in V8.0 history, so why do you want to squash them in V9.0 history ?
(Just a question, I don't have a fixed idea on that point.)

pedrobaeza:
The goals after squashing are:
- Reduce the number of commits in this branch.
- Reduce the repository overall size, as these commits add and remove the same lines a lot of times.
- Ease the traceability.

As they are not meaningful for the history, and there's no conflicts in doing that, it's the best solution for me.

hbrunn:
@legalsylvain I also don't expect problems from the squashing. @pedrobaeza done, this means we can merge, right?

pedrobaeza:
Yeah, that's it, merging.

lmignon:
@hbrunn @pedrobaeza @legalsylvain I don't share the POV of @pedrobaeza to always request to squash the commits and it would be nice to have an official statement of OCA on this subject....
Personally I sometimes use the the commit history to check what changes are missing between two branches. When you squash the commit you lost this possibility. There are plenty of others cases where the history is useful. I've also experienced a lot of troubles when commits are squashed  before merging a PR and you have others PRs based on the first one ....Once again this kind of requests in the review process don't ease to contribute to OCA and can be seen as a barrier for potential new contributors.

pedrobaeza:
@lmignon, @hbrunn is not a kind of new contributor :wink: Anyway, you have the same problem squashing or not, as this is not a cherry-pick of the exact commits from 8.0, due to the extraction made with the process of https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0 to let only changes of the involved module. You should do other type of comparison, like a git blame or git diff one.

I can ask for that, and as in your case, if you don't want to do it, I don't insist too much, but if the contributor wants to do it, why not?

PR #470, opened 2016-11-08
9.0-mig-web_widget_one2many_tags

susport:
9.0 Mig web_widget_one2many_tags.


jjscarafia:
Thanks! I've tried it on runbot and locally and it breaks interface. Setps to reproduce:
1. install this module
2. reload tab (nothing is displayed)

jjscarafia (changes requested):
Thanks! I've tried it on runbot and locally and it breaks interface. Setps to reproduce:
1. install this module
2. reload tab (nothing is displayed)

PR #473, opened 2016-11-08
9.0 web tree dynamic colored field

hbrunn:
ping @jbeficent


web_tree_dynamic_colored_field/__openerp__.py
```diff
@@ -1,37 +1,22 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# © 2015 Camptocamp SA, Damien Crier
```
JordiBForgeFlow, line 2: Add therp

web_tree_dynamic_colored_field/demo/res_users.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <record id="view_users_tree" model="ir.ui.view">
+            <field name="model">res.users</field>
+            <field name="inherit_id" ref="base.view_users_tree" />
+            <field name="arch" type="xml">
+                <xpath expr="." position="attributes">
+                    <attribute name="color_field">lang</attribute>
```
JordiBForgeFlow, line 9: Is this a valid example? I cannot see how the field 'lang' of the user can change the color of the record. For example, if the language is "EN"(English).

JordiBForgeFlow (commented):
Tested on runbot. Works as expected. Could not evaluate the color_field. Not sure how it works. Should it be connected to a field such as "color" that returns "red / black,..." and would paint the whole line accordingly?

web_tree_dynamic_colored_field/__openerp__.py
hbrunn: copyrights in files apply to the author of the file, and we didn't change anything meaningful there

web_tree_dynamic_colored_field/demo/res_users.xml
hbrunn: no, but you can verify it works as intended by inspecting the dom, it will have written the language in the color property of the style attribute. This also answers your previous question: The content of the `color_field` will stupidly be written into the style attribute. You can even use this to hack the whole thing by returning there `red; whatever_other_css_property: some_value`

JordiBForgeFlow (approved):
Ok, tested all options on a separate module. Works nicely!
image

pedrobaeza:
@jbeficent, that hurts a bit my view. You should reconsider the abuse of colors in tree views...

JordiBForgeFlow:
@pedrobaeza haha! well, I put all lines in red, just to test `color_field` :)

pedrobaeza:
Ah, OK then...

AaronHForgeFlow:
I'm testing on runbot and the color_field attribute works perfectly. However I'm getting errors using fg_color and bg_color attributes. Is this syntax correct?: <field name="display_name" fg_color="yellow"'

AaronHForgeFlow:
The problem was that I was not setting the condition. The correct syntax would be: <field name="display_name"  fg_color="yellow:1==1"/

AaronHForgeFlow (approved):
Tested on runbot :+1:

susport (commented):
👍

PR #474, opened 2016-11-09
[ADD] web_x2m_filter

hbrunn:
Filters for x2many fields
=========================

This module was written to allow developers to add filter buttons to an x2many field.

Usage
=====

#. add a key ``web_x2m_filter`` to the field's ``options`` dictionary
#. this key should contain a list if dictionaries which describe a button by the properties ``name`` and ``domain``::
   
        {'web_x2m_filter': [{'name': 'Only admins', 'domain': [('groups_id', '=', %(base.group_erp_manager)s)]}]}
#. to have one filter being active on form load, add a key named ``default`` with value ``True``
#. the string in ``name`` is subject to the standard translation mechanism
#. the domain is evaluated with the user's context and the current form values


pedrobaeza:
Maybe you can add a screenshot?

hbrunn:
done

pedrobaeza (approved):
I tried on previous runbot. I noticed that the first filter is activated by default. Maybe default should be to show all registers. Another improvement is to put a label on the "All" button, as the arrows are not very descriptive.

hbrunn:
The first filter is activated by default because the button definition says so - if you don't set the `default` key in the dict, there's no filtering by default.
Good point about the 'All' button.

pedrobaeza:
Right, I didn't see it the default one!

NL66278 (approved):
I can not say to much about the code due to my lack of experience with Odoo javascript. But I installed the module, and edited the invoice view to add a filter button. Worked without any problems.

pedrobaeza:
I think we can merge this one, and if there's any other incident like the JSON parsing, make another PR.

PR #475, opened 2016-11-09
[8.0][IMP][web_dashboard_tile] Added groups to control who can see the tile.

ivantodorovich:
This PR adds a new field `group_ids`.

If this field is set, only users of this group can view this tile. 
Please note that it will only work for global tiles (that is, when User field is left empty)

**This allows to share tiles with a specific groups of users.**

ping @legalsylvain

PR #476, opened 2016-11-10
[IMP][8.0][web_dashboard_tile] Refactor (see changes in description)

ivantodorovich:
~~**This PR is based on changes made in https://github.com/OCA/web/pull/475**~~

~~*Although I could just close the other PR and leave only this one with all of the changes. 
It's your call, just tell me.*~~ **Just did that.**

----

This PR refactors `web_dashboard_tile` to support the following features:

- **Posibility to set a format string to alter the value.**
For instance: `{:,} kgs` will display *1,000 kgs* if value is `1000`

- **Posibility to define a function (sum/max/min/..) as a primary value. (instead of just `count`)**
Now, both primary and secondary values can be functions or just count. It's up to the user.
*Note: As a sideeffect of this, I had to remove the image indicators from the tile. But since we can now format the value we can just use a format string like this: `SUM: {:,}` if we want to clarify where the value comes from.*
sin titulo

- **Handles errors due to malformed domains and displays `ERR!` instead of throwing**
Now, the UI won't crash if the user sets a wrong domain declaration.
An error description will show up on the tile form, if any.
  - On the UI:

  - On the form:

- **Simplified templates and stylesheets by relying more on css**

- **Dynamically reduce font-size based on the length of the value**
*It's done in the QWeb view, to make sure large values do not overflow the tile*

---

**Why?**
We were creating tiles just to show the results of a `sum` or `avg` functions, and most of the times the big number (`count`) was irrelevant. With this changes, it's possible to create a Tile that only shows the results of a function.


ivantodorovich:
@legalsylvain I would love to hear your thoughts

ivantodorovich:
Ok this is ready for review now

ivantodorovich:
@pedrobaeza, I ping you because you seem to be the guy to call when some PR needs reviewing 😝 
Would you mind adding the `needs review` label? I'm looking forward to get some reviews and see this merged.

web_dashboard_tile/models/tile_tile.py
```diff
@@ -14,18 +14,56 @@
...
 
+def median(vals):
+    # https://docs.python.org/3/library/statistics.html#statistics.median
+    # TODO : refactor, using statistics.median when Odoo will be available
+    #  in Python 3.4
+    even = (0 if len(vals) % 2 else 1) + 1
+    half = (len(vals) - 1) / 2
+    return sum(sorted(vals)[half:half + even]) / float(even)
+
+FIELD_FUNCTIONS = {
```
lasley: IMO it would be a good idea to use a collections.OrderedDict here. Otherwise the select values are going to randomly change places due to dicts not being ordered

web_dashboard_tile/models/tile_tile.py
```diff
@@ -14,18 +14,56 @@
...
+    'count': {
+        'name': 'Count',
+        'func': False,  # its hardcoded in _compute_data
+        'help': 'Number of records',
```
lasley: I don't immediately see where these helps are used, but I don't think in ORM. Should probably translate `_('Number of records')`

web_dashboard_tile/models/tile_tile.py
```diff
@@ -46,70 +84,113 @@ def _context_today():
...
-                elif self.field_function == 'sum':
-                    self.computed_value = sum(vals)
-                elif self.field_function == 'avg':
-                    self.computed_value = sum(vals) / len(vals)
-                elif self.field_function == 'median':
-                    self.computed_value = self.median(vals)
+            domain = self.domain or '[]'
+            count = model.search_count(eval(domain, eval_context))
+            if any([
+                self.primary_function and
```
lasley: Why the additional boolean guards? `self.primary_function != 'count'` implies that `self.primary_function` is also `True`ish

web_dashboard_tile/models/tile_tile.py
```diff
@@ -135,19 +216,28 @@ def _search_active(self, operator, value):
...
     # Constraints and onchanges
     @api.one
-    @api.constrains('model_id', 'field_id')
+    @api.constrains('model_id', 'primary_field_id', 'secondary_field_id')
     def _check_model_id_field_id(self):
-        if self.field_id and self.field_id.model_id.id != self.model_id.id:
-            raise ValidationError(
-                _("Please select a field from the selected model."))
+        if any([
+            self.primary_field_id and
```
lasley, line 233: Same thing with double guarding here, you can just perform the comparisons.

web_dashboard_tile/models/tile_tile.py
```diff
@@ -46,70 +84,113 @@ def _context_today():
...
+                    func = FIELD_FUNCTIONS[self[f+'function']]['func']
+                    if func and self[f+'field_id'] and count:
+                        vals = [x[self[f+'field_id'].name] for x in records]
+                        value = func(vals)
+                if self[f+'function']:
+                    self[f+'value'] = \
+                        (self[f+'format'] or '{:,}').format(value)
+                else:
+                    self[f+'value'] = False
+        except Exception as e:
```
lasley, line 154: A large try/except like this with a catch for all exceptions is bad practice. You should try to minimize the size of your guards, as well as target specific exceptions that you are expecting.

web_dashboard_tile/models/tile_tile.py
```diff
@@ -46,70 +84,113 @@ def _context_today():
...
+            if any([
+                self.primary_function and
+                self.primary_function != 'count',
+                self.secondary_function and
+                self.secondary_function != 'count'
+                    ]):
+                    records = model.search(eval(domain, eval_context))
+            for f in ['primary_', 'secondary_']:
+                value = 0
+                if self[f+'function'] == 'count':
```
lasley: These string concats are being performed a lot. Why not make an intermediary variable & use instead?

lasley (changes requested):
Thanks @ivantodorovich - comments inline

Additionally, please add test coverage for your added feature-set.

web_dashboard_tile/models/tile_tile.py
ivantodorovich: Since `primary_function` and `secondary_function` are not required, they might be `False`.
Am I missing something?

web_dashboard_tile/models/tile_tile.py
ivantodorovich: I'm translating them afterwords, at the time of usage. But you're right, I should translate it here! Ty!

web_dashboard_tile/models/tile_tile.py
lasley: But if they're false, they're also != 'count' right? There's no need to explicitly check for false basically

web_dashboard_tile/models/tile_tile.py
ivantodorovich: Well, it is kinda the point to handle any possible exception inside the tile, instead of passing through to the ui, because they would break the tile form and it makes it impossible to solve without deleting the tile.

web_dashboard_tile/models/tile_tile.py
ivantodorovich: We want only the cases where `***_function` is `True`ish **and** not `count`.

The following comparisson should result as following

```
False -> False
'count' -> False
'any_other_func' -> True
```

If the comparisson is only `!= 'count'` then it would be as follows:

```
False -> True
'count' -> False
'any_other_func' -> True
```

lasley: Ahh yup, you're right. Ignore this and the other comment on that

ivantodorovich:
@lasley thank you for your review. Please check the submitted changes.

web_dashboard_tile/README.rst
```diff
@@ -36,6 +41,8 @@ Roadmap
 * Add icons.
 * Support client side action (like inbox).
 * Restore original Domain + Filter when an action is set.
+* Posibility to hide the tile based on a field expression.
```
jjscarafia, line 44: Are this two options new? Where can I found them?

web_dashboard_tile/README.rst
```diff
@@ -36,6 +41,8 @@ Roadmap
 * Add icons.
 * Support client side action (like inbox).
 * Restore original Domain + Filter when an action is set.
```
jjscarafia, line 43: I would add here something about the "Format" with an example

jjscarafia:
Good work, :+1: functional testing. (I've add some notes about readme)

web_dashboard_tile/README.rst
ivantodorovich: These are just comments under the Roadmap section. I just added them because I thought they might be a good addition to this module, but I haven't had the time to implement them right now. Also I didn't want to make a larger PR.

web_dashboard_tile/README.rst
ivantodorovich: There's usage help on the field: `Python Format String valid with str.format()`
I think it's a good idea to add some examples there, though.

ivantodorovich:
@jjscarafia thank you for your review 😄 
Just pushed a few more commits based on your suggestion.

jjscarafia (approved):
Thanks!

lasley (approved):
Thanks @ivantodorovich

PR #484, opened 2016-11-16
[MIG] web_m2x_options to 10.0

jlzhou:
#435 

There is an issue need help:

In ir/ir_ui_view.py file:
node.set('can_create', 'true' if Comodel.check_access_rights('create', raise_exception=False) else 'false')
node.set('can_write', 'true' if Comodel.check_access_rights('write', raise_exception=False) else 'false')

and then in form_common.js file:
var CompletionFieldMixin = {
init: function() {
this.limit = 7;
this.orderer = new utils.DropMisordered();
this.can_create = this.node.attrs.can_create || true;
this.can_write = this.node.attrs.can_write || true;
},
......

It seems can_create and can_write always returns true, I wonder if it is a bug?

I need this return value to change the visibility of "open" icon next to drop down.


pedrobaeza:
Please cherry-pick the commit from the latest 9.0 patch https://github.com/OCA/web/pull/543

pedrobaeza:
Superseeded by #543

PR #485, opened 2016-11-17
[FIX] spurious dirty flag

hbrunn:


yajo (changes requested):
Explain yourself please.

hbrunn:
the filters are only meant to be an ui adaption. But because we manipulate the dataset, the field will be marked as dirty after applying a filter which actually changes the list of records shown. If you have a default filter, it's even more annoying, because then this also happens in readonly mode for every form you open

yajo:
So is there a way to check this actually fixes something? Sorry for these dumb questions, but I'm no seer :yum:

hbrunn:
on runbot, open the group contact creation. Then go somewhere else. Without this patch, Odoo will tell you that the form has changed and if you want to save

yajo:
I tried and it did not happen, could you specify better?

hbrunn:
ah, that's a many2many fields, there, this doesn't happen. If you want to test, define a default filter on a one2many field, there you'll have the problem.
But it's simple to see in the code what's going on: FieldOne2Many's dataset is a `BufferedDataSet` (https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_form.js#L4423) and those set the dirty flag if `alter_ids` actually alters the ids: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/data.js#L1049
And we don't want that, because we don't want to change the data, only what's displayed.

yajo (approved):
OK I get it now. You want to change the displayed IDS, but not the ones actually found in the field.

Only code review, due to still not having clear STR (*ehem* 😋)

pedrobaeza (approved):

PR #489, opened 2016-11-17
9.0-mig-web_ir_actions_act_window_none

susport:
Mig 9.0 web_ir_actions_act_window_none


hbrunn:
this works just like that because of the compatibility wrapper. but you should migrate this to the `require` api. which will also be a better argument to add yourself to the list of contributors.

and would you care to add js tests?

web_ir_actions_act_window_none/static/src/js/web_ir_actions_act_window_none.js
```diff
@@ -0,0 +1,19 @@
+//-*- coding: utf-8 -*-
+//?? 2016 Therp BV <http://therp.nl>
+//License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+odoo.web_ir_actions_act_window_none = function(instance)
```
hbrunn: this is not yet v9 api. https://www.odoo.com/documentation/9.0/reference/javascript.html#web-client should give you the idea

web_ir_actions_act_window_none/README.rst
```diff
@@ -0,0 +1,62 @@
...
+    :target: https://runbot.odoo-community.org/runbot/162/9.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/web/issues>`_.
```
tarteo, line 35: Please update this section to the latest version of the template.

web_ir_actions_act_window_none/static/src/js/web_ir_actions_act_window_none.js
```diff
@@ -0,0 +1,24 @@
+//-*- coding: utf-8 -*-
+//?? 2016 Therp BV <http://therp.nl>
+//License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+odoo.define('web_ir_actions_act_window_none', function(require) {
+
```
tarteo, line 7: "use strict"; is for all javascript files recommended.

tarteo (approved):
I only have some nitpicking. 👍

web_ir_actions_act_window_none/README.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Usage
+=====
+
+To use this module, depend on it in your module and return::
+
+    {'type': 'ir.actions.act_window.none'}
+
+The button will trigger a reload of the popup's form, in case you want to
+suppress this, set key ``reload`` to ``False`` in the action you return.
```
yajo, line 22: Assuming you set a `none` action with `reload=False`...... What's this for? An addon that lets buttons do nothing?

You should drop support of `reload=False` and call this `web_action_reload`, I'm not in favor of merging an addon that adds a feature that does nothing, honestly 😕.

yajo (changes requested):

PR #493, opened 2016-11-21
[FIX] 7.0 Fix JavaScript error: undefined variable 'widget.view'

### LINKED ISSUE #466: m2x options for view
An error is showing when opening crm_claim form view, after merging [IMP] 7.0 add m2x options for view: https://github.com/OCA/web/pull/305

To reproduce it:
- Go to runbot http://3271007-7-0-64d3bd.runbot1.odoo-community.org/?db=3271007-7-0-64d3bd-all
- Install crm_claim
- Go to Sales -> After-Sale-Services-> Claims.
- Open a claim or click on "Create" button.

Then you get the following:

image

The error is triggered in this line:

https://github.com/initOS/web/blob/c89ab90c056f116d033b19ab1467d6bdfc0701f7/web_m2x_options/static/src/js/form.js#L40

I realize that the problem in this form is related to the field ref which is defined using the selection=crm._links_get 

It also happens in other form views where the field defined using that property as crm_helpdesk for example.

azoellner:
This fixes #466

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -37,11 +37,11 @@ openerp.web_m2x_options = function (instance) {
             if (p < 0) return;
             var key = option.substring(p + 1); // w/o 'web_m2x_options' prefix
             // ... hence set from context ...
-            if (!_.isUndefined(widget.view.dataset.context[key])) {
+            if (!_.isUndefined(widget.view) && !_.isUndefined(widget.view.dataset.context[key])) {
                 widget.view.ir_options[option] = widget.view.dataset.context[key];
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
Thank you for the job!

Don't know the reason but runbot was not built correctly. When I try to access I get the following message:
OperationalError: FATAL:  database "3272958-493-018b2b-all" does not exist 

May it be fixed with a rebuild?

njeudy (approved):

AaronHForgeFlow (approved):
:+1:

hbrunn:
ci error unrelated. merging

PR #496, opened 2016-11-23
[8.0][FIX] web_dashboard_tile security rule

ivantodorovich:
Rule was not being updated because of `<data noupdate="1">`, hence the groups field of the tile was not working.

This is something I missed in https://github.com/OCA/web/pull/476, and didn't realize up until now. 
Sorry for this.

### DIFF
--- a/web_dashboard_tile/__openerp__.py
+++ b/web_dashboard_tile/__openerp__.py
@@ -5,7 +5,7 @@
 {
     "name": "Dashboard Tile",
     "summary": "Add Tiles to Dashboard",
-    "version": "8.0.3.0.0",
+    "version": "8.0.4.0.0",
     "depends": [
         'web',
         'board',
--- a/web_dashboard_tile/migrations/8.0.4.0/post-migration.py
+++ b/web_dashboard_tile/migrations/8.0.4.0/post-migration.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# © 2016 Iván Todorovich <ivan.todorovich@gmail.com>
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+
+def migrate(cr, version):
+    if version is None:
+        return
+
+    # Update ir.rule
+    cr.execute("""
+        SELECT res_id FROM ir_model_data
+        WHERE name = 'model_tile_rule'
+        AND module = 'web_dashboard_tile'""")
+    rule_id = cr.fetchone()[0]
+    new_domain = """[
+        "|",
+        ("user_id","=",user.id),
+        ("user_id","=",False),
+        "|",
+        ("group_ids","=",False),
+        ("group_ids","in",[g.id for g in user.groups_id]),
+        ]"""
+    cr.execute("""
+        UPDATE ir_rule SET domain_force = '%(domain)s'
+        WHERE id = '%(id)s' """ % {'domain': new_domain, 'id': rule_id})


legalsylvain:
Hi @ivantodorovich . 
I don't think that set noupdate to 0 is a good way to fix this problem. ir.rule should be noupdate set to 1 allways. 
This should be fixed better by migration script. Don't you think ?

ivantodorovich:
That sounds better, indeed!
I'll fix it in a moment

ivantodorovich:
@legalsylvain have you seen another case like this?

I have something like this but I don't like it:
```py
    # Update ir.rule
    domain = """[
        '|',
        ('user_id','=',user.id),
        ('user_id','=',False),
        '|',
        ('group_ids','=',False),
        ('group_ids','in',[g.id for g in user.groups_id]),
        ]"""
    cr.execute("""UPDATE ir_rule SET domain_force = "%s" 
                  WHERE name = 'tile.owner' """ % domain)
```

- I don't like `WHERE name = 'tile.owner'` because there might be more one rule with the same name.
I'd like to only edit `web_dashboard_tile.model_tile_rule`, but I don't know how to make the query to filter by external id.

lasley:
@ivantodorovich why not just create an environment and use ORM? Otherwise, a join on `ir_model_data` against the `complete_name` would do it

lasley:
Env generation example can be seen in [Contribution docs](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction)

```
  with openerp.api.Environment.manage():
      with openerp.registry(self.env.cr.dbname).cursor() as new_cr:
          # Create a new environment with new cursor database
          new_env = api.Environment(new_cr, self.env.uid, self.env.context)
          # with_env replace original env for this method
          self.with_env(new_env).write({'name': 'hello'})  # isolated transaction to commit
          new_env.cr.commit()  # Don't show a invalid-commit in this case
      # You don't need close your cr because is closed when finish "with"
  # You don't need clear caches because is cleared when finish "with"
```

ivantodorovich:
@lasley @legalsylvain thank you both!!

I've pushed the changes. Didn't change version because it's a fix of the last PR that's very recent.
Although, If you think it's best to use a new version number, I'll change it :)

web_dashboard_tile/migrations/8.0.3.0/post-migration.py
```diff
@@ -11,3 +11,21 @@ def migrate(cr, version):
...
+        "|",
+        ("user_id","=",user.id),
+        ("user_id","=",False),
+        "|",
+        ("group_ids","=",False),
+        ("group_ids","in",[g.id for g in user.groups_id]),
+        ]"""
+    cr.execute("""
+        UPDATE ir_rule SET domain_force = '%(domain)s'
+        WHERE id = '%(id)s' """ % { 'domain': new_domain, 'id': rule_id })
```
susport: Please correct space error in } in travis.

ivantodorovich:
Done

legalsylvain:
Hi thanks a lot @ivantodorovich. The code seems perfect.
About version number, as this module is maybe used in production without this fix, I guess you should move the added code  in a new file migrations/8.0.4.0/post-migration.py and set 8.0.4.0 as new version in the manifest file.

kind regards.

ivantodorovich:
@legalsylvain done :) thank you for your comments.

susport (commented):
👍

legalsylvain:
:+1: . Thaks a lot !

legalsylvain:
PR : quick merged, as it is quite simple and fix a production bug.

PR #497, opened 2016-11-24
[FIX][10.0] web_responsive: Add class to dropdown

lasley:
* Add `oe_menu_toggler` class to secondary menus


lasley:
Don't review/merge this just yet - I think there's still a missing piece here. Trying to make it so our menus have the same classes as the Odoo ones; it turns out that some incompatibilities with some tours exist otherwise.

pedrobaeza:
Is this still valid?

lasley:
It is not! Closing

PR #506, opened 2016-12-12
[IMP] Add possibility to send notification in new environement.

adrienpeiffer:
In some cases, It may be necessary to notify the user directly after the instruction.
For instance, in my case I run an asynchronous process that will take a long time and I want to inform the user that the process is running.

### DIFF
--- a/web_notify/models/res_users.py
+++ b/web_notify/models/res_users.py
@@ -4,6 +4,8 @@
 
 from odoo import api, fields, models, _
 
+import odoo
+
 
 class ResUsers(models.Model):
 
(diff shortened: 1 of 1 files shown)


web_notify/models/res_users.py
```diff
@@ -23,24 +25,35 @@ def _compute_channel_names(self):
...
                          for record in self]
-        self.env['bus.bus'].sendmany(notifications)
+        if commit:
+            with odoo.api.Environment.manage():
+                with odoo.registry(self.env.cr.dbname).cursor() as new_cr:
+                    new_env = api.Environment(new_cr, self.env.uid,
+                                              self.env.context)
+                    self.env['bus.bus'].with_env(new_env).sendmany(
+                        notifications)
+                    new_env.cr.commit()
```
yajo, line 57: [This is implicit](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/odoo/sql_db.py#L391), no need for it.

OTOH, [quoting from guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#never-commit-the-transaction):

> All cr.commit() calls outside of the server framework from now on must have an explicit comment explaining why they are absolutely necessary, why they are indeed correct, and why they do not break the transactions. Otherwise they can and will be removed!

So, although you will remove this line and don't have a handwritten call to commit, this `with` block deserves a comment on why is it needed, or this should be rejected according to guidelines.

yajo (changes requested):

web_notify/models/res_users.py
adrienpeiffer: Hi @Yajo 

Thanks for your comment.

> This is implicit, no need for it.

I wrote this code like the exemple in the last paragraph of the section in the guidelines :

> You can isolate a transaction for a valid cr.commit using Environment:
```
with openerp.api.Environment.manage():
      with openerp.registry(self.env.cr.dbname).cursor() as new_cr:
          # Create a new environment with new cursor database
          new_env = api.Environment(new_cr, self.env.uid, self.env.context)
          # with_env replace original env for this method
          self.with_env(new_env).write({'name': 'hello'})  # isolated transaction to commit
          new_env.cr.commit()  # Don't show a invalid-commit in this case
      # You don't need close your cr because is closed when finish "with"
  # You don't need clear caches because is cleared when finish "with"
```

yajo: Then you found a bug in the guidelines 😆. Let's be honest, committing twice is not a huge bug, but given `with cursor()` does it automatically, makes not much sense, don't you think?

In any case please document what you are doing 😉

PR #507, opened 2016-12-13
9.0 mig web dashboard tile

nicomacr:


ivantodorovich:
Tested on runbot, seems ok.

PR #509, opened 2016-12-20
[10.0] Add module web_no_bubble

ghost:


pedrobaeza:
Isn't any configuration for this?

ghost:
@pedrobaeza there is the Consume Tour button in debug mode which could do the trick. It consumes all notification bubbles. This module does not remove the notifications but only the odd bubbles that come with it.

pedrobaeza (approved):
I also see them very annoying, so goodbye!

rafaelbn (approved):
Thanks!

guewen (approved):

PR #511, opened 2016-12-27
Scroll control panel with page using CSS instead of jQuery.

danieltorres7:
jQuery scrolling caused scrollbar malfunction.

Scrollbar incresing size while scrolling down and shaking at the end of scroll.

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -88,12 +88,6 @@ odoo.define('web_responsive', function(require) {
                                 'a.oe_logo, ' +
                                 'i.oe_logo_edit'
                                 );
-            $('.o_content').scroll(function() {
-                $('.o_control_panel').css(
-                    'margin-top',
-                    -$(this).scrollTop() + 'px'
-                );
-            });
             $clickZones.click($.proxy(this.handleClickZones, this));
             core.bus.on('resize', this, this.handleWindowResize);
             core.bus.on('keydown', this, this.handleNavKeys);
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -44,3 +44,15 @@ main {
 .o_tooltip.active {
     z-index: 1051;
 }
+
+.o_web_client {
+    >.o_main {
+        overflow: auto;
+        > .o_main_content {
+            overflow: initial;
+            > .o_content {
+                overflow: initial;
+            }
+        }
+    }
+}


danieltorres7:
@lasley, what do you think?

lasley:
@danieltorres7 Oh snap I noticed the shaking issue the other day! I'll review this sometime today, thanks for the fix

pedrobaeza:
Is this also valid for v9?

pedrobaeza (approved):
Tried on a runbot without the patch and with the patch, and I have found the issue, and it's solved with this.

moylop260 (approved):
@oscarolar ping

lasley (approved):
LGTM code + functional, thanks @danieltorres7 

@pedrobaeza - no this isn't applicable to v9

pedrobaeza:
Oh, what a pity! @lasley, did you have time to check the problem in version 9 that seems to not happen in this version (two scrollbars and a increasing offset in the right part)?

lasley:
Yeah I just saw the two scollbars when I tested to see if this was applicable. TBH I had no idea that one was there, I've been in v10 mode since about the time this was released. I'll take a look though because I think a very similar fix may apply, just not a direct 1:1. 

Putting this tab in my todo tabs 😉 

@danieltorres7 if you're feeling adventurous, a v9 port would be amazing! I think the CSS is the only that will apply, and it will be the old web client (oe_webclient IIRC).

lasley:
Oh also @pedrobaeza yeah it seems like the two scrollbar thing was replaced with the Javascript bounce in v10 instead

yajo:
You can see in known issues:

> If you resize the window, body gets a wrong overflow: auto css property and you need to refresh your view or open/close the app drawer to fix that.

If I don't remember wrong, this is caused by a some hardcoded stuff found in the drawer library.

lasley:
@Yajo - I'm actually getting the additional scrollbars in 9 when in any tree view, with no resizing (Chrome 55.0.2883.95 (64-bit)). Do you think these two issues are related then?

yajo:
Yes, I noticed recently it does not only happen at resizing, but some other times, although I'm not sure we're using the last version at production (@pedrobaeza?). But most likely both are the same. This is one of the reasons you see my crusade against inline and `!important` CSS styles all around OCA 😆.

The library should add a custom class instead of adding a custom style, and then we could override that class' style. Maybe [a patch to them](https://github.com/blivesta/drawer/search?l=JavaScript&q=overflow+auto&utf8=%E2%9C%93) (and consecutive lib upgrade in our addon) would be the best.

pedrobaeza:
@Yajo, we're using last version indeed.

leorochael:
> The library should add a custom class instead of adding a custom style, and then we could override that class' style. Maybe a [patch to them](https://github.com/blivesta/drawer/search?l=JavaScript&q=overflow+auto&utf8=%E2%9C%93) (and consecutive lib upgrade in our addon) would be the best.

@Yajo, Now there is another version of `drawer` available, but it doesn't fix the issue with using a custom class instead of a style. They already had a class appropriate for that, though, so a PR was easy: blivesta/drawer#36.

Let's just hope they act on it. I made available the same change on #513, though.

yajo:
@pedrobaeza will love you for that 😉

pedrobaeza:
Can @Yajo or any backport this to v9?

pedrobaeza:
I'm making a blind try in https://github.com/OCA/web/pull/527

PR #513, opened 2016-12-31
Web responsive menu improvements

danieltorres7:
Added apptitle before oe_submenu
Increased user avatar size to maximum
Fullscren dropdown menus on mobile mode

### DIFF
--- a/web_responsive/static/lib/js/drawer.3.2.0.js
+++ b/web_responsive/static/lib/js/drawer.3.2.0.js
@@ -136,7 +136,11 @@
       return $this
         .removeClass(__.settings.class.open)
         .addClass(__.settings.class.close)
-        .css({ 'overflow': 'auto' })
+        // TODO: remove this local patch once PR below is approved:
+        // https://github.com/blivesta/drawer/pull/36
+        //.css("overflow", "auto")
+        .css("overflow", "")
+        // end local patch
         .drawerCallback(function triggerClosedListeners() {
           __.settings.state = false;
           $this.trigger(__.settings.events.closed);
--- a/web_responsive/static/lib/js/drawer.3.2.0.js
+++ b/web_responsive/static/lib/js/drawer.3.2.0.js
@@ -121,7 +121,10 @@
       return $this
         .removeClass(__.settings.class.close)
         .addClass(__.settings.class.open)
-        .css({ 'overflow': 'hidden' })
+        // XXX: local patch waiting for:
+        // https://github.com/blivesta/drawer/pull/36
+        //.css({ 'overflow': 'hidden' })
+        // end local patch
         .drawerCallback(function triggerOpenedListeners() {
           __.settings.state = true;
           $this.trigger(__.settings.events.opened);
@@ -136,10 +139,9 @@
       return $this
         .removeClass(__.settings.class.open)
         .addClass(__.settings.class.close)
-        // TODO: remove this local patch once PR below is approved:
+        // XXX: local patch waiting for:
         // https://github.com/blivesta/drawer/pull/36
         //.css("overflow", "auto")
-        .css("overflow", "")
         // end local patch
         .drawerCallback(function triggerClosedListeners() {
           __.settings.state = false;
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -56,3 +56,10 @@ main {
         }
     }
 }
+
+// Remove z-index from CP buttons so it doesn't overlap the menu
+.o_control_panel {
+     .btn-group > .btn.active {
+        z-index: initial;
+    }
+}
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -63,3 +63,7 @@ main {
         z-index: initial;
     }
 }
+
+.o_chat_window {
+    z-index: 1000;
+}
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -4,7 +4,6 @@
 body {
     width: 100%;
     height: 100%;
-    overflow: hidden!important; // Important is to override base template
 
     // Do not fix the search part, it's too big for small screens
     @media (max-width: @screen-sm-max) {
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -4,6 +4,7 @@
 body {
     width: 100%;
     height: 100%;
+    overflow: hidden!important; // Important is to override base template
 
     // Do not fix the search part, it's too big for small screens
     @media (max-width: @screen-sm-max) {
--- a/web_responsive/static/src/less/navbar.less
+++ b/web_responsive/static/src/less/navbar.less
@@ -18,12 +18,159 @@
         padding: @app-drawer-navbar-padding-vertical @app-drawer-padding-horizontal;
     }
     
+    // App name before oe_submenu
+    .app-name {
+        > .oe_menu_text {
+            font-size: 20px;
+            position: relative;
+            padding: 0 8px;
+            top: 1px;
+        }
+    }
+
+    // Fitt's law: Stick menu to the top of the screen
+    @media (min-width: @screen-sm) {
+        .navbar-nav {
+            margin-top: -2px;
+        }
+    }
+
+    .navbar-header {
+        // Navbar header mobile styling
+        margin-left: -11px;
+        margin-top: 5px;
+        margin-bottom: 5px;
+        margin-right: -11px;
+        .navbar-toggle {
+             margin: 0;
+        }
+
+        // Fitt's law: Stick app-drawer button to the top-left of the screen
+        @media (min-width: @screen-sm) {
+            margin-left: -16px;
+            margin-right: 0px;
+            margin-bottom: 0px;
+            margin-top: 0px;
+            padding: 0px;
+            padding-right: 0px;
+        }
+    }
+
+    @media (max-width: @screen-xs-max) {
+        ul.nav.navbar-nav.navbar-right.navbar-systray-item.oe_systray {
+            margin-top: 1.5px;
+        }
+    }
+
+    @media (min-width: @screen-sm) {
+        .oe_secondary_menu {
+            > li {
+                > a {
+                    margin-top: 2.5px;
+                    height: 25px;
+                }
+            }
+        }
+    }
+
+    .o_sub_menu > .o_sub_menu_content > .oe_secondary_menu > li.open > .dropdown-menu {
+
+        // Drop-down menu should be scrollable on small screens
+        overflow: auto;
+
+        // Headers at dropdown-menu shoud look different
+        > li.dropdown-header {
+            color: white;
+            text-decoration: none;
+            background-color: #8E8E8E;
+            font-weight: bold;
+        }
+
+    }
+
+   // Systray on the top menu bar
+   .navbar-systray {
+       position: absolute;
+       top: 0;
+       right: 32px;
+       white-space: nowrap;
+
+       @media (max-width: @screen-xs-max) {
+           top: 3px;
+           right: 76px;
+           .fa {
+               margin-top: 3px;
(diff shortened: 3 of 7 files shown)


pedrobaeza:
There are conflicts in the branch

danieltorres7:
@pedrobaeza, all conflicts solved.

web_responsive/static/src/less/main.less
```diff
@@ -45,6 +45,15 @@ main {
     z-index: 1051;
 }
 
+
+// Remove z-index to preserve menu
+.o_control_panel {
+    .btn {
+        z-index: initial!important;
```
pedrobaeza: Please don't use important, as it overwrites everything else. Use a more specific selector instead.

pedrobaeza:
In general, you must avoid to use `!important` in CSS styling

danieltorres7:
@pedrobaeza. Succesfully replaced all important decorators by more specific selectors.

danieltorres7:
Check it out @rousseldenis, @lasley and @leorochael. Everything seems to be working with the new menus.

pedrobaeza (approved):
Tested on runbot

pedrobaeza:
@Yajo, can you review technically?

danieltorres7:
I've completely rewritten the commits, removing redundant stuff and adding better comments and commit messages, but the behavior should remain the same (save for a small animation bug that I fixed).

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+        .navbar-nav {
+            margin-top: -2px;
+        }
+    }
+
+    .navbar-header {
+        // Navbar header mobile styling
+        margin-left: -11px;
+        margin-top: 5px;
+        margin-bottom: 5px;
```
lasley: Should we maybe use the Bootstrap vars here for consistency? - https://github.com/odoo/odoo/blob/10.0/addons/web/static/lib/bootstrap/less/variables.less#L361

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+    .o_sub_menu > .o_sub_menu_content > .oe_secondary_menu > li.open > .dropdown-menu {
+
+        // Drop-down menu should be scrollable on small screens
+        overflow: auto;
+
+        // Headers at dropdown-menu shoud look different
+        > li.dropdown-header {
+            color: white;
+            text-decoration: none;
+            background-color: #8E8E8E;
```
yajo: I think you'd better use the corresponding [Odoo](https://github.com/odoo/odoo/blob/10.0/addons/web/static/src/less/variables.less) or [Bootstrap](https://github.com/odoo/odoo/blob/10.0/addons/web/static/lib/bootstrap/less/variables.less) variables for color and background-color.

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+        > .oe_menu_text {
+            font-size: 20px;
+            position: relative;
+            padding: 0 8px;
+            top: 1px;
+        }
+    }
+
+    // Fitt's law: Stick menu to the top of the screen
+    @media (min-width: @screen-sm) {
```
yajo: `@screen-sm` is deprecated. Use `@screen-sm-min`.

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+        margin-left: -11px;
+        margin-top: 5px;
+        margin-bottom: 5px;
+        margin-right: -11px;
+        .navbar-toggle {
+             margin: 0;
+        }
+
+        // Fitt's law: Stick app-drawer button to the top-left of the screen
+        @media (min-width: @screen-sm) {
```
yajo: Deprecated.

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+        }
+    }
+
+    @media (max-width: @screen-xs-max) {
+        ul.nav.navbar-nav.navbar-right.navbar-systray-item.oe_systray {
+            margin-top: 1.5px;
+        }
+    }
+
+    @media (min-width: @screen-sm) {
```
yajo: Deprecated.

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+       }
+   }
+
+    // Force mobile mode dropdown-menu style
+   .navbar-right {
+        float: right;
+        margin: 0px;
+        margin-right: -16px;
+
+        @media (min-width: @screen-sm) {
```
yajo: Deprecated.

web_responsive/static/src/less/navbar.less
```diff
@@ -18,12 +18,159 @@
...
+            @media (max-width: @screen-xs-max) {
+                position: fixed;
+                top: 46px;
+                bottom: 0;
+                left: 0;
+                right: 0;
+                overflow: auto;
+                float: left;
+                margin-top: 2px;
+                background-color: #ffffff;
```
yajo: Again does any available variable fit in?

yajo (changes requested):
Besides some technical details below, I think the top bar should be thick in sm size like it is in xs size, to make it more usable in tablets.

danieltorres7:
@Yajo and @pedrobaeza. Requested changes implemented.

danieltorres7:
> Besides some technical details below, I think the top bar should be thick in sm size like it is in xs size, to make it more usable in tablets.

@Yajo, I agree on you, anyway, how can we make thick top bars for tablets not changing desktop ones? I mean, resolution is higher than 768px on both and there is not available decorators (at least I don't know at the moment) to easily check if the device is a tablet.

yajo:
There must be a media query that makes the bar thick in xs screens (or thin in >=sm). All you have to do is find it and change the width variable from xs to sm.

Oh, and about tablet or small PC, as long as tablet UI is usable in a small PC, you shouldn't care about desktop. At least until the [pointer media query](https://developer.mozilla.org/en/docs/Web/CSS/@media/pointer) lands mainstream.

danieltorres7:
Thick top menus are already available for all sm screens, not only xs.

leorochael:
@Yajo, do you want the top menubar to remain thick, no matter how large the screen?

This way the only difference between larger and smaller screens would be the presence of the menus directly on the top bar, or inside the menu button.

Is this what you want?

danieltorres7:
I will refactor changes, so top menu height will remain the same on all modes, is that what you want? Always thick no matter what mode is active?

pedrobaeza:
He will answer tomorrow, as he works only up to 15:00 CET

lasley:
> Always thick no matter what mode is active?

@danieltorres7 @leorochael - I think he is saying always thick when on XS and SM, but leave it for the others. I think I agree, that would be a great change.

rousseldenis (approved):
:+1: Tested on runbot

web_responsive/static/src/less/navbar.less
```diff
@@ -18,16 +18,166 @@
         padding: @app-drawer-navbar-padding-vertical @app-drawer-padding-horizontal;
     }
     
+    // App name before oe_submenu
+    .app-name {
+        > .oe_menu_text {
```
lasley: Somewhat nitpicky, but maybe a `cursor: default` here to override `text`

lasley (approved):
@danieltorres7 Excellent improvement, few comments but otherwise 👍 

Looking at this, the navbar changing @Yajo is referring to is controlled with the [navbar-collapse, navbar-toggle](https://github.com/OCA/web/blob/10.0/web_responsive/views/web.xml#L43) classes.

I think maybe we should add this one to the roadmap. It's going to be getting into changing the way Bootstrap works, so I think it'll cause some cascading changes that'll blow this PR into a rabbit hole.

@Yajo - correct me if I'm wrong on my inferences.

yajo (dismissed):
I'm afrraid not, friend. Just open the runbot and resize the window to sm size (768-991px width) and you'll see the top bar is thinner than xs (<=767px).

In any case, I'm not blocking this for that reason, but if you want to push that, it would be great 😉 

Note to PSC: **squash** at merge.

yajo (approved):
OK now I have been able to review properly. Indeed I meant what @lasley says. This looks good to me, but I'll wait for @danieltorres7 feedback before merging: do you want to address those changes @lasley said before we merge?

And about the thick bar, if it is too complicated, then maybe just a roadmap note would work for now.

Thanks!

leorochael:
Ok, but considering that we can't tell large tablets apart from desktops before the pointer media query lands, wouldn't it be better (and simpler in the code) to just keep a wide top bar, no matter the resolution?

That's what Odoo does by default, BTW

danieltorres7:
Top menu now is always thick, simplified some styles for that.

web_responsive/static/src/less/main.less
```diff
@@ -4,6 +4,7 @@
 body {
     width: 100%;
     height: 100%;
+    overflow: hidden!important; // Important is to override base template
```
yajo: You knew this was going to happen... Please change this 😆 

A simple trick for these kind of overriding is:

```
html & {
  [syles...]
}
```

`html body` is more concise than `body` selector, so it will win the fight 😉

yajo (changes requested):
You know what? It looks awesome. Just one more little change to merge 🎉

web_responsive/static/src/less/main.less
leorochael: What @danieltorres7 mean by "base template", is that there's a `style="overflow: auto;"` in the `<body>` tag ` *directly in the DOM*.

There's no way to add a more specific selector without `!important`

leorochael: Apparently this `style` attribute is being added by The `drawer.css` library because `<body>` has a `drawer` class, added by the `web_responsive` module. We're still investigating.

leorochael: This happens on the `drawer close` event of the app drawer, which happens automatically on the closing of the app drawer, or on window resize (triggers the drawer close).

At this point we would help suggestions on how to deal with the issue and how to remove the `!important`.

danieltorres7: Just removed overflow: hidden!important from main.less. Library drawer.3.2.0.js adds "overflow: auto" on close, also triggered on resize like @leorochael told us. This causes body to have it's own scrollbar and at small height screens we got 2 scrollbars at the same time. Committing that change on lib separately now...

web_responsive/static/lib/js/drawer.3.2.0.js
```diff
@@ -136,7 +136,11 @@
       return $this
         .removeClass(__.settings.class.open)
         .addClass(__.settings.class.close)
-        .css({ 'overflow': 'auto' })
+        // TODO: remove this local patch once PR below is approved:
```
leorochael: Given [@Yajo's comment on 511](https://github.com/OCA/web/pull/511#issuecomment-269595936), This line should probably be just removed, and the following block added to `web_responsive/static/lib/css/drawer.3.2.0.css`:

    .drawer-open {
        overlay: hidden;
    }

After doing the same in a PR for drawer, of course (which means, I should probably [update my own PR](https://github.com/blivesta/drawer/pull/36)...)

danieltorres7:
@Yajo, requested changes and final fixes applied. We are ready for the final review.

leorochael:
Please, let us know when it's ok to squash/reorder the commits.

lasley:
@Yajo - the changes look good to you? 

@leorochael - Once he confirms, you're good to squash/reorder.

rousseldenis:
Come back from holidays and changes seem great !
Thanks for the job !

yajo (approved):

yajo:
Please squash to merge 😉

pedrobaeza:
I have squashed on merge.

pedrobaeza:
Can you please backport this to v9 or isn't direct?

PR #514, opened 2017-01-02
Web responsive form line breaks

danieltorres7:
On screens with less than 768px content is limited to 2 columns
On mobile titles have its own line

Depends:
- [x] #513

### DIFF
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -62,4 +62,22 @@
             }
         }
     }
+
+    // Make image editing controls always available, instead of depending on resolution or hover
+    .o_form_field_image > .o_form_image_controls {
+        position: initial;
+        opacity: 1;
+        > .fa {
+            width: 50%;
+            padding: 6px;
+            margin: 0px;
+            text-align: center;
+        }
+        > .fa.o_select_file_button {
+            background: @odoo-brand-primary;
+        }
+        > .fa.o_clear_file_button {
+            background: @brand-danger;
+        }
+    }
 }
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -1,0 +1,65 @@
+/* Copyright 2016 Ponto Suprimentos Ltda.
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+.o_form_view {
+    > .o_form_sheet_bg > .o_form_sheet {
+         @media (max-width: @screen-xs-max) {
+             min-width: 0px; // allow thinner form_sheet width for smaller screens
+         }
+    }
+    @media (max-width: @screen-xs) {
+        .o_form_field > .o_form_input_dropdown {
+            width: 80%;
+        }
+    }
+    .o_group {
+        &.o_inner_group > tbody > tr > td {
+            // prevent td content from breaking into lines
+            white-space: nowrap;
+            .note-editor > .note-toolbar {
+                // prevent wysiwyg editor buttons from expanding the screen
+                white-space: initial;
+            }
+        }
+        // reduce form maximum columns for smaller screens
+        @media (max-width: @screen-xs-max) {
+            .o-generate-groups(12);
+            .o-generate-groups(@n, @i: 1) when (@i =< @n) {
+                .o_group_col_@{i} {
+                    width: 100%;
+                }
+                .o-generate-groups(@n, @i + 1);
+            }
+        }
+        // break field label into a separate line from field on small screens
+        @media (max-width: @screen-xs) {
+            &.o_inner_group {
+                display: block;
+                > tbody {
+                    display: block;
+                    > tr {
+                        margin-top: 8px;
+                        .o-flex-display();
+                        .o-flex-flow(row, wrap);
+                        > td {
+                            .o-flex(1, 0, auto);
+                            padding: 0;
+                            display: block;
+                            padding: 0;
+                            // odoo adds a `style="width: 100%"` by javascript
+                            // directly on the tag so we need `!important` here:
+                            width: auto!important;
+                            max-width: 100%;
+                            &.o_td_label {
+                                border-right: 0;
+                                // keep 6% space on line to fit checkboxes
+                                // see above about `!important`
+                                width: 94%!important;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -80,4 +80,9 @@
             background: @brand-danger;
         }
     }
+
+    // Adapt chatter widget to small viewports
+    .oe_chatter {
+        min-width: inherit;
+    }
 }
--- a/web_responsive/views/assets.xml
+++ b/web_responsive/views/assets.xml
@@ -23,6 +23,9 @@
                   href="/web_responsive/static/src/less/app_drawer.less"
                   />
             <link rel="stylesheet"
+                  href="/web_responsive/static/src/less/form_view.less"
+                  />
+            <link rel="stylesheet"
                   href="/web_responsive/static/src/less/variables.less"
                   />
             <script type="application/javascript"


danieltorres7:
What do you think @lasley @pedrobaeza @leorochael. Web responsive form views should adapt to better fit smaller screens.

danieltorres7:
Before:
===
captura de tela de 2017-01-02 15-49-54

After:
===
captura de tela de 2017-01-02 15-46-37

web_responsive/static/src/less/form_view.less
```diff
@@ -0,0 +1,49 @@
+.o_form_view {
+    > .o_form_sheet_bg > .o_form_sheet {
+        max-width: 1150px; // increased form_sheet width for larger screens
```
pedrobaeza: Will this collide with _web_sheet_full_width_ module?

danieltorres7: Don't know, will test...

danieltorres7:
Just added better screenshots above...

pedrobaeza:
Down arrows from right part are lost on mobile view:

seleccion_009

danieltorres7:
@pedrobaeza, just tested on last commit runbot: http://3276840-514-f788ea.runbot2.odoo-community.org/?db=3276840-514-f788ea-all and got these screenshots (both with down arrows):
captura de tela de 2017-01-02 16-08-09
captura de tela de 2017-01-02 16-07-54

danieltorres7:
@pedrobaeza, where did you see the arrows dissapearing? I'm trying to reproduce that...

pedrobaeza:
I'm using Firefox. That should be the problem. I'm also not seeing that cool buttons for editing or removing the image.

pedrobaeza:
OK, buttons appear at a very low resolution, but that's not useful I think.

web_responsive/static/src/less/form_view.less
danieltorres7: I've removed `max-width`, should be OK now...

danieltorres7:
> OK, buttons appear at a very low resolution, but that's not useful I think.

This is default Odoo behaviour, I'm not changing that...

pedrobaeza:
OK, I didn't know, but maybe we can change class for showing it at a higher resolution, as mobile views are greater than that I think.

danieltorres7:
> OK, I didn't know, but maybe we can change class for showing it at a higher resolution, as mobile views are greater than that I think.

OK, I'll consider that for another PR. 

BTW, I would prefer that this PR is merged only after rebasing onto #513

lasley:
> BTW, I would prefer that this PR is merged only after rebasing onto #513

To mark that a PR is dependent on another, put this in your original comment with a check list. Something like:

```
Depends:
- [ ] #513
```

Then once it's merged and you rebase, you can just check it off the list 🚀

leorochael:
> OK, I didn't know, but maybe we can change class for showing it at a higher resolution, as mobile views are greater than that I think.

@pedrobaeza: You're right about mobile devices having higher resolution, and the resolution will only get higher with time.

The problem is that it can be argued that what Odoo does here is wrong: it uses screen resolution as a proxy for the absence of a mouse, as it would require having a mouse to hover over the image before seeing the edit icons in high resolution.

In discussions of the design community ([like this](https://css-tricks.com/forums/topic/conditional-css-for-touch-screen-devices/)), the consensus seems to be that hover events should only ever be used to add optional embellishments to the design, but not be necessary for the functionality, no matter the resolution.

So I guess we could have `web_responsive` *fix* Odoo by making the buttons always be present.

But like @danieltorres7 said, this should go into another PR.

danieltorres7:
Since image editing controls are related to forms I just committed an addition to make them always visible.

pedrobaeza:
@Yajo, please review

yajo:
Seems a good IMP. I'll mark as WIP until the other PR is merged and then review 😉

leorochael:
@Yajo, the behaviour of this PR is not really affected by the presence of #513. It just looks prettier when the menus are also better adapted to mobile.

You can review it at your convenience, regardless of #513.

web_responsive/static/src/less/form_view.less
```diff
@@ -0,0 +1,66 @@
...
+                    > tr {
+                        margin-top: 8px;
+                        .o-flex-display();
+                        .o-flex-flow(row, wrap);
+                        > td {
+                            .o-flex(1, 0, auto);
+                            padding: 0;
+                            display: block;
+                            padding: 0;
+                            width: auto!important; // important is to override width fixed by element style on base
```
yajo: Please replace `!important`s for more concise CSS selectors.

web_responsive/static/src/less/form_view.less
```diff
@@ -0,0 +1,66 @@
...
+                            max-width: 100%;
+                            &.o_td_label {
+                                border-right: 0;
+                                width: 94%!important; // keep 6% space on line to fit checkboxes
```
yajo: Replace.

web_responsive/static/src/less/form_view.less
```diff
@@ -0,0 +1,66 @@
...
+        position: initial;
+        opacity: 1;
+        > .fa {
+            width: 50%;
+            padding: 6px;
+            margin: 0px;
+            text-align: center;
+        }
+        > .fa.o_select_file_button {
+            background: #7c7bad;
```
yajo: Use color variables from where you now know 😉

web_responsive/static/src/less/form_view.less
```diff
@@ -0,0 +1,66 @@
...
+        }
+        > .fa.o_clear_file_button {
+            background: #d9534f;
```
yajo: Color variables.

yajo (changes requested):
This seems much better than it was indeed. However it seems incomplete.

See, this is 458×575px window (FF):

captura el 2017-01-04 a las 10 02 23

Do you see the horizontal scroll bar? See what happens when I scroll to the right:

captura el 2017-01-04 a las 10 03 07

I think the problem is in the chatter widget. it gets wider than its parent container. Can you check that please?

danieltorres7:
> I think the problem is in the chatter widget. it gets wider than its parent container. Can you check that please?

Your right, #513 should fix that after merging.

web_responsive/static/src/less/form_view.less
danieltorres7: This import is to override the base template where td's have width: 100%!important. Precise selector can't do the job.

web_responsive/static/src/less/form_view.less
danieltorres7: This important is to override the previous important on the same less file.

web_responsive/static/src/less/form_view.less
danieltorres7: Changing...

web_responsive/static/src/less/form_view.less
danieltorres7: Changing...

mileo:
@lasley "Other than that though, general rule is one final commit per PR - unless multiple contributors or migrations" Is this recorded in any document written in the OCA? (sorry for the off topic)

lasley:
@mileo [Contribution Docs, Commit guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#git). "One logical changeset" typically works out to one per PR, but obviously can vary depending on context - such as multiple authors.

pedrobaeza:
What it still not reflected in guidelines is the agreement we made between several regular contributors for the reviewing process: don't squash commits until the PR is approved. This way, if you have reviewed full PR and make comments, you only need to review again the new commits to see if your comments have been attended. Once the review is done and you approve the PR, that new extra commits like "PEP8", "Requested changes" and so on must be squashed before merging. If all the commits come from the same person and it's only one logical changeset, this can be done using the GitHub feature "Squash and merge". If not, any contributor with write access on the repository can do it (if enabled "Allow to edit to maintainers" on the PR), or the author of the PR otherwise.

This can be included in the guidelines for being clearer if any volunteer.

moylop260:
The best comment about squash commit @pedrobaeza

I can help but my bad english make many typo feedback and you finish writing all again haha

What about create your PR to guidelines?

pedrobaeza:
OK, I'll do it sometime in the day.

web_responsive/static/src/less/form_view.less
yajo: Same case as https://github.com/OCA/web/pull/513#discussion_r94752243

leorochael: Like in #513, there's a `style="width: 100%"` attribute directly on the element (`<td>` in this case) I don't know how to make a selector more specific than the `style` attribute.

We haven't figured out yet where this `style` attribute on the `<td>` is coming from.

Does anyone have any idea?

leorochael: Found [the culprit](https://github.com/odoo/odoo/blob/3d085f632d4c5a8aece92731494e1dbbc8923049/addons/web/static/src/js/views/form_view.js#L1499):

     $td.css("width", (nonlabel_width*parseInt($td.attr('colspan')))+"%");

@Yajo what do you suggest?

danieltorres7:
@Yajo, requested changes and final fixes applied. We are ready for the final review.

pedrobaeza:
Dependency merged. Should you rebase?

web_responsive/static/src/less/form_view.less
yajo: Hmm a tricky one... What about this?

1. Make sure your solution does not break when there are colspans.
2. If it works fine, then I suggest you put an explanation above this line, seems like it would be hard to fix this upstream, and after all they're gonna open the responsive client for v11.

yajo (changes requested):
Rebase, I want to check if nothing breaks now that #513 is merged.

leorochael:
@danieltorres7 is till the end of the week. If he adds me as a developer to his fork I can rebase for him.

pedrobaeza:
I have rebased the branch.

@Yajo, OCA/web maintainers can edit PR branches.

@leorochael, what we do when working collaboratively is to push our PR branches to the team, so anyone in the team can modify them. It's only a suggestion, obviously :smile:

yajo (changes requested):
As you can see below in a screenshot of a partner form, there's still the `.oe_chatter` problem. It has a `min-width` that must be removed for it to not break the layout:

captura de pantalla de 2017-01-17 09-14-38

yajo (approved):

pedrobaeza:
Please squash the first 4 commits, Daniel.

danieltorres7:
Squashed last 4 commits into 2 commits by relevant changes on each.

Ready for merging in my opinion.

rafaelbn (approved):
Thanks!

PR #515, opened 2017-01-03
[9.0][FIX][web_export_view] ReferenceError: instance is not defined

yajo:
There was a bug that could be easily reproduced by:

1. Go to *Sales > Products*.
2. Choose list view.
3. Select all.
4. *Export current view > Excel*.

It raised this message to the user:

    ReferenceError: instance is not defined

This is all due to a wrong migration from v8 code. `instance` exists no more in
v9, instead we need to use some methods and variables from correct modules.

@Tecnativa

### DIFF
--- a/web_export_view/__openerp__.py
+++ b/web_export_view/__openerp__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'Export Current View',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'category': 'Web',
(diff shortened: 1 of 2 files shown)


rafaelbn:
@Yajo I think you should change module version. @simahawk could be interested in reviewing this fix

simahawk (approved):
not tested but LGTM

pedrobaeza (approved):

rafaelbn (approved):

PR #518, opened 2017-01-09
[9.0][FIX][web_widget_x2many_2d_matrix] Correctly bind events.

yajo:
Event binding was applied in the wrong places, causing `xy_value_change()` never to be executed.

@pedrobaeza This should fix point 2 of https://github.com/OCA/product-variant/pull/49#issuecomment-271107004.

@Tecnativa


hbrunn (approved):

pedrobaeza (approved):
Fast-tracking it

PR #522, opened 2017-01-12
[MIG] web_widget_float_formula: Migrate to 10.0

kellylougheed:


lasley (approved):
Thanks @kellylougheed 

Not sure why CLA bot is spazzing out, Oleg has been covered under LasLabs ECLA for a while. I think it actually meant to say that Kelly's wasn't in, and just got confused on rebase or something. We should have her's too now, under SMDrugstore ECLA.

lasley (changes requested):
Now that Runbot has rebuilt though, I can see that this doesn't work functionally. Please double check.

kellylougheed:
Ready for review @lasley

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -52,12 +53,7 @@ odoo.define('web_widget_float_formula', function(require) {
             },
 
             _process_formula: function(formula) {
-                try{
```
lasley: Why is this catch being removed? This was a bug fix for an edge case #431 abfea0fff77466e268dfad8dd007a45c3ffbc0c0

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -85,12 +81,13 @@ odoo.define('web_widget_float_formula', function(require) {
...
                 this._clean_formula_text();
 
-                var formula = this._process_formula(this.$el.find('input').val());
+                var formula = this._process_formula(this.$el.val());
                 if (formula !== false) {
                     var value = this._eval_formula(formula);
                     if (value !== false) {
                         this._formula_text = "=" + formula;
                         this.set_value(value);
+                        this.$el.val(value);
```
lasley: This is circumventing the widget value set, which will not play nicely with inherited modules that may want to do other things with the value. Please remove - I'm almost certain nothing will break.

web_widget_float_formula/static/tests/js/test_web_widget_float_formula.js
```diff
@@ -13,6 +13,7 @@ odoo.define_section('web_widget_float_formula', ['web.form_common', 'web.form_wi
         self.field = new form_widgets.FieldFloat(field_manager, filler);
         self.$element = $('<input>');
         self.field.$el.append(self.$element);
+        self.field.$label = $('<label>');
```
lasley: Why was this needed?

web_widget_float_formula/static/tests/js/test_web_widget_float_formula.js
```diff
@@ -116,10 +117,10 @@ odoo.define_section('web_widget_float_formula', ['web.form_common', 'web.form_wi
     test('._compute_result() should behave properly when the current value of the input element is a valid formula',
         function(assert, form_common, form_widgets, core) {
             window.test_setup(this, form_common, form_widgets, core);
-            this.$element.val('=2*3');
+            this.field.$el.val('=2*3');
```
lasley: `this.$element` and `this.field.$el` are functionally equivalent, but the former is more explicit. Please switch back.

The reasoning behind this is that `this.field.$el` is a jQuery collection (array basically) that contains `this.$element`. jQuery operates on collections as a package, so `.val` is just setting all elements in the collection (`[this.$element]`) to the value.

lasley (changes requested):
THanks @kellylougheed 

Looks like this doesn't work in products now though. Sales => Sales => Products => Any Product => Edit => Cost to `=2+3` => Save => Error on Cost

web_widget_float_formula/static/src/js/web_widget_float_formula.js
kellylougheed: Oops, I must have accidentally deleted that with my console logs. Putting it back.

web_widget_float_formula/static/tests/js/test_web_widget_float_formula.js
kellylougheed: For the last two tests (which tested the change on blur and focus), this function was called: https://github.com/odoo/odoo/blob/6466aa499fc37bba5f4feeeecac99448ef88e62e/addons/web/static/src/js/views/form_common.js#L700 but _toggle_label failed because there was no this.$label in the test, so I put one in.

kellylougheed:
Changes made @lasley

lasley (approved):

felixvillafranca (approved):
Great job ¡¡¡

PR #524, opened 2017-01-12
[9.0][FIX] No create options when user has no create rights on relational record

ghost:
When no create options was specified using field attributes or config parameter and when user has no create rights on relational record, create options were still available.

Removed get_search_result from many2manytags because this method is not used. Indeed, this widget uses many2one widget.

### DIFF
--- a/web_m2x_options/__openerp__.py
+++ b/web_m2x_options/__openerp__.py
@@ -2,7 +2,7 @@
 
 {
     "name": 'web_m2x_options',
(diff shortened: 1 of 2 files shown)


jesusVMayor:
Added on https://github.com/OCA/web/commit/e37db6412aa4ca85c8ad1635c49e9150a202f5af

yajo:
Yes, fixed in https://github.com/OCA/web/pull/643.

PR #527, opened 2017-01-14
[FIX] web_responsive: fix for the double scrollbar

pedrobaeza:
Scroll control panel with page using CSS instead of jQuery.

jQuery scrolling caused scrollbar malfunction.

Blindly backported from https://github.com/OCA/web/pull/511, but it should work I think.

### DIFF
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -25,3 +25,15 @@ main {
 div.oe_footer {
     display: none;
 }
+
+.o_web_client {
+    >.o_main {
+        overflow: auto;
+        > .o_main_content {
+            overflow: initial;
+            > .o_content {
+                overflow: initial;
+            }
+        }
+    }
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I still have the double scrollbar, so no luck with this patch:

seleccion_003

Any ideas, @lasley @Yajo @danieltorres7 @leorochael ?

lasley:
I was able to resolve this by changing the `html` tag's `height: 100%` to `height: 99%`. Seems like a ghetto hack though

image

pedrobaeza:
Well, if there's no better solution, I prefer that instead of having the double bar.

lasley:
Agreed. Looks like in v10 there is no static height, which is why we weren't experiencing the issue there

We want to just use this PR for the change?

pedrobaeza:
OK, feel free to propose the patch then :smile:

PR #528, opened 2017-01-14
[9.0][FIX] web_responsive: HTML Height

lasley:
* Update HTML height to 99% in order to fix double scrollbar issue

### DIFF
--- a/web_responsive/__openerp__.py
+++ b/web_responsive/__openerp__.py
@@ -18,5 +18,6 @@
     "data": [
         'views/assets.xml',
         'views/web.xml',
+        'views/web_layout.xml',
     ],
 }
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I'm afraid I still have the double scrollbar. You have to try in other view different from module list, as that one has its own layout (and inside Odoo they are even not sure why), but if you try for example in current runbot in the inventory app showing all the pickings and reducing a bit your browser window, you will find the problem.

lasley:
I've actually been using sale tree view, but I'm not able to replicate in inventory anymore either - even if I play around with browser size. You're a Firefox guy right?

pedrobaeza:
Yes, totally, I'm making the test with Firefox. Is it then a browser problem?

lasley:
Yeah in FF I can replicate it in basically every screen actually. In Chrome this was only happening on the list view & it definitely wasn't as pronounced. 

I can get it fixed in FF by reducing the HTML height to 96% through DOM inspector. Would you mind confirming on your end before I pursue some sort of JS something?

pedrobaeza:
Closing in favor of #529

PR #529, opened 2017-01-16
[9.0][web_responsive] Backport new features from 10.0

yajo:
This one should close #528.

It gets all new features from 10.0 branch to 9.0, including a fix for the double-scrollbar problem.

@Tecnativa.


pedrobaeza:
Yajo, this is not working as expected. Screenshot of our instance:

seleccion_005

yajo:
Why did you try in production instead of runbot? 🤔 

Well in any case it seems I had a little mess with git commits and browser cache. I marked as WIP for now, still waiting for #514 to be merged and backport that too.

pedrobaeza:
Because I'm anxious to have it, hehe!! OK, waiting your fix.

yajo:
Rebased, fixed, this should be working now. Includes #514.

pedrobaeza:
Improvements from #514 are not working:

seleccion_001

yajo:
This should be working now. It turns out #514 needed a complete refactor to make it work in v9.

Note to self: cherry-pick 84e0323 for v10 too.

Also, squashed by relevant steps.

pedrobaeza (approved):
Working perfectly, but version on README is not correct (it puts 10.0)

pedrobaeza:
For me, there are only 2 remaining things for having a complete backend responsive:

* Show kanban views with class `o_kanban_mobile` when available on XS devices.  For example: https://github.com/odoo/odoo/blob/3c9937388d422846c52600e3a1a99a24a7d2e757/addons/account/views/account_invoice_view.xml#L413. This should be easy, isn't it?
* Provide another many2one widget for selecting the option. Odoo, even with web_enterprise module, doesn't allow to select correctly for example the second row on the dropdown. I think this is harder to solve.

@Yajo, @lasley, what are your impressions about this?

rafaelbn (approved):
Thanks! In production

yajo:
I fixed the README :bug: and rebased to get it clean for merging.

About the other stuff...

> Show kanban views with class o_kanban_mobile when available on XS devices

It would be easy, although I'm not sure what's the purpose of that class, [it seems to be visible already](https://github.com/odoo/odoo/blob/9.0/addons/web_kanban/static/src/less/kanban_view.less#L198), so not sure what you mean. Maybe some screenshots could help.

> Provide another many2one widget for selecting the option. Odoo, even with web_enterprise module, doesn't allow to select correctly for example the second row on the dropdown. I think this is harder to solve.

You're right, it is harder. However I tested the m2o widget (for example selecting a country in the partner view) in Android Firefox and it works flawlessly. Maybe we could have bigger rows for nightstick-sized fingers users...

In any case, none of these seem a blocker for the PR IMHO.

lasley (approved):
Very nice! This version looks so much better in so many different ways ❤️ 

I agree that the outstanding points aren't necessarily blockers. Add to roadmap perhaps?

FYI I may write our search plugin here pretty soon given all this interest in `web_responsive`. I think that's still the one thing Enterprise UI has on us at this point.

pedrobaeza:
@Yajo, about the kanban mobile view, there should be any kind of JS switch or similar that reads the resolution and put this in the enterprise module:

seleccion_001

yajo:
But look here: https://github.com/odoo/enterprise/search?utf8=%E2%9C%93&q=o_kanban_mobile

pedrobaeza:
No, correct search is https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=o_kanban_mobile, and you shouldn't put here links to enterprise. What is done in module _web_enterprise_ is to arise that view instead of the normal one on mobiles (I don't know how because looking at the code will mean that we can't copy the solution in open code).

yajo:
What I mean is that there's no patch in enterprise that makes this appear or not, so it should work out of the box the same way in both versions.

pedrobaeza:
There should be a patch in web_enterprise module, which is the equivalent to our web_responsive one, because it doesn't happen on community. You can try if you want.

PR #530, opened 2017-01-18
[FIX] Bugfix in web_sheet_full_width for enterprise edition

BT-fgarbely:
#449

### DIFF
--- a/web_sheet_full_width/README.rst
+++ b/web_sheet_full_width/README.rst
@@ -9,6 +9,8 @@ Form view sheet at full width
 This module was written to extend the functionality of the web client
 to get full width in the form view sheet.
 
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Thanks for the PR.

I have tried on runbot and it's working also in community, and I suppose you have tried also that works in enterprise.

Two minor things that can be done:

* Change version in the manifest to reflect the update
* Include a comment in the README saying that it also supports _web_enterprise_ module.

bistaray:
I just tried this today with a brand new database (only this module installed):





You can see the sheet doesn't change size - it just sits aligned on the left.

bistaray:
WAIT.

I just realized I had the old version in .local and it wasn't seeing the new version.

It works!!

PR #531, opened 2017-01-19
[MIG][10.0] web_tree_many2one_clickable

SodexisTeam:
Minor changes only
cc: @pedrobaeza @antespi @Yajo

### DIFF
--- a/web_tree_many2one_clickable/__manifest__.py
+++ b/web_tree_many2one_clickable/__manifest__.py
@@ -1,38 +1,26 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright
-#      (C) 2013 Therp BV (<http://therp.nl>).
-#      (c) 2015 Serv. Tecnol. Avanzados (http://www.serviciosbaeza.com)
-#               Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


web_tree_many2one_clickable/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Please remove this file

atchuthan: @pedrobaeza removed the migration

pedrobaeza:
Better to remove the commit that add it (it's separated) and then the repo final size will be smaller

atchuthan:
@pedrobaeza removed the commit

web_tree_many2one_clickable/__manifest__.py
```diff
@@ -2,16 +2,18 @@
...
 {
     "name": "Clickable many2one fields for tree views",
     "summary": "Open the linked resource when clicking on their name",
-    "version": "9.0.1.0.0",
+    "version": "10.0.1.0.0",
     "category": "Hidden",
     "website": "https://odoo-community.org/",
     "author": "Therp BV, "
               "Tecnativa, "
+              "Sodexis, "
```
pedrobaeza: The changes you have made don't make that you should put it on the author list

pedrobaeza (changes requested):
Tried on runbot and working as expected, but the author thing is blocking IMO

web_tree_many2one_clickable/__manifest__.py
yajo: I don't think contributions should always be counted as number of LOC. Migrating an addon involves time (test it, fix it, send PR...). IMHO it's fair to let the contributors be honoured by these little credits.

yajo (approved):

web_tree_many2one_clickable/__manifest__.py
pedrobaeza: @Yajo, I repeat, it's not the case (see diff), and this is not contribution, it's authorship.

tarteo (approved):
Thanks!

pedrobaeza:
Please change authorship and I'll merge.

dvdhinesh:
@pedrobaeza 

@atchuthan is off for a week. He will continue to work on this by next week. Thanks.

atchuthan:
corrections done

pedrobaeza (approved):

PR #533, opened 2017-01-20
[FIX][web_responsive] Form padding, titles and buttons OK in XS

yajo:
Small makeup to make it look good in XS screens.

Before:

captura el 2017-01-20 a las 11 50 09

After:

captura el 2017-01-20 a las 11 50 23

@Tecnativa


pedrobaeza (approved):
Tried on runbot. This looks better and beter!

rafaelbn (approved):

rousseldenis (approved):
:+1: tested on runbot

PR #536, opened 2017-01-24
[9.0][IMP] web_responsive: Load default kanban view for small screen and One2many related fields.

sergio-teruel:
@Yajo please can you review??, For the moment I have made these changes, load default kanban view if exists and if the view has a One2many field and it has a kanban view, this view is showed.

cc @Tecnativa

### DIFF
--- a/web_responsive/README.rst
+++ b/web_responsive/README.rst
@@ -12,6 +12,8 @@ Features:
 
  * New navigation with an App drawer
  * Keyboard shortcuts for easier navigation
+ * Display kanban views for small screens format if an action or field One2x
+   have a kanban view
 
 
 Installation
@@ -52,6 +54,8 @@ this module.
 * Adding ``oe_main_menu_navbar`` ID to the top navigation bar triggers some
   great styles, but also `JavaScript that causes issues on mobile
   <https://github.com/OCA/web/pull/446#issuecomment-254827880>`_
+* The kanban view not is shown if an action window not has kanban mode view
+
 
 Bug Tracker
 ===========
@@ -75,6 +79,7 @@ Contributors
 
 * Dave Lasley <dave@laslabs.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
+* Sergio Teruel <sergio.teruel@tecnativa.com>
 
 Maintainer
 ----------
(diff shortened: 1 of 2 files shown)


web_responsive/static/src/js/web_responsive.js
```diff
@@ -292,10 +295,39 @@ odoo.define('web_responsive', function(require) {
...
+                default_view != 'kanban' &&
+                this.views['kanban']){
+                    default_view = 'kanban';
+            };
+            return default_view;
+        },
+    });
+
+    // if we are in small screen change One2many field view to kanban if exists
+    FieldOne2Many.include({
```
pedrobaeza: Is this really needed?

pedrobaeza (changes requested):
Working for one2many field:

seleccion_010

but not for tree view:

seleccion_011

pedrobaeza (approved):
OK, it's due to kanban view is not indicated on the action window browsing from sales team.

web_responsive/static/src/js/web_responsive.js
sergio-teruel: Yes... mhhu but i would like that @Yajo says your feeling ...

yajo: I think @pedrobaeza means that with the above code in `ViewManager` should work. Try to remove this chunk and see if it works. Otherwise, it's perfectly OK to have this here.

I'd thank you a lot if you can document this behavior in the addon's README, since it's somewhat hidden and unexpected otherwise.

yajo (changes requested):

sergio-teruel:
@pedrobaeza @Yajo you are rigth,  is not necessary rewrite the field, so i remove it and update the readme file.

web_responsive/README.rst
```diff
@@ -52,6 +54,8 @@ this module.
 * Adding ``oe_main_menu_navbar`` ID to the top navigation bar triggers some
   great styles, but also `JavaScript that causes issues on mobile
   <https://github.com/OCA/web/pull/446#issuecomment-254827880>`_
+* The kanban view not is shown if an action window not has kanban mode view
```
lasley: if an action window does not have a kanban mode view

web_responsive/README.rst
```diff
@@ -12,6 +12,8 @@ Features:
 
  * New navigation with an App drawer
  * Keyboard shortcuts for easier navigation
+ * Display kanban views for small screens format if an action or field One2x
+   have a kanban view
```
lasley: has a kanban

lasley (approved):
Few comments in the ReadMe. Functional looks good too, thanks for the improvement!

yajo (approved):

PR #537, opened 2017-01-25
[FIX][web_responsive] Remove hidden floating input over images

yajo:
Without this patch, when you clicked on a partner's name field next to its image, it asked you to change its image.

See attached video for the fixed behavior.
[Screencast_25-01-17_14:22:12.zip](https://github.com/OCA/web/files/729777/Screencast_25-01-17_14.22.12.zip)

@Tecnativa

### DIFF
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -82,6 +82,9 @@
         display: block;
         opacity: 0.7;
     }
+    .oe_hidden_input_file {
+        display: none;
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Is that class used in any other place?

yajo:
I hope not. I did not check too much because its name seems to imply it should be invisible.

pedrobaeza (approved):

lasley (approved):

PR #541, opened 2017-02-01
[FIX][web_responsive] Show upload control in binary widgets.

yajo:
We only want to hide it in image widgets, to avoid breaking binary widgets.

This is how they were before:

captura de pantalla de 2017-02-01 09-27-53

And after the fix:

captura de pantalla de 2017-02-01 09-28-28

@Tecnativa


pedrobaeza (approved):
Uhm, indentation also counts in CSS?

pedrobaeza:
Ah, no, it's inside the other bracket!

I fast-track this important bug.

yajo:
And it's Less, not CSS 🙊

PR #543, opened 2017-02-02
[9.0][FIX]web_m2x_options: s/ev.handleObj.selector/currentTarget

zakiuu:
The current behavior in the Many2many widget with `options={'open': true}`, when clicking on the selected elements, always the first element is the one opening. this done by `var id = parseInt($(ev.handleObj.selector).attr('data-id'));`.
With this fix the right id will be returned and the popup will open the right one.


zakiuu:
@sbidoul @lmignon

njeudy (approved):
Tested ok for me

lmignon (approved):
Tested on runbot.

pedrobaeza (approved):

PR #544, opened 2017-02-02
[10.0] [MIG] Migrate web_search_with_and from 8.0 to 10.0

Creamaster:
This module isn't ported in version 9.0 but is find it useful and propose to reintegrate it in 10.0 (as I didn’t find any alternative to combine search item with an “AND” operator natively)


pedrobaeza:
You should migrate JS code to new API and respect commit history. Follow migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0#technical-method-to-migrate-a-module-from-90-to-100-branch (change 9 by 8 when applicable)

web_search_with_and/static/src/js/inherit_search_view.js
```diff
@@ -0,0 +1,53 @@
+odoo.define('horanet_web_search_with_and', function (require) {
```
rousseldenis, line 1: Shouldn't it be company neutral ???

web_search_with_and/templates/web_assets_backend.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <template id="assets_backend" name="Horanet backend sources" inherit_id="web.assets_backend">
+            <!-- Backend javascript sources for altering the search method-->
+            <xpath expr="//script[last()]" position="after">
+                <script src="/horanet_web_search_with_and/static/src/js/inherit_search_view.js"
```
rousseldenis, line 7: Should be web_search_with_and (couldn't work like this - tested on runbot!) and company neutral in name

rousseldenis (changes requested):

web_search_with_and/static/src/js/inherit_search_view.js
Creamaster: You are absolutly right, I migrated this module some time ago (for my company) and today decide to make a PR for OCA, it's an omission form our custom module

web_search_with_and/templates/web_assets_backend.xml
Creamaster: OK, to recap and avoid wasting your time:
In order to respect commit history, I can migrate from the branch 8.0 even if the module isn’t present in 9.0
I remove this PR to create a new one
Sorry for the inconvenience

PR #545, opened 2017-02-02
[10.0][MIG] Migrate web_search_with_and from 8.0 to 10.0 

Creamaster:
This PR should replace #544


pedrobaeza:
Please sign CLA

rousseldenis (approved):
:+1: Code review + tested on runbot

Creamaster:
CLA signed

web_search_with_and/README.rst
```diff
@@ -0,0 +1,53 @@
...
+Usage
+=====
+
+* Enter your value in omnibar search field
+* Press and hold Shift key
+* Select field with mouse or keyboard to perform search on
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/8.0
```
lasley: s/8.0/10.0

web_search_with_and/README.rst
```diff
@@ -0,0 +1,53 @@
...
+
```
lasley: Missing bug tracker section: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

simahawk (approved):
code review: LGTM

paulius-sladkevicius:
:+1:

pedrobaeza:
Can you please squash all the initial commits from @andrius-preimantas? They are meaningless in the final branch.

lasley (changes requested):

web_search_with_and/static/src/js/search.js
```diff
@@ -0,0 +1,53 @@
...
+
+                this.query.add(ui.item.facet, {shiftKey: true});
+            } else {
+
+                return this._super.apply(this, arguments);
+            }
+        }
+    });
+
+    SearchView.SearchQuery.prototype = SearchView.SearchQuery.extend({
```
tarteo, line 25: Is it possible to change this to `SearchView.SearchQuery.include({});` or `SearchView.SearchQuery = SearchView.SearchQuery.extend({});` ?

web_search_with_and/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# © 2015 Andrius Preimantas <andrius@versada.lt>
```
tarteo: Can you change the copyright character to 'Copyright' to comply the template, please? https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

tarteo (changes requested):

web_search_with_and/static/src/js/search.js
Creamaster: I don't think so, this part was the real trick to migrate the module in v10.
I don't remember exactly the code but it was something like that : 
-  native v10 framework : method1 inherit method2
-  my goal : method1 inherit method_custom inherit method2 (place my code between existing inheritance)

This is how I usually deal with this problem, but if you have a better way to do it, please let me know :)

web_search_with_and/__manifest__.py
Creamaster: done

web_search_with_and/README.rst
Creamaster: done

web_search_with_and/README.rst
Creamaster: done

web_search_with_and/static/src/js/search.js
lasley: Hmmmmm I'm not so sure about this. Are you sure you're not going to `include` here instead? AFAIK you're emulating that behavior by assigning to the prototype.

Creamaster: To be honest, before using prototype, I tried 

1. `SearchView.SearchQuery.include({}); `
2. `SearchView.SearchQuery = SearchView.SearchQuery.extend({});`

But it doesn't work, for I need to place the code (in the inheritance stack) before the extend in odoo/web/static/src/js/views/search_view.js, extract : 

```
var SearchQuery = Backbone.Collection.extend({
    ...
    add: function (values, options) {
        ...
        _(values).each(function (value) {
            ...
            Backbone.Collection.prototype.add.call(this, model, options);

...
_.extend(SearchView, {
    SearchQuery: SearchQuery,
    Facet: Facet,
});

return SearchView;
```
Not sure if I make sense here, but it was crystal clear when I implemented it

lasley: Alright sounds like a good enough reason for me, thanks for the elaboration @Creamaster.

Because this is non-standard, can you add a docblock indicating the why so that future module maintainers don't have to solve the same problem repeatedly?

simahawk (changes requested):
LGTM, I agree w/ Dave on dropping some doc lines to explain the JS thing.

web_search_with_and/static/src/js/search.js
Creamaster: I added some lines, but I will value your feedback, see if it's helpful

lasley (approved):
Excellent, thanks @Creamaster

PR #554, opened 2017-02-08
[9.0][FIX]web_m2x_options: prevent popup to open web badger is deleted

zakiuu:
The **current behavior** is when you try deleting an element from a `many2many_tag` that has `options={'open': true}` the popup still show up even the element is deleted.
The `$(ev.currentTarget).data('id')` always returns a valid ID, hence the action is executed.
With this fix I've used the `($(ev.target).data('id')` instead and checked wether I've a value in `id` or not.
So the expected bahavior now: when deleting the tag the popup should no show up any more.


zakiuu:
@sbidoul @lmignon

PR #555, opened 2017-02-15
[9.0][MIG] web_advanced_search_x2x module

cubells:
- Updated README
- Updated views
- Updated js

@Tecnativa


web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
```diff
@@ -180,7 +187,7 @@ openerp.web_advanced_search_x2x = function(instance)
         popup_domain_selection: function()
         {
             var self = this,
-                popup = new instance.web_advanced_search_x2x.SelectCreatePopup(this);
+                popup = new form_common.SelectCreatePopup(this);
```
hbrunn: I think you lose the "use criteria" function this way, which is one of the most powerful features of this module imho

hbrunn (changes requested):

yajo:
This is finally fixed & working. We now make use of the brand new domain field widget.

pedrobaeza:
There is an error making this sequence:

* Go to Invoicing/Accounting > Customer Invoices
* Open "Add custom filter"
* Select "Partner"
* Select "Is in selection"
```
TypeError: this.getParent(...) is undefined

http://3280294-555-fd4370.runbot1.odoo-community.org/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js:119
Traceback:
X2XAdvancedSearchPropositionMixin.operator_changed@http://3280294-555-fd4370.runbot1.odoo-community.org/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js:119:13
Widget<.proxy/<@http://3280294-555-fd4370.runbot1.odoo-community.org/web/static/src/js/framework/widget.js:398:20
jQuery.event.dispatch@http://3280294-555-fd4370.runbot1.odoo-community.org/web/static/lib/jquery/jquery.js:4640:50
jQuery.event.add/elemData.handle@http://3280294-555-fd4370.runbot1.odoo-community.org/web/static/lib/jquery/jquery.js:4309:41
```

pedrobaeza:
It's very uncomfortable to have the dropdown hidden when you click directly on the result for "Is equal to"/"Is not equal to". Isn't there any workaround?

hbrunn (approved):
nice. There's still Pedro's issue, and for perfection, you could style the selection widget to match the other two widgets above, but that's a minor issue

yajo:
Thanks for your reviews. I spent the whole morning trying to remove that item from the known issues, but no way.

If anyone wants to help, you're welcome, otherwise I think it's easy enough to use the keyboard and up/down keys.

The exception is fixed.

pedrobaeza:
@hbrunn maybe you can take a look and give us a hand as you are the Odoo JS UI master.

nicomacr (approved):

hbrunn:
@pedrobaeza at your service: https://github.com/Tecnativa/web/pull/1
@Yajo Chromium's DOM breakpoints are very helpful to debug this kind of stuff. I set one on the `ul` that is the dropdown, with this I found https://github.com/OCA/OCB/blob/9.0/addons/web/static/lib/bootstrap/js/dropdown.js#L154 to be the culprit because it simply closes all open dropdowns. Then a few WTFs later after understanding how upstream uses this, this is the best I managed. A bit convoluted, but i wouldn't know how to do this in a more elegant way.

If you guys want to return the favor, you can review https://github.com/OCA/server-tools/pull/738, I think this has a lot of promise

hbrunn:
when I cloned this, it still contained the link to the stylesheet, this is missing in this branch. Can you add that back? Then the one2many combobox looks more like the others

pedrobaeza:
Oh, I was putting a large explanation about some problems that maybe are solved with this. Let me add it and re-check.

pedrobaeza:
Thanks again for the patch, Holger. My comments still applies even with the CSS.

Trying the module in runbot, I find these little problems:

* When using the domain widget for getting the "Is in selection", there are several visual problems applying custom filters (but I'm not sure if it's due to this module):
  seleccion_001
  * Trash button is not in the correct place.
  * Apply button is hidden. I work-around this clicking on Filters button to hide the dropdown, unfolding again it, and clicking on 
* When you put the criteria in the "is in selection" popup, you need to select all the records clicking on the header check and click on "Select". The previous button "Use criteria" was more intuitive and less error-prone.
* With the new domain widget, now filters appears with the technical domain tuple, which is excellent for developers, but not so good for users. Previous text "Customer is in selection "Country is China"" was better for them. Is there any possibility of having this text on normal mode and the other on developer mode?
  seleccion_002
* And this one is the worst: when I select the criteria for the in selection, automatically the criteria is applied in the upper window. This is good for most cases, as it prevents you to make another click on "Apply" that also provokes some users questions about that. However, this is preventing the possibility to apply "OR" filters with advanced searches. Maybe having 2 buttons (the one mentioned before "Use criteria" or similar and another that is "Add criteria and continue filtering" or similar.

yajo:
> When using the domain widget for getting the "Is in selection", there are several visual problems applying custom filters (but I'm not sure if it's due to this module):

Wrapping the element in a `div.oe_form` was enough, not sure why the CSS was required again. I'd have to check that.

> When you put the criteria in the "is in selection" popup, you need to select all the records clicking on the header check and click on "Select". The previous button "Use criteria" was more intuitive and less error-prone.

IMHO it's better that users get used to the new style because it's the standard domain widget for Odoo. End users after all don't care about "domains" (nor should they). They simply click all and select all, or click several and select several. For more details they can still read the docs, so this seems good enough.

If you still want this difference, I think the place to fix that would be Odoo itself.

> With the new domain widget, now filters appears with the technical domain tuple, which is excellent for developers, but not so good for users. Previous text "Customer is in selection "Country is China"" was better for them.

Not sure if we could use the get_label from the domain selection window. If so, that would be easy, otherwise a maintenance hell possibly.

> Is there any possibility of having this text on normal mode and the other on developer mode?

Possibility, yes, but worth the effort? Developers can inspect the request parameters and get the exact domain without many problems...

> this is preventing the possibility to apply "OR" filters with advanced searches.

Maybe you could add the OR to the domain window itself.

> Maybe having 2 buttons (the one mentioned before "Use criteria" or similar and another that is "Add criteria and continue filtering" or similar.

The main reason for this was actually not losing so much more hours in trying to circumvent the dropdown disappearing problem that @hbrunn managed to fix with the many2one filter.

yajo:
Hi, after a little further research, there's a problem to fix this:

> when I select the criteria for the in selection, automatically the criteria is applied in the upper window.

[The dialog is open in a sandboxed method that exposes nothing](https://github.com/odoo/odoo/blob/9.0/addons/web/static/src/js/views/form_widgets.js#L487), so there's no access to it.

I'm wondering if a simple `many2many_tags` widget would fill all the gaps and be much simpler to implement, use and maintain... 🤔 Just one widget for all cases.

BTW @hbrunn, I took the freedom to squash your commits in one and remove all the unneeded things (mostly CSS, that I added latter in another commit), just to keep all thin & useful. I hope you don't mind.

pedrobaeza (approved):
OK for me now.

yajo:
I opened https://github.com/odoo/odoo/pull/15579, if they merge we'd be able to fix what I said above.

rafaelbn:
I've tested and it works but some usability points must be fixed in future as commented about.

Only one question: is it possible to have button "select criteria"? see screenshot:

rafaelbn (approved):
Functionally tested

yajo:
Hi @rafaelbn, as we explained above, that feature comes from a custom widget that existed in previous versions. We are switching to the upstream char_domain widget here, and that's its behavior. We are plannin further updates on this, like backporting the widget from master, which improved immensely. But for now there's no chance to add that button. However, if you click on all registries (top-left checkbox), the result is the same.

luismontalba (approved):
Tested on runbot

PR #557, opened 2017-02-15
[10.0] [MIG] web_duplicate_visibility

lmignon:


lmignon:
I've spent a lot of time trying to fix these JS tests but without success... On the functional point of view the module works as expected.
With this commit all tests are OK on my machine https://github.com/acsone/web/commit/ab75a3ee9ed68a1917a87b5b30c65af804a9c7f5

simahawk:
@lmignon dumb question probably: the fix you applied on ``10.0-mig-web_duplicate_visibility-lmi-bis`` can't be included in this PR to fix tests (which is on ``10.0-mig-web_duplicate_visibility-lmi``)?

simahawk:
@lmignon ping :)

lmignon:
@simahawk I'll include the commit from 10.0-mig-web_duplicate_visibility-lmi into this one. But I'm not sure that everything will be ok after//

lmignon:
@simahawk done and rebased on 10.0

Fenkiou:
@lmignon the browser cache (i.e odoo caching policy) is making tests fail..

Just tried on Firefox, with console open:
- without cache -> tests pass
- with cache -> tests fail

Hope it will help to get this merged. :)

Fenkiou:
Hey guys, did you knew that odoo already had this implemented?

My colleague searched in odoo source code and found [this](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/ir/ir_model_view.xml#L224).

andreparames:
Yes, it seems they added it (after this PR was made) just to solve a specific problem, but should work generally: https://github.com/odoo/odoo/commit/b6a08dc982d3f27006f62b6b2ba2e75547b63e75

PR #559, opened 2017-02-17
[PORT] 9.0 web_listview_custom_element_number

legalsylvain:
* port to V9.0, web_listview_custom_element_number that allow users to set a custom quantity of item to display, in any tree view
* simplify header files
* apply OCA conventions (readme files)
* update screenshots
* improve UI (if user select a predefined item, the tree view will refresh, without need to press return).

For more information, see readme file.

CC

* original reviewers in V7.0 : @pedrobaeza, @OSguard 
* @florian-dacosta


florian-dacosta:
Functional test : it works like a charm!
Thank you Sylvain :+1:

pedrobaeza:
Have you checked if there's new translations in the 8.0 branch with the method from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#technical-method-to-migrate-a-module-from-80-to-90-branch?

legalsylvain:
Hi @pedrobaeza. Thanks for your review. Indeed, i did not follow this process. Thanks for pointing that ! Is it a good practice to squach all commits I recovered ? Or is it better to keep it, for historical reasons ?

Thanks.

pedrobaeza:
No reason for keeping the exact commit history, as it's already in the previous branch. You have to group by author and logic set (if there are pieces very decoupled that deserves to be kept separately). In this case, you can squash all the commits, as they are from Transbot or a fix that Stéphane applied also to translations. Follow https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests for the Howto.

legalsylvain:
@pedrobaeza. done. thanks for this advice.

pedrobaeza (approved):

PR #560, opened 2017-02-20
[FIX][web_responsive] Fit view to screen.

yajo:
This patch makes the `.oe_web_client` element to fit exactly inside the viewport, instead of overflowing it as before.

Before:
 captura el 2017-02-20 a las 13 43 31 

After: 
captura el 2017-02-20 a las 13 44 46

@Tecnativa

### DIFF
--- a/web_responsive/README.rst
+++ b/web_responsive/README.rst
@@ -10,11 +10,10 @@ This module provides a mobile compliant interface for Odoo Community web.
 
 Features:
 
- * New navigation with an App drawer
- * Keyboard shortcuts for easier navigation
- * Display kanban views for small screens format if an action or field One2x
-   has a kanban view
-
+* New navigation with an App drawer
+* Keyboard shortcuts for easier navigation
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
I confirm it works

lasley (approved):
Wow this `display: flex` is some serious voodoo. I think I have some reading to do!

Can also confirm this fixes it

pedrobaeza:
Is this also happening on v10?

lasley:
Not that I can see (Chrome & FireFox). 

This same code is in v10 though, so I would say that it would be good to unify as long as we don't see degradation.

pedrobaeza:
Merging as it has been confirmed by other customer that works correctly.

yajo:
Indeed [flex](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) is awesome! And, best of all, [mainstream](http://caniuse.com/#feat=flexbox).

PR #561, opened 2017-02-21
[FIX][web_m2x_options] Fix Qweb templates.

yajo:
They were trying to replace non-existing elements, and this was being logged to console:

    Can't find "a.oe_m2o_cm_button" when extending template FieldMany2One
    Can't find "span.badge" when extending template FieldMany2ManyTag

@Tecnativa

### DIFF
--- a/web_m2x_options/__openerp__.py
+++ b/web_m2x_options/__openerp__.py
@@ -2,7 +2,7 @@
 
 {
     "name": 'web_m2x_options',
-    "version": "9.0.1.0.1",
+    "version": "9.0.1.1.0",
     "depends": [
         'base',
         'web',
--- a/web_m2x_options/__openerp__.py
+++ b/web_m2x_options/__openerp__.py
@@ -2,7 +2,7 @@
 
 {
     "name": 'web_m2x_options',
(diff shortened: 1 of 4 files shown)


web_m2x_options/__openerp__.py
```diff
@@ -2,7 +2,7 @@
 
 {
     "name": 'web_m2x_options',
-    "version": "9.0.1.0.0",
+    "version": "9.0.1.0.1",
```
pedrobaeza: Don't you need to update?

yajo: Update what? The addon?

pedrobaeza: Yes, I mean that if it's the case, you need to put 9.0.1.1.0

yajo:
It turns out that making the addon "work" makes `many2many_tags` widget fail. Gonna fix.

yajo:
This should be OK now. Tested it works in production.

pedrobaeza (approved):

njeudy (approved):
- test and code review

rafaelbn (approved):
thanks

PR #576, opened 2017-03-09
[FIX][web_responsive] Make calendar invitations work again

yajo:
This patch displays the event acceptance template to the end user without failing with a server error.

Before this patch:

1. Create a calendar event.
2. Invite Mr. Foo.
3. Enter Mr. Foo's mail inbox.
4. Press "Accept" in the invitation email.
5. Get a beautiful `Internal Server Error`.

After this patch, replace step 5 by "See the accepted event details".

This is the **real** fix for https://github.com/odoo/odoo/pull/15762.
@Tecnativa

### DIFF
--- a/web_responsive/views/web.xml
+++ b/web_responsive/views/web.xml
@@ -27,14 +27,13 @@
 
             <t t-set="body_classname" t-value="'drawer drawer--left o_web_client'" />
 
-            <header role="banner">
+            <header role="banner" groups="base.group_user,base.group_portal">
                 <nav id="odooAppDrawer" class="app-drawer-nav drawer-nav" role="navigation">
                     <t t-call="web.menu" />
                 </nav>
 
                 <nav class="navbar navbar-default main-nav"
                      role="navigation"
-                     groups="base.group_user,base.group_portal"
                      >
                     <div class="container-fluid">
 


pedrobaeza:
Is this reproducible in runbot? Maybe you should indicate if you have to logout or access via incognito window in any step.

rafaelbn:
Before this patch:

1. Create a calendar event as an Odoo user (admin for example)
2. Invite Mr. Foo. which is not a Odoo user
3. Enter Mr. Foo's mail inbox.
4. Press "Accept" in the invitation email. As is not a Odoo user Mr. Foo. is not logged
5. Get a beautiful Internal Server Error.

After this patch, replace step 5 by "See the accepted event details"

cubells (approved):
:+1: 
LGTM

sergio-teruel (approved):
:+1:

rafaelbn (approved):
Quite difficult issue. Take care @lasley , nobody know what could be broken changing totally the web interface ;-)
👍

lasley:
@Yajo thanks for the fix!

@rafaelbn Hahah yeah I figured there would be a massive amount of bugs with this module. TBH it's been a lot less than expected, so at least that's a plus 😆 🚀

yajo:
@lasley maybe you could take a look at v10 to see if it happens there too, you're more into v10 than me for now 😅

lasley:
@Yajo - Good call, we're converting our prod right now - I'll install this and find out.

PR #578, opened 2017-03-13
[8.0][ADD][web_x2many_add_button_position] Move Add an item button

tarteo:
This module moves the _Add an item_ button to the top of the list if attribute _editable_ is equal to _top_


pedrobaeza:
Why not try to make a PR to Odoo for accepting this in core?

tarteo:
@pedrobaeza Maybe have both? For companies with an older revision of Odoo 8.0 and don't want to update the base?

jvOrsouw:
I really like this module. Really want it to be available in Odoo8 to prevent a lot of scrolling.

@tarteo 👍

tarteo:
@pedrobaeza I tried to make a PR to Odoo. but they wouldn't accept it because 8.0 is already stable and end of life. https://github.com/odoo/odoo/pull/15847

pedrobaeza:
OK, we have tried...

We can indeed go for both options now: try to merge this on master, and create this module.

tarteo:
@pedrobaeza I already created the module it's for 8.0 (this PR)... When I've spare time I can make it for version 10 including the module for 10.

tarteo:
@jvOrsouw Please use the GitHub review system.

jvOrsouw (approved):
Really nice functionality, prevents a lot of scrolling.

👍

web_x2many_add_button_position/templates/assets.xml
```diff
@@ -0,0 +1,13 @@
...
+
+<openerp>
+    <data>
+        <template id="assets_backend" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/web_x2many_add_button_position/static/src/js/web_x2many_add_button_position.js"></script>
+            </xpath>
+        </template>
+    </data>
+</openerp>
```
pedrobaeza: Line at the end.

tarteo: @pedrobaeza Thanks Pedro. Fixed it :)

pedrobaeza (approved):

Issue #586, opened 2017-03-21
[wish] web_environment_ribbon display database name by default


zeroheure:
This is a very little wish: 
It would be nice if web_environment_ribbon could display database name by default (instead of "TEST")

lukebranch:
this would be pretty handy actually. I have been doing this manually myself.

moylop260:
Database name could be too much big for the ribbon
Do you prefer another option?
Maybe a environment variable

lukebranch:
you're right. yes, I think an environment variable could be handy. Running runbots for customers, this could be really useful, as they tend to get a bit lost sometimes without the ribbon telling them which DB they are in.

moylop260:
In fact, IMHO we should install this module by default for all builds of runbot from MQT

pedrobaeza:
Closing as there's now a PR.

PR #588, opened 2017-03-22
add session name in web_environment_ribbon

zeroheure:
See issue (wish) #586

### DIFF
--- a/web_environment_ribbon/models/__init__.py
+++ b/web_environment_ribbon/models/__init__.py
@@ -1,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2017 Xavier Brochard
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+
+from . import ribbon
--- a/web_environment_ribbon/models/ribbon.py
+++ b/web_environment_ribbon/models/ribbon.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2017 Xavier Brochard
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+
+import os
+from odoo import models
+
+# brouillon OCA/server-tools/blob/10.0/auto_backup/models/db_backup.py
+class DbName(models.Model):
+    folder = fields.Char(
+        default=lambda self: self._default_name(),
+        help='Text printed on the Ribbon',
+        required=True
+    )
+    @api.model
+    def _default_name(self):
+        """Default to database name."""
+        return self.env.cr.dbname
--- a/web_environment_ribbon/models/ribbon.py
+++ b/web_environment_ribbon/models/ribbon.py
@@ -7,6 +7,7 @@ from odoo import models
 
 # brouillon OCA/server-tools/blob/10.0/auto_backup/models/db_backup.py
 class DbName(models.Model):
+    
     ribbon.name = fields.Char(
         default=lambda self: self._default_name(),
         help='Text printed on the Ribbon',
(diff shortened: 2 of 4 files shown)


pedrobaeza:
You shouldn't change ribbon size.  And the method should be to put that name by default (in the system parameter), not to have it always in screen.

zeroheure:
Text is 2 lines now. If I don't change the ribbon size, text doesn't fit in or ribbon's upper border is visible. Admitedly I'have increased the size a lot, but database name can be very long (see the form size in database manager).
As discussed in #586 displaying database name is useful.

pedrobaeza:
Please read again my comment.

zeroheure:
I think I don't understand. I'm sorry. And the fact that my server is curently down doesn't help me.
Do you mean that db name should be in the parameter form in the ribbon.name field to let user choose what he want?

lasley (commented):
@zeroheure you can use Runbot to functionally review your code in the event of your server being down. 

I think one of Pedro's main points is that increasing the side of the ribbon makes it overlap with other elements in the UI - particularly with web_responsive installed:

image

lasley:
Oh actually nope I just double checked & the banner is exactly the same on prod. Sorry I don't use this plugin, just trying to help 😛

zeroheure:
@lasley thanks anyway

pedrobaeza:
I am saying that you should have the freedom to choose whatever ribbon text, but you can put as default text for this ribbon (instead of TEST), the DB name. This can be done in a post_init_hook after the installation of the module in the DB. From there, you can change it for what you want.

zeroheure:
Thanks for the explanation. You're right about the freedom. 
But I'm not a dev, I currently don't know how to do that. I will try tomorrow. Anyway I think  Ribbon is better even if DB name is always on the screen.

jcdrubay:
It's a very nice and useful module :)

Here are some ideas I wished to share for a while. Quite often, we need to restore a production database to local environment for debug. Therefore, I think it would be great not to rely on the system parameter which is stored in database but instead to rely on new configuration in the configuration file of Odoo which I believe is specific per environment: (https://github.com/odoo/odoo/blob/10.0/odoo/tools/config.py).

Then in the config file we could specify on of these 3 options none OR database OR test:
 - none: don't display the ribbon, this is for production environment
 - database: display the database name
 - anything else: display as a text, if none specified but the module is installed then display based on the default value (TEST or we could then use the config_parameter)

sbidoul:
@jcdrubay to rely on the configuration file you can use https://github.com/OCA/server-tools/tree/9.0/server_environment_ir_config_parameter.

This module could detect a string such as '{db}' in the config parameter to replace it with the dbname.

sbidoul:
FYI, v10 migration of server_environment_ir_config_parameter in https://github.com/OCA/server-tools/pull/640

sbidoul:
A better solution for having the session in the ribbon and make it configurable would be to make a backend model with a `get_environment_ribbon()` method, so it is easy to override.

The default implementation of the method could apply `format(db=...)` to the string obtained from the system parameters.

ThomasBinsfeld:
I'll propose a solution in a new PR, based on @sbidoul idea (with a backend).

EDIT: #621

sbidoul:
@zeroheure would you like checking #621, which provides a more extensible approach?

zeroheure:
@sbidoul  it's fine for me, it's a better solution. I can not approve the changes as I'm not a developper.

sbidoul:
@zeroheure thanks for the feedback. I close this PR so the review can continue in #621

PR #591, opened 2017-03-28
[FIX] web_responsive: Change edit accesskey

### LINKED ISSUE #587: [10.0] web_responsive: accesskey conflict
On v10, global acceskey for editing a record is 'A' (Ctrl+Shift+A usually in most browsers), but that's the one we are using for the AppDrawer. So we have 2 options:

1) Change our AppDrawer accesskey to other
2) Change standard Odoo edit accesskey back to 'E', which I don't know why really they have changed it without any meaning.

I would be in favor of 2.

lasley:
* Change accesskey for `edit` in form view back to `e` to fix #587 

Disclaimer - I haven't actually tested this yet. I'll update tag with a Ready for Review once I have (after Runbot build)

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2016 LasLabs Inc.
+# Copyright 2016-2017 LasLabs Inc.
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Working as expected

pedrobaeza:
I think we can fast-track this for having this repaired.

PR #592, opened 2017-03-29
[FIX][web_export_view] Remove monetary formatting

### LINKED ISSUE #558: web_export_view - null values for Monetary fields
When a field is defined as 'fields.Monetary' and currency symbol is placed before the value (i.e. '$ 55.67', '£ 11.23') then resulting value parsed from QWeb (web_export_view/static/src/js/web_export_view.js:on_sidebar_export_treeview_xls) to controller is null.

For the values where currency symbol is after value (i.e. '0.55€') is parsed and exported correctly.

yajo:
Monetary fields were hard to work with in Excel after exporting because having the currency symbol made Excel treat those cells as text.

This patch removes those symbols and thus allows Excel to treat those values as numbers, allowing easier management.

It works no matter the symbol is before or after the value, so in any case it fixes #558.

Took the chance to make this code chunk a little bit easier with jQuery.

@Tecnativa @DarkChamp

### DIFF
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -73,23 +73,24 @@ Sidebar.include({
             var checked = $row.find(row_selector).is(':checked');
             if (children && checked === true) {
                 $.each(export_columns_keys, function () {
-                    var cell = $row.find('td[data-field="' + this + '"]').get(0);
-                    var text = cell.text || cell.textContent || cell.innerHTML || "";
-                    if (cell.classList.contains("oe_list_field_float")) {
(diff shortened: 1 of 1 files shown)


simahawk (approved):

pedrobaeza (approved):
Tried on runbot. It still don't put as a monetary field (it's only as number), but it's enough for most cases.

pedrobaeza:
@Yajo v10 requires other solution as the code has differed. Please check it make also the PR for that version.

yajo:
v10 patch is #594.
BTW I directly added https://github.com/OCA/web/commit/551715b791186929710396ed326a9a96d9398a3a to fix a bug I saw while migrating patch.

PR #595, opened 2017-04-01
[10.0][MIG] web_widget_darkroom: Migrated to 10.0

kellylougheed:
Work in progress


lasley:
@obulkin please review

web_widget_darkroom/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016-2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Web DarkroomJS Image Editing',
+    'summary': 'Provides web widget for image editing and adds it to standard'
+               ' image widget as modal',
+    'version': '10.0.1.0.1',
```
lasley: 10.0.0.0

web_widget_darkroom/static/lib/darkroomjs/core/darkroom.js
```diff
@@ -0,0 +1,356 @@
+/**
+*    Copyright 2013 Matthieu Moquet
+*    Copyright 2016-2017 LasLabs Inc.
```
lasley, line 3: We don't hold a copyright to this

web_widget_darkroom/static/lib/darkroomjs/core/darkroom.js
```diff
@@ -0,0 +1,356 @@
...
+*    License MIT (https://opensource.org/licenses/MIT)
+**/
```
lasley: Hmmm I wonder what version darkroom this is? We should update then add the version number into the header

lasley (changes requested):

web_widget_darkroom/__manifest__.py
elicoidal: I think usually it is 10.0.1.0.0

web_widget_darkroom/views/assets.xml
```diff
@@ -0,0 +1,28 @@
...
+<!--
+    Copyright 2016-2017 LasLabs Inc.
+    License LGPL-3 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+
+<odoo>
+    <template id="assets_darkroom" name="web_widget_darkroom Assets" inherit_id="web.assets_backend">
+        <xpath expr="//script[last()]" position="after">
+            <link href="/web_widget_darkroom/static/src/less/darkroom.less" rel="stylesheet" type="text/less"/>
+            <script src="https://cdnjs.cloudflare.com/ajax/libs/fabric.js/1.5.0/fabric.require.min.js"/>
```
lasley: Bring [this](https://github.com/kangax/fabric.js/blob/v1.7.15/dist/fabric.js) into `static/lib/js/fabric.js` and source from there instead.

web_widget_darkroom/wizards/darkroom_modal.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from openerp import api, fields, models
+from openerp.exceptions import MissingError
```
obulkin: These imports need to use ``odoo``.

web_widget_darkroom/static/src/js/widget_darkroom.js
```diff
@@ -0,0 +1,224 @@
...
+**/
+
+odoo.define('web_widget_darkroom.darkroom_widget', function(require) {
+    'use strict';
+
+    var core = require('web.core');
+    var common = require('web.form_common');
+    var session = require('web.session');
+    var utils = require('web.utils');
+    var _ = require('_');
```
obulkin: This line can be dropped. It doesn't work in v10, and `_` should just be available as a global.

obulkin (changes requested):
There are a couple missing migration changes. I did not review the v9 code as I was the last person to work on it. I also could not perform a functional review since the widget won't load in v10 without the Fabric.js change requested [here](https://github.com/OCA/web/pull/595#pullrequestreview-56713400).

web_widget_darkroom/static/lib/darkroomjs/core/darkroom.js
obulkin: I added the LasLabs copyrights as I wound up having to make a lot of minor changes in all the DarkroomJS files to satisfy our linter.

web_widget_darkroom/static/lib/darkroomjs/core/darkroom.js
obulkin: The version is listed in the README, but I don't know if it makes sense to include the version number here as the code has been modified to satisfy our linter and wouldn't exactly match the original source for this version number.

lasley: Yeah it's still best to have it here - it's relevant in the context of the code to know what version you're dealing with.

Also hmmm I wonder if we should ignore `static/lib` from our linter. We don't really want to have to modify libraries like this.

moylop260: @lasley FYI our linter is ignoring `static/lib` folder, you can see that we are not emitting message from this folder:
https://travis-ci.org/OCA/web/jobs/264532412#L638

obulkin: Nice. I could swear this wasn't the case when I worked on this module in v9, but I was linting everything locally, so maybe I just missed this exception.

lasley: Thanks for the confirmation @moylop260!

hughesbm:
I have made the requested/suggested changes. I also updated some css classes that apparently changed in v10.

obulkin:
My previous comments have been addressed, but when I test the widget on Runbot, I'm running into a weird cropping bug. The grey background layer created during cropping is moving together with the cropping reticle rather than staying put:

image

I can't seem to recreate this in v9 or in the sample on the DarkroomJs page, so I think it's an issue specific to this branch.

lasley:
It seems that the desync is being caused in the click event. Clicking again synchronizes it (at least in my case)

image

hughesbm:
The cropping bug was caused by a change to Fabric.js in version 1.7.0 that is not compatible with DarkroomJS. Downgrading to 1.6.7 appears to have restored normal functionality.

lasley:
Can you add the https://github.com/MattKetmo/darkroomjs/issues/112 to the ReadMe under known issues please

lasley (approved):
LGTM

obulkin (approved):
The version change does address the cropping bug in my last comment, so this should be good to go now.

rami-wafaie (approved):
Review + Functional testing: Looks fine to me
:+1:

rami-wafaie:
@pedrobaeza can you please have a look at this??

lasley:
@hughesbm - please squash for merge

lasley:
Oh derp you're one of mine! Squashed myself

PR #601, opened 2017-04-19
[UPDATE]web_readonly_bypass:migrate from 9 to 10

blutecsolutions:
1.Repair Travis.
2.Make this to work.

### DIFF
--- a/web_readonly_bypass/__manifest__.py
+++ b/web_readonly_bypass/__manifest__.py
@@ -6,7 +6,7 @@
     'name': 'Read Only ByPass',
     'version': '10.0.1.0.0',
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
-    "maintainer": "ACSONE SA/NV,Odoo Community Association (OCA),Kalpana Hemnani,Blutec Solutions,Kamil Muhammad",
+    "maintainer": "ACSONE SA/NV,Odoo Community Association (OCA),Kalpana Hemnani,Blutec Solutions",
     "website": "http://www.acsone.eu",
     "license": "LGPL-3",
     'category': 'Technical Settings',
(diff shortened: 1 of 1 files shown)


blutecsolutions:
@pedrobaeza 

Can we continue using this PR

pedrobaeza:
But is this working at all?

blutecsolutions:
Changes will be updated ASAP. 👍

web_readonly_bypass/__manifest__.py
```diff
@@ -1,33 +1,14 @@
...
+# Copyright (c) 2015 ACSONE SA/NV (<http://acsone.eu>)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
 {
     'name': 'Read Only ByPass',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
-    "maintainer": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "maintainer": "ACSONE SA/NV,Odoo Community Association (OCA),Kalpana Hemnani,Blutec Solutions,Kamil Muhammad",
```
pedrobaeza: You can't put 3 co-authors here, and less if the changes are so minimal.

blutecsolutions: Then we can leave it at two as JS changes will be updated

pedrobaeza: No, it should be only one author, and it the changes deserve to be here. The place for putting all the involved persons is in README.rst file, in the section contributors. There you can put several persons if they have worked on that.

blutecsolutions: Okay :)

yvaucher:
This PR seems incomplete, I'm closing it in favor of the other most recent ones #743 and #673

PR #604, opened 2017-04-21
10.0 web_widget_color: Init widget on kanban and tree views

omar7r:
Hi,

This module works fine on create event from form view but it doesn't work on create event in tree and kanban views, this PR initializes the widget in these views too.

Regards.

### DIFF
--- a/web_widget_color/static/src/js/widget.js
+++ b/web_widget_color/static/src/js/widget.js
@@ -72,19 +72,7 @@ odoo.define('web.web_widget_color', function(require) {
      * Init jscolor for each editable mode on view form
      */
        FormView.include({
-        on_button_edit: function () {
-            this._super();
-            jscolor.init(this.$el[0]);
-        },
-        on_button_create: function () {
-            this._super();
-            jscolor.init(this.$el[0]);
-        },
-        on_button_new: function () {
-            this._super();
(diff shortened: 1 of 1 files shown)


jesusVMayor:
Modified to init the widget in all triggers of form edit mode.

pedrobaeza (approved):

daramousk (approved):
While this works, perhaps the init should better be done in the `do_show` in case the load_record is not called, right now multiple inits take place (every time anything is changed)

web_widget_color/static/src/js/widget.js
```diff
@@ -71,15 +71,11 @@ odoo.define('web.web_widget_color', function(require) {
     /*
      * Init jscolor for each editable mode on view form
      */
-    FormView.include({
-        on_button_edit: function () {
+       FormView.include({
+        on_form_changed: function () {
             this._super();
             jscolor.init(this.$el[0]);
```
Tardo, line 77: jscolor is not correctly used in this module.
init doesn't accept any parameter... when you call init, jscolor initialize all elements with 'color' class.

Line 61 is wrong too...
See https://github.com/OCA/web/blob/10.0/web_widget_color/static/lib/jscolor/jscolor.js#L27

Tardo (changes requested):

PR #613, opened 2017-04-28
[FIX] we need to parse the parameter to an int

hbrunn:
without this, configuring '0' triggers an infinite loop on some browsers

### DIFF
--- a/web_menu_navbar_needaction/static/src/js/web_menu_navbar_needaction.js
+++ b/web_menu_navbar_needaction/static/src/js/web_menu_navbar_needaction.js
@@ -36,7 +36,7 @@ openerp.web_menu_navbar_needaction = function(instance)
         },
         refresh_navbar_needaction: function(timeout)
         {
-            if(parseInt(timeout))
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

tarteo:
@hbrunn Is the if statement triggering an infinite loop or the setTimeout when timeout is '0'? If it's the setTimeout triggering it you'll need to reset the variable. like this `timeout = parseInt(timeout)`

hbrunn:
the timeout. In the conditional, `'0'` is coerced to boolean `true`, in `setTimeout`, it's coerced to int `0`, and then the browser freezes

tarteo (approved):
@hbrunn Thanks, I understand the problem now.

hparfr:
If the intent is to trigger a callback periodically, [setInterval()](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval) should be used instead.

yajo (approved):

PR #614, opened 2017-05-01
8.0 web search autocomplete prefetch

hbrunn:
the latter commit improves usability for big databases/slow server a lot, as the original would trigger a search for every keystroke, which can lead to very annoying delays

### DIFF
--- a/web_search_autocomplete_prefetch/static/src/js/web_search_autocomplete_prefetch.js
+++ b/web_search_autocomplete_prefetch/static/src/js/web_search_autocomplete_prefetch.js
@@ -84,6 +84,7 @@ openerp.web_search_autocomplete_prefetch = function(instance)
     });
 
     instance.web.search.AutoComplete.include({
+        keypress_timeout: 200,
         select_item: function()
         {
             if(!this.current_result)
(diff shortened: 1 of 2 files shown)


web_search_autocomplete_prefetch/static/src/js/web_search_autocomplete_prefetch.js
```diff
@@ -83,6 +84,7 @@ openerp.web_search_autocomplete_prefetch = function(instance)
     });
 
     instance.web.search.AutoComplete.include({
+        keypress_timeout: 350,
```
legalsylvain, line 87: Hi. It could be interesting to have a setting for that value (ir.config;parameter) ? or not ? just a question.

hbrunn: @legalsylvain already proposed this to the customer, they agreed, it's in https://github.com/OCA/web/pull/614/commits/08b3b724f042f3269b0ba60ce6ca343fd3e26a58

jesusVMayor (approved):

pedrobaeza:
Merging this old one.

PR #619, opened 2017-05-12
[9.0][MIG] Ported `web_polymorphic_many2one`

katyukha:
Seems to be working now)

But i am not expert in JS.


chermed:
Hello,

Why not use reference instead, can you sepcify some CONS and PROS of using polymorphic over reference fields.

katyukha:
Hello,

Example of code that use this widget

```python
    condition_id = fields.Many2one(
        'generic.condition', 'Condition', required=True, ondelete='cascade')
    res_model = fields.Char(
        string='Object Model',
        related='condition_id.model_id.model',
        readonly=True)
    res_id = fields.Integer(
        'Object ID', help='ID of object to test condition on')
```

```xml
                <field name="condition_id"/>
                <field name="res_model" invisible="0"/>
                <field name="res_id"
                       widget="polymorphic" polymorphic="res_model"/>
```

How to use *reference* field in this case?

----------

So, polymorphic widget is best solution i have found for cases, that use separate fields to store *model* and *object\_id*.
For example, how simpler would be *Aeroo reports* wizard *Test report* with this widget...

Also few more things about reference field:
- it seems to be unfinished (i have tried to use it in odoo 6.0, and i disliked it after)
- I think, mixing two different values in one field, is bad idea. Reasons for this:
  - need to serialize / deserialized on *read* / *write*
  - difficult to search values (not standard way)
- How to change dinamically list of models for reference field? For example in onchange? And what if this list varies for each document (depending on other fields)?
- Odoo's internal structure (ex. workflows, actions, etc) does not use *reference* field!
- Are there any real cases when *reference* field is used? Is it used there clean?

chermed:
@katyukha it's just a question, and the reference fields is used on the models `ir.ui.menu` and `ir.actions.server`

It's a 👍  for filters and for loading dynamilcally the records.

katyukha:
Thanks for notes about real cases of usage of *reference* field.
As i see it is good in cases with a small set of predefined models, that could be used in this field.

May be, it is a good idea to add some notes to readme, with differences betwen *reference* field and *polymorphic* widget?

katyukha:
Code rebased on last 9.0 branch

PR #620, opened 2017-05-15
[FIX][web_responsive] Make it compatible with l10n_es_toponyms

yajo:
Before this patch, there was an incompatibility between this addon and l10n_es_toponyms, caused by the lack of wrapping of the special address field that was being added:

captura de pantalla de 2017-05-15 14-51-21

Now layout works as expected in that case. In any other case, it seems sensible to wrap inputs so no more weird overflows happen and less chances of needing horizontal scrolling happen:

captura de pantalla de 2017-05-15 14-51-31

@Tecnativa

### DIFF
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -32,8 +32,6 @@
     }
     .o_group {
         &.o_inner_group > tbody > tr > td {
-            // prevent td content from breaking into lines
-            white-space: nowrap;
             .note-editor > .note-toolbar {
                 // prevent wysiwyg editor buttons from expanding the screen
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@lasley do you remember why did you put that?

pedrobaeza (approved):
Tentatively until Dave's answer

lasley (approved):
@pedrobaeza - I actually didn't put that, it was @danieltorres7 in f744d2c30f2fc2cf4d40b687a6c11912a8cfb24f

Seems fine to me though - I don't notice any issues on smaller screens with this removed.

rafaelbn (approved):

PR #628, opened 2017-05-17
[MIG] Migration from 9.0 to 10.0 web_access_rule_buttons

daramousk:


pedrobaeza:
It would be interesting to rename to singular as stated in the OCA guidelines.

StefanRijnhart:
@pedrobaeza Not sure in this case, as it does not refer to a data model but to the standard edit/create/delete __buttons__ in the interface.

StefanRijnhart (approved):

web_access_rule_buttons/static/src/js/web_access_rule_buttons.js
```diff
@@ -2,43 +2,39 @@
...
-    }
-
-  });
+odoo.define("web_access_rule_buttons.main", function(require) {
+    "use strict";
+
+    var core = require("web.core");
+
+    var FormView = require("web.FormView");
+    var ListView = require("web.ListView");
```
felixvillafranca: This line could be removed

felixvillafranca (changes requested):
It's Ok; but there is a variable unused.

web_access_rule_buttons/static/src/js/web_access_rule_buttons.js
```diff
@@ -2,43 +2,39 @@
...
+    FormView.include({
+
+	load_record() {
```
yajo: Sorry friend, we can't allow this syntax until it is supported by PhantomJS, or all JS tests would potentially break. You have to use the old, ugly and safe `load_record: function () {...` syntax.

web_access_rule_buttons/static/src/js/web_access_rule_buttons.js
```diff
@@ -2,43 +2,39 @@
...
+	    var self = this;
+	    return this._super.apply(this, arguments).then(function() {
+		self.show_hide_buttons()
```
yajo: Feels better IMHO to proxy the function like `this._super.apply(this, arguments).then($.proxy(this.show_hide_buttons, this))`

web_access_rule_buttons/static/src/js/web_access_rule_buttons.js
```diff
@@ -2,43 +2,39 @@
...
+	    });
+	},
+
+	show_hide_buttons() {
+	    var self = this;
+	    this.dataset.call("check_access_rule_all", [ [ this.datarecord.id ], [ "write" ] ]).then(function(accesses) {
+		console.log(accesses)
```
yajo: IMHO this deserves a `console.debug` instead. Please add semicolon.

web_access_rule_buttons/static/src/js/web_access_rule_buttons.js
```diff
@@ -2,43 +2,39 @@
...
+		self.show_hide_edit_button(accesses.write);
+	    });
+	},
+
+	show_hide_edit_button(access) {
+	    if (this.$buttons) {
+		var button = this.$buttons.find(".o_form_button_edit");
+		if (button) {
+		    button.prop("disabled", !access);
```
yajo: 4 spaces, no tabs.

yajo (changes requested):

daramousk:
@Yajo Changes implemented

felixvillafranca (approved):
Change Ok.
Funtional Ok.

yajo (approved):

pedrobaeza:
I would like to have this renamed to singular, but too late... Maybe on v10?

pedrobaeza:
Ouch, I see that this is v10. Uf, too bad. Can you maybe make a PR renaming it?

PR #630, opened 2017-05-18
[New] widget product selector variant

fmdl:


pedrobaeza:
Please make a proper README following template https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst for seeing what is the feature that this adds.

fmdl:
Dear Pedro, I have add the description, but the screenshot do not appear, do you know why ?

I have made a PR for Odoo to fix the the random order of variante, if you like can put an 👍 : https://github.com/odoo/odoo/pull/17034

pedrobaeza:
I think this is better solved with https://github.com/OCA/product-variant/pull/47, as no specific widget is required, and it allows you more options.

fmdl:
I think is not for the same use.  This module is more universal, it work in bom, purchase, ...

pedrobaeza:
Well, I see a mix of concepts in this widget, developing a full window that can be done with ORM, but let's see what others think. Anyway, in English it's `variant`, not `variante`.

fmdl:
Dear @pedrobaeza, 
The main difference is that the dimension is not limited, the OCA/product-variant#47 is limited at two.

pedrobaeza:
Then you have this module for a similar thing: https://github.com/OCA/product-variant/tree/9.0/purchase_variant_configurator

fmdl:
Yes it is similar, but for me it is not the best integration. In purchase order line you need open a form to add line, even if there no variante. And you need to developpe one new module for each case. My module is more generique.

What do you think that my module use https://github.com/OCA/product-variant/blob/9.0/product_variant_configurator/models/product_configurator.py in the popup configurator ?

pedrobaeza:
You're right, but it's also requires to have all the possible variants created, which sometimes is not feasible as the number of possible combinations is very high. Well, I think we have several different cases that can co-exist, so let's go with the PR and see if people see this interesting.

Please check Travis status.

PR #631, opened 2017-05-18
10.0 web confirm window close

StefanRijnhart:
I could not use the bus functionality anymore because that now uses async promises which don't work well with a synchronous event like beforeunload.


guewen (approved):

guewen:
Code review LGTM, but on runbot it didn't seem to have any effect?

StefanRijnhart:
@guewen seems to work for me. Note that Odoo only sets the 'dirty' flag when your focus leaves the field that you edited. It does not work when you close the window when your cursor is still in that field.

StefanRijnhart:
Added the above as a Known issue in the README.rst.

legalsylvain (approved):
Thanks a lot @StefanRijnhart for porting this module.
code review. no test.

Issue #151, opened 2015-06-23
[8.0] web_ckeditor4 does not work in email templates/mako


RoelAdriaans:
With the web_ckeditor4 enabled, editing an email template breaks the template.

The code before saving is, for example:

``` mako
       % if object.origin:
       &nbsp;&nbsp;Order reference: ${object.origin}<br />
       % endif
```

and after saving:

``` mako
% if object.origin: &nbsp;&nbsp;Order reference: ${object.origin}<br />
% endif %
```

The missing enters break the mako formatting engine, you have to edit the template by hand.. 
This is in 8.0. 
(Ninja edit, `%endif` is not removed)

pedrobaeza:
And `% endif` is also removed?

@antespi, you have also this issue, doesn't it?

RoelAdriaans:
@pedrobaeza %end if is not removed, it's on a new line with the next % if next to it.
I've updated my paste.. 

https://github.com/0k/web_ckeditor/issues/4 seems to be the same issue

pedrobaeza:
I think that has to be treated on the ckeditor4 JS widget itself, not here, but @hbrunn should say about this

hbrunn:
To fix this, we'll have to dive into ckeditor's internals, I think. http://docs.ckeditor.com/#!/api/CKEDITOR.htmlDataProcessor seems to be a good starting point. Or this looks even better: http://docs.ckeditor.com/#!/api/CKEDITOR.config-cfg-protectedSource

antespi:
@pedrobaeza yes, we have the same issue.
@hbrunn yes, protectedSource regex maybe fix this.

rafaelbn:
@hbrunn did you detect the same issue using https://github.com/OCA/crm/tree/8.0/newsletter when creating/editing emails templates Under Sales/Configuration/Newsletter type? (as newsletter installs ckeditor)

hbrunn:
@RoelAdriaans-B-informed @pedrobaeza @antespi @rafaelbn 

I think @lfreeke mentioned this problem once.

Should be fixed by #158

RoelAdriaans:
#158 did not fix this issue

rafaelbn:
As #158 didn't fix this issue, are any of you interested in fixing it? Are you using this module en production enviroments? We are still not using this module as corrupts many standards functionalities.

rafaelbn:
Check https://github.com/OCA/web/pull/213/ @hbrunn update ckeditor version

[...]

PR #158, opened 2015-06-30
[FIX] protect jinja2 expressions from being messed with

hbrunn:
fixes #151

PR #634, opened 2017-05-20
[10][UPD] web_ckeditor4: Make it compatible Odoo V10

dntq83:
This PR makes module web_ckeditor4 compatible with odoo 10


hbrunn (changes requested):
you should follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0#howto in order to cherry pick changes here from 9/8.
I take it you also took the opportunity to update ckeditor? If so, do this in a separate commit, otherwise, reviewing your adaptions is a lot more effort than necessary.

Did you also check if the update fixes #151? I tried to fix it in #158, but the users never came back to tell if it works for them.

dntq83:
I changed and make another pull request #639, please check again

PR #639, opened 2017-05-23
Change as reuqest fixing #634

dntq83:
Change as reuqest fixing #634


pedrobaeza:
Why do you close the other PR and open another one with worst title and the same things not fixed? Please reuse the same PR. Anyway, please attend @hbrunn's comments from the other PR.

PR #641, opened 2017-05-29
Fix [8.0] Perf issue with web_m2x_options #615 

### LINKED ISSUE #615: Perf issue with web_m2x_options
Hello,

Steps to reproduce: 

On a clean base, create a new partner, in Company field enter one character then enter another one, then another one.
6 requests are sent 
- name_search (3 times)
- check_access_rights (3 times)

Expected: 
4 requets: 
- name_search (3 times)
- check_access_rights (1 time)

Depending of the config set, there can be one more extraenous requests after each key press
https://github.com/OCA/web/blob/8.0/web_m2x_options/static/src/js/form.js#L113

Before working on a patch, Is there a reason why this kind of work are processed client side ?

Ping @codingforfun

hparfr:
Before : 
if no options specified after each keystroke, a call to disable_quick_create and/or check_access_rights are made

After:
a call is done once per field

Each request saved improves perf for user and load on the server.

Extra
+ some fixes on `parseInt` : always specify a radix
+ refactor  `is_option_set` 

Fixes #615

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -40,12 +40,17 @@ openerp.web_m2x_options = function (instance) {
         },
 
         is_option_set: function(option) {
-            //return option in ('True', 'true', true)
-            return {
-                'true': true,
-                'True': true,
-                true: true,
-            }[option] || false;
+            if (_.isUndefined(option)) {
+                return false
+            }
+            var is_string = typeof option === 'string'
+            var is_bool = typeof option === 'boolean'
+            if (is_string) {
+                return option === 'true' || option === 'True'
+            } else if (is_bool) {
+                return option
+            }
+            return false
         },
 
         show_error_displayer: function () {
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -40,17 +40,12 @@ openerp.web_m2x_options = function (instance) {
         },
 
         is_option_set: function(option) {
-            if (_.isUndefined(option)) {
-                return false
-            }
-            var is_string = typeof option === 'string'
-            var is_bool = typeof option === 'boolean'
-            if (is_string) {
-                return option === 'true' || option === 'True'
-            } else if (is_bool) {
-                return option
-            }
-            return false
+            //return option in ('True', 'true', true)
+            return {
+                'true': true,
+                'True': true,
+                true: true,
+            }[option] || false;
         },
 
         show_error_displayer: function () {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@Yajo please check

web_m2x_options/static/src/js/form.js
```diff
@@ -40,17 +40,12 @@ openerp.web_m2x_options = function (instance) {
...
-                return option === 'true' || option === 'True'
-            } else if (is_bool) {
-                return option
-            }
-            return false
+            //return option in ('True', 'true', true)
+            return {
+                'true': true,
+                'True': true,
+                true: true,
```
yajo: This is not doing what you expect. Object keys are never bools, just a `"true"` string.
What you want to do is `return $.inArray(option, [true, "true", "True"])`.

web_m2x_options/static/src/js/form.js
```diff
@@ -70,7 +65,7 @@ openerp.web_m2x_options = function (instance) {
             if (_.isUndefined(this.view))
                     return this._super.apply(this, arguments);
                 if (!_.isUndefined(this.view.ir_options['web_m2x_options.limit'])) {
-                this.limit = parseInt(this.view.ir_options['web_m2x_options.limit']);
+                this.limit = parseInt(this.view.ir_options['web_m2x_options.limit'],10);
```
yajo, line 73: According to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Parameters), please revert this change.

web_m2x_options/static/src/js/form.js
```diff
@@ -291,7 +287,7 @@ openerp.web_m2x_options = function (instance) {
             // returned.
 
             if (!_.isUndefined(this.view.ir_options['web_m2x_options.limit'])) {
-                this.limit = parseInt(this.view.ir_options['web_m2x_options.limit']);
+                this.limit = parseInt(this.view.ir_options['web_m2x_options.limit'], 10);
```
Yajo: As said above, revert this.

web_m2x_options/static/src/js/form.js
```diff
@@ -401,7 +397,7 @@ openerp.web_m2x_options = function (instance) {
                     .css('cursor', 'pointer')
                     .click(function(e)
                     {
-                        var id = parseInt(jQuery(this).attr('data-id'));
+                        var id = parseInt(jQuery(this).attr('data-id'), 10);
```
Yajo: Revert this.

yajo (changes requested):
The idea is good, but you still need to change some things.

web_m2x_options/static/src/js/form.js
hparfr: According to MDN: 
>Always specify this parameter to eliminate reader confusion and to guarantee predictable behavior.

yajo: Oops sorry, I mistaked the diff, somehow it was in my mind that you were *removing* it, while you were adding it. Then for get my comments about this 😅

web_m2x_options/static/src/js/form.js
hparfr: True, I will revert it.

yajo (approved):
Will squash at merge.

jesusVMayor (approved):

pedrobaeza:
Does it happen also in 9.0/10.0?

hparfr:
Thanks guys :)

PR #642, opened 2017-05-30
[IMP] web_m2x_options: Perf issue https://github.com/OCA/web/issues/615

jesusVMayor:
* fix unecessary calls

Do check_access_rights and disable_quick_create calls only once per field.

https://github.com/OCA/web/pull/641 for 9.0

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -113,7 +113,7 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
             if (_.isUndefined(this.view))
                     return this._super.apply(this, arguments);
                 if (!_.isUndefined(this.view.ir_options['web_m2x_options.limit'])) {
-                this.limit = parseInt(this.view.ir_options['web_m2x_options.limit']);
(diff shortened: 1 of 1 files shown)


web_m2x_options/static/src/js/form.js
```diff
@@ -136,13 +136,14 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
...
             if (!(self.options && (self.is_option_set(self.options.create) || self.is_option_set(self.options.create_edit)))) {
-                create_rights = new Model(this.field.relation).call(
+                this.create_rights = this.create_rights || (function(){
+                return new Model(self.field.relation).call(
                     "check_access_rights", ["create", false]);
+                })();
             }
 
-            $.when(search_result, create_rights).then(function (data, can_create) {
+            $.when(search_result, this.create_rights).then(function (data, can_create) {
```
yajo, line 146: I feel like it could happen that `this.create_rights` can be `undefined` in some circumstances (if the above `if` never executes). Would that be a problem?

yajo (changes requested):

PR #643, opened 2017-05-30
[9.0][fix]web_m2x_options: creation access rules

jesusVMayor:
In 8.0 when the user does not have creation permission, the buttons create/create and edit... are hidden. We recover that functionality.


pedrobaeza (approved):
Reading 8.0 changes, seems correct, although the scope of this one is bigger.

yajo (approved):
This makes the addon do what its README says it does, and improves performance, so great 👍

pedrobaeza:
Merging as well

PR #651, opened 2017-06-12
[FIX] Search more always visible even if results below limit

chienandalu:
**Current behaviour:**

- With `web_m2x_options.search_more: True` set, _Search more..._ option will be available only if the results of the query are above `web_m2x_options.limit` (or Odoo's default).

**Expected behaviour:**

- With `web_m2x_options.search_more: True` set, _Search more..._ option should be always visible no matter how many ocurrences found in the field.

cc @Tecnativa

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -186,7 +186,7 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
                     search_more_undef = _.isUndefined(self.options.search_more) && _.isUndefined(self.view.ir_options['web_m2x_options.search_more']),
                     search_more = self.is_option_set(self.view.ir_options['web_m2x_options.search_more']);
 
-                if ((values.length > self.limit && (can_search_more || search_more_undef || search_more)) || search_more) {
+                if ((values.length > self.limit && (can_search_more || search_more_undef)) || search_more) {
                     values = values.slice(0, self.limit);
                     values.push({
                         label: _t("Search More..."),
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -186,7 +186,7 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
                     search_more_undef = _.isUndefined(self.options.search_more) && _.isUndefined(self.view.ir_options['web_m2x_options.search_more']),
                     search_more = self.is_option_set(self.view.ir_options['web_m2x_options.search_more']);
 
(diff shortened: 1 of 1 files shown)


web_m2x_options/static/src/js/form.js
```diff
@@ -187,7 +187,7 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
                     search_more_undef = _.isUndefined(self.options.search_more) && _.isUndefined(self.view.ir_options['web_m2x_options.search_more']),
                     search_more = self.is_option_set(self.view.ir_options['web_m2x_options.search_more']);
 
-                if (values.length > self.limit && (can_search_more || search_more_undef || search_more)) {
+                if ((values.length > self.limit && (can_search_more || search_more_undef || search_more)) || search_more) {
```
carlosdauden: I do not understand the usefulness of this but in any case I think that the first `search_more` variable is unnecessary...

carlosdauden (changes requested):

web_m2x_options/static/src/js/form.js
yajo: Indeed redundant

carlosdauden: Why not use limit = 1?

chienandalu: - First `search_more` is necesary in case is set to False (so the option is disabled).
- Limit = 1 limits also the displayed ocurrences. So this keeps that option independent.

yajo: I think your first point is exactly the opposite: it's an OR (`||`), so it will only have effect when it is `true`, which is already the case with the little code you added.

rousseldenis:
Hi all,

I'm interested in this module.

But I'm wondering : If we want to activate the search_more in field options and not in ir.config parameters, and if we want to show it event if values.length < limit, is it possible (as now it is not)? Is that behaviour wanted ?

rousseldenis:
Ok, combination with limit=0 and search_more=True


UPDATE : It could be more convenient to be able to force visibility on just one field. My question is open

chienandalu:
Hi, @rousseldenis. Yes, that's the wanted behaviour :wink: but I have to fix it yet. I think it could also be possible to have it at a field level.

yajo (approved):
Travis fails per linter, but not your problem, so ok for me.

rousseldenis (approved):
LGTM

pedrobaeza (approved):
It's an acceptable change of behavior.

PR #659, opened 2017-06-28
[8.0][ADD] web search 'starts with'

hparfr:
Adds 'Starts with' operator in advanced search.


pedrobaeza:
https://github.com/OCA/web/tree/8.0/web_advanced_search_wildcard already does this

hparfr:
Thanks @pedrobaeza 

I searched for it, do not know how I manage to not see it. 😢  

This module is simplier: users do not have to remember to put '%' in the query.

pedrobaeza:
Well, that's true, but the other is more powerful

PR #660, opened 2017-06-29
[8.0][ADD] web_search_nospace: remove redundant spaces when pasting content into search box

daramousk:
When you paste content in the search bar that has whitespace characters before or after the search string thus causing the search to fail. 

This module fixes that


web_search_nospace/static/src/js/search.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * © 2017 Therp BV <http://therp.nl> License AGPL-3.0 or later
+ * (http://www.gnu.org/licenses/agpl.html).
+ */
+openerp.web_search_nospace = function(instance) {
+	instance.web.search.InputView = instance.web.search.InputView.extend({
```
hbrunn: don't do this. This way, you actually *replace* the `InputView` object, this can cause very hard to debug problems. Say use [include](https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/openerpframework.js#L148) instead, this modifies the existing object

hbrunn (changes requested):
if you painted the icon yourself, add its source too, if not, add a reference to where you got it and check that it has a free license in the first place

web_search_nospace/__openerp__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Search no space",
+    'author': "Therp BV, Odoo Community Association (OCA)",
+    'category': "web",
+    'summary': """
+    This module trims the whitespace from the search bar in the end of your
+    search term.
+    """,
+    'version': "1.0",
```
hbrunn: use [OCA's versioning scheme](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers)

daramousk:
@hbrunn Implemented the changes

hbrunn (approved):

daramousk:
@NL66278 A review here as well would be nice.

web_search_nospace/static/src/js/search.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * © 2017 Therp BV <http://therp.nl> License AGPL-3.0 or later
```
NL66278: Make it the full word Copyright. and use https://

web_search_nospace/static/src/js/search.js
```diff
@@ -0,0 +1,16 @@
...
+ * (http://www.gnu.org/licenses/agpl.html).
+ */
+openerp.web_search_nospace = function(instance) {
+	instance.web.search.InputView.include({
+
+		onPaste : function() {
```
NL66278: Why the large indents? Another :retab needed?

web_search_nospace/static/src/js/search.js
```diff
@@ -0,0 +1,16 @@
...
+			this._super()
+			setTimeout(function() {
```
NL66278: I am not an expert in Odoo javascript programming. But is the timeout really needed here? Can't we have simply:  ```this.$el.text(this.$el.text().trim());``` directly? Just asking...

web_search_nospace/views/assets_backend.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data>
+		<template id="assets_backend" name="web_search_nospace.assets_backend"
```
NL66278: More indentation problems.

web_search_nospace/static/src/js/search.js
NL66278: @daramousk Please answer this question as well

daramousk: @NL66278 Basically there were some timing/race condition issues from what I remember, you can see an explanation here https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful.

NL66278: @daramousk Did you actually experience these conditions? Because if I read up on the discussion on stackoverflow, especially the extended answer by DVK, this does not seem to apply here, but maybe I am mistaken.

daramousk: @NL66278 Ok I just went back to this to remember why I used it. Basically I used it because it was used on the super call and for some reason the the changes to `$el` are not propagated unless you force it with setTimeout. If you remove the setTimeout the whole thing will not work at all. 

Now why this happens I do not know, it's probably some obscure event handling mechanism. You can check the super call for more data.

NL66278: OK, lets keep it as you made it.

web_search_nospace/README.rst
```diff
@@ -0,0 +1,40 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
simahawk: as this is a new module pls follow the new readme structure https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

web_search_nospace/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
```
simahawk: copy is not needed in `__init__`, no mandatory to remove tho... but pls, replace :copyright: w/ "Coypright"

web_search_nospace/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Search no space",
+    'author': "Therp BV, Odoo Community Association (OCA)",
+    'category': "web",
+    'summary': """
+    This module trims the whitespace from the search bar in the end of your
```
simahawk: "at the end"

web_search_nospace/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Search no space",
```
simahawk: wouldn't be better "Search no _trailing_ space"?

simahawk (changes requested):
LG, just minor changes required

web_search_nospace/__openerp__.py
```diff
@@ -0,0 +1,18 @@
...
+    'author': "Therp BV, Odoo Community Association (OCA)",
```
nikul-serpentcs, line 6: @daramousk Add a `website` key as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11)

nikul-serpentcs (commented):
minor change

web_search_nospace/README.rst
daramousk: @simahawk Doesn't this readme follow the new structure? (some non necessary sections are omitted)

simahawk: if you follow the link you see that the README.rst is not there anymore and is auto-generated w/ per-section files from `/readme`. 
I don't see the dir so.... nope :)

web_search_nospace/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
```
NL66278: Another copyright symbol to replace

NL66278 (approved):
:+1: LGTM. Please still remove last copyright symbol. Not blocking though. For the rest OK.

tarteo (changes requested):
Can you take a look at the lint issues https://travis-ci.org/OCA/web/jobs/480883417#L1686

NL66278:
@tarteo These lint errors are from a totally different module.

tarteo:
@NL66278 Oops, I meant https://travis-ci.org/OCA/web/jobs/480883417#L1822

tarteo (approved):

daramousk:
@simahawk Simone, anything else that needs to be done here?

PR #662, opened 2017-07-03
[FIX] Bad setup file for version 10.0

adrienpeiffer:


sbidoul:
Fast-tracking ovious fix.

PR #664, opened 2017-07-05
[10.0][ADD] New module which allow to hide measures in pivot view

Olageibol:


daramousk (changes requested):
Hello Janire and thank you for this PR,

From what I see, this module requires the end user to define the `fields_get` and adds some js to remove a field if it has the `selectable=False`.

Now, from what I understand we could just define a `fields_get` and check `self.env.context.get('view_type') == 'pivot'` and if True we remove the fields from there (or just change the call here which we can achieve the same with minor changes https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/js/framework/data_manager.js#L153. Which fields to remove (rather not fetch) could be determined by perhaps fetching the context from an action or the view definition itself.

I am saying this so that we could avoid the end user actually having to write any code rather than change their views by adding a context.

What is your opinion?

pedrobaeza:
FYI, this is now possible in core in v12.

PR #671, opened 2017-07-11
[FIX] web_widget_float_formula: Fix Travis Build

lasley:
My bet's on this being the issue. We'll see once Travis comes back.

* Add debug param to UI test runner to hopefully fix OCA#637

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="web_widget_float_formula"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="web_widget_float_formula"
   - TESTS="1" INCLUDE="web_widget_float_formula"
 
 virtualenv:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,9 @@ env:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
   - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="web_widget_float_formula"
-  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="web_widget_float_formula"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="web_readonly_bypass"
   - TESTS="1" INCLUDE="web_widget_float_formula"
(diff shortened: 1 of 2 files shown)


lasley:
Seems it worked. Sorry about the lag everyone, 9.0 is basically dead to me so I totally forget to look at anything relating to it!

lasley:
Hmmm maybe it doesn't. It's passing on my [Travis](https://travis-ci.org/LasLabs/web/builds/252459506), but not in [OCA](https://travis-ci.org/OCA/web/builds/252459977). I'll keep digging - not sure why there would be a disparity

lasley:
This is being caused by conflicting tests in `web_widget_float_formula` and some mystery module. I spent a good deal of time trying to identify said mystery module with no success. 

My solution is to isolate the Travis builds, unless someone has a better idea on how to find which module is incompatible that is not just submitting random module combinations to Travis. Seems like a crappy solution, but I'm out of ideas.

Interestingly enough, I have a similar issue [in `oauth_provider`](https://github.com/OCA/server-tools/pull/889#issuecomment-314596175)

pedrobaeza (approved):
OK, for me it's enough right now. Fast-tracking this

Garamotte:
@lasley There is also an issue with the report-print-send repo, which has red builds in 10.0 since yesterday.
I think something has been broken in the last week in Travis configuration...

pedrobaeza:
It should be something different. There was a change on the requirements.txt installed file, but nothing to do with this fix I think. report-print-send repo can be affected by that change. @moylop260 can confirm this.

lasley:
Hmmm the `report-print-send` looks like a weird error too - [a mock isn't being called](https://travis-ci.org/OCA/report-print-send/jobs/252325245#L942). I don't think that could be a requirements thing.

We've been internally tracking some Travis issues over the past maybe 3 weeks - maybe this is now the time to say something. Unfortunately, we haven't been able to nail down any correlation or causation, but we've been seeing
* Builds failing on OCB but not Odoo (or vise versa), only to pass on a subsequent rerun
* Builds failing on our fork consistently, but not on OCA (or vice versa, as we saw here with my fork passing and OCA failing - I even reset our fork to OCA/9.0 and we still passed)
* Builds failing when modules are declared with each other, such as here and https://github.com/OCA/server-tools/pull/889#issuecomment-314596175

Unfortunately all I have are observations though. It's all been so totally random that it's pretty much been a waiting game for it to either resolve itself or explode to the point where we can figure it out.

cc @laslabs

moylop260:
@sylvain-garancher 
Now [odoo-10.0 use mock==2.0.0](https://github.com/odoo/odoo/blob/68934c6b8/requirements.txt#L13)
But your project is compatible just with mock==1.0.1 

I'm using the ssh runbot connection to reproduce the travis error and the result was:
 - screen shot 2017-07-12 at 13 28 27


I have created the fix in the following PR: https://github.com/OCA/report-print-send/pull/96

Issue #505, opened 2016-12-08
[9.0] web_readonly_bypass: Missing features


pedrobaeza:
The commit e4e0b72ea07f57074551411cbe5a9e9f79efd1f2 is not included in current module for v9.

@JonathanNEMRY, as you are the original author of that module, can you tell me if you can forward port that feature to version 9?

pedrobaeza:
@JonathanNEMRY any news?

pedrobaeza:
Closing as this is not used by me anymore.

PR #451, opened 2016-10-18
[IMP]web_readonly_bypass:migrated from 9 to 10

hemnanikalpana:

PR #673, opened 2017-07-17
[10.0][MIG] web_readonly_bypass : Migration to 10.0

Ehtaga:


pedrobaeza:
There's already another migration in https://github.com/OCA/web/pull/451. Have you checked it?

Ehtaga:
@pedrobaeza Sorry I didn't see this one. I saw #601 01 and it was not working so I started from the beginning. I can add the 2 other authors if you want to (in the README, not in the manifest as they did...).

pedrobaeza:
If their contribution is valuable for this PR, then add them on contributors list. If it's not working, then we simply dispose that PR. Have you seen also https://github.com/OCA/web/issues/505?

Ehtaga:
@pedrobaeza I added one of the contributor who did a similar migration.
I also cherry picked 2 commits from the 8.0 version.

Ehtaga:
@pedrobaeza We noticed a bug with this module in v10. It works, but we can't load the Settings dashboard anymore.
The error is: `odoo could not find client action 'web_settings_dashboard.main'.`
For the moment we choose not to use this module as our project needed to be deployed quickly. If we have some time later to investigate, we will do it.
Can you add the 'need fixing' tag?
This PR is the most advanced among those you referred to.
It could also be interesting to look at the stylistic guidelines, as you suggested in #451 .
Thanks

web_readonly_bypass/static/src/js/readonly_bypass.js
```diff
@@ -138,4 +138,47 @@
         },
 
     });
+
+    instance.web.ProxyDataSet.include({
```
yajo, line 142: You have to migrate the whole file to the new module-based JS API.

yajo (changes requested):

JayVora-SerpentCS:
@Ehtaga and @blutecsolutions need to work this out together as #601 was PRed earlier.

web_readonly_bypass/static/src/js/readonly_bypass.js
anhvu-sg: Hello, do we have any update?

simahawk: JS has been ported here https://github.com/OCA/web/pull/743

yajo: Can we close this one then?

PR #683, opened 2017-07-28
[8.0][FIX] help_online: XMLRPC Call problem

renzo1974:
The import method has no return value so it is not possible to call the import wizard from XMLRPC.


pedrobaeza (approved):
Fast-tracking this simple patch. Please propose the same patch for newer versions.

renzo1974:
wow thanks for the very fast response! 
Im sorry but I currently do not have odoo 9.0 and 10.0 available for development so I can't test it.

pedrobaeza:
It should be the same solution...

PR #685, opened 2017-07-31
9.0 web widget datepicker fix

### LINKED ISSUE #684: [9.0] web_widget_datepicker_options does not set datepicker widget options
The module does not set options on the related date widget of a date field.

I will add a PR to this issue to show my proposed changes.

StephanRozendaal:
Fix #684

### DIFF
--- a/web_widget_datepicker_options/static/src/js/datepicker.js
+++ b/web_widget_datepicker_options/static/src/js/datepicker.js
@@ -43,15 +43,13 @@ odoo.define('web_widget_datepicker_options', function (require) {
     });
 
     DateWidget.include({
-        start: function(parent, options) {
+        init: function(parent, options) {
             this._super.apply(this, arguments);
             var self = this;
             if (this.__parentedParent.options.datepicker) {
                 var options = this.__parentedParent.options.datepicker;
                 $.each(options, function(value, key) {
                     self.options[value] = key;
-                    self.picker[value] = key;
-                    self.picker.options[value] = key;
                 });
             }
         },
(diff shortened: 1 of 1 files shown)


web_widget_datepicker_options/static/src/js/datepicker.js
```diff
@@ -24,34 +24,16 @@ odoo.define('web_widget_datepicker_options', function (require) {
...
-                    self.picker[value] = key;
-                    self.picker.options[value] = key;
-                });
-            }
-        },
-    });
-
     DateWidget.include({
-        start: function(parent, options) {
+	init: function(parent, options) {
```
yajo: Use 4 spaces indentation please.

yajo (changes requested):

yajo:
Could you explain how to test the thing that this is fixing?

web_widget_datepicker_options/static/src/js/datepicker.js
StephanRozendaal: Looks like I used a tab, it's fixed now.

StephanRozendaal:
@Yajo This module adds two demonstration fields on a res.users form. both fields have " options="{'datepicker':{'calendarWeeks': false}}" " parameters, in the current 9.0 version setting calendarWeeks to true does nothing to the calendar display of the field when you try to pick a date. With my change, when setting calendarWeeks to true, when picking a date it will display extra weeknumbers.

Which leads to the conclusion that options on the field were not being transfered to the JQuery datepicker widget, and with this change they are.

StephanRozendaal:
I've fixed that date fields, when viewed in a list view caused a javascript error.

PR #686, opened 2017-08-04
[10.0][FIX] web_notify tests: Fix an arg check

houzefa-abba:
Fix a check when comparing a user count with items within a mock call.

The previous method was succeeding by pure luck because OCA test
databases contain 2 users, which happens to be the amount of items
within a mock "call_args" (it contains args + kwargs).


lmignon (approved):

pedrobaeza:
FYI, I needed to fix also the tests in v9 this way: https://github.com/OCA/web/commit/9094d80da0109a33ed212ccd2f69d922a11d3a86

PR #688, opened 2017-08-08
[FIX][web_responsive] Long-named multicompanies nice in XS

yajo:
Before this patch:
captura el 2017-08-08 a las 14 00 20

After it:
captura el 2017-08-08 a las 14 00 25
captura el 2017-08-08 a las 14 17 37

@Tecnativa

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -6,7 +6,7 @@
     "name": "Web Responsive",
     "summary": "It provides a mobile compliant interface for Odoo Community "
                "web",
-    "version": "10.0.1.0.1",
+    "version": "10.0.1.1.0",
     "category": "Website",
     "website": "https://laslabs.com/",
     "author": "LasLabs, Tecnativa, Odoo Community Association (OCA)",
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -6,7 +6,7 @@
     "name": "Web Responsive",
     "summary": "It provides a mobile compliant interface for Odoo Community "
                "web",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Website",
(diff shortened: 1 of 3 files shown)


web_responsive/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "name": "Web Responsive",
     "summary": "It provides a mobile compliant interface for Odoo Community "
                "web",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
```
pedrobaeza: This requires to update the module, so put `10.0.1.1.0`.

pedrobaeza (changes requested):
Please hide the building icon on large displays.

web_responsive/__manifest__.py
yajo: Are you sure? I think that's not needed to update a client-side qweb template, isn't it?

pedrobaeza: Well, I'm not sure either, but anyway, it doesn't hurt

pedrobaeza:
What about the other request?

yajo:
Ah sorry, I think it's more consistent if we display the same icon in all sizes. This way when using XS viewport you are already familiar with that icon and its meaning.

pedrobaeza:
Well, there are other elements that change notably when they are in small size, and it's very easy to discover for what is the building with the dropdown arrow, so I prefer to keep big screen without that icon.

yajo:
I feel that's anti-UX, but will do.

pedrobaeza (approved):

pedrobaeza:
@lasley @chienandalu please review

chienandalu (approved):

rousseldenis (approved):
:+1: Looks nice

lasley:
Bit late to the party, but 👍 !

PR #690, opened 2017-08-10
[FIX][WIP] Avoid to save sign every time form is saved

cubells:
Bug detected:
- Every time form is saved, sign is saved too, even if the sign is not modified.

With these changes, sign only is saved when sign is modified.

cc @Tecnativa


yajo (approved):

pedrobaeza (commented):
Why have you rewritten entire jSignature library? You should keep it as is, because tomorrow we update that lib and all your effort goes away. Is there's something you can't do outside of this file?

pedrobaeza:
@cubells you haven't answered my question

rafaelbn (approved):
:+1:

PR #700, opened 2017-08-18
[PORT] Ports web_compute_domain_x2many to 10.0

daramousk:


pedrobaeza:
This is about to change in https://github.com/odoo/odoo/pull/18239 and https://github.com/odoo/odoo/pull/18815, so I don't think this deserves the migration of this module.

Issue #702, opened 2017-08-21
[BUG][10.0][web_advanced_search_x2x] Fields not in res.user language


rafaelbn:
Hi,

When you user this module in other language than English

![2017-08-21_11-22-04](https://user-images.githubusercontent.com/8736623/29512668-1fa8ef98-8663-11e7-987e-9bf6cde7d00e.png)
 fields are not translated to the res.user language

cc @Yajo cc @tecnativa

rafaelbn:
Solved in https://github.com/OCA/web/pull/704

PR #704, opened 2017-08-21
[FIX][web_advanced_search_x2x] UI failures

### LINKED ISSUE #703: [BUG][10.0][web_advanced_search_x2x] "add filter" button close window from the second time you clic on it
Hi:

- Go to runbot and try to make an advance search the same filter more than one time
- You will see how window is closed after 2nd time

cc @Yajo @tecnativa

yajo:
Fix #703 and #702.

@Tecnativa


rafaelbn (approved):
tested in runbot, both issues are fixed. thanks! 👍

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #710, opened 2017-08-29
[10.0][FIX] web_access_rule_buttons

espo-tony:
This fixes a bug occurring when creating, with the same operation, a record plus one or more _o2m_, _m2m_ or _m2o_ records related to it. In this scenario, invoking _check_access_rule()_ will crash reporting an error like _DataError: Invalid input syntax for integer "one2many_v_id_42"_ trying to execute a query using a string instead of an integer.

### DIFF
--- a/web_access_rule_buttons/models.py
+++ b/web_access_rule_buttons/models.py
@@ -25,6 +25,17 @@ def check_access_rule_all(self, operations=None):
             # operations
             result[operation] = True
             continue
+        # If we're writing on a new o2m, m2m or m2o record related to a new
+        # record (for example an invoice line created altogether with the
+        # invoice to which it belongs), ids contains a string. Calling
+        # check_access_rule() in this condition will crash, so we just
+        # blindly allow the operations.
+        try:
+            for id in self.ids:
+                int(id)
+        except Exception:
+            result[operation] = True
+            continue
         try:
             self.check_access_rule(operation)
         except exceptions.AccessError:


web_access_rule_buttons/models.py
```diff
@@ -25,6 +25,17 @@ def check_access_rule_all(self, operations=None):
...
+        # If we're writing on a new o2m, m2m or m2o record related to a new
+        # record (for example an invoice line created altogether with the
+        # invoice to which it belongs), ids contains a string. Calling
+        # check_access_rule() in this condition will crash, so we just
+        # blindly allow the operations.
+        try:
+            for id in self.ids:
+                int(id)
+        except Exception:
+            result[operation] = True
```
yajo, line 37: I personally don't use this addon, but this code seems a very beautiful attack vector. Can I send just `"somefield_ids": "muahahahahaaaa"` and skip security rules?

espo-tony: @Yajo It's not actually the case. This method is only called from the js when rendering the form view, in order to decide if the create / edit buttons should be visible or not. The dictionary returned by this method lists the user permissions to perform Odoo CRUD operations, and it's used to decide what buttons should be visible in the form view.

Therefore sending some junk data as in your example would only get you a dictionary saying that you're indeed allowed to write/create records and that you should see the related buttons. This doesn't affect at all what will happen when you'll try to perform an operation on the record, in which case security provided by Odoo will apply as usual.

LeartS: I would only catch `ValueError`, not all exceptions

PR #711, opened 2017-08-29
[9.0][FIX]web_advanced_search_x2x:check logical operator before adding element to the domain

zakiuu:
**Actual Behavior:** the `get_domain` method updates the domain returned by the search on the field, by _blindly_ prefixing each first element of a tuple with the field name. In another word if we have a domain a `product_id` field such `[('name', 'ilike', 'something'), ('default_code', 'ilike', 'something')]` the `get_domain` will return `[('product.name', 'ilike', 'something'), ('default.default_code', 'ilike', 'something')]`.
**The issue:** Sometimes the domain doesn't just contain tuples but it may contain logical operators such `|, &, !` ( exemple: `['|', ('product.name', 'ilike', 'something'), ('default.default_code', 'ilike', 'something')]`).
**The solution:** I've added a utilty method `is_logical_operator` to check wether a domain element is a logical operator or not. if the element is a logical operator add it as it is. Else prefix the first element of a tuple with the field name.

### DIFF
--- a/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
+++ b/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
@@ -123,18 +123,33 @@ odoo.define('web_advanced_search_x2x.search_filters', function (require) {
                 event.stopPropagation();
             });
         },
+        is_logical_operator: function(element){
+            // test whether a element is a logical operator ('|' or '&', '!')
+            if (element === '|' || element === '&' || element === '!') {
+                return true;
+            }
+            else{
+                return false;
+            }
+        },
         get_domain: function () {
             // Special way to get domain if user chose "domain" filter
+            var self = this;
             if (this.get_operator() == "domain") {
                 var value = this._x2x_field.get_value();
                 var domain = new data.CompoundDomain(),
                     name = this.field.name;
                 $.map(value, function (el) {
-                    domain.add([[
-                        _.str.sprintf("%s.%s", name, el[0]),
-                        el[1],
-                        el[2],
-                    ]]);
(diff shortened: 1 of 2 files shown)


hbrunn (approved):

web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
```diff
@@ -123,18 +123,33 @@ odoo.define('web_advanced_search_x2x.search_filters', function (require) {
...
                 var value = this._x2x_field.get_value();
                 var domain = new data.CompoundDomain(),
                     name = this.field.name;
                 $.map(value, function (el) {
-                    domain.add([[
-                        _.str.sprintf("%s.%s", name, el[0]),
-                        el[1],
-                        el[2],
-                    ]]);
+                    if (self.is_logical_operator(el)){
```
yajo, line 143: I'd have gone the `$.isArray(el)` way, but this seems OK too.

yajo (approved):

pedrobaeza:
Please change module version number.

@Yajo is this applicable also to v10 (I don't think so) or v8 (probable)?

yajo:
v8 uses the `isArray` system:
https://github.com/OCA/web/blob/d3a0e8b92ae3bc47c86bdca106c6f847949d320a/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js#L178

v10 is affected.

pedrobaeza:
Please make the corresponding PR then to v10.

zakiuu:
@pedrobaeza done!, and thx for the review

ThomasBinsfeld (approved):

Issue #697, opened 2017-08-16
[9.0][10.0] widget_x2many_2d_matrix: Widget doesn't refresh totals when changing amounts


pedrobaeza:
![imagen](https://user-images.githubusercontent.com/7165771/29371414-94ca201c-82a8-11e7-8884-4ec1a2bd62ca.png)

pedrobaeza:
Closed with #712

PR #712, opened 2017-08-30
[FIX] [10.0] 2d widget fixes

richard-willdooit:
Fixes display and input bugs in 2d widgets.

- dataset.read_ids needs fields otherwise it returns no data
- renderElement was loosing the change event handler
- add standard odoo classes for read / edit mode hiding of alternate values
- effective_readonly_change was switching same nodes twice

### DIFF
--- a/web_widget_x2many_2d_matrix/__manifest__.py
+++ b/web_widget_x2many_2d_matrix/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "2D matrix for x2many fields",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "author": "Therp BV, "
               "Tecnativa,"
               "Odoo Community Association (OCA)",
--- a/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
@@ -131,7 +131,7 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                 })
                 .then(function()
                 {
-                    return self.dataset.read_ids(self.dataset.ids).then(function(rows)
+                    return self.dataset.read_ids(self.dataset.ids, self.fields).then(function(rows)
                     {
                         // setup data structure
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Thanks for the patches. Maybe you can consider also the issue #697

hbrunn (approved):
how easy it is to overlook stuff in reviews...

richard-willdooit:
@pedrobaeza I actually don't use it with totals in my implementation.  I don't think the totals render very well in version 10, but decided not to try and add this "enhancement", and I just did the fix to ensure correct recomputing.

pedrobaeza:
Thanks again, Richard, can you please make a last change that is modifying module version number for reflecting the evolution of the module?

pedrobaeza (approved):
Can I squash at merge? As the only way to try here on runbot is to merge the PR and rebuild runbots that use it, I will merge. I would like to have demo data inside the module, but that would require to add code that can't be avoided when no demo data. @hbrunn do you have any possible idea?

hbrunn:
first I thought you mean something like [this](https://github.com/OCA/server-tools/blob/8.0/field_rrule/tests/test_field_rrule.py#L40), but now I understand you mean adding something that causes a field/wizard/view exist with demo=1 so that we can check stuff in runbot simpler?

Never tried this, but I had to [detect runbot](https://github.com/OCA/knowledge/blob/8.0/document_sftp/hooks.py#L17) before, and maybe we can say

```python
import socket
import os
if socket.getfqdn().endswith('odoo-community.org'):
    import my_demo_model
```

in `models/__init__.py`?

pedrobaeza:
@hbrunn thanks for the suggestion, but I don't really like a lot that because that will mean that I can't try in other DB except OCA runbot. I will think in this in my spare time and see if there's other option.

hbrunn:
if you always want to add stuff in demo mode, using a hack like https://github.com/OCA/server-tools/pull/876/files#diff-ded7038d0dcfffc218c449b206263ebfR38 to get the current database's cursor, and then check if it has demo data enabled might work

pedrobaeza:
That's something I like more :smile: but as said, to be tried on my spare time, which is very little...

PR #714, opened 2017-08-31
[IMP] web_responsive: Add search feature

lasley:
* Add feature to search menus in the app drawer

Rough sketch at the moment, but it does work

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -6,6 +6,7 @@ odoo.define('web_responsive', function(require) {
 
     var Menu = require('web.Menu');
     var Class = require('web.Class');
+    var DataModel = require('web.DataModel');
     var SearchView = require('web.SearchView');
     var core = require('web.core');
     var config = require('web.config');
@@ -61,10 +62,32 @@ odoo.define('web_responsive', function(require) {
 
     var AppDrawer = Class.extend({
 
+        /* Provides all features inside of the application drawer navigation.
+
+        Attributes:
+            directionCodes (str): Canonical key name to direction mappings.
+            deleteCodes
+         */
+
         LEFT: 'left',
         RIGHT: 'right',
         UP: 'up',
         DOWN: 'down',
+
+        // These keys are ignored when presented as single input
+        MODIFIERS: [
+            'Alt',
+            'ArrowDown',
+            'ArrowLeft',
+            'ArrowRight',
+            'ArrowUp',
+            'Control',
+            'Enter',
+            'Escape',
+            'Meta',
+            'Shift',
+            'Tab',
+        ],
 
         isOpen: false,
         keyBuffer: '',
@@ -72,8 +95,10 @@ odoo.define('web_responsive', function(require) {
         keyBufferTimeoutEvent: false,
         dropdownHeightFactor: 0.90,
         initialized: false,
+        searching: false,
 
         init: function() {
+
             this.directionCodes = {
                 'left': this.LEFT,
                 'right': this.RIGHT,
@@ -84,6 +109,10 @@ odoo.define('web_responsive', function(require) {
                 '+': this.RIGHT,
                 '-': this.LEFT,
             };
+            this.$searchAction = $('.app-drawer-search-action');
+            this.$searchAction.hide();
+            this.$searchResultsContainer = $('#appDrawerSearchResults');
+            this.$searchInput = $('#appDrawerSearchInput');
             this.initDrawer();
             this.handleWindowResize();
             var $clickZones = $('.odoo_webclient_container, ' +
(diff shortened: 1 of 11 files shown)


pedrobaeza:
I see that you are not too much confident about Odoo moving its enterprise interface to community, eh!

lasley:
😆  yep definitely not holding my breath for that. Plus I noticed we're at over 2k downloads in the apps store if you combine v9 & 10, so I figure it's a worthy cause

pedrobaeza:
And that's without counting source code downloads. I think it's probably the most popular addon.

lasley:
My thoughts, exactly. Last I checked, we're included in nearly all of the backend themes on the market - free & paid. 

I'm watching and waiting to see what Flectra has to offer in terms of their UI. We may be able to take some inspiration from the _hopefully_ compatible licensed code

pedrobaeza:
Let's see, but I see today a teaser video of the interface and I don't like too much (colorful icons instead of plain ones).

lasley:
True true, and they didn't show what I'm truly interested in - the responsiveness of the views.

lasley:
Alright so functionally I think I have all the pieces here now, it just looks like crap and isn't tested. I'll knock that out in the next few days

yajo (changes requested):
I have no idea what to test in runbot, please add usage instructions 😅

yajo (changes requested):
I found some UX problems:

---

In Firefox, if you have [Search text while I type](https://support.mozilla.org/en-US/kb/search-contents-current-page-text-or-links?redirectlocale=en-US&redirectslug=Searching+within+a+page#w_quick-find) enabled, it does not trigger the filtering. You could say it's expected, but actually you can use gmail/github shortcuts while that option is enabled.

---

It always writes in upper case, so I instinctively pressed shift lock and got this result:

captura de pantalla de 2017-09-04 11-34-19

Pressing ctrl+backspace to delete a whole word does the same.

---

When I search for "users", I get both *Settings/Users* and *Settings/Users/Users*. If I click on the 1st element, nothing happens. It's because it's a menu item without an attached action, so IMHO those results should not appear:

captura de pantalla de 2017-09-04 11-34-41

pedrobaeza:
@lasley thanks for improving the module with this useful feature. The only problem I find, apart from @Yajo's comments, is that the textbox doesn't appear at the beginning, but only when you type the first letter, so this makes that you can't use the feature on mobiles, as you don't have the virtual keyboard available.

There's another problem, but I think not directly related with this PR: the scroll of the menu doesn't work correctly on mobiles: dragging only scroll a bit the menu.

yajo:
Well... I don't think people using phones prefer typing over clicking...

pedrobaeza:
@Yajo you can't tap if you don't have the option to tap: you only have access to main apps. There can be a button in mobiles for showing the textbox.

dreispt:
Great feature Dave!

I did a test drive using Firefox desktop and noticed a few usability issues:

* The text in the search box is all upper case
* Press backspace with an empty text box and you go back to the website homepage (very annoying)
* If you press "Caps Lock", "SHIFT+Tab" or "CTRL+anykey" (such as CTRL+V) you get a unicode square symbol instead of the intended action.

lasley:
Thanks for the testing everyone! You all unearthed basically every problem I had swept under the rug and hoped nobody would notice, and some that I had no idea existed.

I'm going to switch the strategy up a bit here. At the moment, the text box is just superficial & is actually being filled by the key buffer that was already being maintained in order to allow for navigating between the app icons. 

I'm going to instead make the keys pass directly through to the text box & trigger my events off of that. That's admittedly the better route from the start, but it was easier to cheat with the keybuffer as a PoC. But then after I got to PoC, I got the whole lazy engineer curse and shrugged it off!

Mobile - we have some grey area on the right 15%ish of the screen. I think the solution here is that I will add the standard magnifying glass glyph at the top-right to serve as a search action and trigger the unhiding/focusing of the search field. Easy enough!

Thanks for the tip about the lacking actions @Yajo. I was totally wondering why some of the links weren't working, just hadn't had time to walk through the specifics.

pedrobaeza:
What is the status of this?

lasley:
Status is: it pissed me off so I tabled it to come back with a level head. Planning on hitting it again sometime this week

pedrobaeza:
Thanks for the update.

lasley:
New status: Oops I forgot

lasley:
Alright this is functional again, and works significantly better than the PoC. All noted issues in above reviews have been fixed (I think/hope). Still to go: tests

mgielissen:
Great feature, I've tested it with backend_theme_v10.

screenshot-10 0 10 147-8069-2017-10-06-23-45-05-490

lasley:
@mgielissen - thanks for the functional review! 

Is that additional padding at the top of the search results? I had that same issue here and solved in with https://github.com/OCA/web/pull/714/commits/37810e16cc4b5aab7ee35432dbf2c23f14857d0f#diff-5601720e3b0255535e4075cc46503e13R87 - but I find it interesting it came back up in your theme. What browser are you in?

mgielissen:
I'm using Chrome. The default in web_responsive is padding-top: 54px; I use 80px to add more space between the navbar and icons.

With 54px it looks like this:

screenshot-10 0 10 147-8069-2017-10-07-00-02-52-847
screenshot-10 0 10 147-8069-2017-10-07-00-02-17-641

mgielissen:
Video preview: https://youtu.be/Z5njVc7gWyQ

yajo (changes requested):
Tested on runbot:

1. When app drawer is closed, then re-opened and you start typing again, it does not start from scratch, but continuing with last search. It's confusing.
1. In Firefox, if you enable *Settings > General > Navigation > Search while I type*, the feature does not work fine.

lasley:
> When app drawer is closed, then re-opened and you start typing again, it does not start from scratch, but continuing with last search. It's confusing.

Hmmm so I thought a lot about this (you're talking about a "feature" I implemented right now) - I was somewhat mirroring the way that we see this done in OS search features (Windows Search, Mac Spotlight). I consistently drop out of my search to look at something, just to go back in to complete the search. 

I think this would be useful here, particularly because I plan on making the search modular before this merges. Basically I want the ability for users to create glue modules that add into this search feature in order to add functionality - think everything from calculators, to searching for records.

So maybe my implementation is off? I'm playing with the OS search features, and:

* Clicking outside of the search does not clear it
* Pressing "ESC" inside of search clears it

Maybe we need some similar logic? 

> In Firefox, if you enable Settings > General > Navigation > Search while I type, the feature does not work fine.

Hrmmmm isn't this feature by definition at odds with what we're trying to do here? I assume that this issue doesn't happen in, say, website editor?

I effed my Firefox install the other day while making a malicious add-on PoC, so I guess I'll need to fix that and test 😆

yajo:
> I consistently drop out of my search to look at something, just to go back in to complete the search.

Not sure what you mean... But I guess your search input remains on screen, right?

I use GNOME here and when I search anything and go back with no action, the search bar gets emptied for the next time. It makes most sense IMHO, since each search is either targeted to go somewhere or nowhere, but not to prefill next search.

Right now in the addon:

1. I'm on "Inventory".
2. I want to go to "Sale", so `Shift+Alt+A` and type "Sal" but then I remember I had to do something else before leaving. Go back.
3. I finish my stuff, so my muscle memory repeats: `Shift+Alt+A` and type "sal" but then I get "salsal". Confusing.

If the search box + results were there when I opened the drawer again, then it would make some sense, but in any case it seems more natural (at least to me) to start from scratch each time I open the drawer.

> Hrmmmm isn't this feature by definition at odds with what we're trying to do here? I assume that this issue doesn't happen in, say, website editor?

Indeed, and it doesn't happen either in Github when you're on a repo and press `t` to go to a file or `s` to search.

pedrobaeza:
I also agree that consecutive searches when you close drawer shouldn't contain previous typed characters. I have also noticed that access key + A is no longer working for unfolding the drawer.

lasley:
> I have also noticed that access key + A is no longer working for unfolding the drawer.

Can you double check that this is still an issue in the last commit? Assuming it happens again, can you open your web console, try to trigger the app drawer open by keyboard, and send me the output from the console? It should just be a bunch of keys looking something like:

```
web_responsive.js:232 Redirecting keyup for å
web_responsive.js:219 å
web_responsive.js:219 Alt
web_responsive.js:219 Control
```

pedrobaeza:
Thanks for the changes. Now it's working flawlessly on computers, but on my Chrome Android mobile, these are the issues I find:

* Although there's an icon of the glass and I tap on it for activating search, there's no text box nor virtual keyboard is open when I tapped on the icon.
* The scroll doesn't work correctly when there are more menus than the admitted by the screen size. There's a little movement, but not fully and it returns back to the original position.
* Now the header misplacement issue of #756 is more critical, as you may need to tap on the icon. But not only that, it's held on the background instead on top.

lasley:
> The scroll doesn't work correctly when there are more menus than the admitted by the screen size. There's a little movement, but not fully and it returns back to the original position.

To confirm, you mean the menus that are being searched right? Like if the list spans further vertically than the screen?

pedrobaeza:
Yeah, exactly. It serves as example when you haven't type anything, that all menus are shown. That won't really fit in the screen for sure.

pedrobaeza:
Any news with this?

lasley:
Ummm yeah so I'm having a lot of trouble getting this to work with the `Search while I type` feature without completely JavaScript-ifying the menu elements to create a fake focus mechanism. I think the entire apps/menu navigation will need to be rewritten too.

The issue at hand is that the menu navigation by arrow key requires that the menu element itself has actual focus. This is what creates the highlight on the menus, and allows the enter key to actually select it. 

With this accessibility feature enabled, there seem to be some certain conditions in which Firefox internally decides to disable it. These conditions all seem to be around the actual object that has browser focus, and whether the browser deems it to be editable. I cannot find any documentation to confirm any of this, it's just been through a bunch of testing.

The problem is that we need two elements focused at a time then - one is the input that needs to receive the keyboard events for searching, and the other are the menu items for navigation. Event redirection does not seem to be captured in the disable conditions, either. I've been stuck at trying to tackle this problem, but I just end up pissed off and working on something else. Short of new external ideas, this cycle will be rinse/repeated until I hopefully come up with something.

We could optionally just add to roadmap and I can finalize this. I doubt that it gets fixed once merged though.

lasley:
Pretty sure I solved the Search While I Type, as well as the lack of onscreen keyboard in the last commit 0e8998b

pedrobaeza:
Is everything ready to be tried?

lasley:
Still left on the to-do is getting the search results to scroll, but I'd love some outside confirmation on the following:
* Does the Android keyboard show up when the search link is clicked in the App Drawer?
* Did I screw up any of the keybindings?
* Anything else you find not related to search results scrolling

Still left for me to test, but external test is good too:
* Does Firefox behave properly with `Search While I Type`?

pedrobaeza:
> Does the Android keyboard show up when the search link is clicked in the App Drawer?

I'm afraid not: buttons appear misaligned, and when I get to tap in the glass, the textbox appears, but when I tap inside, the keyboard appear for a second to disappear afterwards and folding the app switcher.

> Did I screw up any of the keybindings?

This appears to be correct

> Anything else you find not related to search results scrolling

Can't try more on mobile, but on Firefox desktop, the mouse scroll doesn't work and there's no scrollbar.

> Does Firefox behave properly with Search While I Type?

It seems so, except one thing: first letter you type is not considered for the search.

lasley:
Thanks for the review Pedro - I'm pretty close here I think. I should have a few more commits today that should resolve everything & will notify when ready 🚀

lasley:
To anyone curious, or potential Googlers - the fix to prevent the Firefox  `Search While I Type` feature from triggering is simple and completely obvious in hindsight: `event.preventDefault();`

lasley:
Alright pretty sure I've fixed all outstanding issues, as well as a few ones that hadn't been identified. This is ready for review.

pedrobaeza:
Cool! Testing then

pedrobaeza:
@lasley what a great improvement right now!

There are still a problem with Android, when clicking on the textbox still makes to disappear the keyboard after a flash and folding the app switcher.

And 2 things for making it perfect:

* On mobiles, the layout of the icons is very ugly:
  <details>

  screenshot_20171130-002549

  </details>

  Maybe you can hide the term "Applications" on mobiles or make the logo smaller on small devices

* This is not from the search, but I have found it right now: if an app title takes 2 lines, there's a missing tile on the app switcher:

  <details>

  screenshot_20171130-002619

  </details>

lasley:
> There are still a problem with Android, when clicking on the textbox still makes to disappear the keyboard after a flash and folding the app switcher.

Just to confirm, this is after clicking the search icon and then subsequently clicking the search query text input?
 
Is it also accurate to infer that the text box is not focusing by default on Android, immediately after clicking the search link?

Is this Chrome on Android?

> On mobiles, the layout of the icons is very ugly:

Good idea - done

> This is not from the search, but I have found it right now: if an app title takes 2 lines, there's a missing tile on the app switcher

Done.

pedrobaeza:
> Just to confirm, this is after clicking the search icon and then subsequently clicking the search query text input?

Yeah, that's it

> Is it also accurate to infer that the text box is not focusing by default on Android, immediately after clicking the search link?

Indeed, that's not happening on Firefox on Android, but it does on Chrome on Android

> Is this Chrome on Android?

Chrome and Firefox

I have remembered another thing that can be improved: on a mobile, you unfold menu list, and tap on a menu, then the menu list is not unfolded, being a bit uncomfortable.

lasley:
> I have remembered another thing that can be improved: on a mobile, you unfold menu list, and tap on a menu, then the menu list is not unfolded, being a bit uncomfortable.

Done 76cfa4d

pedrobaeza:
Wow, this is great, my own wishlist! :stuck_out_tongue_winking_eye:

web_responsive/static/src/js/web_responsive.js
```diff
@@ -61,19 +62,43 @@ odoo.define('web_responsive', function(require) {
 
     var AppDrawer = Class.extend({
```
yajo, line 63: Shouldn't this inherit from `Widget`?

web_responsive/static/src/less/main.less
```diff
@@ -100,3 +100,48 @@ main {
...
+
+.app-drawer-scrollbar > div {
+	position:absolute;
+	z-index:100;
+	width:100%;
+
+	/* The following is probably what you want to customize */
+	background:-webkit-gradient(linear, 0 0, 100% 0, from(#f00), to(#900));
+	background-image:-moz-linear-gradient(top, #f00, #900);
+	background-image:-o-linear-gradient(top, #f00, #900);
```
yajo: I miss the standard form

yajo (approved):
Some little code details, while reading I realised we should have made a `Widget` in the first place, which should in turn allow proper unit testing in v11+. 😞 

Well, not asking for that, just pointing it out. It could be a ton of job.

web_responsive/static/src/js/web_responsive.js
lasley: I'm almost positive I tried this as a widget originally, but I can't remember the problem. I don't think I want to expand this scope to that though TBH. 

Roadmap?

web_responsive/static/src/less/main.less
lasley: I'm not following?

Note though, I'll be deleting all of this (Vertical Scrollbar LESS section). It was testing code that I forgot to remove once I figured out how to actually make the scrollbar appear

lasley:
I'm having a lot of trouble with the Android side of things. I've spent a few hours trying to get Android emulator to actually start a damned emulator, with horrible results. Basically all Google docs are incorrect because they are out of date (typical Google). I feel sorry for Android developers and anyone with an Android.

The above said- I cannot fix the issue that I cannot troubleshoot. I'm unfortunately going to just add this to the roadmap.

_Edit - iPhone works:_

  <details>

image

</details>

pedrobaeza:
Ouch, that's bad, but if you don't have possibilities... @Yajo, maybe you can check.

lasley:
Yeah I'm actually quite mad at Google for this right now. It seems there's a bunch of hidden dependencies and licensing and stuff, with basically zero real indication of why the emulator fails to launch. It's almost hilarious how bad this all is, particularly coming from iPhone development.

I edited my above comment to include a screenshot of the search working on iPhone. Best I can do is prove it works on my own devices unfortunately 😆

rousseldenis:
@lasley I've not followed the whole thread but I've just tested on runbot with chrome on Android device (on nougat) and the search menu disappear directly. In order: click the search button, the Android virtual keyboard appears and then the whole responsive menu disappears.

So, impossible to search.

lasley:
Hey @rousseldenis - thanks for the info. Unfortunately I can't replicate this on any of my devices & the Android Emulator is horribly broken, so I have to leave this issue up to someone in the community with an Android and the ability to fix it. 

For lack of a PR fixing this, I am proposing a merge with the issue in place and the roadmap items that were added in 45af3c4

web_responsive/static/src/js/web_responsive.js
yajo: Roadmap 👍 

If we can't do it is because we don't know or have time, but for sure it should be.

yajo:
I'll do the Android job 😊

lasley:
Thanks for taking this one on @Yajo. I'm really curious what the fix will be.

lasley:
Hrmmm it looks like the OCA/10.0 was merged into this branch or something, which was causing me major issues with my deployer. I had to do some git wizardry, but I think we're good here. I squashed at the same time seeing as I destroyed the timeline anyways.

melroy89:
Nice! Could you also merged it later to 11.0 branch?

melroy89 (approved):

pedrobaeza:
@Yajo are you going to work soon in fixing Android view? If not, I propose to merge this and put the known issue annotation.

lasley:
I already added a notation in the ReadMe when I gave up, so we could merge this as is IMO

melroy89:
Yea! Nice

pedrobaeza:
This can be put in v11/v9.

melroy89:
@lasley Can be merged to 11.0 branch 👍

PR #716, opened 2017-09-05
[FIX][web_advanced_search_x2x] Fix "false" label

### LINKED ISSUE #717: web_advanced_search_x2x doesn't work with contains filter
I am using this module and I really like it but it seems that it contains a bug.

On a model if you want to search on a many2one with a filter "contains" a string.
The string is always replaced by false.

Does anyone have the same problem ?

yajo:
When you were adding a custom filter in a x2x field using "contains", it always became "contains 'false'". Now it is fixed.

Fixes #717 

@Tecnativa

### DIFF
--- a/web_advanced_search_x2x/__manifest__.py
+++ b/web_advanced_search_x2x/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "Search x2x fields",
-    "version": "10.0.2.0.0",
+    "version": "10.0.2.0.1",
     "author": "Therp BV, "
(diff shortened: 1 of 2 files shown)


hbrunn (approved):
this fixes https://github.com/OCA/web/issues/717, right?

rafaelbn (approved):
tested in runbot 👍

pedrobaeza (approved):
Tested on runbot

Cedric-Pigeon:
:+1: Thanks

PR #732, opened 2017-09-18
[ADD] Add support for swipe between form records on mobile devices

wpichler:


gonzaloruzafa (commented):
Hi, tried to swipe between forms on iphone 6 SE with runbot instance, could not make it work, any suggestion? I am reviewing

wpichler:
please try the used js lib directly here: http://hammerjs.github.io/#try-it - does it work on iPhone.

I do not own an iphone - so not tested with iphone - but it should work

wpichler:
i've checked the runbot installation - module is not installed there - and not available.

How to fix this ?

web_gestures/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
wtaferner: Editor specific stuff should be removed

web_gestures/static/src/js/gestures.js
```diff
@@ -0,0 +1,29 @@
+odoo.define('web_gestures', function (require) {
```
wtaferner: The whole module should be named in singular, please rename it to ```web_gesture```

wtaferner (changes requested):
Nice contribution IMO, the reason for not working on runbot is the missing ```__manifest__.py```

web_gestures/__init__.py
wpichler: done

web_gestures/static/src/js/gestures.js
wpichler: done

PR #735, opened 2017-09-27
[10.0][IMP] web_timeline improvements

Creamaster:
* Update the vis.js framework to latest stable version (v4.20.1) to add new functionalities and save almost
400ko of source code in odoo assets (the vis dev team has isolated the code for timeline widget from the rest of their widgets)

* Add the possibility to display events with no duration (instantaneous) as the same time has regular events, exemple:
image

* Add the possibility to choose between two scroll mode :
  -  zoom in time resolution (default)
  or
  -  vertical scroll (web browser default behaviour), exemple on heavy clustured view:
scroll_vertical_exemple


Creamaster:
In the code, I stumbled on a variable 'all_day' the seems to never be initialize and consequently the test based on this variable could never be triggered, is this an old feature ? Sould this code be updated ?

web_timeline/static/src/js/web_timeline.js line:200
```javascript
all_day = this.all_day ? evt[this.all_day] : false,
```

Creamaster:
Also, I parsed the javascript file with my linter, I could revert all the non important changes caused by the reformat, If needed

web_timeline/static/lib/vis/vis-timeline-graph2d.min.js
```diff
@@ -0,0 +1,40 @@
+/**
```
lmignon, line 1: @Creamaster PLZ We must keep the minified version of the JS and CSS files to ease the debug. All the js and css files are automatically minified by Odoo when the application run in normal mode. But once you activate the debug mode, you get the original file and it's very useful to ease the debug.

web_timeline/static/src/js/web_timeline.js
```diff
@@ -62,12 +62,17 @@ odoo.define('web_timeline.TimelineView', function (require) {
         //     });
         // },
 
-        parse_colors: function(){
-            if(this.fields_view.arch.attrs.colors) {
-                this.colors = _(this.fields_view.arch.attrs.colors.split(';')).chain().compact().map(function(color_pair) {
+        parse_colors: function () {
```
lmignon, line 65: @Creamaster PLZ you must avoid to change the formatting of the file even if this one is not conform to your criteria. It's hard to dissociate cosmetic changes from the changes required by your PR. These cosmetic changes should at least be done into a separate commit.

lmignon (changes requested):
@Creamaster Thank you for this enhancement. Comments done into the code

web_timeline/static/lib/vis/vis-timeline-graph2d.min.js
Creamaster: That was my intent, but the non-minified version isn't distributed anymore : 
https://github.com/almende/vis/tree/master/dist

web_timeline/static/src/js/web_timeline.js
pedrobaeza: I think this is for complying with JS linter, which is correct, but as said, better to keep it in a separate commit.

web_timeline/static/lib/vis/vis-timeline-graph2d.min.js
lmignon: @Creamaster Ok

web_timeline/static/src/js/web_timeline.js
Creamaster: As requested, done in separate commit

pedrobaeza:
@Creamaster thanks for the welcome improvement. Can you check https://github.com/OCA/project/issues/317 to see if you can fix it at the same time with the improvement?

lmignon (approved):
Thank you for the split of your commits!
:+1:  Code Review

Creamaster:
@pedrobaeza I will update the code to fix the [issue 317](https://github.com/OCA/project/issues/317)

pedrobaeza:
Thank you, @Creamaster 

https://github.com/OCA/project/issues/316 is also related and give you some hints I think.

Creamaster:
A (long) gif to demonstrate the timeline with the added 'date_delay' attribute

I added two datetime fields on the defaut project.task model, and changed the views as follow :

| Calendar view  | Timeline view |
| ------------- | ------------- |
| date_start="x_date_start" date_delay="planned_hours"  | date_start="x_date_start" date_stop="x_date_stop" date_delay="planned_hours" |


demo_timeline

pedrobaeza (approved):
You have made an excellent work, thanks.

davinci41:
I noticed a mistake. How to set the start time and duration of the task, and in the calendar and in the timeline view is displayed well. In the timeline you can edit start time and duration. The calendar is displayed correctly. As in the calendar, I will move the task to the next day in the timeline so it displays:
image
Editing in the calendar, the program does not change the end date

PR #740, opened 2017-10-06
[MIG] [11.0] web_responsive

angelmoya:
Hi @lasley I only update module version and tested functionaly works, and no errors logs.

It's works fine


mgielissen:
The module works fine under 11.0.

angelmoya:
Fails javascript pylint, I will check it

lasley:
Hooray I love it when the changes are minor. Guess I need to get around to finishing #714 so we can have the nifty search too.

web_responsive/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "name": "Web Responsive",
     "summary": "It provides a mobile compliant interface for Odoo Community "
                "web",
-    "version": "10.0.1.2.0",
+    "version": "11.0.1.o.0",
```
yajo: o != 0

yajo (changes requested):

yajo (approved):

yajo:
Approved, but you have to fix branch conflicts and runbot.

web_responsive/views/web.xml
```diff
@@ -0,0 +1,254 @@
...
+-->
+
+<odoo>
+
+    <template id="webclient_bootstrap"
+              inherit_id="web.webclient_bootstrap"
+              name="App Drawer - Web Client"
+              >
+
+        <xpath expr="//div[@class='o_sub_menu']" position="replace" />
```
lasley: Looks like Odoo added some linting to the Qweb:

>2017-10-07 08:32:01,969 150 WARNING openerp_test odoo.addons.base.ir.ir_ui_view: Error-prone use of @class in view App Drawer - Web Client (): use the hasclass(*classes) function to filter elements by their classes

Simple change, because it seems they also added a `hasclass` shortcut:
```
//div[@hasclass('o_sub_menu')]
```

vs it used to be:

```
//div[@contains('class', 'o_sub_menu')]
```

lasley: IMO we should port this change backwards too

pedrobaeza:
This should include also search feature that is in progress right now in #714

mgielissen:
The V11.0 section is available in the Odoo appstore. I want to add a theme which depends on web_responsive but it is not yet available: 

> backend_theme_v11: unmet dependency web_responsive for series 11.0

Is there a workaround?

lasley (approved):
I think we can add search in a subsequent update. There's still some ways to go there, I think.

@angelmoya - can you squash your commits please?

pedrobaeza:
@mgielissen this afternoon I wasn't able to add a repository for 11.0. Any of the OCA repos is still available for 11.0 thus.

mgielissen:
It's active since a couple of hours

pedrobaeza:
OK, merging this one as is and adding this repo to apps platform right now.

lasley:
I'll take responsibility for the search cross ports when they happen. I'd like it in 9 as well

pedrobaeza:
@angelmoya Travis is giving an error on branch 11.0 after merging this module. Can you please check?

PR #745, opened 2017-10-10
[8.0][ADD] web_form_sticky_header

daramousk:


pedrobaeza:
Fixed = sticky?

Any screenshot?

daramousk:
@pedrobaeza Yes, the header is static, attached a screenshot

pedrobaeza:
I prefer then the name `web_form_sticky_header`.

daramousk:
@pedrobaeza Renamed the module

web_form_sticky_header/static/src/js/web_form_sticky_header.js
```diff
@@ -0,0 +1,24 @@
+//-*- coding: utf-8 -*-
```
NL66278: This is only for python files, not javascript

daramousk: @NL66278 Yes wrong vim keymap, I will reimplement this module

web_form_sticky_header/static/src/js/web_form_sticky_header.js
```diff
@@ -0,0 +1,24 @@
...
+
+openerp.web_form_sticky_header = function(instance)
+{
+    instance.web.FormView.include({
+        start: function() {
+            self = this;
+            return this._super().then(function() {
+                var header = self.$el.find('header')
+                header.css({'background-color': '#ededed;'})
+                header.insertAfter(self.$el.parents().children('.oe_view_manager_header'))
```
NL66278: Please give some comments here on what is happening here, and how this leads to a fixed header.

daramousk:
@NL66278 Reimplemented using CSS.

web_form_sticky_header/static/src/css/sticky_header.css
```diff
@@ -0,0 +1,8 @@
+.oe_form header {
```
NL66278, line 1: The original specifies it must be within an openerp class. Should that not be also done for this one?

NL66278 (approved):
👍 LGTM and tested. One remark to look at, but not critical. Big leap in usability!

daramousk:
@NL66278 This still has one issue, if you click on the More menu the bar overlaps because of the z-index, will need to find a way around this

NL66278:
And it is the same problem for print and attachment menu's. So calls for
some general solution.

web_form_sticky_header/views/templates.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <template id="assets_backend" name="web_form_sticky_header assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
```
gfcapalbo, line 5: according to what we discussed today tthis would place it last. equivalent to  expr="//link[@href=/base/static/src/css/modules.css]"   , you are injecting in https://github.com/OCA/OCB/blob/8.0/addons/web/views/webclient_templates.xml#L90 .   looks ok.

web_form_sticky_header/static/src/css/sticky_header.css
gfcapalbo: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/css/base.css#L1922

I believe ronald is right here

change with .openerp .oe_form header {

web_form_sticky_header/static/src/css/sticky_header.css
```diff
@@ -0,0 +1,8 @@
...
+    position: sticky;
```
hbrunn, line 2: I was not so sure about this one, but https://caniuse.com/#search=sticky says this is a thing. TIL, thanks!

daramousk:
@hbrunn @gfcapalbo @NL66278  Everything should work perfectly now

hbrunn (approved):
nice!

In the chat, I also proposed to add

```css
div.modal-header {
    position: sticky;
    bottom: 0px;
    background: white;
    right: 0px;
    left: 0px;
}
```

to have a similar function for big popup windows.

but as that's a feature request, take or ignore at your convenience

hbrunn:
PS: don't forget to rebase

lfreeke:
:+1: (test)

yajo (approved):
Code OK

yajo:
I feel this will harm phone users, but since it's already unusable, merging...

PR #749, opened 2017-10-11
[10.0][IMP] web_timeline change the option to initialize the timeline window on display.

Creamaster:
Change the attribute '**default_window**' to '**mode**' like the Odoo calendar view (to be iso functional)

### DIFF
--- a/web_timeline/static/src/js/web_timeline.js
+++ b/web_timeline/static/src/js/web_timeline.js
@@ -55,13 +55,6 @@ odoo.define('web_timeline.TimelineView', function (require) {
             }
         },
 
-        // set_default_options: function(options) {
-        //     this._super(options);
-        //     _.defaults(this.options, {
-        //         confirm_on_delete: true
-        //     });
-        // },
-
         parse_colors: function () {
             if (this.fields_view.arch.attrs.colors) {
                 this.colors = _(this.fields_view.arch.attrs.colors.split(';')).chain().compact().map(function (color_pair) {
@@ -101,7 +94,6 @@ odoo.define('web_timeline.TimelineView', function (require) {
             this.$el.addClass(attrs['class']);
 
             this.info_fields = [];
-            this.mode = attrs.mode;
 
             if (!attrs.date_start) {
                 throw new Error(_t("Timeline view has not defined 'date_start' attribute."));
@@ -111,7 +103,7 @@ odoo.define('web_timeline.TimelineView', function (require) {
             this.date_delay = attrs.date_delay;
             this.no_period = this.date_start == this.date_stop;
             this.zoomKey = attrs.zoomKey || '';
-            this.default_window = attrs.default_window || 'fit';
+            this.mode = attrs.mode || attrs.default_window || 'fit';
 
             if (!isNullOrUndef(attrs.quick_create_instance)) {
                 self.quick_create_instance = 'instance.' + attrs.quick_create_instance;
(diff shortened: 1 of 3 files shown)


Creamaster:
This PR contains a commit I initially intended to include in a previous PR ([#735](https://github.com/OCA/web/pull/735)). But was forgotten in my local branch, while I was working on a rework of the group_by functionality, to address the issue [OCA/project/#316.](github.com/OCA/project/issues/316)

This PR is a very small improvement/correction and can be rejected as the time needed to validate a PR could be greater than it's gain.

With this change, now a display mode 'day' show the current day, not the next 24h

pedrobaeza:
Please change module version number, and of course this is very welcome!

Creamaster:
Module version number updated 10.0.1.1.0 -->  10.0.1.2.0

pedrobaeza (approved):
@Yajo please review

web_timeline/static/src/js/web_timeline.js
```diff
@@ -111,7 +104,7 @@ odoo.define('web_timeline.TimelineView', function (require) {
             this.date_delay = attrs.date_delay;
             this.no_period = this.date_start == this.date_stop;
             this.zoomKey = attrs.zoomKey || '';
-            this.default_window = attrs.default_window || 'fit';
+            this.mode = attrs.mode || 'fit';
```
yajo: The rest of the patch is OK, but I see no reason for breaking backwards compatibility by changing the option name. Can't we leave the `default_window` name as it was before?

web_timeline/static/src/js/web_timeline.js
```diff
@@ -172,25 +165,27 @@ odoo.define('web_timeline.TimelineView', function (require) {
                 onRemove: self.on_remove,
                 zoomKey: this.zoomKey
             };
-            if (this.default_window) {
-                var start = new moment();
-                var end;
-                switch (this.default_window) {
+            if (this.mode) {
+                var start, end = false;
```
yajo: Any reason why `start` defaults to `undefined` and `end` defaults to `false`?

yajo (changes requested):

web_timeline/static/src/js/web_timeline.js
Creamaster: I added the attribute 'default_window' just before this PR. this PR intend to change the 'default_window' behavior, to a more usable one.
The new parameter is the same as the default Odoo one (same functionality), and to facilitate the use of the timeline view, it seems best to use the same attribute name for the same behavior in similar view ([see Odoo calendar view default view mode](https://github.com/odoo/odoo/blob/10.0/addons/web_calendar/static/src/js/web_calendar.js#L127))

I will change the attribute to be backwards compatible, using both syntax, like : 
```javascript
this.mode = attrs.mode || attrs.default_window || 'fit';
```
Is it okay ?

web_timeline/static/src/js/web_timeline.js
Creamaster: No particular reasons, I often declare "null" variables like that for readability.
I will change it with :
```javascript
var start = false, end = false;
```
To be more explicit

web_timeline/static/src/js/web_timeline.js
yajo: In such case, then please break backwards compatibility 😊

web_timeline/static/src/js/web_timeline.js
yajo: `undefined` is a falsey value, so if it doesn't really matter, just do `var start, end;`

simahawk (approved):

simahawk:
@Yajo sounds like your comments were addressed: can you check again?

yajo (approved):

PR #753, opened 2017-10-12
[IMP] Up version to work with version 10

jo541:
Upgrade the JS and XML code to work with V10.

### DIFF
--- a/web_widget_datepicker_options/__manifest__.py
+++ b/web_widget_datepicker_options/__manifest__.py
@@ -32,6 +32,4 @@
     ],
     "author": "Vincent Vinet, "
               "Odoo Community Association (OCA)",
-    'installable': False,
-    "active": False,
 }
--- a/web_widget_datepicker_options/static/src/js/datepicker.js
+++ b/web_widget_datepicker_options/static/src/js/datepicker.js
@@ -18,36 +18,34 @@
 */
 /*global openerp, _, $ */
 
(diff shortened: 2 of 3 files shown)


yajo (changes requested):
Hi there @jo541, thanks for your contribution, although AFAICS this addon is simply no more needed in v10, since you can just add any options you want directly in the `options` attribute of the `<field>`. [Those get expanded with default values](https://github.com/odoo/odoo/blob/25725eae578d126f79c69f72a7a4e3b4136b4e5d/addons/web/static/src/js/widgets/date_picker.js#L25-L39) and [then passed to the datetimepicker BS widget](https://github.com/odoo/odoo/blob/25725eae578d126f79c69f72a7a4e3b4136b4e5d/addons/web/static/src/js/widgets/date_picker.js#L43), so there's nothing that this addon is actually adding.

I'd be glad to merge the PR if it just removes the addon 😊

jo541:
Hi @Yajo, Thank you for you answer :). Yes, I see the options passed directly by the standard of Odoo. But [In this piece of code](https://github.com/odoo/odoo/blob/25725eae578d126f79c69f72a7a4e3b4136b4e5d/addons/web/static/src/js/views/form_widgets.js#L477), When the Datepicker class is initialized, No options is passed in the constructor of the object. Before changed this repository, I try to passed options in the xml field. But all the time options are undefined.

yajo (approved):
OK I get it now

PR #755, opened 2017-10-16
[MIG] web_sheet_full_width: migration to 11.0

Jerther:
Since the chatter is now on the right side of the screen, the form cannot take the whole screen as it used to, but when there is no chatter or when it is small, this addon is still useful.

I could try and make the chatter sizable with the mouse if that's still within the scope of this module. The chatter width could then be saved and used globally. Thoughts welcome :)


pedrobaeza:
For smallers screens where the chatter is not at the right side, it still should be full screen.

Jerther:
Oh, responsive design! Nice!

mgielissen:
@Jerther Chatter on the right screen is only in EE

Jerther:
And it needed the extra CSS :) Thanks @mgielissen

oihane (approved):
LGTM

yajo:
I'd be glad to see this integrated this into `web_responsive` and delete this addon.

pedrobaeza:
No, because there's people work enterprise interface that also wants it.

web_sheet_full_width/README.rst
```diff
@@ -0,0 +1,61 @@
...
+to get full width in the form view sheet.
+
+This module works in community edition and in enterprise edition.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/10.0
```
yajo, line 19: Wrong version

web_sheet_full_width/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "version": "11.0.1.0.1",
+    "author": "Therp BV, Sudokeys, GRAP, Métal Sartigan, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "summary": "Use the whole available screen width when displaying sheets",
+    "category": "Tools",
+    "depends": [
+        'web',
+    ],
+    "data": [
+        "view/qweb.xml",
```
yajo, line 14: I think this should be `templates/assets.xml` instead.

yajo (changes requested):
Missing commit history. Check out https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

web_sheet_full_width/__manifest__.py
Jerther: I thought it was weird... I'll sure rename this.

web_sheet_full_width/README.rst
Jerther: That of web_responsive is also wrong

pedrobaeza: Indeed, changed in https://github.com/OCA/web/commit/ebdbd58e2b7ff0dc6285e4a0fa301387fa674609

PR #766, opened 2017-10-20
[FIX] Ensure that domain field can be evaluated before creating x2x field

benwillig:
If a domain has been declared on a field (in python) and if this domain depends on an other model field, there will be an error when trying to perform ```name_search```. 

Example:
```python
partner_id = fields.Many2one(...)
invoice_id = fields.Many2one(..., domain="[('partner_id', '=', partner_id)]")
```
To fix this, we try to evaluate given domain and if it crash, we override it.

### DIFF
--- a/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
+++ b/web_advanced_search_x2x/static/src/js/web_advanced_search_x2x.js
@@ -12,6 +12,7 @@ odoo.define('web_advanced_search_x2x.search_filters', function (require) {
     var SearchView = require('web.SearchView');
     var data = require('web.data');
     var core = require('web.core');
+    var pyeval = require('web.pyeval');
 
     var X2XAdvancedSearchPropositionMixin = {
         template: "web_advanced_search_x2x.proposition",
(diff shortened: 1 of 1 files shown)


adrienpeiffer (approved):
:+1: Thanks for the fix ! Tested on runbot

pedrobaeza:
@Yajo please review

yajo:
It seems this is an alternate solution to #767, right? @adrienpeiffer have you checked if the bug you are fixing still happens after that fix?

benwillig:
``` web_advanced_search_x2x``` doesn't depend on ```web_m2x_options```, right?

pedrobaeza:
Yeah, that's it

pedrobaeza:
But having both installed exposes the problem

benwillig:
There is a problem if you only installed ```web_advanced_search_x2x``` too.

pedrobaeza:
Which problem? Are you talking with any of the patches provided? With both?

benwillig:
@Yajo said this PR is an alternate solution of #767. It's not true. If you installed both modules (including the fix of web_m2x_options), there isn't any error when trying to evaluate given field domain. If you installed only web_advanced_search_x2x (without this PR), there will be an error. Both PRs fix the same error, but both must be merged because web_advanced_search_x2x doesn't depend on web_m2x_options.

pedrobaeza:
OK, thanks for clarifying.

@Yajo please review this

yajo (approved):

pedrobaeza (approved):
Merging and bumping version number later in other commit

PR #770, opened 2017-10-21
[MIG] web_dashboard_open_action: Migration to 10.0

spacemonkeythe:


web_dashboard_open_action/__manifest__.py
```diff
@@ -20,7 +20,7 @@
 ##############################################################################
```
daramousk, line 20: This can be removed now and replaced with the shorter version like this https://github.com/OCA/maintainer-tools/blob/master/template/module/controllers/main.py#L1 on all the *.py files

daramousk: Also, remove the `test`, `external_dependencies` and turn `installable=True`

web_dashboard_open_action/static/src/js/web_dashboard_open_action.js
```diff
@@ -19,17 +19,22 @@
 //
 //############################################################################
 
-openerp.web_dashboard_open_action = function(instance)
-{
-    instance.web.form.DashBoard.include({
+odoo.define('pos_all_orders_reprint.chrome_extensions', function(require) {
```
daramousk, line 22: Why the rename here? From what I understand, it should be `odoo.define('web_dashboard_open_action.bashboard_action'...)`

web_dashboard_open_action/views/templates.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<openerp>
+<odoo>
     <data>
         <template id="assets_backend" name="web_dashboard_open_action assets" inherit_id="web.assets_backend">
             <xpath expr="." position="inside">
-                <script type="text/javascript" src="/web_dashboard_open_action/static/src/js/web_dashboard_open_action.js"></script>
                 <link rel="stylesheet" href="/web_dashboard_open_action/static/src/css/web_dashboard_open_action.css"/>
+                <script type="text/javascript" src="/web_dashboard_open_action/static/src/js/web_dashboard_open_action.js"></script>
```
daramousk, line 7: Is there a need for this change?

web_dashboard_open_action/static/src/xml/web_dashboard_open_action.xml
```diff
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
daramousk, line 1: While adding xml declarations is a good idea, preferably on ports it would be better if we kept the diff minimal.

daramousk (changes requested):

PR #771, opened 2017-10-21
[MIG] web_context_in_colors: Migration to 10.0

spacemonkeythe:


web_context_in_colors/view/qweb.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<openerp>
+<odoo>
     <data>
         <template id="assets_backend" name="web_context_in_colors" inherit_id="web.assets_backend">
             <xpath expr="." position="inside">
                 <script type="text/javascript" src="/web_context_in_colors/static/src/js/web_context_in_colors.js"></script>
             </xpath>
         </template>
     </data>
```
Jerther, line 9: The data tags could be removed as well.

Jerther (changes requested):
The module should be marked as installable in the manifest.

PR #772, opened 2017-10-22
[MIG] web_responsive test to 11.0

### LINKED ISSUE #759: [11.0] web_responsive: odoo.define_section is undefined
Current 11.0 builds will all fail due to:

 https://github.com/OCA/web/blob/d62b1b563af449317ab9cada65998df4f0707902/web_responsive/static/tests/js/web_responsive.js#L4

Odoo 11.0 changed the way we write widget tests, and in this change, it removed `odoo.define_section`. That's why we are getting [errors such as this one](https://travis-ci.org/OCA/web/jobs/287908508#L942-L945):

```
` AssertionError: ("TypeError: undefined is not a function (evaluating 'odoo.define_section')\n"
`  ' at http://127.0.0.1:8069/web_responsive/static/tests/js/web_responsive.js:4 '
`  '(in global code)\n'
`  '(leaf frame on top)')
```

We have to fix that, and it would be good to add the instructions to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0 once discovered how to do it.

The good news is that we have lots of good examples available and some introduction in slides 35-41 of @gde-odoo's [presentation at Odoo Experience 2017](https://www.odoo.com/es_ES/slides/slide/the-odoo-js-framework-492).

To start, it seems you have to just use `odoo.define` to run the tests, but I'm not sure the depth of the changes in that file to adapt to new tests, to be honest...

cc @Tecnativa @lasley @angelmoya

hgminh95:
Fix #759 

@Yajo can you take a look?

### DIFF
--- a/web_responsive/static/tests/js/web_responsive.js
+++ b/web_responsive/static/tests/js/web_responsive.js
@@ -1,278 +1,318 @@
 /* Copyright 2016 LasLabs Inc.
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
 
-odoo.define_section('web_responsive', ['web_responsive'], function(test) {
+odoo.define('web_responsive.test', function(require) {
     "use strict";
-    
-    // It provides a base drawer compatible interface for testing
-    self.initInterface = function(AppDrawer) {
-        
-        var $el = $('<div class="drawer drawer--left">');
-        $el.append(
-            $('<header role="banner">')
-                .append(
-                    $('<button class="drawer-toggle"><span class="drawer-hamburger-icon">')
-                )
-                .append(
-                    $('<nav class="drawer-nav"><ul class="drawer-menu"><li class="drawer-menu-item">')
-                )
-                .append(
-                    $('<div class="panel-title" id="appDrawerAppPanelHead">')
-                )
-        ).append($('<main role="main">'));
-        
-        self.$clickZone = $('<a class="oe_menu_leaf">');
-        
-        self.$secondaryMenu = $('<div><div class="o_sub_menu_content">');
-        
-        self.$dropdown = $('<div class="dropdown-scrollable">');
-    
-        $el.append(self.$clickZone)
-            .append(self.$secondaryMenu)
-            .append(self.$dropdown);
-        
-        var $document = $("#qunit-fixture");
-        $document.append($el);
-        
-        self.drawer = new AppDrawer.AppDrawer();
-        
-        return $document;
-        
-    };
-    
-    self.linkGrid = function() {
-        for(var i=0; i < 3; i++){
-            self.drawer.$el.append(
-                $('<div class="row">').append(
-                    $('<a class="col-md-6" id="a_' + i + '"><span class="app-drawer-icon-app /></a>' +
-                      '<a class="col-md-6" id="b_' + i + '"><span class="app-drawer-icon-app /></a>'
-                      )
-                )
+
+    var core = require('web.core');
+    var responsive = require('web_responsive');
+
+    QUnit.module('web_responsive', {
+        beforeEach: function() {
+            var $el = $(
+                '<div class="drawer drawer--left">' +
+                    '<header role="banner">' +
+                        '<button class="drawer-toggle">' +
+                            '<span class="drawer-hamburger-icon"/>' +
+                        '</button>' +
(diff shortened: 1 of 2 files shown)


lasley (changes requested):
Neat, QUnit is being exposed directly now. I actually take back some of my crap talking about them changing things then, this is totally for the best.

Can you please increase the tab level on the internals though? This would be a stylistic decision that I'm sure @Yajo and I could go back and forth on endlessly. I made the decision here to have the tab level where it's at though, so please respect that. Otherwise, the diff is quite large and it is hard to understand the changes that are being made.

hgminh95:
Sure, I have changed the style. However, the diff is still quite big as there are some minor changes like `self` become `this`, ...

web_responsive/static/tests/js/web_responsive.js
```diff
@@ -1,278 +1,317 @@
...
+                    .append(
+                        $('<button class="drawer-toggle"><span class="drawer-hamburger-icon">')
+                    )
+                    .append(
+                        $('<nav class="drawer-nav"><ul class="drawer-menu"><li class="drawer-menu-item">')
+                    )
+                    .append(
+                        $('<div class="panel-title" id="appDrawerAppPanelHead">')
+                    )
+            ).append($('<main role="main">'));
```
yajo: I think it's better to just bundle all of this appends in the single first `var $el = ...` definition. KISS.

web_responsive/static/tests/js/web_responsive.js
```diff
@@ -1,278 +1,317 @@
...
+
+            this.$clickZone = $('<a class="oe_menu_leaf">');
+
+            this.$secondaryMenu = $('<div><div class="o_sub_menu_content">');
+
+            this.$dropdown = $('<div class="dropdown-scrollable">');
+
+            $el.append(this.$clickZone)
+                .append(this.$secondaryMenu)
+                .append(this.$dropdown);
```
yajo: Same about this. It seems easier to read when you drop in the full HTML and then use `$el.find()` to get the specific elements you need IMHO.

web_responsive/static/tests/js/web_responsive.js
```diff
@@ -1,278 +1,317 @@
...
+        linkGrid: function() {
+            for(var i = 0; i < 3; i++){
+                this.drawer.$el.append(
+                    $('<div class="row">').append(
+                        $('<a class="col-md-6" id="a_' + i + '"><span class="app-drawer-icon-app /></a>' +
+                          '<a class="col-md-6" id="b_' + i + '"><span class="app-drawer-icon-app /></a>'
+                          )
+                    )
+                );
+                this.drawer.$appLinks = $('a.col-md-6');
```
yajo: Better use a namespaced `$el.find()` instead of whole-page `$()`

yajo (changes requested):
Beautiful

hgminh95:
Thanks for your review. I have updated the code.

lasley (approved):
Thank you @hgminh95 - absolutely amazing

web_responsive/static/tests/js/web_responsive.js
```diff
@@ -1,278 +1,318 @@
...
-                .append(
-                    $('<nav class="drawer-nav"><ul class="drawer-menu"><li class="drawer-menu-item">')
-                )
+
+    var core = require('web.core');
+    var responsive = require('web_responsive');
+
+    QUnit.module('web_responsive', {
+        beforeEach: function() {
+            var $el = $('<div class="drawer drawer--left">')
```
yajo: Still all of these appends seem overly complex. Can't you write normal HTML and put it into this first call?

yajo (changes requested):

web_responsive/static/tests/js/web_responsive.js
lasley: I'm not saying it won't work, but I really hope I wouldn't have done something this incredibly complex and stupid for no reason. (this is just following the pattern of my original code)

hgminh95:
Done :smile:

web_responsive/static/tests/js/web_responsive.js
lasley: So yeah guess I did do it for no reason. Thanks @hgminh95!

yajo (approved):

PR #780, opened 2017-10-28
web_tree_image:

kurniawanlucky:
* Change 16" to 32" because 16" is very small

### DIFF
--- a/web_tree_image/static/src/xml/widget.xml
+++ b/web_tree_image/static/src/xml/widget.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
     <img t-name="ListView.row.image"
-         t-att-height="widget.height || 16"
+         t-att-height="widget.height || 32"
          t-att-src="src" />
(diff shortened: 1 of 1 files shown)


yajo:
Please explain how to test the patch :blush:

StefanRijnhart:
I originally wrote this module to display __icons__ in the tree view, plus it fits the size of the standard Odoo row without an image, so the small size is on purpose. Then again, the size can be managed by the height attribute so I have no principle qualms with this change. I'll abstain from reviewing.

pedrobaeza:
I agree with Stefan that it's not needed to change the default. Just put in your view definition the desired height.

:-1:

legalsylvain (changes requested):
> so the small size is on purpose

agree with @StefanRijnhart. I think that default value should be the default Odoo value, so 16.

If you want to globally change this value (as your PR does), an acceptable improvment could be to add an extra settings default 16 to the ir.config.parameter, that could be changed.

In the xml file, something like

`t-att-height="widget.height || widget.default_tree_image_height"`

but it is a more complex PR.

PR #781, opened 2017-10-29
web_notify: migration to 11.0

bouvyd:
Still have to check the js code to see if everything is still needed.


pedrobaeza:
Marking as "Work in progress" until you tell me about the JS.

bouvyd:
Removed the `on_logout` function which was never called AFAICT.

Note: not completely sure why the polling is started in `show_application` override, which seems unrelated. Perhaps an override of the `start` checking for a bound sessions and returning a promise would be more appropriate.

However this is a migration, not a refactoring - if it ain't broken, don't fix it. Unless you want me to.

pedrobaeza:
@lmignon you were the original author. Maybe you can clarify @dbo-odoo's comments.

lmignon:
@pedrobaeza I'm not at work this week. I'll take a look at tis PR the next week

yajo (changes requested):
> However this is a migration, not a refactoring - if it ain't broken, don't fix it. Unless you want me to.

It would be nice, given the JS system is new and this could serve as an example for OCA users 😊 

In any case, please squash all translation commits

bouvyd:
@Yajo 
I'm not a JS guru, not sure I should be taken as an example :-p

bouvyd:
@Yajo said

> It would be nice, given the JS system is new and this could serve as an example for OCA users :blush:

I'm gonna have to disappoint you, since I'm not part of the JS Framework team, I'm not really familiar with the new webclient architecture (neither the new one nor the old one)...

I squashed the translations commits and moved the error map to a method as requested :)

bouvyd:
Small thing to notice: I'm unable to make this work in the runbot (tests work but a manual test does not). I think this might be a runbot server configuration issue because it works fine locally (and in tests) and because the network activity shows some strange things; e.g.:

```
SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-2ubuntu2.8
Protocol mismatch.
```

as the response for the `poll` request sent by the bus...

lmignon (approved):
@dbo-odoo Thank you for the migration LGTM (code review only). If I remember well, the polling is started into the 'show_application' method to be sure that all the required stuff to  show the notifications are in place.  It's maybe no more the most appropriate place but if it works I agree with you that we can leave it there.

yajo (changes requested):
OK, not a problem @dbo-odoo 😉 

Please squash the migration commits to merge

web_notify/models/res_users.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
```
phatnguyenuit: In python3 `# -*- coding: utf-8 -*-` it should be removed.

Same for the others python file in this pull request.

phatnguyenuit (changes requested):

web_notify/models/res_users.py
pedrobaeza: It's not strictly needed, but optional when you have already encoded your file.

jcdrubay:
@dbo-odoo thanks for the migration. I hope you can find time to take into account the last few comments ;)

phatnguyenuit (approved):

bouvyd:
@jcdrubay I forgot about this, thanks for the reminder :)

Commits squashed, encoding removed as per @phatnguyenuit's suggestion.

Hope you all had a nice easter weekend 🐇 🥚  (assuming you're somewhere it gets celebrated, otherwise I hope you had a nice completely normal weekend :p) 

Cheers!

jcdrubay:
@Yajo Your comments have been taken into account :) Can you approve? Maybe merge? Thanks

yajo (approved):

PR #783, opened 2017-10-31
Update web_no_bubble.css

aman-g202:


pedrobaeza:
Why this change?

aman-g202:
Just going through your project..felt like changing it

pedrobaeza:
But what it improves?

yajo:
No sense on aligning a hidden element, don't you think? 🤔

simahawk:
so... anything to do here? :)
@Yajo seems like that element is displayed by tooltip JS.
@darkshadow237 can you provide more info on what you want to do, possibly w/ screenshots?

pedrobaeza:
I don't think so.

PR #785, opened 2017-10-31
[ADD] [web_export_view] export all records if all checkbox is selected

hbrunn:
this allows to export all records when the all records checkbox is selected.

I wasn't very happy with the implementation choice of fishing out the data from the DOM, that's why I changed this to read from the underlying dataset. Commit 5ca37d4 contains a version that manipulates the DOM to do what we want, but that's slow and very unwieldy. What do @eLBati @simahawk @StefanRijnhart @lepistone (current contributors) think?

### DIFF
--- a/web_export_view/__openerp__.py
+++ b/web_export_view/__openerp__.py
@@ -21,7 +21,7 @@
 
 {
     'name': 'Export Current View',
-    'version': '8.0.1.2.0',
+    'version': '8.0.1.3.0',
     'category': 'Web',
     'author': "Agile Business Group,Odoo Community Association (OCA)",
     'website': 'http://www.agilebg.com',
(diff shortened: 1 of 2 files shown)


lepistone:
Hi @hbrunn ! I haven't been involved with this for a while now. I'll let the others speak out. Thanks!

StefanRijnhart (approved):
Thanks! I tested this change succesfully exporting both individual records and all records. I also checked the correct formatting of floats with different decimal separators.

eLBati (approved):
Thanks.
Should the same be applied to the more recent versions of odoo?

hbrunn:
if we agree this is a cleaner implementation and if I didn't destroy any feature this way, yes, I think we should (being unsure if this tacitly removes some feature is the reason I harassed all contributors I could find)

StefanRijnhart:
I'd merge, but it needs a version bump

hbrunn:
right, thanks. Any reasons for not having a CI check for this? (it's my understanding that actually applying the rules means a version bump for a changed comma)

PR #787, opened 2017-11-01
[ADD] web_listview_custom_column

hbrunn:
Custom columns in listview
==========================

This module was written to allow users to rearrange columns in list views. This can be done organization wide, for certain groups or just for the user herself.

Configuration
=============

To configure this module, you need to add users supposed to be able to customize columns to the group `Customize list views`. Note that this permission allows all sorts of mischief up to privilege escalation under certain circumstances, so this is only for trusted users.

Usage
=====

- go to some list and click the columns symbol right of the view switcher
- use the buttons appearing in column headers to remove and rearrange columns
- use the dropdown appearing next to the columns symbol to add columns

Known issues
============

- this addon creates custom views and (for group/all changes) standard tree view overrides. The latter are created with priority 10000 to avoid side effects, but the views will break if you remove fields from the database. Uninstalling the module will remove all view customizations.
- when rearranging columns, invisible columns count. So if it seems like nothing happens, you'll probably have some invisible columns

Roadmap
=======

- support different levels of customization: Everyone, some group, just current user


elicoidal (approved):
LGTM. I tried to test but Runbot is awfully slow on my side of the world...

andreschenkels:
:+1:  looking good Holger

hbrunn:
- [x] this module must manipulate view validation on updates such that customized views are validated only at the end of the update (or not at all, let's see)

lfreeke:
:+1: (test)

andhit-r (approved):

PR #776, opened 2017-10-23
[10.0][FIX] web_responsive: Menu Height on Init

lasley:
* Add a call to the window resize method during init to fix OCA#773

PR #791, opened 2017-11-04
[FIX] web_responsive: Menu Height on Init

lasley:
* Add a call to the window resize method during init to fix OCA#773

Forward port of #776


pedrobaeza (approved):
Fast-tracking also.

PR #792, opened 2017-11-07
[10.0][FIX] web_widget_x2many_2d_matrix: Sort matrix by incoming lines

YarnSeemannsgarn:
- So far the matrix is only sorted by the ids of the incoming lines, because JavaScript dictionaries are used
- This commit transfers the corresponding dictionaries to Map objects, which provide ordered hashes


yajo:
I don't think we can use `Map` (ES6 feature) within the PhantomJS testing framework. Can you double-check that please?

PR #807, opened 2017-11-20
[FIX] [web_export_view] Exporting records from a grouped view causes data rows have an empty cell at the beginning.

ELNOGAL:
Hello, I have detected that exporting records from a grouped view by some field, now the data rows have an empty cell at the beginning, which makes the values of the columns do not correspond to their headers.
With this fix I think it is resolved.


Abranes:
Tryed and is working.

Abranes (approved):
Tested and works ok.

yajo:
Why are you changing the file's permissions?

ELNOGAL:
I am sorry, It has been a mistake. The file permissions are now restored.

hbrunn (changes requested):
shouldn't you better check this in https://github.com/OCA/web/pull/807/files#diff-5dec7c0bd7346476fc0e51de4bdd1c97R57 and in the rest of the code use this as the leading list?

ELNOGAL:
Ok @hbrunn, if you mean is changing (this.type != null) by (this.tag == 'field') then I think a good idea (I can change it).
On the other hand, you have a PR #816 opened for this module. If you want to include these changes in the mentioned PR, you can do it and I close this.

yajo (approved):

ELNOGAL:
Hello,
Any news about this PR?
Should I change something else?

ELNOGAL (commented):
Changes are made

ELNOGAL:
Someone could tell me if there is something else that should be corrected to be able to merge this PR? 
Am I doing something wrong?

hbrunn (approved):

hbrunn:
nothing wrong, it's just that you should squash your commits to one, and during squashing you'll probably solve the current conflict

ELNOGAL:
Thanks!
I take into account it for the future.

Issue #808, opened 2017-11-22
[11.0] web_responsive: header is still appear when print directly from browser.


fanha99:
![header](https://user-images.githubusercontent.com/1172848/33109917-019dc34a-cf77-11e7-9dff-22e60424b5ca.png)
I want to print directly from the browser instead of create and save PDF before print. but when I print, the header is still appear.  Is there a way to hide it?

the screenshot is from the sale order / print

thx

PR #809, opened 2017-11-22
Hide header when print directly from browser issues 808

fanha99:
this is for: https://github.com/OCA/web/issues/808

### DIFF
--- a/web_responsive/static/src/less/navbar.less
+++ b/web_responsive/static/src/less/navbar.less
@@ -6,7 +6,6 @@ header {
     padding: 0;
     
     @media print {
-        visibility: hidden;
         display: none;
     }
 
--- a/web_responsive/static/src/less/navbar.less
+++ b/web_responsive/static/src/less/navbar.less
@@ -4,6 +4,11 @@
 header {
     margin: 0;
     padding: 0;
+    
+    @media print {
+        visibility: hidden;
+        display: none;
+    }
 
     > .main-nav {
         display: block;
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Next time, you can put the issue text directly in the PR.

Please change module version number.

@lasley, can you check?

web_responsive/static/src/less/navbar.less
```diff
@@ -4,6 +4,11 @@
 header {
     margin: 0;
     padding: 0;
+    
+    @media print {
+        visibility: hidden;
```
lasley: Remove this one. Visibility hidden means that the element will still take the same amount of space, it simply won't show up. `display: none` is what you're going for, and I assume because it's later it's actually overriding this and removing the space

lasley (changes requested):
Thanks @fanha99 - one comment inline

web_responsive/static/src/less/navbar.less
fanha99: hi, thanks for review.
updated.
please help to review again

lasley (approved):
Perfecto, thanks @fanha99

pedrobaeza (approved):

lasley:
@fanha99 - do you plan on backporting this?

fanha99:
hi @lasley,
have just checked the older version:
[9.0]: system open a new windows for report, and there is no header in this windows, so I think there is no need to fix.
[10.0]: same as [11.0], a new pull request was created: https://github.com/OCA/web/pull/813
thanks

PR #810, opened 2017-11-23
[ADD][10.0] web_kanban_sortable

fuentes73:
In Odoo v8.0 it was possible to sort any kanban view if the model has a field named sequence.
This module enable again this functionality in v10.

It just enable drag&drop sorting if the model has a field named sequence.


yajo:
It's not that odoo 10 doesn't support this, it's that it only supports it when rendering a grouped view. To add this to ungrouped views, I think the proper place would be a fix in odoo core, if you really need this.

fuentes73:
Well... that's true I should have explain it better. I need it at v10 and I'm pretty sure that this change won't be accepted at a stable branch, don't you?

yajo:
There are a couple of workarounds you can use:

a. Set a default filter for the user that needs this that includes a group by clause. This way he'll have grouping enabled by default.
b. Configure list view as the default one, where you can do ungrouped sorting.

Both are officially supported and just some clicks away, and you save to maintain one extra addon with 119 LOC. Then you could close this. Just trying to KISS. What do you think?

rousseldenis:
@fuentes010

web_kanban_sortable/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# Copyright 2017 Studio73 - Pablo Fuentes <pablo@studio73.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Web Kanban Sortable',
+    'category': 'Web',
+    'version': '10.0.1.0.0',
+    'author': 'Studio73, '
+              'Odoo Community Association(OCA)',
+    'website': "https://www.studio73.es",
```
nikul-serpentcs, line 11: URL should be `https://github.com/OCA/web`

nikul-serpentcs (changes requested):
Minor Change

fuentes73:
We don't need this anymore. If someone needs it feel free to contact.

PR #812, opened 2017-11-27
[FIX] web_export_view: Error when export a view with a many2many field.

Abranes:
Hello,
With the latest version when you try to export a current view where is a m2m field we get an error.
To reproduce go to Sales --> Products --> Product variants, select any number of records and 'Export Current View', then:
export_current_view_error
(Tryed on runbot)
To fix this we implemented a case in the web_export_view.js for the m2m fields and in the controller.py we get the correct value acording to the m2m values.

### DIFF
--- a/web_export_view/controllers/controllers.py
+++ b/web_export_view/controllers/controllers.py
@@ -40,7 +40,14 @@ class ExcelExportView(ExcelExport):
         data = json.loads(data)
         model = data.get('model', [])
         columns_headers = data.get('headers', [])
-        rows = data.get('rows', [])
+        rows = []
+        for row in data.get('rows', []):
+            row = [
+                ', '.join(map(lambda c: c.name_get()[0][1], request.env[cell['relation']].browse(cell['ids'])))
+                if isinstance(cell, dict) else cell
+                for cell in row
+            ]
+            rows.append(row)
 
         return request.make_response(
             self.from_data(columns_headers, rows),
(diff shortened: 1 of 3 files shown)


web_export_view/static/src/js/web_export_view.js
```diff
@@ -85,12 +85,20 @@ openerp.web_export_view = function (instance) {
                     var export_row = [],
                         record = new instance.web.list.Record(record).toForm();
                     $.each(view.visible_columns, function() {
-                        export_row.push(
-                            this.type != 'integer' && this.type != 'float' ?
```
hbrunn, line 89: why don't you just add your exclusion here?

web_export_view/controllers/controllers.py
```diff
@@ -40,7 +40,19 @@ def export_xls_view(self, data, token):
...
+        rows = []
+        for row in data.get('rows', []):
+            row = [
+                ', '.join(
+                    map(
+                        lambda c: c.name_get()[0][1],
+                        request.env[cell['relation']].browse(cell['ids'])
+                    )
+                )
+                if isinstance(cell, dict) else cell
```
hbrunn, line 52: don't guess stuff. Extend the protocol that it also passes the field names being exported. From this you can decide where to do anything at all and that relation to use. And don't copy around data, the id lists are references, you you can just replace those.

hbrunn (changes requested):
I left comments what to change if you insist on this approach.

I think we'll be better off if you do what the client does (that's the trick of this module: Generate it the way the client presents it): https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_list.js#L1109 You might have to slip in an inbetween step where you gather all ids of all many2many fields in the export to read

hbrunn:
in https://github.com/OCA/web/pull/816 I fix this the way I propose and a few other issues, so I'd suggest to close this and review the latter.

PR #817, opened 2017-11-29
[10.0] Add module web_listview_number_right

AMhadhbi:


hbrunn:
are you sure this shouldn't be a bugfix? We have https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/less/tree_view.less#L26 and https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/less/list_view.less#L102, so if those don't work (didn't check myself), I think it's a bug

PR #818, opened 2017-11-30
[10.0][FIX] web_widget_x2many_2d_matrix: Styles

houzefa-abba:
Hi,

I set out by wondering why cells in the matrix widget were no longer showing up with their blue "required field" background color: It appears various CSS classes (but not all of them) have been renamed in Odoo 10.

I then looked into other CSS classes used by the `web_widget_x2many_2d_matrix` addon and made them match regular Odoo lists.
In editable mode, we are still quite far from editable lists which put `input` fields inside a fake `div` so as to simulate an actual `form`, but styles do match now at least.

Don't hesitate to shout if anything is broken - I will be using this widget quite a bit so I will try to watch for issues reported in here.

Commit message:

Update CSS classes used by the "web_widget_x2many_2d_matrix" widget to
match base Odoo ones.

In particular, cells now show up with a "required field" style (as
intended).

Remove custom read-only field management in the process that was no
longer working and is not needed anymore (all handled via CSS classes).

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
@@ -14,7 +14,7 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
 
     var WidgetX2Many2dMatrix = FieldOne2Many.extend({
         template: 'FieldX2Many2dMatrix',
-        widget_class: 'oe_form_field_x2many_2d_matrix',
+        widget_class: 'o_form_field_x2many_2d_matrix',
 
         // those will be filled with rows from the dataset
         by_x_axis: {},
@@ -145,7 +145,6 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                             self.setup_many2one_axes();
                             self.$el.find('.edit').on(
                                 'change', self.proxy(self.xy_value_change));
-                            self.effective_readonly_change();
                         }
                     });
                 });
@@ -207,13 +206,27 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                 this.field_label_y_axis, true);
         },
 
-        // return the class(es) the inputs should have
+        // return the class(es) inputs should have
         get_xy_value_class: function()
         {
-            var classes = 'oe_form_field oe_form_required';
-            if(this.is_numeric)
-            {
-                classes += ' oe_form_field_float';
+            var classes = 'edit o_form_input o_form_required oe_edit_only';
+            if (this.is_numeric) {
+                classes += ' o_form_field_number';
+            }
+            return classes;
+        },
+
+        // return the class(es) read-only inputs should have
+        get_xy_value_read_only_class: function()
+        {
+            var classes = 'read o_form_input';
+            if (this.is_numeric) {
+                classes += ' o_form_field_number';
+            }
+            if (!this.get('effective_readonly')) {
+                // When the whole matrix is read-only but shown inside of an editable form, don't
+                // add "oe_read_only" so our read-only <span>s always show up.
+                classes += ' oe_read_only';
             }
             return classes;
         },
@@ -350,9 +363,6 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                 'change', self.proxy(this.xy_value_change));
             this.compute_totals();
             this.setup_many2one_axes();
-            this.on("change:effective_readonly",
-                    this, this.proxy(this.effective_readonly_change));
-            this.effective_readonly_change();
             return this._super();
         },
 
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Can you include some screenshots of before/after?

houzefa-abba:
Good idea!

Here is what I get with the latest Odoo 10, `OCA/web`, `OCA/product-variant` after installing `sale_order_variant_mgmt`.

- Default look - notice the blue "required field" background is missing:
variants-default

- Default read-only look (I have added `readonly="1"` by hand) - notice all content is missing:
variants-default-readonly

- With my fixes:
variants-fixed

- With my fixes in read-only mode:
variants-fixed-readonly

These shots are with the default theme; I have been testing with [another theme](https://www.odoo.com/apps/themes/10.0/backend_theme_v10/) and it works fine there as well.

Also for each CSS class change, I `grep`ped through the `web` addon; usually the only result was in old docs that haven't been updated.

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -207,13 +206,27 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
...
 
-        // return the class(es) the inputs should have
+        // return the class(es) inputs should have
         get_xy_value_class: function()
         {
-            var classes = 'oe_form_field oe_form_required';
-            if(this.is_numeric)
-            {
-                classes += ' oe_form_field_float';
+            var classes = 'edit o_form_input o_form_required oe_edit_only';
```
pedrobaeza, line 212: But what if the field is not required?

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
```diff
@@ -380,17 +390,6 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
 
         },
 
-        effective_readonly_change: function()
```
pedrobaeza, line 383: Why this is not working?

pedrobaeza (changes requested):

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
houzefa-abba: When testing this event was never fired; I think that is because the base `FieldX2Many` struct already catches it via:
```javascript
        this.is_loaded.done(function() {
            self.on("change:effective_readonly", self, destroy);
        });
```
(`web/.../form_relational_widgets.js`)

Since the whole widget is re-rendered on a `readonly` attr change, what this method was doing by hand is also no longer necessary.

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
houzefa-abba: Would be easy to conditionally add the attribute based on `this.get('required')` but I'd rather not until the widget actually supports this; currently when I set the matrix to be required, it still allows being saved without any actual input.

This to me would be a separate improvement than these simple style fixes.

pedrobaeza: OK

web_widget_x2many_2d_matrix/static/src/js/web_widget_x2many_2d_matrix.js
pedrobaeza: OK

yajo:
Can't we make the style of 1 match the style of 2 with the default theme?

33428760-6fb22a34-d5ca-11e7-83b1-c6351e864418

houzefa-abba:
Good catch @Yajo, that is because our `input` fields are missing a `type="text"` attribute which all other `input` fields have!

houzefa-abba:
Here is how it now looks.

I have also attempted to fix the disabled field bg color via CSS as doing it from the JS code would be annoying (the "disabled" attr gets added separately from CSS classes).

variants-fixed-2

yajo (approved):
code looks good, no functional test done

pedrobaeza (approved):

pedrobaeza:

ping @hbrunn

web_widget_x2many_2d_matrix/static/src/xml/web_widget_x2many_2d_matrix.xml
```diff
@@ -14,20 +14,22 @@
...
+                            <span t-att-class="widget.get_xy_value_read_only_class()"><t t-esc="widget.format_xy_value(widget.get_xy_value(x, y))" /></span>
                         </td>
-                        <td t-if="widget.show_row_totals" class="row_total oe_number" t-att-data-y="y"/>
+                        <td t-if="widget.show_row_totals" class="row_total o_form_field_number" t-att-data-y="y"/>
                     </tr>
                 </tbody>
                 <tfoot t-if="widget.show_column_totals">
                     <tr>
                         <th>Total</th>
-                        <td t-foreach="widget.get_x_axis_values()" t-as="x" class="oe_list_footer oe_number column_total" t-att-data-x="x" />
```
simahawk, line 29: `oe_list_footer` should be moved to `o_list_footer`. Is not used anywhere as far as I see but all the list-related classes are no `o_list_*`

simahawk (changes requested):
LGTM, one small non-blocking change in class name

ovnicraft:
any update for this ?

PR #820, opened 2017-12-04
[10.0][MIG] web_ckeditor4: migration to 10.0

daramousk:


simahawk (changes requested):
LGTM. Even if is still a WIP... not sure if this is already in your TODO list but here are some remarks on README and version:

1. "This addon provides a widget for editing html fields via CKEditor 4.6.6" -> then in credits it mentions "4.4.6"
2. in this regard would be better to change the version of the module to the version of the lib
3. I see 4 widgets options: would you mind describing them in README?
4. Maybe it's me but: sounds like is not clear if this addon replaces current HTML fields editor or you are supposed to do it yourself
4.1:  might be worth to add ad demo view w/ the widget in place to make it easier to evaluate
5. Update the README to match latest template https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
6. Commit history is gone

ecino:
I agree with @simahawk that you should describe the widget options in the README and how to use it. I tested the module and noticed that the editor shows up even when the view is not in edit mode which is confusing for the user. Also, when editing and saving, the editor is showing up several times like shown on the screenshot:
image
@daramousk I am willing to review more if needed. Would you mind taking a look at my PR and leave a review please? https://github.com/OCA/social/pull/185

simahawk:
@daramousk @ecino I fixed the view only appearance and eased some customization here https://github.com/daramousk/web/pull/1

ecino:
Now it's better but still doesn't completely work. I tested it on Chromium. When I read it's good, then I edit and when I save, a new instance of the editor appears below the html field. If I edit several times and save, another widget appears each time. Example screenshot:
image

web_ckeditor4/static/src/js/web_ckeditor4.js
```diff
@@ -214,11 +220,18 @@ openerp.web_ckeditor4 = function(instance)
...
         {
             return value;
         }
     });
-}
-
+    core.form_widget_registry.add('text_ckeditor4', FieldCKEditor4);
+    core.form_widget_registry.add('text_ckeditor4_raw', FieldCKEditor4Raw);
+    core.form_widget_registry.add('text_html', FieldCKEditor4);
+    core.form_widget_registry.add('html', FieldCKEditor4);
```
ecino, line 233: Why overriding html default widget? Would it be better to only use text_ckeditor4 and text_ckeditor4_raw widgets so that default html widget is still working? Or maybe add another widget `default_html` that allows someone to keep the default editor where needed.

web_ckeditor4/static/src/js/web_ckeditor4.js
```diff
@@ -214,11 +220,18 @@ openerp.web_ckeditor4 = function(instance)
...
-    instance.web_ckeditor4.FieldCKEditor4Raw = instance.web_ckeditor4.FieldCKEditor4.extend({
+    var FieldCKEditor4Raw = FieldCKEditor4.extend({
         filter_html: function(value)
         {
             return value;
         }
     });
-}
-
+    core.form_widget_registry.add('text_ckeditor4', FieldCKEditor4);
```
ecino, line 230: I would document in the README what are those widgets and what difference there is between them.

web_ckeditor4/models/ckeditor_monkeypatch.py
```diff
@@ -27,7 +27,7 @@ class CkeditorMonkeypatch(models.AbstractModel):
     _name = 'ckeditor.monkeypatch'
     _description = 'Monkeypatches for CKEditor'
 
-    def _register_hook(self, cr):
+    def _register_hook(self):
         marker = self._name.replace('.', '_')
         if not hasattr(clean, marker):
             # monkey patch lxml's html cleaner to allow image data urls
```
hbrunn, line 33: @daramousk this one was necessary because lxml cleaned away all dataurls, but meanwhile, [my patch](https://github.com/lxml/lxml/pull/162) for this is merged in lxml, so I don't think we need this in the new version

simahawk:
@ecino are you sure you still see it duplicated? Maybe a bad refresh? AFAIK my patch fixed that too... we are using it on prod now and sounds like is working fine.

ecino:
@simahawk Yes I updated the module with the latest changes, forced refresh and still having this issue both on Chromium and Firefox on a Linux workstation.

web_ckeditor4/models/ckeditor_monkeypatch.py
daramousk: @hbrunn I see that your commit exists from lxml 3.6 onwards while v10 has version 3.5.0 pinned. 

git branch  -a --contains 763a7c4d05a6a9031e22488d834600610ce28762
* master
  remotes/origin/HEAD -> origin/master
  remotes/origin/LP1595781_adopt_external_doc
  remotes/origin/lxml-3.6
  remotes/origin/lxml-3.7
  remotes/origin/master
  remotes/origin/pypy4

https://github.com/OCA/OCB/blob/10.0/requirements.txt#L10

Would it still be correct to change it?

hbrunn: no. Thanks for checking!

simahawk:
@daramousk can you include this https://github.com/daramousk/web/pull/2 and squash it w/ my fix commit? @ecino this one should fix that issue once for all ;)

ecino:
@simahawk Thanks, indeed it works now.

ecino (approved):

daramousk:
@simahawk Is there something else here? If not please approve so that it can be possibly merged

simahawk:
@daramousk there's an issue that we are facing: 
1. go to a form where you have the editor enabled
2. click on edit
3. click on ANY translate icon (not necessarily the on of the field w/ ck4 editor widget) 
4. click "yes I want to go away from the form"
5. get to the translation page
6. click on the breadcrumb/path menu to get back to the record -> :boom: 

JS is broken because the editor was not destroyed. We have half a patch for it. 
I can get back on it next week hopefully.

simahawk:
@daramousk this should fix the above issue https://github.com/daramousk/web/pull/3

/cc @ecino @tschanzt

ecino:
Now I have the following error when creating a record, preventing the editor to show up. It's working when editing an existing record. 
```
http://localhost:8069/web_ckeditor4/static/lib/ckeditor/ckeditor.js:243
Retraçage :
TypeError: Cannot read property 'destroy' of undefined
    at a.destroy (http://localhost:8069/web_ckeditor4/static/lib/ckeditor/ckeditor.js:243:334)
    at Class._cleanup_editor (http://localhost:8069/web_ckeditor4/static/src/js/web_ckeditor4.js:209:29)
    at Class.destroy_content (http://localhost:8069/web_ckeditor4/static/src/js/web_ckeditor4.js:221:18)
    at Object.<anonymous> (http://localhost:8069/web_ckeditor4/static/src/js/web_ckeditor4_formview.js:15:29)
    at fire (http://localhost:8069/web/static/lib/jquery/jquery.js:3119:58)
    at Object.add [as done] (http://localhost:8069/web/static/lib/jquery/jquery.js:3165:49)
    at Class.can_be_discarded (http://localhost:8069/web_ckeditor4/static/src/js/web_ckeditor4_formview.js:11:31)
    at Class.can_be_discarded (http://localhost:8069/web/static/src/js/framework/class.js:122:38)
    at Object.<anonymous> (http://localhost:8069/web/static/src/js/views/form_view.js:84:29)
    at Object.<anonymous> (http://localhost:8069/web/static/lib/jquery/jquery.js:3276:89)
```

simahawk:
@ecino I have a possible fix here https://github.com/simahawk/web/commit/faecea67d7c247233fdf23fa3453b2804bfa649b

We still have to do some tests. Would mind trying this change on your instance?

ecino:
@simahawk Yes it seems to work with your fix!

simahawk:
@ecino tnx! @daramousk would you mind cherrypicking this? https://github.com/simahawk/web/commit/faecea67d7c247233fdf23fa3453b2804bfa649b

ecino:
@simahawk With further testing I still have the same error showing up after this scenario:
1. View has the editor field
2. Edit mode
3. Add a x2many record by creating a new one

I recorded a video for better understanding:
https://youtu.be/GcJML4GHp6w

tschanzt:
@ecino could you check again? we added a commit that should fix this.

ecino:
@tschanzt It works!

web_ckeditor4/static/src/js/web_ckeditor4.js
```diff
@@ -105,59 +99,70 @@ openerp.web_ckeditor4 = function(instance)
...
+    var FieldCKEditor4 = core.form_widget_registry.get('text').extend({
+        ckeditor_config: function () {
+          return {
+              removePlugins: this._getRemovePlugins(),
+              removeButtons: this._getRemoveButtons(),
+              filebrowserImageUploadUrl: 'dummy',
+              extraPlugins: 'filebrowser',
+              // this is '#39' per default which screws up single quoted text in ${}
+              entities_additional: ''
+          };
```
ecino, line 122: Is it possible to add this line in the config ?
`allowedContent: true`
My problem is that ckeditor is removing classes from my html, but by adding this parameter we allow the attributes.
See https://stackoverflow.com/questions/15659390/ckeditor-automatically-strips-classes-from-div for reference

tschanzt: @ecino i think that would be a bad idea, since it disables filtering completely. The odoo html field sanitizes the input so its not terrible, but i think we should keep it because it is stripped immediately and not only on save. This keeps the user from entering stuff which looks amazing only to have it removed when he saves.

hbrunn: @ecino if you really want this, your use case is covered by `text_ckeditor4_raw` - but as @tschanzt points out, you'll also have to turn off filtering in the field to make this effective. and you have to trust your users a lot of course.

ecino: Thanks for the tips, then I think you are right, we shouldn't add this.

ecino:
One another problem I found, but I don't know if it comes from the editor or from the module `web_responsive` which is installed on my base: when you open the editor in full screen, then it's impossible to bring back the normal view, unless you refresh the page.

ecino:
What's the state of this PR? Would be nice to have this finalized and merged.

pedrobaeza:
@daramousk should be attend comments and fix Travis or someone can take over the PR and finish it

daramousk:
@pedrobaeza I am not sure on how to fix the issues mentioned by travis to be honest. They do not seem to come from this module

pedrobaeza:
I have restarted the build as it seems a Travis glitch

daramousk:
@pedrobaeza Is there something else for me to do here or can this be merged?

daramousk:
@tschanzt 
@hbrunn
@simahawk

Any other changes needed here?

hbrunn:
I'd like to see an update to the latest incarnation of ckeditor4 (not 5, that's just as dumbed down as the standard editor)

simahawk:
current version is working fine for us since months :+1:

daramousk:
@simahawk Then please mark this PR as approved :smile:

simahawk (approved):
squash/commit rephrase needed

tschanzt (approved):
works well for us.

daramousk:
@hbrunn At this point, maybe this should be merged and the changes you request can be taken care off in a new PR? Otherwise if I push any changes they will invalidate any reviewing that took place above.

hbrunn:
good one. Just fix https://travis-ci.org/OCA/web/jobs/429542114#L7210, then I'll merge this. But please follow up with an update for the library code, I've some hope that an update will make it possible to improve the patching code a little bit. Two things are important there:

- image 'upload' must still work. Note that this is not super efficient, it takes the image and embeds it as a data uri. Sometimes we want that, but not in the general case. Much more javascript that lets us select an existing ir.attachment or create a new one from an upload would be much better, but that's out of scope for a migration. I also never found anyone to fund this as I expect this to be some hours of work,
- this editor is known to mess up jinja2 expressions when cleaning html. In https://github.com/OCA/web/pull/158 I made an approach to fix this, maybe you can integrate that in your update and check if there were api improvements?

daramousk:
@hbrunn Fixed, and opened #1075 for the update

PR #825, opened 2017-12-14
[FIX] Added missing touch support to signature widget

ghost:
On iPad and Android the mouse events are not working, therefore added the same event behaviour for touchXXX-Events.


yajo (changes requested):
On one hand, the feature is cool.

OTOH, I am generally against these kind of patches, because it means *we have to maintain our own fork of the plugin*.

It would be best to contribute this improvement directly to https://github.com/kbwood/signature itself, and then when they merge it, just update the lib version here.

Another option would be to add these improvements in a separate file, outside the lib dir.

yajo:
BTW this reminds me to #842.

PR #828, opened 2017-12-18
[MIG] web_decimal_numpad_dot: Migration to 11.0

chienandalu:
Numpad Dot as decimal separator
===============================

Allows using numpad dot to enter period decimal separator even in localizations
where comma is used instead of period.

Usage
=====

Whenever on a float or monetary input field pressing numpad dot produces the
proper decimal separator for the active localization.

cc @Tecnativa


hbrunn (approved):
finally! I used to carry around a patch to core for that. Extra points for handling float_time.

pedrobaeza (approved):
Working as expected in runbot.

luismontalba (approved):
Tested on runbot

PR #830, opened 2017-12-20
[10.0] Allow complex color expressions in web_timeline

npiganeau:
Current implementation of `colors` only allows for simple evaluation of kind ` field operator value` and does not work for more complex conditions such as `a != value1 and b in (value2, value3)`.

This PR reimplements the colors evaluation the same way as in tree views.

Drawback is that fields used in complex conditions must be declared, like in a kanban view. However, fields in a simple `field operator value` are still fetched automatically for backward compatibility.

### DIFF
--- a/web_timeline/README.rst
+++ b/web_timeline/README.rst
@@ -59,7 +59,7 @@ Example:
                           default_group_by="user_id"
                           event_open_popup="true"
                           zoomKey="ctrlKey"
-                          colors="#ec7063:user_id == false;#2ecb71:kanban_state=='done';">
+                          colors="#ec7063:user_id == False;#2ecb71:kanban_state=='done';">
                     <field name="user_id"/>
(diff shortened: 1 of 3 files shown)


web_timeline/README.rst
```diff
@@ -59,6 +60,8 @@ Example:
                           event_open_popup="true"
                           zoomKey="ctrlKey"
                           colors="#ec7063:user_id == false;#2ecb71:kanban_state=='done';">
```
pedrobaeza: Shouldn't this be now Python syntax?

npiganeau: Yes, definitely. I've pushed the correction. But actually it was already pyevaled before...

hparfr (approved):

hparfr:
Code read, LGTM

PR #719, opened 2017-09-06
[9.0][ADD] web_widget_bokeh_chart

LoisRForgeFlow:
Web Widget Bokeh Chart
==================

This module add the posibility to insert Bokeh charts into Odoo standard views.

Example and use case (ddmrp):

example

PR #834, opened 2017-12-27
[10.0][MIG] web_widget_bokeh_chart

LoisRForgeFlow:
Web Widget Bokeh Chart
==================

This module add the posibility to insert Bokeh charts into Odoo standard views.

example

--------

Migration done based on open PR: https://github.com/OCA/web/pull/719


hparfr (approved):

hparfr:
It's may be better to declare js dependencies in a packages.json and rely on js tools (like npm) to fetch them.

JordiBForgeFlow:
We discussed in the past with @hbrunn that it is better to always include the code of external libraries in the module, in order to avoid the dependency. Also, it is safer from a security perspective.

hparfr:
Thanks for the clarification

JordiBForgeFlow (approved):

PR #843, opened 2018-01-22
[MIG] web_action_conditionable: Migration to 11.0

andreparames:
Just had to change the JS module definition.


web_action_conditionable/README.rst
```diff
@@ -0,0 +1,83 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+* go to ...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/web/8.0
```
simahawk: $REPO_ID/11.0

simahawk (changes requested):
history missing. You should checkout the module from older branch. Follow instructions here https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

andreparames:
Thanks for your help, I've pushed the updated branch.

web_action_conditionable/README.rst
```diff
@@ -0,0 +1,84 @@
...
+.. branch is "8.0" for example
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/web/issues>`_.
```
simahawk: please, update to latest [README](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

web_action_conditionable/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
```
simahawk: remove coding headers

simahawk (changes requested):

simahawk:
Also, would you mind squashing these commits under `[add] web_action_conditionable`  or alike?
image
Beside they are not really clear on which module they are applied sounds like they are all part of the same work.

andreparames:
@simahawk  I don't mind at all, but then the history will diverge (those were the commits ported from the previous branch), is that OK?

simahawk:
@andreparames history does not diverge as this module was not here yet and this is a good moment to clean it up :wink:

web_action_conditionable/README.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Contributors
+------------
+
+* Cristian Salamea <cristian.salamea@gmail.com>
+* André Paramés <github@andreparames.com> (https://www.acsone.eu/)
+
+Do not contact contributors directly about support or help with technical issues.
+
+Funders
```
simahawk: same remark as of #845 ;)

andreparames:
@simahawk thanks, I've applied the same fixes and squashed my commits for merging.

web_action_conditionable/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    "name": 'web_action_conditionable',
+    "version": "11.0.1.0.0",
+    "depends": [
+        'base',
+        'web',
+    ],
+    'data': ['views/view.xml'],
+    "author": "Cristian Salamea,Odoo Community Association (OCA)",
```
simahawk, line 9: 'website': 'https://github.com/OCA/web'

andreparames: Thanks, I've made the change, though seems like Github lost track of it after the force-push.

StefanRijnhart: It's in the line below ;-)

StefanRijnhart (approved):
I checked the diff. Literally only the version number was changed. Thanks!

web_action_conditionable/__manifest__.py
andreparames: @StefanRijnhart yeah, the line is there, but when I click on the file name (above the piece of code), it sends me [here](https://github.com/OCA/web/pull/843/files/345c1e9d2b5a6948980b658e0e905579be65f901#diff-1c9df7f23577f3809216942b00713b51), which is the old commit. No worries, though, I don't think it'll affect the merge.

andreparames:
@StefanRijnhart thank you for the review!

web_action_conditionable/__manifest__.py
StefanRijnhart: Should be fine, indeed. It's the commit from when @simahawk reviewed the code.

simahawk (approved):

andreparames:
Thanks @simahawk

yajo (approved):

PR #844, opened 2018-01-22
add load argument to ensure correct context split

tschanzt:
Fixes the display of the inactive languages in the translation overlay. Although the field is already translated correctly it is not reflected in the dialog since the default value is loaded for languages other than the one currently selected by the user.

Before:
screenshot from 2018-01-22 17-57-51
Here we can see that the correct string for German is displayed but French and English are the same, since the non translated value is loaded. 

After:
screenshot from 2018-01-22 17-46-01
Here we now can see, that the French string is now translated and no longer uses the untranslated value.

The problem was, that in [split_context](https://github.com/odoo/odoo/blob/10.0/odoo/api.py#L290) it is checked if we provide more arguments than the method actually needs. If this is the case the last element of the argument list is treated as the context. Since [read](https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L2975) takes the load argument we need to provide it, so that the context is split correctly.

it would also be possible to use self.view.dataset._model.call which takes where we could specify the context as a kwarg.

### DIFF
--- a/web_translate_dialog/__manifest__.py
+++ b/web_translate_dialog/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Web Translate Dialog",
     "summary": "Easy-to-use pop-up to translate fields in several languages",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Web",
     "website": "https://odoo-community.org/",
     "author": "Camptocamp, "
--- a/web_translate_dialog/static/src/js/web_translate_dialog.js
+++ b/web_translate_dialog/static/src/js/web_translate_dialog.js
@@ -141,7 +141,7 @@ var translateDialog = Dialog.extend({
                 deff.resolve();
             } else {
                 self.view.dataset.call('read',[[self.view.datarecord.id],
-                    self.translatable_fields_keys,
+                    self.translatable_fields_keys, '_classic_read',
                     self.view.dataset.get_context({
                         'lang': lg.code })]).done(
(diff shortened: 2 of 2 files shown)


simahawk (changes requested):
LGTM. Just one nitpicking thing that makes reading logs easier: would you mind changing the commit message with something like:
```
[fix] web_translate_dialog: correct context split

Long description here (what you put in the PR description basically)```

pedrobaeza:

tschanzt:
@simahawk @pedrobaeza I increased the version number and reworded my commit message.

simahawk (approved):

pedrobaeza (approved):

grindtildeath (approved):

hbrunn:
any reason not to just pass the context by [name](https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/js/framework/model.js#L41) as it should be anyways?

pedrobaeza:
Agh, too late, I have merged. Do you see this a problem?

hbrunn:
not really, people usually (hopefully!) are careful with changing the signature of the read function.

You'd then also have to call the function on the model, not the dataset, because as written above, the dataset version only allows positional arguments.

tschanzt:
@hbrunn I wasn't sure which one would be better. So I did the one which looked like the smaller change to me.

hbrunn:
next time, better go for the bigger diff but safer call. But that's fine as it is I think.

PR #846, opened 2018-01-23
[8.0][FIX] ActWindow from Python Method after Popup

Maartincm:
Allow python methods called from an ir.action.act_window.message
action to return window actions without specifying the ``views``
key, as they would be if called from a button defined in xml.
``view_mode`` key is used for creating the views array as it is required anyway.

### Problem:

* After calling a popup message defined like

        confirmation = {
            'type': 'ir.actions.act_window.message',
            'title': conf_vals['title'],
            'message': conf_vals['message'],
            'close_button_title': conf_vals['close_title'],
            'buttons': [
                {
                    'type': 'method',
                    'name': conf_vals['accept_title'],
                    'model': self._name,
                    'method': 'apply_method',
                    'args': [self.ids, expenses.ids],
                }
            ]
        }
        return confirmation

the action returned from the method `apply_method` which was something like

        action = {
            'name': 'Moves',
            'type': 'ir.actions.act_window',
            'res_model': 'account.move',
            'view_type': 'form',
            'view_mode': 'tree,form',
            'target': 'current',
        }

ended up in a JS error.
image

### DIFF
--- a/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
+++ b/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
@@ -78,6 +78,29 @@ openerp.web_ir_actions_act_window_message = function(instance)
                             {
                                 if(_.isObject(result))
                                 {
+                                    if (! ('views' in result) && 'view_mode' in result){
+                                        types = result.view_mode.split(",");
+                                        if (types.length > 1){
+                                            if (types.includes('tree') && types.includes('form')){
+                                                if (types[0] === "tree"){
+                                                    views = [[false, "list"], [false, types[1]]];
+                                                    result.view_mode = "list" + ',' + types[1];
+                                                }
+                                                else {
+                                                    views = [[false, types[0]], [false, "list"]];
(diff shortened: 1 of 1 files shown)


hbrunn (changes requested):
if you really want to fix this here instead of the offending button handler, your code should do exactly what the server does: https://github.com/OCA/OCB/blob/8.0/addons/web/controllers/main.py#L373

I personally would be in favor of fixing the method you call.

Maartincm:
I used the confirmation popup from a wizard button that was formerly calling my python method, and the action was working ok, but after I call the same method from the popup button itself it does not.
That's why I though it would be nice to fix this problem from here.
I'd be glad to check the backend code and modify this PR. Thank you for the suggestion.

hbrunn (approved):

hbrunn:
thanks! I think you have a peculiar way of formatting your code, but given we have no formal conventions about js style, I won't discuss js style

Maartincm:
Haha Im a Python programmer, and not even that good to talk about conventions either, but being able to help is always nice :ok_hand:

StefanRijnhart (approved):

PR #622, opened 2017-05-16
10.0 mig web m2x options

chienandalu:
Just some migration tweaks.

cc @Tecnativa

PR #853, opened 2018-01-31
[IMP] Taking into account core options:

BT-aestebanez:
- In Odoo 10.0 the flags 'no_create_edit' and 'no_create' can be used. Without this improvement, they are ignored, and all the views which use this flags are ignored

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -239,12 +239,23 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
                     }
                 }
 
+                // TODO: Remove create_edit,create,no_create_edit and no_create options when migrating to 11.0 -> Odoo's core already support them by default
                 // create...
                 var create_edit = self.is_option_set(self.options.create) || self.is_option_set(self.options.create_edit),
-                    create_edit_undef = _.isUndefined(self.options.create) && _.isUndefined(self.options.create_edit),
+                    create_edit_undef = _.isUndefined(self.options.create) && _.isUndefined(self.options.create_edit) && _.isUndefined(self.options.no_create) && _.isUndefined(self.options.no_create_edit),
                     m2x_create_edit_undef = _.isUndefined(self.view.ir_options['web_m2x_options.create_edit']),
-                    m2x_create_edit = self.is_option_set(self.view.ir_options['web_m2x_options.create_edit']);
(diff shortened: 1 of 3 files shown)


BT-aestebanez:
@pedrobaeza I saw you merged this PR: https://github.com/OCA/web/pull/622
can you merge this too?

pedrobaeza:
It has to be reviewed by at least 2 persons and 5 working days delay or 3 persons. You need also to change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

BT-aestebanez:
Thanks for the information and sorry for the inconvenience

pedrobaeza:
You can review other PRs and ask in exchange that they review yours.

web_m2x_options/static/src/js/form.js
```diff
@@ -241,10 +241,20 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
 
                 // create...
                 var create_edit = self.is_option_set(self.options.create) || self.is_option_set(self.options.create_edit),
-                    create_edit_undef = _.isUndefined(self.options.create) && _.isUndefined(self.options.create_edit),
+                    create_edit_undef = _.isUndefined(self.options.create) && _.isUndefined(self.options.create_edit) && _.isUndefined(self.options.no_create) && _.isUndefined(self.options.no_create_edit),
```
yajo: So now we have `create_edit` and `no_create_edit` options? It seems a little bit absurd... IMHO, If Odoo core supports some of this addon's features, they should just be dropped in this addon, to DRY.

BT-aestebanez: Yes, I agree, it's absurd and they could be removed from this addons.
However, I'd suggest to keep them to make the module compatible with both attributes from the core, and attributes created before this commit.

But anyway, I'd refactor it completely in v11

yajo: In such case, you should please add a note in Known Issues (in the README) so we don't forget dropping that in the migration.

pedrobaeza: But you can put it here also as comment with `TODO`

BT-aestebanez: good idea :+1:

yajo (approved):

feketemihai (approved):

BT-aestebanez:
Hello, I realised this PR is not merged yet, are there any problem?
Thanks and regards

njeudy (approved):

PR #854, opened 2018-01-31
[IMP] web_dashboard_tile: reduce computation time

legalsylvain:
This PR reduces computation time for tile values.

before this PR, the value needed are loaded one by one.

after this PR, all the values are loaded by a search_read function.

regards.

CC : 
@ivantodorovich , @pedrobaeza.

### DIFF
--- a/web_dashboard_tile/models/tile_tile.py
+++ b/web_dashboard_tile/models/tile_tile.py
@@ -173,7 +173,10 @@ class TileTile(models.Model):
             elif self[f_function]:
                 func = FIELD_FUNCTIONS[self[f_function]]['func']
                 if func and self[f_field_id] and count:
-                    vals = [x[self[f_field_id].name] for x in records]
+                    field_name = self[f_field_id].name
+                    read_vals = records.search_read(
+                        [('id', 'in', records.ids)], [field_name])
+                    vals = [x[field_name] for x in read_vals]
                     value = func(vals)
             if self[f_function]:
                 try:
--- a/web_dashboard_tile/models/tile_tile.py
+++ b/web_dashboard_tile/models/tile_tile.py
@@ -157,7 +157,8 @@ class TileTile(models.Model):
             return
         fields = [f.name for f in [
             self.primary_field_id, self.secondary_field_id] if f]
-        read_vals = model.search_read(eval(domain, eval_context), fields)
+        read_vals = fields and\
+            model.search_read(eval(domain, eval_context), fields) or []
         for f in ['primary_', 'secondary_']:
             f_function = f+'function'
             f_field_id = f+'field_id'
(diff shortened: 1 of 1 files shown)


hparfr (approved):

mourad-ehm (approved):

ivantodorovich (commented):
To be honest, I'm not sure I understand how it improves performance.

Shouldn't you remove this block to do so?

```py
    if any([
        self.primary_function and
        self.primary_function != 'count',
        self.secondary_function and
        self.secondary_function != 'count'
            ]):
            records = model.search(eval(domain, eval_context))
```

And then

```py
    field_name = self[f_field_id].name
    read_vals = model.search_read(eval(domain, eval_context), [field_name])
    vals = [x[field_name] for x in read_vals]
```

-------

Or better yet, change it like this, so it only search_reads once for both fields:

```py
        if any([
            self.primary_function and
            self.primary_function != 'count',
            self.secondary_function and
            self.secondary_function != 'count'
                ]):
                fields = [f.name for f in [self.primary_field_id, self.secondary_field_id] if f]
                read_vals = model.search_read(eval(domain, eval_context), fields)
```

and then

```py
	if func and self[f_field_id] and count:
		vals = [x[self[f_field_id].name] for x in read_vals]
```


**Haven't tested any of it, but I think you'll get the point**

legalsylvain:
> To be honest, I'm not sure I understand how it improves performance.

well, 
- [x.field for x in records] generate many reads in the DB.
- search_read generate 1 read in the DB.

I just migrated from 7.0 to 8.0 a database, and the 8.0 module is very slower than the 7.0. (same tiles, same data). The current PR restore original performance.

About your proposal, I'll implement it. Nice idea. Thanks.

legalsylvain:
@ivantodorovich. 

I merge the two search_read as proposed, and reduce complexity code. Thanks for updating your review.
regards.

web_dashboard_tile/models/tile_tile.py
```diff
@@ -155,35 +155,30 @@ def _compute_data(self):
...
-        if any([
-            self.primary_function and
-            self.primary_function != 'count',
-            self.secondary_function and
-            self.secondary_function != 'count'
-                ]):
-                records = model.search(eval(domain, eval_context))
+        fields = [f.name for f in [
+            self.primary_field_id, self.secondary_field_id] if f]
+        read_vals = model.search_read(eval(domain, eval_context), fields)
```
ivantodorovich: What happens if there's no field set? (ie: `fields = []`)
How does `search_read` behave then?

ivantodorovich (commented):
Just a minor comment. Otherwise, LGTM 👍

legalsylvain:
Nice catch ! 
Well no bug, bug it will read all the fields, that is not the objective of this PR !
I fixed this. Ready to merge ?
thanks for your review !

ivantodorovich (approved):
:+1:

PR #863, opened 2018-02-12
[FIX-MIG][9.0][web_dashboard_tile] fix dashboard tile migration

ivantodorovich:
This fixes some issues with web_dashboard_tile migration to 9.0

- **CSS styles where ignored because 9.0 changed some html structure and css classes.**

- **Fixed 'add to dashboard' from the searchview menu**
Instead of inheriting board.board views, it now adds a standalone menu next to the board.board menu

Also...

- **Added a dropdown menu to edit the tile from the dashboard**
  - It's only visible in 'developer mode'.
  - Had to include form view to the dashboard action.

- **Added tile categories**
  - Using kanban group
  - Usefull for grouping tiles
  - Following odoo-style

---

ping @nicomacr


nicomacr (approved):

legalsylvain:
Hi @ivantodorovich. Thanks for this improvment. 
About the CSS / JS improvment, LGTM.
About the categories, maybe some implementation will loose in a UI PoV, because they have a lot of uncategorized tiles, so scrolling the view will be required. But well, this new implementation is more standard, that's good ! 

One thing I justed tested on runbot. If I move a tile in the kanban view ** in the same category** (up / down)

- the color of the tile is changing. Did you faced the same problem ? I don't know if it is an imcompatibility with another installed module. (like web_responsive...)
- if I refresh the view, the move of the tile is not taken into account.
image

- Could you cherry-pick https://github.com/OCA/web/pull/854 ? ;-)

One idea : could you add a link in the tile, in the kanban view, to have the possiblity to edit the configuration ? (like an image of a wheel).

kind regards.

ivantodorovich:
Hi @legalsylvain 

> One thing I justed tested on runbot. If I move a tile in the kanban view ** in the same category** (up / down)
>the color of the tile is changing. Did you faced the same problem ? I don't know if it is an imcompatibility with another installed module. (like web_responsive...)

Yes, I faced the same problem. I'm not sure how to solve this, though. Will look into it.

---

> if I refresh the view, the move of the tile is not taken into account.

Also noticed it, but didn't find a way to solve it. Any ideas?

---

> Could you cherry-pick #854 ? ;-)

Sure

---

> One idea : could you add a link in the tile, in the kanban view, to have the possiblity to edit the configuration ? (like an image of a wheel).

I added exactly that. But, in order to see it, you need to be on 'developer mode'. Just seemed something that should be hidden from regular users. I can make it visible to everyone, if you think it's best.

legalsylvain (approved):
> Yes, I faced the same problem. I'm not sure how to solve this, though. Will look into it.

You could maybe just add a Know issue section in the readme, regarding this weird bug. Did you try to play with the default colors ? maybe the text is written in white with white background color...

> Also noticed it, but didn't find a way to solve it. Any ideas?

I'm not a kanban expert, but we should maybe introduce a new 'sequence' field to indicate to odoo how to order the tiles into a category ?
Well, in fact, this is occuring with other native models.

> I added exactly that. But, in order to see it, you need to be on 'developer mode'. Just seemed something that should be hidden from regular users. I can make it visible to everyone, if you think it's best.

OK ! No it's good ! sorry for the noice. It works like a charm. thanks ! 
(Well in fact, I don't like the new 'developper mode' feature, but this is another story, that should be fixed in Odoo !)

👍, once cherry pick will be done.

ivantodorovich:
I've looked into the white background issue. When reordering inside the same group, the tile DOM element losses it's `style` attribute, where the `background-color` is set. It's weird. I'm not sure why.

As for maintining the sequence order, the `tile.tile` model already has a sequence, and it's working on the tree view (`widget="handle"`). I don't know how to make it work on the kanban view.. Maybe there's an attribute that should be set somewhere on the view definition that I'm not aware of.


- Added both issues to the known issues section
- Cherry-picked those commits

legalsylvain:
Hi. Maybe we could change default font-color (white for the time being) by a light gray ? If yes, the bug is less embarassing.

image

I just checked, installing stock on the same instance. The bug doesn't occure in stock picking type kanban view.
image

I suspected that something is hard-coded with the name 'color'. All the other models has the field background-color named color. Did you tried to change the name ? 

Note : I uninstalled web_responsive but it did'nt solve the problem.

ivantodorovich:
Fixed it! No more white background 😄 !

ivantodorovich:
This is ready to merge, if approved.

legalsylvain:
It works like a charm. Thanks for all the improvments.

regards.

PR #858, opened 2018-02-07
[FIX] web_m2x_options: fix ignored field options

matt454357:
The module was ignoring the local field options 'create' and 'create_edit'.
For example, specifying options="{'create': true, 'create_edit': true}" on a
field would not override the web_m2x_options system parameters.

PR #865, opened 2018-02-14
[FIX] web_m2x_options: fix ignored field options

matt454357:
The module was ignoring the local field options 'create' and 'create_edit'.
For example, specifying options="{'create': true, 'create_edit': true}" on a
field would not override the web_m2x_options system parameters.

Replaces PR #858

### DIFF
--- a/web_m2x_options/__manifest__.py
+++ b/web_m2x_options/__manifest__.py
@@ -2,7 +2,7 @@
 
 {
     "name": 'web_m2x_options',
-    "version": "10.0.1.1.0",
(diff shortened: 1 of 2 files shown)


jabibi (approved):
For the next time you could reset your previous branch to main 10.0 and submit your commits again. If you make the changes in the same branch there is no need to make a new PR. Hope it helps for the next time

As previous one, this is :+1:  for me

pedrobaeza (approved):
As it was already approved in the previous PR, I merge.

PR #871, opened 2018-02-20
[11][MIG] web_widget_x2many_2d_matrix

simahawk:
The widget has been refactored from scratch according to new v11 MVC architecture (heavily inspired by ListRenderer).
Now the widget is responsible only to prepare data for the matrix renderer which in turns takes care of rendering every field with its own widget and render aggregations on cols and rows.

**TODO**
- [x] cleanup and document code
- [x] update README (especially roadmap)
- [x] update screenshots
- [x] fix case where m2o fields are used as axis
- [x] test on runbot via new module `web_widget_x2many_2d_matrix_example`

**DESIDERATA**
- [ ] be able to nav from cell to cell w/ tabs and arrows (there's something similar in the list renderer) 

Example w/ stock_demand_estimate:
stock_demand_estimate


pedrobaeza:
You can consider to review #818 and include it here.

simahawk:
@pedrobaeza as stated in the description above :wink:

hbrunn:
are you planning to work on https://github.com/OCA/web/pull/871/files#diff-b6c895567d57ff79c5a7e594164a9e6dR143 too? I think 11's widget revamping should make this more or less straightforward, and then we could more or less use arbitrary field types in the matrix, which would be awesome of course

simahawk:
@hbrunn judging from the nice decoupling that we have now it should be easy to obtain (see https://github.com/OCA/web/issues/872). I'll see what I can do. We've already bumped on an issue w/ transient records whereas we don't have real ids in res_ids but xmlids of not yet existing records :/

grindtildeath (approved):
I didn't review the code but it seems to work on this PR : https://github.com/OCA/stock-logistics-warehouse/pull/381

pedrobaeza (changes requested):
There are very bad things done in this PR:

* You have lose the possibility to use the widget for a many2many.
* And more critical, you are writing values when changing a value in screen without pressing on Save!!

Please manipulate correctly the widget.

simahawk:
@pedrobaeza oh wow... didn't noticed that :thinking: :stuck_out_tongue_winking_eye: 
I know, this is why is still a WIP and I don't have time now to work full time on it.

Please wait correctly for the widget :wink:

simahawk:
@hbrunn I made it :wink: 

As stated in the updated description: the widget has been refactored from scratch according to new v11 MVC architecture (heavily inspired by ListRenderer). Now the widget is responsible only to prepare data for the matrix renderer which in turns takes care of rendering every field with its own widget and render aggregations on cols and rows.

It still needs some cleanup (as per description) but having your review will be much appreciated meanwhile :smile:

LoisRForgeFlow:
What is the status of this one? is it ready?

simahawk:
@lreficent sounds like it works fine w/ https://github.com/OCA/stock-logistics-warehouse/pull/381

We need some cleanup, docs and more testing. If you could help w/ more tests on stock demand, it would be more than great :wink: 

(FYI I handed it over to @tschanzt as I have no time to work on this now)

web_widget_x2many_2d_matrix/views/assets.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <template id="assets_backend" name="web_widget_x2many_2d_matrix assets" inherit_id="web.assets_backend">
+            <xpath expr="." position="inside">
+                <script type="text/javascript" src="/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js"></script>
```
MiquelRForgeFlow: `<script .... />`

simahawk:
@mreficent @lreficent should be ready now.

LoisRForgeFlow (approved):
Just locally tested with `stock_demand_estimate` and it works like a charm :+1: 

Runbot issue is not related to the module I think.

Thanks!

MiquelRForgeFlow:
A rebase might fix the runbot

simahawk:
rebased and squashed

MiquelRForgeFlow (approved):

simahawk:
@hbrunn gentle ping... when you have time... :wink:

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -0,0 +1,172 @@
...
+        this.by_x_axis = {};
+        this.by_y_axis = {};
+        this.field_x_axis = node.field_x_axis || this.field_x_axis;
+        this.field_y_axis = node.field_y_axis || this.field_y_axis;
+        this.field_label_x_axis = node.field_label_x_axis || this.field_x_axis;
+        this.field_label_y_axis = node.field_label_y_axis || this.field_y_axis;
+        this.x_axis_clickable = this.parse_boolean(node.x_axis_clickable || '1');
+        this.y_axis_clickable = this.parse_boolean(node.y_axis_clickable || '1');
+        this.field_value = node.field_value || this.field_value;
+        // TODO: is this really needed? Holger?
```
hbrunn, line 48: yes, I'd have asked about this at the end of the review: This seems to be a bit orthogonal to using the renderer, but this property allows you to pull some nice tricks like making specific fields in the matrix readonly and the like. This worked because before, you could set attributes directly on the html input element. I doubt there's an elegant way to do this with the matrix renderer (maybe something with simulating an attrs dict?), if so, please update the readme and move this to the roadmap, maybe someone has a good idea how to implement this. I wouldn't block on this I think

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -0,0 +1,172 @@
...
+        this.show_row_totals = this.parse_boolean(node.show_row_totals || '1');
+        this.show_column_totals = this.parse_boolean(node.show_column_totals || '1');
+        this.init_matrix();
+    },
+    /**
+     * Initializes the Value matrix.
+     * Puts the values in the grid. If we have related items we use the display name.
+     */
+    init_matrix: function(){
+      var self = this,
```
hbrunn, line 65: do i understand this correctly that this will yield the expected result even if the matrix is partial? if so, great, but please update the readme accordingly then

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
hbrunn: PS: I think if you can query those values and make them attributes of the table cell of the data point in question, this should be rather simple to do and we don't lose features in the migration. The good place would be https://github.com/OCA/web/pull/871/files#diff-d6a628d507f069353237db3c362415ccR218 - I think in most cases this was used to either set a class or make fields readonly/disabled. As the disabled property doesn't cascade, it wouldn't be a solution for this case, but where that's necessary an author could mark cells that are supposed to be disabled and put some small script on top of the widget to do so. Or patch an extra disabled field into the widget, but I don't think we need this in version 1

simahawk: yep, adding data attrs in there is fine. I'll add it to the roadmap and put a comment here.

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
simahawk: yep

PR #877, opened 2018-02-23
[ADD][11.0] Add module web_ir_actions_act_multi

pnajman-modoolar:
Hello,

This module gives us option to invoke list of actions on ActionManager.

Kind Regards,
Petar


web_ir_actions_act_multi/README.rst
```diff
@@ -0,0 +1,52 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
elicoidal: please check https://github.com/OCA/web/pull/881#pullrequestreview-99094143
Many remarks apply here

elicoidal (changes requested):
:+1: (some details)

simahawk (changes requested):
Same remarks as for https://github.com/OCA/web/pull/881

bealdav:
Thanks for this module

web_ir_actions_act_multi/README.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Usage
+=====
+
+To use this functionality you need to return following action with list of actions to execute:
+
+.. code-block:: python
+
+      @api.multi
+      def foo()
```
zaoral: example has a python error: missing : and self argument. Should be like
```python
def foo(self):
```

web_ir_actions_act_multi/README.rst
```diff
@@ -0,0 +1,52 @@
...
+=================
+
+This module provides a way to trigger more than one action on ActionManager
+
+Usage
+=====
+
+To use this functionality you need to return following action with list of actions to execute:
+
+.. code-block:: python
```
zaoral, line 16: You use standard rst, this WARNING appears in odoo log when running

```bash
2018-03-15 19:31:37,578 105 WARNING oca_modules odoo.addons.base.module.module: docutils' system message present: <system_message level="2" line="16" source=
"<string>" type="WARNING"><paragraph>Cannot analyze code. Pygments package not found.</paragraph><literal_block xml:space="preserve">.. code-block:: python
```

Use normal block of code instead

zaoral (commented):
LGTM +1.

However, will be nice if you consider the next comments. Also will be good to have demo data that let us to test this functionality, I think that more than a unit tests will be good to have a new module created as an example to show how the functionality works

pnajman-modoolar:
My apologies for such a long idle time.

web_ir_actions_act_multi/README.rst
pnajman-modoolar: Hi @zaoral, what's normal block of code? Can you please provide me an example. Thanks :)

web_ir_actions_act_multi/README.rst
pnajman-modoolar: Fixed

pnajman-modoolar:
Can we merge this with 11.0?

pedrobaeza:
2 positive reviews are needed

web_ir_actions_act_multi/README.rst
zaoral: > Hi @zaoral, what's normal block of code? Can you please provide me an example. Thanks :)

`code-block` is used by sphinx, is not standard rst directive. In rst you can create role to simulate this. you can find an example [here](https://stackoverflow.com/questions/10870719/inline-code-highlighting-in-restructuredtext)

But that is not the important issue, the reason why I was asking you to change it was because it was throwing a WARNING in the server log that was making this PR a fake "red runbot". But as far as I can see now odoo has integrated the rst package needed for support code-block directve since the last time you push. So now  you do not need to change the directive you can use ..code-block

pnajman-modoolar: Great, thanks. Can you then please give thumbs up so we can have two positive reviews and merge this PR with 11.0 branch ...

zaoral (approved):

LoisRForgeFlow (approved):
LGTM

PR #884, opened 2018-03-05
[10.0][MIG] migrate/rewrite web_polymorphic_many2one->web_widget_many2one_polymorphic

hbrunn:
you can test this on attachments, there, res_id will behave like a many2one field


pedrobaeza:
As the field type over you apply the widget is an integer, maybe a better module name is `web_widget_integer_polymorphic_m2o` and widget name `integer_polymorphic_m2o` or `integer_polymorphic_many2one`

hbrunn:
I disagree. All many2ones are ints, and this is an adaption of the many2one widget. If you really want to, this will work with a float, char or text field too, thanks to javascript's lose typing.

Probably I'll accompany this with its own field type in a while that sets up partial indexes for an int/char field combination, and then this is even more reference/many2one like as it is now.

pedrobaeza:
OK, thanks for the explanation. If this can be used with other field types, then the name is OK.

web_widget_many2one_polymorphic/static/src/js/web_widget_many2one_polymorphic.js
```diff
@@ -0,0 +1,44 @@
...
+    var core = require('web.core'),
+        form_relational = require('web.form_relational');
+
+    var FieldMany2OnePolymorphic = form_relational.FieldMany2One.extend({
+        init: function(field_manager, node) {
+            var self = this;
+            this._super.apply(this, arguments);
+            // reset our value when the model field changes
+            field_manager.on(
+                _.str.sprintf('field_changed:%s', this.options.model_field),
```
Tardo: Why not use string templates here?
https://github.com/lukehoban/es6features#template-strings

Yajo: We can't use ES6 features as long as we use phantomjs in our toolchain, which has a very old JS engine that would break tours and tests silently.

AFAIK in the upcoming v12, Odoo will use chrome headless to run tests, and thus we can start using XXI century JS! But for now, it's not possible, sorry :confused:

Tardo:
Ops I deleted the comment accidentally sry :/ thx for explain @Yajo 



Deleted messages:
 Tardo 31 minutes ago Member

Why not use string templates here?
https://github.com/lukehoban/es6features#template-strings
@Yajo
Yajo 20 minutes ago Member

We can't use ES6 features as long as we use phantomjs in our toolchain, which has a very old JS engine that would break tours and tests silently.

AFAIK in the upcoming v12, Odoo will use chrome headless to run tests, and thus we can start using XXI century JS! But for now, it's not possible, sorry confused

simahawk:
FYI: you can use [_.template](http://underscorejs.org/#template)

web_widget_many2one_polymorphic/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# Copyright 2010-2014 Elico Corp. All Rights Reserved.
+# Augustin Cisterne-Kaas <augustin.cisterne-kaas@elico-corp.com>
+# Copyright 2018 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Polymorphic many2one widget',
+    'summary': 'Allows developers to split reference fields in two fields',
+    'version': '10.0.1.0.0',
+    'category': 'Hidden/Dependency',
+    'depends': ['web'],
```
simahawk, line 11: website: https://github.com/OCA/web/

web_widget_many2one_polymorphic/README.rst
```diff
@@ -0,0 +1,74 @@
...
+    :alt: License: AGPL-3
+
+===========================
+Polymorphic many2one widget
+===========================
+
+The polymorphic field allows to dynamically store an id linked to any model in
+Odoo instead of the usual fixed one in the view definition, allowing comfortable
+editing in cases where you'd have to craft a reference field.
+
```
simahawk, line 12: time to switch to new README layout maybe? :)

web_widget_many2one_polymorphic/static/src/js/web_widget_many2one_polymorphic.js
```diff
@@ -0,0 +1,44 @@
...
+                this,
+                function() {
+                    // TODO: this resets too early
```
simahawk, line 19: still WIP then?

web_widget_many2one_polymorphic/views/templates.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
```
simahawk, line 2: `<odoo />` and remove `<data />`

web_widget_many2one_polymorphic/README.rst
```diff
@@ -0,0 +1,74 @@
...
+This way, the user gets the full featured many2one widget for the model in question.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. depend on it
+#. use the widget in the integer field and pass the field to be used for the model in the options dictionary
```
simahawk, line 31: does it work if you populate `res_model` via onchange? You could put a select on `res_model` for instance.

simahawk (changes requested):

ecino:
Hi, is this something ready to merge?

hbrunn:
I still have to follow up on @simahawk's comments

PR #885, opened 2018-03-06
[FIX] web_tree_many2one_clickable: Check if the node is a field

guewen:
If a list contains a node which is not a field (e.g. a button), it will
not be found in the fields so we'll have an error trying to get 'type'
from undefined.


simahawk (approved):

pedrobaeza (approved):

guewen:
@pedrobaeza done (directly amended, I amended only the version)

pedrobaeza:
OK, I think we can fast-track this.

Issue #842, opened 2018-01-18
Create a new repository for js lib


legalsylvain:
hi all, 

## Findings

For a while, there are a lot of js libraries that are included in oca "web" modules. (for graphical purpose, or to make light web apps) and I guess that the quantity of such modules will increase. 

sample : 
- stock-logistics-barcode / mobile_app_inventory : web app to make an inventory. [V7 version](https://github.com/OCA/stock-logistics-barcode/tree/7.0/mobile_app_inventory) ([PR V8](https://github.com/OCA/stock-logistics-barcode/pull/118) by @hparfr and [PR V10](https://github.com/OCA/stock-logistics-barcode/pull/120) by @alexis-via ) It is using angular, angular-cookie, angular-translate and ionic lib 
-  web / web_widget_bokeh_chart : widget to use bokeh lib to make nice graph. [V10](https://github.com/OCA/web/tree/10.0/web_widget_bokeh_chart) by @lreficent

Regarding [that discussion](https://github.com/OCA/web/pull/834#issuecomment-358619751).

The current deployment of js lib is very simple 
- add the js / css code in the static folder of the module
- include a new path of js / css lib in odoo xml file) 

That's cool and very easy to deploy. It works anywhere. (runbot, odoo.sh, appstore, etc...)

### Limitations

but in other hand, 
- it make huge PR that are not simple to review
- it is not possible to share the same lib between two modules. (**duplicated code**)
- and above all, it is a possible **fail of security**, because if a contributor propose a JS lib, he can alter it without real peer control.

### Step 1

- creating a new repository ~~`js_librairy`~~ `web_library` ([comment](https://github.com/OCA/web/issues/842#issuecomment-358652617)) under OCA umbrella
- in this repository, create a module per lib `lib_bokeh`, that contains only the lib (js & css files).
- If many major incompatible versions are available (for exemple, for angular JS), create 1 module per major version. `lib_angular_1`, `lib_angular_2`. ([comment](https://github.com/OCA/web/issues/842#issuecomment-358671132))

If a module has to use a lib, developper just have to : 
- add an odoo dependency
- add a path in a xml file 
    - `lib_bokeh/static/lib/bokeh.js`
    - `lib_bokeh/static/lib/bokeh.css`

### Step 2

In a second time, we can add a bot that check that the code is secured (not altered) up-to-date, ...

What do you think ?

CC : @hbrunn, @jbeficent, @sebastienbeau, @gurneyalex.

hbrunn:
[...]

PR #889, opened 2018-03-08
[10.0][FIX] web_widget_bokeh_chart: force the version of bokeh when using pip

LoisRForgeFlow:
@sbidoul is this correct?

### DIFF
--- a/setup/web_widget_bokeh_chart/setup.py
+++ b/setup/web_widget_bokeh_chart/setup.py
@@ -2,5 +2,11 @@ import setuptools
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        'external_dependencies_override': {
+            'python': {
+                'Bokeh': 'bokeh==0.12.7',
+            },
+        },
+    },
 )


sbidoul:
@lreficent the syntax is correct. But you should not pin a specific version. It's better to state a minimum version (>=0.12.7 for instance). Otherwise if another part of a project requires a more recent version of bokeh, you will have conflict.

LoisRForgeFlow:
@sbidoul The problem is that this module specifically uses the 0.12.7 version ([see libraries on the addon](https://github.com/OCA/web/tree/10.0/web_widget_bokeh_chart/static/src/lib/bokeh)). So when installing ddmrp using pip, the version installed is the latest (I think 0.12.14 or so) and the bokeh charts fail.

sbidoul:
Ah I see. In this case, ideally you should "vendor" the python part like you do with the js part. 

BTW, do you really need to include the js part in the Odoo addon? I thought bokeh was generating self contained js snippets to include in pages.

LoisRForgeFlow:
What do you mean with "vendor"?

Well, that was a fair point. Just locally tested, it seems that they are required.

sbidoul:
> What do you mean with "vendor"?

include the bokeh lpython lib inside the module in a sub namespace of `odoo.addons.web_widget_bokeh_chart`: not easy to do, probably.

sbidoul:
When I generate an HTML chart with bokeh, I get something like this in the HTML headers:

image

You should be able to do the same here.

LoisRForgeFlow:
>include the bokeh lpython lib inside the module in a sub namespace of odoo.addons.web_widget_bokeh_chart: not easy to do, probably.

what would be the benefit of doing so instead of just forcing the version?

>When I generate an HTML chart with bokeh, I get something like this in the HTML headers:

The way we are using it on DDMRP the js is inserted inline. Something like:

```<div class="bk-root">
    <div class="bk-plotdiv" id="dab7b716-25b8-4321-9105-5aeae263fbb8"></div>
</div>
<script type="text/javascript">
    (function() {
  var fn = function() {
    Bokeh.safely(function() {
    (...)
     });
  };
  if (document.readyState != "loading") fn();
  else document.addEventListener("DOMContentLoaded", fn);
})();

</script>
```
We are using this method `from bokeh.embed import components` that after the "bokeh magic" return you the html div and js script in a html tag. Honestly didn't try other options, which one do you use?

sbidoul:
Forcing the version will generate conflicts in bigger projects where bokeh would be used for something else.

> which one do you use

TBH I never mixed Odoo and bokeh in the same process so far.

LoisRForgeFlow:
Oh, yes that is true. But it someone use a more recent version of bokeh within odoo, they should also make a similar module for the libraries, like `web_widget_bokeh_v12.14`. And that reminds me this issue: https://github.com/OCA/web/issues/842. The problem is there but the solution seems not easy :sweat_smile:

hbrunn:
is this PR also the right place to fix https://travis-ci.org/OCA/web/jobs/350784115#L607-L614 (that's a bokeh dependency if I read teh buid correctly) or is this a general travis problem?

sbidoul:
@hbrunn the default python 2.7.6 used by travis is getting pretty old, could you try using 

```yaml
python:
  - "2.7.13"
```
in travis.yml?

LoisRForgeFlow:
@hbrunn @sbidoul I opened an issue some days ago for that problem: https://github.com/OCA/web/issues/893

hbrunn:
this one nearly drove me crazy, but with https://github.com/OCA/web/pull/907/commits/606158d02316f7d58c504b488346e703e5a7afbb, the build only crashes on 'normal' stuff. I'd suggest we go with this for the time being for 9, as the issue https://github.com/travis-ci/travis-ci/issues/8097 seems to be quite old; cf https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2

Don't remember why we want system site packages in the first place.

sbidoul:
> Don't remember why we want system site packages in the first place.

I think it was to get the apt installed lxml because compiling it was slow. Now that pypi has a binary wheel for it its not necessary anymore.

hbrunn:
https://github.com/OCA/web/pull/909

MiquelRForgeFlow:
@lreficent could you rebase?

PR #909, opened 2018-04-10
[FIX] higher python version, no system site packages

hbrunn:


pedrobaeza:
Please change module version number... No, it's a joke :yum: 

Let's see Travis.

hbrunn:
it will fail because there's some syntax error in javascript which makes phantomjs choke. But I've trouble finding the place where this happens

pedrobaeza:
No, it's green! I think you had the phantom PhantomJS failure (today I'm comical, hehe).

hbrunn:
interesting, will have a look then in my PR. Given runbot seems to have trouble (https://runbot2.odoo-community.org/runbot/static/build/3322475-909-3913e6/logs/job_20_test_all.txt), I think we can fasttrack this so that other 10 PRs get green, right?

In my native language and then the 80ies, I'd say you had a clown for breakfast today ;-)

pedrobaeza:
Yes, fast-tracking this

pedrobaeza:
I'm thinking that maybe the reason for runbot to fail is not having `system_site_packages: true`... @moylop260 can you confirm?

moylop260:
I saw the following error in the [runbot base log](https://runbot2.odoo-community.org/runbot/static/build/3322475-909-3913e6/logs/job_10_test_base.txt)
 - `/.repo_requirements/virtualenv/python2.7.13/bin/activate: No such file or directory`

Runbot don't have python2.7.13 version pre-loaded in the base image...
I will fix it

hbrunn:
@moylop260 which ones are there? (and where do I have to look that up to avoid the next stupid question?) we don't specifically need .13, I think from .8 or .9 we should be fine

moylop260:
@hbrunn 
The file binary of python is not named with the minor version and so it is not found it.

 - screen shot 2018-04-10 at 12 23 24

But if the python2.7.12 works for you then we can remove that subversion for the case of runbot for now.

I just used py2.7 for py2.7.12 from travis2docker package: 
 - https://github.com/Vauxoo/travis2docker/pull/115

I tested and now this project is building the image fine.

 - screen shot 2018-04-10 at 12 26 09


Note: The ideal case is added all minor version of python in the base image but is a lot of work... for now this quick fix resolve the main issue.

moylop260:
@gurneyalex 
Could you help us to update `travis2docker` python package latest version (4.0.6) for OCA's runbot server, please?

hbrunn:
wow, what a service. thank you so much!

gurneyalex:
@moylop260  Update done.

moylop260:
Thanks @gurneyalex
@hbrunn Now the 10.0 build is green in runbot
https://runbot.odoo-community.org/runbot/build/3322643

hbrunn:
I had no doubt that when the three of us combine we can generate greenness :-) Thanks everyone!

Do we have an opinion if I should PR my change to MQT?

moylop260:
Are you talking about bumpversion of python from [sample_files/.travis.yml](https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml#L9)?

hbrunn:
and removing system site packages, my understanding is we did this for speed (lxml), but now with the binary python package stuff (wheels) that's not really an issue any more

moylop260:
I got it
+1
It was for old Travis architecture. In fact, for py3.5 we need remove `system-packages` too.

And we should use minimal postgresql 9.4 in order to avoid other issues related like as:
 - https://github.com/OCA/maintainer-quality-tools/issues/432

Issue #739, opened 2017-10-03
Migration to version 11.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

# Modules to migrate

- [ ] help_online - By @qtheuret - #874 
- [ ] help_popup
- [ ] support_branding
- [x] web_access_rule_buttons - By @espo-tony - #1243 
- [x] web_action_conditionable - By @andreparames - #843 
- [x] web_advanced_search_x2x > web_advanced_search - By @Yajo #999
- [ ] web_advanced_search_wildcard - By @JBF91 - #985
- [ ] web_chatter_paste
- [ ] web_ckeditor4
- [ ] web_confirm_window_close
- [ ] web_context_in_colors
- [ ] web_dashboard_open_action
- [ ] web_dashboard_tile
- [x] web_decimal_numpad_dot - By @chienandalu - #828 
- [x] web_dialog_size - By @qtheuret - #869 
- [ ] web_dom_model_classes
- [ ] web_domain_field - By @Maartincm - #1019 
- [x] web_drop_target - By @grindtildeath - #974
- [ ] web_duplicate_visibility
- [x] web_editor_background_color - By @Yajo - #1074
- [x] web_environment_ribbon - By @xavierjimenez  - #796
- [x] web_export_view - By @JBF91 - #798 
- [x] web_favicon - By @misern2 - #800 
- [ ] web_graph_improved
- [x] web_group_expand - By @aldo774 - #928 
- [x] web_hide_db_manager_link - ~~By @JBF91 - #801~~ Already included in core
- [ ] web_ir_actions_act_window_message
- [ ] web_ir_actions_act_window_page
- [ ] web_last_viewed_records
- [x] web_listview_invert_selection - By @MeetKD - #1004 
- [x] web_listview_range_select - By @aldo774 - #910 
- [x] web_m2x_options - By @ernestotejeda - #952 
- [ ] web_menu_navbar_needaction
- [x] web_no_bubble - By @oihane - #765 
- [x] web_notify - By @dbo-odoo - #781 
- [ ] web_option_auto_color
- [ ] web_polymorphic_many2one
- [x] ~~web_readonly_bypass~~ Not needed as Odoo includes a new mechanism for this: (odoo/odoo@3b3f6f0)
- [ ] web_recipients_uncheck
- [x] web_responsive - By @angelmoya - #740  
- [ ] web_search_alphabetic
[...]

PR #911, opened 2018-04-13
11.0 mig web_timeline

Maartincm:
Migration of the Module **web_timeline** from V10 to V11.

I'm not a JS developer so any suggestion is more than welcome.

In the Issue #739 @pedrobaeza recommended the use of a new JS Library for this module. I'm working on  that rigth now. Anyway, this version of the module is working on my end and uses the same lib that was used in V10.

P.D: I'm also uploading a **PR** for the module **project/project_timeline**


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

PR #875, opened 2018-02-22
[11.0][MIG] web_tree dynamic_colored_field

ghost:

PR #638, opened 2017-05-22
Migration of web_tree_dynamic_colored_field to 10.0

jesusVMayor:
Added 9.0 commits from @hbrunn  and @sbidoul , changed version to 10.0 and openerp xml tags to odoo tags.

PR #913, opened 2018-04-19
[10.0][MIG] web_tree_dynamic_colored_field

TDu:
Only set it to be installable, and so far works as expected


pedrobaeza:
The default value is that one, and it's already migrated in #875

pedrobaeza:
Sorry, in #638 for v10.

PR #916, opened 2018-04-23
Import from odoo instead of openerp

jladage:
I think it's better to import from odoo instead of openerp on branch 11.0.

### DIFF
--- a/web_environment_ribbon/__manifest__.py
+++ b/web_environment_ribbon/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': "Web Environment Ribbon",
-    'version': '11.0.1.0.1',
+    'version': '11.0.1.0.2',
     'category': 'Web',
     'author': 'Francesco OpenCode Apruzzese, '
               'Tecnativa, '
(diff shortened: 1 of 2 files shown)


simahawk (changes requested):
LGTM, but, can you please:
1. change module version to `11.0.1.0.1`
2. rephrase the commit msg to include module name, like: `web_environment_ribbon: import 'odoo' instead of 'openerp'`

jladage:
Is it OK like this? Since I already pushed I don't know how to squash the two commits into one with the correct commit message. This is my first contribution, please let me know if I should do something else ;)

simahawk:
@jladage 1st of all: tnx for the contrib :)
Next time to replace the history you could simply do `git commit --amend` and then `git push -f`.
Now, as you've already pushed you can:
1. git rebase -i HEAD~2
2. follow the instructions. You can move the 1st commit below the second one and replace `pick` w/ `fixup`
When you save the commits will be squashed together. See http://gitready.com/advanced/2009/03/20/reorder-commits-with-rebase.html or any other git rebase doc :)

jladage:
I think I managed with your feedback, thanks alot!

dreispt (approved):

simahawk:
@jladage :+1:

simahawk (approved):

PR #917, opened 2018-04-25
[IMP] support any list view

hbrunn:
this way, we also have this on customized list views like the one for move lines added by the `account` module


pedrobaeza:

hbrunn:
done

tarteo (approved):

PR #919, opened 2018-05-02
[11.0][FIX] Make matrix editable depending of the status of the parent

MiquelRForgeFlow:
It doesn't make sense to always have the matrix in edit mode. Obviously, if you have the matrix in a wizard, it always will be in edit mode. But if you put the matrix in a normal model (not transient), then the expected behavior is that the matrix is editable if you are in edit mode and the matrix is not editable if you are in read mode.

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
@@ -238,7 +238,7 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
       if (modifiers.invisible && !(options && options.renderInvisible)) {
           return $td;
       }
-      options.mode = 'edit';  // enforce edit mode
+      options.mode = this.getParent().mode;  // enforce mode of the parent
       var widget = this._renderFieldWidget(node, record, _.pick(options, 'mode'));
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow:
@grindtildeath @simahawk @lreficent @jbeficent

JordiBForgeFlow:
cc @hbrunn

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -238,7 +238,7 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
       if (modifiers.invisible && !(options && options.renderInvisible)) {
           return $td;
       }
-      options.mode = 'edit';  // enforce edit mode
+      options.mode = this.__parentedParent.mode;  // enforce mode of the parent
```
simahawk: wouldn't be better to use `this.getParent().mode`?

simahawk (changes requested):
also, version update missing

simahawk (approved):

hbrunn (approved):
travis failure is unrelated

JordiBForgeFlow:
Thanks! We have a nice surprise coming from @mreficent shortly..

PR #926, opened 2018-05-15
[FIX][11.0][web_dialog_size] Fix discard

angelmoya:
On any form, if you edit and try to discard you get an error

check this lines:
https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/views/basic/basic_controller.js#L82
https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/core/dialog.js#L235

### DIFF
--- a/web_dialog_size/static/src/js/web_dialog_size.js
+++ b/web_dialog_size/static/src/js/web_dialog_size.js
@@ -32,7 +32,8 @@ Dialog.include({
                 handle: '.modal-header',
                 helper: false
             });
-        }
+        };
+        return this;
     },
 
     close: function() {


pedrobaeza:
There's already #927. Isn't the same?

angelmoya:
No, my PR  fix error when discard any form in edit mode.

The other PR is for fix discard modal.

qtheuret:
Thanks @angelmoya to have fixed this issue.

rafaelbn:
Thanks @angelmoya ! Is useful too put a tittle with the module name like [FIX][11.0][web_dialog_size] :smile:

rafaelbn (approved):
Functionally tested in runbot :+1:

angelmoya:
Thanks @rafaelbn it was a quick fix directly on github, I fixed the tittle

pedrobaeza (approved):
It lacks module version bump, but as there's 2 consecutive PRs, I amend it after merging both

Issue #921, opened 2018-05-05
[11.0]web_dialog_size Click Discard button on pop-up dialogs causing error


youring:
## How to reproduce:
* From **web_dialog_size** project page, click **TRY ME ON RUNBOT**
* Create a new sale order, on **Customer** fields, select **Create and Edit...** on drop down, which opens up the dialog
* Click **Discard**, error occurred. 

## Current behavior: 
Click on **Discard** button on any Many2one pop-up dialog will cause this error
```Error:
Uncaught TypeError: Cannot read property 'draggable' of undefined

http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/586-583e4f7/web.assets_backend.js:7108
Traceback:
TypeError: Cannot read property 'draggable' of undefined
    at Class.close (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/586-583e4f7/web.assets_backend.js:7108:820)
    at Class.close (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:3819:371)
    at Class.trigger (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:3915:180)
    at Class.trigger (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:3921:148)
    at Class.destroy (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:4054:589)
    at Class.prototype.(anonymous function) [as destroy] (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:3816:488)
    at Class.close (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:4054:509)
    at Class.close (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/586-583e4f7/web.assets_backend.js:7109:20)
    at Class.close (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:3819:371)
    at Object.<anonymous> (http://3325086-11-0-6feb30.runbot2.odoo-community.org/web/content/443-fba2744/web.assets_common.js:802:681)
```
Expected behavior:
There should be no error. **Discard** button is often used in case of unwanted modification, .

[...]

PR #927, opened 2018-05-15
[FIX] #921 Fixed issue when discard a modal with draggable element

qtheuret:
Fixed issue #921.


pedrobaeza:
There's already #926. Isn't the same?

rafaelbn (approved):
Tested functionally in runbot :+1:

pedrobaeza (approved):

PR #928, opened 2018-05-15
11.0 mig web_group_expand

aldo774:
Migration of the web_group_expand module to v11


mstuttgart:
PR original: https://github.com/multidadosti-erp/others-addons

web_group_expand/static/src/js/web_group_expand.js
```diff
@@ -0,0 +1,76 @@
...
+        var controller = this.active_view.controller;
+        this.toggle_group_records(false, controller)
+    },
+    
+    expand_records: function () {
+        var controller = this.active_view.controller;
+        this.toggle_group_records(true, controller)
+    },
+})
+return ViewManager;
```
tarteo: You don't need to return the ViewManager here. It is already accessible through web.ViewManager

tarteo (commented):
Thanks for the migration! Tested it on Runbot and it works perfectly, I have one small comment though.

web_group_expand/static/src/js/web_group_expand.js
```diff
@@ -0,0 +1,74 @@
...
+            var reset_button = searchview.find('#oe_group_by_reset');
+
+            if (res.groupBy.length > 0) {
+                reset_button.show()
+                expand_button.show()
+            }
+            else {
+                reset_button.hide()
+                expand_button.hide()
+            }
```
MeetKD: You can create a new method for ex. do_toggle_visibility instead of calling  hide/show 3 times.

aldo774: Previously i made some changes in button template, so, could i assume that you propose something like this ?:
```
render_view_control_elements: function (){
// ...
var has_groups = res.groupBy.length > 0
this.do_toggle_visibility(has_groups)
// ...
}

function do_toggle_visibility(show) {
    var buttons = searchview.find('.toggle_buttons');
    if (show) {
        buttons.show()
    }
    else {
        buttons.hide()
    }
}
```

MeetKD: Yes, You are right  but i would  propose to create a helper function like below :
```
do_toggle_visibility: function (show){
        var searchview = this.searchview_elements.$searchview_buttons
        var buttons = searchview.find('.toggle_buttons');
        if (show) {
            buttons.show()
        }
        else {
            buttons.hide()
        }
    }
```

Call this function from render_view_control_elements and _process_search_data method.

This helper function can be used in future as well.  If someone wants to hide/show group by buttons based on some condition than they can directly call this function.

aldo774: I liked, changes made

MeetKD: Great. :+1: from my side.

aldo774: If you liked it and approve, please, also leave your evaluation, needs 3 to be merged

web_group_expand/static/src/js/web_group_expand.js
aldo774: Thanks for your review, changes made

tarteo (approved):

MeetKD (approved):

rodrigets (approved):

PR #933, opened 2018-05-18
encoding declaration to avoid syntax error

fjouatte:
We need to declare encoding in the __init__.py of the web_timeline module to avoid this error : 

"SyntaxError: Non-ASCII character '\xc2' in file web_timeline/models/__init__.py on line 1, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details


pedrobaeza:
That's because you are using Python 2.x, not Python 3.x. In Python 3 this is totally unneeded if the file is already encoded.

PR #934, opened 2018-05-18
[10][ADD] Enhance relational fields dropdown and tags

MicheleZak:
This module allow to display additional fields to the many2one dropdown and the many2many tags.


web_relational_enhancer/__manifest__.py
```diff
@@ -0,0 +1,35 @@
...
+        Usage for M2o:
+
+        ::
+
+            <field name='<your field name>'
+                   options='{
+                       "display_fields": "name,state,description"
+                   }'/>
+    """,
+    'author': "Michele Zaccheddu - LevelPrime Srl",
```
MeetKD: Add OCA as a author.

web_relational_enhancer/__manifest__.py
```diff
@@ -0,0 +1,35 @@
...
+    'website': "http://www.levelprime.it",
```
MeetKD: 'website': 'https://github.com/OCA/web',

web_relational_enhancer/__manifest__.py
```diff
@@ -0,0 +1,35 @@
...
+    'category': 'Web',
+    'version': '1.0',
```
MeetKD: Can you please add proper version number as mentioned in oca guidelines ?

https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

web_relational_enhancer/__manifest__.py
```diff
@@ -0,0 +1,35 @@
...
+    'depends': ['base'],
```
MeetKD: Module should be depend on web module.

web_relational_enhancer/README.rst
```diff
@@ -0,0 +1,99 @@
...
+   :target: https://www.gnu.org/licenses/lgpl-3.0
+   :alt: License: LGPL-3
+
+
+=======================
+Web Relational Enhancer
+=======================
+
+
+This module allow to add information to the 'display_name' on
```
simahawk, line 11: hmm is this whole machinery really required just to change the display name?
I mean, you can use `context="{'use_special_name: True}"` on the field declaration and customize display name/ name_get to achieve the same result.

Can you elaborate on the use case and the benefit of maintaining this big piece of JS? Tnx :)

simahawk (changes requested):

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+            let self = this,
+                _super = this._super,
+                def  = enhanceMany2ManyTag.bind(self)(data)
+
+            if (!Array.isArray(def)) {
+                def.done(res => {
+                    _super.bind(self)(res)
+                })
+            } else
+                this._super(data)
```
tarteo: Here the variable _super can also be used

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+            ids     = _.pluck(_.filter(entryList, entry => {
+                        return ( !shouldSkipEntries || !(entry.label.includes('Create') || entry.label.includes('Search')))
+            }), 'id')
+
+        try {
+            dataset.read_ids(ids, fields).then(res => {
+                def.resolve(res)
+            })
+        } catch(err) {
+            console.log(err)
```
tarteo: Can you use console.error here please?

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+                        }
+                    })
+                })
+
+
+                def.resolve(data)
+            }).fail(err => console.log(`Error getting relational fields. Error: ${err}`))
+
+            return def
+        } else
```
tarteo: Else statement is irrelevant here

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+                            resp(result)
+                    })
+                },
+                select: function(event, ui) {
+                    isSelecting = true
+                    var item = ui.item
+                    if (item.id) {
+                        self.display_value = {}
+                        self.display_value_backup = {}
+                        self.display_value['' + item.id] = item.name
```
tarteo, line 277: Can u use toString here?

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+
+                def.resolve(data)
+            }).fail(err => console.log(`Error getting relational fields. Error: ${err}`))
+
+            return def
+        } else
+            return data
+    }
+
+    function enhanceMany2OneEntryLabel(result) {
```
tarteo: Can you move the functions to the field? This will enable other developers to change the behaviors in other modules.

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+
+        let self = this
```
tarteo, line 68: Missing semicolon

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+            /**
+               If 'display_fields' (comma-separated fields names) is present in 'options'
+               this method adds them in the displayed dropdown entry.
+               We can't use super 'cause of this arch
+             */
+            var self = this
+            this.$input = this.$('input')
+
+            this.init_error_displayer()
+            self.$input.on('focus', function() {
```
tarteo, line 123: You can use `this` here

web_relational_enhancer/static/src/js/field_enhancer.js
```diff
@@ -0,0 +1,309 @@
...
+    })
+
+    relational.FieldMany2One.include({
+        render_editable: function() {
+            /**
+               If 'display_fields' (comma-separated fields names) is present in 'options'
+               this method adds them in the displayed dropdown entry.
+               We can't use super 'cause of this arch
+             */
+            var self = this
```
tarteo, line 119: Why not use let instead of var here?

tarteo (changes requested):

web_relational_enhancer/README.rst
MicheleZak: Yes, because it changes the display name only on the dropdown widget, leaving the return values of those methods untouched.

simahawk: this is exactly what I'm talking about :)

you can set the ctx on the field itself and this will enable name_get override (which is used to show the label on the dropdown) only on this field an only when you set the ctx key...

I'm still looking for an answer then :)

MicheleZak: Yeah, ok. I didn't understand what you meant on the first answer. I thought it was a known feature instead of a keyword which you check on the name_get override. 

Your solution is better on a dynamic use case, but I still think it could be good also to have a static way to do that. It is simpler and faster from the usage point of view, because you only have to specify the fields, you want to see in the label, inside the view field options.

simahawk: I still don't agree... you can achieve what you want like:

```xml
context="{'display_fields': ['a', 'b','c']}"
```
then in this module you can simply have the name_get override and use those fields in sequence.
Or  you can even support a format pattern:
```xml
context="{'display_fields': ['a', 'b','c'], 'display_name_pattern': '{c}: {a} - {b}'}"
```
and you are done. This is the only thing you have to do as a user to enjoy the feature.

Just provide the base logic in this module (or a new one, like "web_display_name_improved" or similar...) to handle this context data.
You'll have a few python lines to maintain instead of this huge JS.

Using ctx keys to manipulate/customize look and feel of forms, fields and such is a common and well-know pattern in Odoo world ;)

MicheleZak: You're right. I already used it, I don't know why I did not think about it this time. I'll do it asap.

Thanks for the support.

PR #942, opened 2018-06-01
[FIX] fixes issue #937 : use lexer on text field only

### LINKED ISSUE #937: web_widget_text_markdown: Lexer passed array instead of string
**Issue:**

Install `web_widget_text_markdown`.
Open any sale order with sale order lines, it will throw js traceback. Issue is with tax_id field inside sale order lines.

An array is passed to the Markdown Lexer instead of string.

**Reproduce:**
The issue can be reproduced on our runbot with this [link](http://3325978-10-0-ed6ffe.runbot1.odoo-community.org/web?debug=assets#id=8&view_type=form&model=sale.order&menu_id=217&action=288).

**Demo:**
I have uploaded [demo video](https://youtu.be/Nc3GPivmxHU) **explaining/debugging** the bug.

Introduced in this [commit](https://github.com/OCA/web/commit/9b515fc408e14c5c375f62e5bef5421970f13a19)

Thank you.

njeudy:
- Fixes #937 

image

### DIFF
--- a/web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
+++ b/web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
@@ -2,137 +2,138 @@
  * Copyright 2017 Komit - <http:///komit-consulting.com>
  * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
 odoo.define("web_widget_text_markdown.bootstrap_markdown",
-            function (require) {
+  function(require) {
     "use strict";
 
     var core = require('web.core');
     var form_common = require('web.form_common');
-    var formats = require ("web.formats");
+    var formats = require("web.formats");
 
     var _lt = core._lt;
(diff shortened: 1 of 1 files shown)


njeudy:
@pedrobaeza travis is red on other module than markdown .. should I fixe here ?

njeudy:
@dek-odoo can you test ?

pedrobaeza:
@njeudy this is because now RST syntax errors are blocking (due to PyPi stricter policy), so if you do the favour of making a PR greenifying the branch, it would be great.

njeudy:
Ok but it should be on an other PR I think .. will try to work on it !

pedrobaeza:
Any news about the PR for fixing READMEs? And this should include the version number change.

njeudy:
@pedrobaeza local test are ok and @dek-odoo validate it (think he as test)

travis error are on an other module (will check if I can help)

PR #907, opened 2018-04-09
[ADD] web_drop_target

hbrunn:
Drop target support
===================

This module was written to extend the functionality of the web client to support dropping local files into the web client.

By default, this will create an attachment when a file is being dropped on a form, other modules can add more features.

Further, this module is meant as a base module supporting other actions after some file is dropped so that we don't get incompatible drag&drop modules within the OCA.

Usage
=====

To use this module, you need to:

- drag a file from your local computer onto an Odoo form view
- it should become an attachment of the currently opened record

Known issues / Roadmap
======================

* dropping on list or kanban views would be nice too
* handle multiple files
* add an upload progress meter for huge files

PR #943, opened 2018-06-02
[ADD][10.0] Added web module to drag-drop the files to create attachment

sudhir-erpharbor:
web module to allow user to drag-drop the files to create attachment


pedrobaeza:
There's already another module proposed in OCA for that: #907. It's not allowed to have several modules for the same goal, so please review the other proposal and propose things that can have yours and not the other for merging both in one.

I'm closing this right now.

PR #950, opened 2018-06-10
[FIX] web_widget_datepicker_options: Fix error when field is undefined

luisg123v:
Currently, when a view is rendered and a date or datetime field is not
defined (e.g. when designing reports), it causes a JS error.

This patch fixes the above error.


moylop260 (approved):

pedrobaeza:

luisg123v:
@pedrobaeza 
Thanks for pointing it out, done.

pedrobaeza (approved):

PR #951, opened 2018-06-11
[IMP][11.0][web_responsive] Add option to select chatter position

Tardo:
Add option in user settings to select chatter position.

Usage:
- Go to user preferences and select "Chatter Position"

### DIFF
--- a/web_responsive/__init__.py
+++ b/web_responsive/__init__.py
@@ -1,1 +1,4 @@
-# -*- coding: utf-8 -*-
+# Copyright 2018  Alexandre Díaz
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from . import models
--- a/web_responsive/models/__init__.py
+++ b/web_responsive/models/__init__.py
@@ -1,0 +1,5 @@
+# Copyright 2018  Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import inherited_res_users
+from . import ir_http
--- a/web_responsive/models/inherited_res_users.py
+++ b/web_responsive/models/inherited_res_users.py
@@ -1,0 +1,13 @@
+# Copyright 2018 Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    chatter_position = fields.Selection([
+        ('normal', 'Normal'),
+        ('sided', 'Sided'),
+    ], string="Chatter Position", default='normal')
--- a/web_responsive/readme/DESCRIPTION.rst
+++ b/web_responsive/readme/DESCRIPTION.rst
@@ -1,0 +1,8 @@
+This module provides a mobile compliant interface for Odoo Community web.
+
+Features:
+
+ * New navigation with an App drawer
+ * Keyboard shortcuts for easier navigation
+ * Display kanban views for small screens if an action or field One2x
+ * Set chatter side (Optional per user)
--- a/web_responsive/readme/USAGE.rst
+++ b/web_responsive/readme/USAGE.rst
@@ -1,0 +1,6 @@
+The following keyboard shortcuts are implemented:
+
+* Toggle App Drawer - `ActionKey <https://en.wikipedia.org/wiki/Access_key#Access_in_different_browsers>` + ``A``
+* Navigate Apps Drawer - Arrow Keys
+* Type to select App Links
+* ``esc`` to close App Drawer
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -9,8 +9,8 @@ odoo.define('web_responsive', function(require) {
     var SearchView = require('web.SearchView');
     var core = require('web.core');
     var config = require('web.config');
-    var FieldOne2Many = core.form_widget_registry.get('one2many');
     var ViewManager = require('web.ViewManager');
+    var Session = require('web.session');
 
     Menu.include({
 
@@ -26,7 +26,7 @@ odoo.define('web_responsive', function(require) {
             this._super(id);
             if (allowOpen) {
                 return;
-            };
+            }
             var $clicked_menu = this.$secondary_menus.find('a[data-menu=' + id + ']');
             $clicked_menu.parents('.oe_secondary_submenu').css('display', '');
         }
@@ -291,18 +291,18 @@ odoo.define('web_responsive', function(require) {
 
     // It inits a new AppDrawer when the web client is ready
     core.bus.on('web_client_ready', null, function() {
-        new AppDrawer();
+        return new AppDrawer();
     });
 
     // if we are in small screen change default view to kanban if exists
     ViewManager.include({
         get_default_view: function() {
-            var default_view = this._super()
+            var default_view = this._super();
             if (config.device.size_class <= config.device.SIZES.XS &&
                 default_view.type !== 'kanban' &&
                 this.views.kanban) {
                 default_view.type = 'kanban';
-            };
+            }
             return default_view;
         },
     });
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -1,14 +1,65 @@
 /* Copyright 2016 Ponto Suprimentos Ltda.
+   Copyright 2018 Alexandre Díaz
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+@sheet-margin: @sheet-padding;
+@chatter-side-width: 30%;
+
+// Sided Chatter
+@media (min-width: @screen-md) {
+    .o_chatter_position_sided {
+        .o_form_view {
+            display: flex;
+            height: 100%;
+
+            .o_form_sheet_bg {
+                border-right: 1px solid @table-border-color;
+                overflow: auto;
+                flex: 1 1 auto;
+            }
+
+            .oe_chatter {
+                overflow: auto;
+                flex: 0 0 @chatter-side-width;
+
+                .o_chatter_topbar {
+                    height: auto;
+                    flex-wrap: wrap;
+
+                    button:last-of-type {
+                        flex: 1 0 auto;
+                        text-align: left;
+                    }
+
+                    .o_followers {
+                        order: -10;
+                        flex: 0 1 100%;
+                    }
+                }
+
+                &:empty {
+                    display: none;
+                }
+            }
+        }
+    }
+}
+
+// Normal Chatter
+.o_chatter_position_normal {
+    .oe_chatter {
+        max-width: initial;
+    }
+}
 
 .o_form_view {
     // Form must fill 100% width in any size
     .o_form_sheet_bg {
-        padding: @sheet-padding;
 
         .o_form_sheet {
             min-width: auto;
             max-width: 100%;
+            margin: @sheet-margin;
         }
 
         @media (max-width: @screen-sm-max) {
--- a/web_responsive/views/inherited_view_users_form_simple_modif.xml
+++ b/web_responsive/views/inherited_view_users_form_simple_modif.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0"?>
+
+<!--
+    Copyright 2018
+    @author Alexanre Díaz <dev@redneboa.es>
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+
+<odoo>
+
+    <record id="view_users_form_simple_modif" model="ir.ui.view">
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form_simple_modif" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='email']" position="after">
+                <field name="chatter_position" />
+            </xpath>
+        </field>
+    </record>
+
+</odoo>
--- a/web_responsive/views/web.xml
+++ b/web_responsive/views/web.xml
@@ -2,6 +2,7 @@
 
 <!--
     Copyright 2016 LasLabs Inc.
+    Copyright 2018 Alexandre Díaz
     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 -->
 
@@ -92,6 +93,10 @@
 
             </header>
 
+        </xpath>
+
+        <xpath expr="//div[hasclass('o_main')]" position="attributes">
+            <attribute name="t-attf-class">o_main o_chatter_position_{{ json.loads(session_info)['chatter_position'] }}</attribute>
         </xpath>
 
     </template>
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -1,5 +1,5 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016-2017 LasLabs Inc.
+# Copyright 2018 Alexandre Díaz
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 {
@@ -9,7 +9,8 @@
     "version": "11.0.1.0.2",
     "category": "Website",
     "website": "https://laslabs.com/",
-    "author": "LasLabs, Tecnativa, Odoo Community Association (OCA)",
+    "author": "LasLabs, Tecnativa, Alexandre Díaz, "
+              "Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "installable": True,
     "depends": [
@@ -18,6 +19,7 @@
     "data": [
         'views/assets.xml',
         'views/web.xml',
+        'views/inherited_view_users_form_simple_modif.xml',
     ],
     'qweb': [
         'static/src/xml/form_view.xml',
--- a/web_responsive/models/ir_http.py
+++ b/web_responsive/models/ir_http.py
@@ -1,0 +1,16 @@
+# Copyright 2018 Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+from odoo.http import request
+
+
+class Http(models.AbstractModel):
+    _inherit = 'ir.http'
+
+    def session_info(self):
+        res = super(Http, self).session_info()
+        res.update({
+            'chatter_position': request.env.user.chatter_position or 'normal',
+        })
+        return res
--- a/web_responsive/readme/CONTRIBUTORS.rst
+++ b/web_responsive/readme/CONTRIBUTORS.rst
@@ -1,0 +1,5 @@
+* Dave Lasley <dave@laslabs.com>
+* Jairo Llopis <jairo.llopis@tecnativa.com>
+* Dennis Sluijk <d.sluijk@onestein.nl>
+* Sergio Teruel <sergio.teruel@tecnativa.com>
+* Alexandre Díaz <dev@redneboa.es>
--- a/web_responsive/readme/ROADMAP.rst
+++ b/web_responsive/readme/ROADMAP.rst
@@ -1,0 +1,12 @@
+Note: Data added to the footer ``support_branding`` is not shown while using
+this module.
+
+* Provide full menu search feature instead of just App search
+* Drag drawer from left to open in mobile
+* Figure out how to test focus on hidden elements for keyboard nav tests
+* If you resize the window, body gets a wrong ``overflow: auto`` css property
+  and you need to refresh your view or open/close the app drawer to fix that.
+* Override LESS styling to allow for responsive widget layouts
+* Adding ``oe_main_menu_navbar`` ID to the top navigation bar triggers some
+  great styles, but also `JavaScript that causes issues on mobile
+  <https://github.com/OCA/web/pull/446#issuecomment-254827880>`_
--- a/web_responsive/static/tests/js/web_responsive.js
+++ b/web_responsive/static/tests/js/web_responsive.js
@@ -1,3 +1,4 @@
+/* global QUnit */
 /* Copyright 2016 LasLabs Inc.
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
 
--- a/web_responsive/tests/test_ui.py
+++ b/web_responsive/tests/test_ui.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 


web_chatter_right/README.rst
```diff
@@ -0,0 +1,13 @@
+WEB CHATTER RIGHT
+=================
+
+Put chatter right
```
yajo: Please use new split readme system

web_chatter_right/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
yajo: Remove all of this.

web_chatter_right/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
yajo: :fire: Use short license headers, as found in the official template.

web_chatter_right/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+
+{
+    'name': 'Web Chatter Right',
+    'version': '11.0.1.0.0',
+    'author': "Odoo Community Association (OCA), Alexandre Díaz <dev@redneboa.es>",
+    'website': '',
```
yajo: Put oca repo website

web_chatter_right/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+    'summary': "Put chatter right",
+    'depends': [
+        'mail',
+    ],
+    'data': [
+        'views/general.xml',
+    ],
+    'qweb': [],
+    'test': [
+    ],
```
yajo: Remove empty keys

web_chatter_right/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+    'version': '11.0.1.0.0',
+    'author': "Odoo Community Association (OCA), Alexandre Díaz <dev@redneboa.es>",
+    'website': '',
+    'category': 'web/theme',
+    'summary': "Put chatter right",
+    'depends': [
+        'mail',
+    ],
+    'data': [
+        'views/general.xml',
```
yajo: Use a more specific name

web_chatter_right/static/src/css/chatter_right.css
```diff
@@ -0,0 +1,23 @@
+@media only screen and (min-width: 1024px) {
+  .o_form_sheet_bg.o_form_sheet_chatter_right {
+    width: calc(100% - 390px);
+    display: inline-block;
+  }
+
+  .oe_chatter.o_chatter_right {
+    display: inline-block !important;
+    width: 390px !important;
+    min-width: 390px !important;
```
yajo: Do not use `!important`. Those rules are extremely hard to theme. Better, use more specific selectors if you need to override some odoo css.

Also, use Less if possible.

web_chatter_right/static/src/css/chatter_right.css
```diff
@@ -0,0 +1,23 @@
...
+    background-color: white;
+  }
+
+  .o_followers {
+    position: absolute;
+    top: 2em;
+  }
+
+  .o_web_client > .o_main > .o_main_content > .o_content .table-responsive {
+    overflow-x: auto !important;
```
yajo: :fire:

web_chatter_right/static/src/js/chatter_right.js
```diff
@@ -0,0 +1,30 @@
+odoo.define('web_chatter_right.Chatter', function (require) {
+  'use strict';
+  /*
+   * Theme Chatter Right
+   * GNU Public License
+   * Alexandre Díaz <dev@redneboa.es>
+   */
```
yajo: Use short headers as in template.

web_chatter_right/static/src/js/chatter_right.js
```diff
@@ -0,0 +1,30 @@
+odoo.define('web_chatter_right.Chatter', function (require) {
```
yajo: 2 options: 

a. call it `web_chatter_side`. Arabic readers may want it left. 

b. Do not use a new addon. Instead, add the features to `web_responsive`. (I prefer this one)

web_chatter_right/static/src/js/chatter_right.js
```diff
@@ -0,0 +1,30 @@
...
+  var FormRenderer = require('web.FormRenderer');
+
+  FormRenderer.include({
+
+      _updateView: function ($newContent) {
+          this._super($newContent);
+          var $chatter = this.$el.find('.oe_chatter');
+          if (this.has_sheet && $chatter.length !== 0) {
+              if ($chatter.css('visibility') === 'hidden' || $chatter.css('display') === 'none') {
+                  this.$el.find('.o_form_sheet_bg').removeClass("o_form_sheet_chatter_right");
```
yajo: Possibly you could do all of this overriding the XML template instead, and it would be more clear.

Also, users should be able to choose these settings.

yajo (changes requested):

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,73 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+==============
+Web Responsive
+==============
+
```
yajo: Remove all until here. The new split readme system generates all this boilerplate.

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,73 @@
...
+This module provides a mobile compliant interface for Odoo Community web.
+
+Features:
+
+ * New navigation with an App drawer
+ * Keyboard shortcuts for easier navigation
+ * Display kanban views for small screens if an action or field One2x
+ * Set chatter side (Optional per user)
+
```
yajo: Remove from here...

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,73 @@
...
+Installation
+============
+
+Configuration
+=============
+
+Usage
+=====
+
```
yajo: ... until here.

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,73 @@
...
+Keyboard Shortcuts
```
yajo: Move this section to `readme/USAGE.rst`

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,73 @@
...
+------------------
+
+The following keyboard shortcuts are implemented:
+
+* Toggle App Drawer - `ActionKey <https://en.wikipedia.org/wiki/Access_key#Access_in_different_browsers>` + ``A``
+* Navigate Apps Drawer - Arrow Keys
+* Type to select App Links
+* ``esc`` to close App Drawer
+
```
yajo: Remove from here to the end.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -307,11 +311,42 @@ odoo.define('web_responsive', function(require) {
...
 
+    // Chatter Side
+    FormRenderer.include({
+
+        _updateView: function ($newContent) {
+            this._super($newContent);
+            var self = this;
+            var $chatter = this.$el.find('.oe_chatter');
+            if (this.has_sheet && $chatter.length !== 0) {
+                Rpc.query({
```
yajo: This is not the good hook to call for this. You're asking for the setting before you render each form. A normal user session renders many forms, so we should try to save requests whenever possible, as they affect directly to the performance. You have 2 options, to avoid doing so many calls:

- Call once, cache request. Similar to what's done in #954.
- Manage this from the backend. Overwrite [template `web.webclient_bootstrap`](https://github.com/odoo/odoo/blob/dbdd742d2046f208c0057fdc60b359e5c88814ee/addons/web/views/webclient_templates.xml#L642), by adding a class such as `chatter-position-#{user.chatter_position}`.

  Then, no JS is needed at all; you can reposition it through Less directly.

IMHO I'd go for 2nd option. I think it will be faster for the user and easier to maintain. But you choose :wink:

web_responsive/views/inherited_view_users_form_simple_modif.xml
```diff
@@ -0,0 +1,14 @@
...
+
+  <record id="view_users_form_simple_modif" model="ir.ui.view">
+    <field name="model">res.users</field>
+    <field name="inherit_id" ref="base.view_users_form_simple_modif" />
+    <field name="arch" type="xml">
+      <xpath expr="//field[@name='email']" position="after">
+        <field name="chatter_position" />
+      </xpath>
+    </field>
+  </record>
```
yajo: Please use 4 spaces for indentation, to be consistent with other files.

yajo (changes requested):
Some comments on code below. But it's awesome!

I've been comparing it with how `web_enterprise` makes it, and I noticed a couple of little UI/UX details that would be IMHO important:



1. Add some separation between the form and the chatter. Just 1px line, like in the picture: 
captura de pantalla de 2018-06-13 11-13-14

2. Separate scroll bars for form and chatter. With your patch, user scrolls both or none, but separate scrolling is useful:
captura de pantalla de 2018-06-13 11-13-33

web_responsive/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Tecnotel - Alexandre Díaz
+# Copyright 2018  Alexandre Díaz
```
yajo: 2 spaces?

web_responsive/models/__init__.py
```diff
@@ -1,4 +1,5 @@
-# Copyright 2018 Tecnotel - Alexandre Díaz
+# Copyright 2018  Alexandre Díaz
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import inherited_res_users
+from . import inherited_ir_http
```
yajo: No need to prefix with `inherited_`

web_responsive/models/inherited_res_users.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Tecnotel - Alexandre Díaz
+# Copyright 2018  Alexandre Díaz
```
yajo: 2 spaces?

web_responsive/readme/USAGE.rst
```diff
@@ -0,0 +1,13 @@
+Keyboard Shortcuts
+------------------
```
yajo: Remove this header please.

web_responsive/readme/USAGE.rst
```diff
@@ -0,0 +1,13 @@
...
+The following keyboard shortcuts are implemented:
+
+* Toggle App Drawer - `ActionKey <https://en.wikipedia.org/wiki/Access_key#Access_in_different_browsers>` + ``A``
+* Navigate Apps Drawer - Arrow Keys
+* Type to select App Links
+* ``esc`` to close App Drawer
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/11.0
```
yajo: Remove the runbot stuff please. It should be autofilled.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -10,11 +9,9 @@ odoo.define('web_responsive', function(require) {
     var SearchView = require('web.SearchView');
     var core = require('web.core');
     var config = require('web.config');
-    var FieldOne2Many = core.form_widget_registry.get('one2many');
+    //var FieldOne2Many = core.form_widget_registry.get('one2many');
```
yajo: :fire:

web_responsive/views/web.xml
```diff
@@ -94,6 +94,10 @@
 
         </xpath>
 
+        <xpath expr="//div[@class='o_main']" position="attributes">
+            <attribute name="t-att-class">'o_main o_chatter_position_%s' % json.loads(session_info)['chatter_position']</attribute>
```
yajo: I'd use `t-attf-class` instead for this.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -291,27 +292,44 @@ odoo.define('web_responsive', function(require) {
...
         },
     });
 
+    // Check Chatter Visbility
+    FormRenderer.include({
+
+        _updateView: function ($newContent) {
+            this._super($newContent);
+            var $chatter = this.$el.find('.oe_chatter');
+            if ($chatter.length === 0 || $chatter.css('visibility') === 'hidden' || $chatter.css('display') === 'none') {
```
yajo: Use `$chatter.is(":hidden")`

web_responsive/static/src/js/web_responsive.js
```diff
@@ -291,27 +292,44 @@ odoo.define('web_responsive', function(require) {
...
+                this.$el.find('.o_form_sheet_bg').removeClass("o_form_sheet_bg_chatter_visible");
+            } else {
+                this.$el.find('.o_form_sheet_bg').addClass("o_form_sheet_bg_chatter_visible");
```
yajo: I'm wondering if do we need all of this now. :thinking: 

You are now adding a class to the whole web client that indicates the chatter position. Can't we use that class directly in Less to change the chatter position, and remove all related JS code?

I see that the point is to have 100% width in forms without chatter, but see suggestions below on Less code.

web_responsive/views/web.xml
```diff
@@ -94,6 +94,10 @@
 
         </xpath>
 
+        <xpath expr="//div[@class='o_main']" position="attributes">
```
yajo: Use the v11 new `hasclass()` method instead.

web_responsive/static/src/less/form_view.less
```diff
@@ -1,14 +1,48 @@
 /* Copyright 2016 Ponto Suprimentos Ltda.
+   Copyright 2018 Alexandre Díaz
  * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
 
+@sheet-margin: @sheet-padding;
+@chatter-side-width: 390px;
```
yajo: I'd use some variable from [current bootstrap theme](https://github.com/odoo/odoo/blob/11.0/addons/web/static/lib/bootstrap/less/variables.less). i.e. `@screen-md / 3`

web_responsive/static/src/less/form_view.less
```diff
@@ -1,14 +1,48 @@
...
+
+// Sided Chatter
+.o_chatter_position_sided {
```
yajo: My suggestion is to use flexbox here. You have a playground in https://demo.agektmr.com/flexbox/. If I'm not wrong, this should do the trick (or almost):

```less
@media (min-width: @screen-md) {
.o_chatter_position_sided {
  .o_form_view {
    display: flex;
    flex-flow: row;
    height: 100%;
    align-items: stretch;

    .o_form_sheet_bg {
      border-right: 1px solid @table-border-color;
      overflow: auto;
      flex: 3 0 auto;
    }

    .o_chatter {
      overflow: auto;
      flex: 1 1 @chatter-side-width;

      .o_chatter_topbar {
        height: auto;
        flex-wrap: wrap;
      }
    }
  }
}
}
```

The benefits:

- No JS needed. Thus, more fast.
- Adaptable to any size.
- Autofills if no chatter available.
- Cleaner, IMHO.

yajo (changes requested):

Tardo:
Thanks for all the help @Yajo , now i know how flex property works xD and like chatter toolbar solution too!

yajo (approved):
Code looks perfect, and functionality too. Thanks for the hard job!

The Travis :x: is due to new linters that affect other files in the addon, so the fault is not yours. If you want to fix it, that'd be OK, but not a requirement IMHO.

elicoidal:
I would suggest to keep travis green otherwise it will pollute the repo

pedrobaeza:
I have cleaned the branch and rebase this one, so it should be all correct now.

pedrobaeza:
Good work with this, @Tardo !

Using the new feature, I see that it's better than enterprise one, not using too much width, but sometimes this can be short for showing large content, like embedded large images. Do you think it's possible to make the width configurable through the typical arrows?

Another thing I have seen. With this width, not everything fits on the first line:

seleccion_005

making a visual inconsistency, not having connected the opened tab with the rest of the content, which is very important IMO for identifying easily if you are sending an internal message or a public one. See the difference with the right normal one:

seleccion_006

Possible solutions to this are:

* Make the width bigger, but this depends on the translation of the elements.
* Put the followers up. My colleague @Yajo said to me that it's very easy with flexbox.

web_responsive/static/src/less/form_view.less
```diff
@@ -1,14 +1,47 @@
...
+@media (min-width: @screen-md) {
+    .o_chatter_position_sided {
+        .o_form_view {
+            display: flex;
+            height: 100%;
+
+            .o_form_sheet_bg {
+                border-right: 1px solid @table-border-color;
+                overflow: auto;
+                flex: 1 1 auto;
```
yajo, line 18: `3 0 auto`, which means: *give him 3/4 of the space when growing, but not let it shrink less than 1/2 of the space*

pedrobaeza:
Ah, another thing abusing a bit :angel: 

On previous versions, if the chatter in bottom, the content was fit to 100%. Is it possible that you change also this thing in this PR?

Thanks again.

web_responsive/static/src/less/form_view.less
```diff
@@ -1,14 +1,47 @@
...
+            }
+
+            .o_chatter {
+                overflow: auto;
+                flex: 0 0 @chatter-side-width;
```
yajo: `1 1 @chatter-side-width`, which means: *give him 1/4 of the space when growing, also when shrinking, and start with `@chatter-side-width` to make computations*.

web_responsive/static/src/less/form_view.less
```diff
@@ -1,14 +1,47 @@
...
+
+                .o_chatter_topbar {
+                    height: auto;
+                    flex-wrap: wrap;
+                }
+            }
+            .oe_chatter:empty {
```
yajo: You can put this inside the `.o_chatter` block and use `&:empty` instead.

web_responsive/static/src/less/form_view.less
```diff
@@ -1,14 +1,47 @@
...
+            }
+
+            .o_chatter {
+                overflow: auto;
+                flex: 0 0 @chatter-side-width;
+
+                .o_chatter_topbar {
+                    height: auto;
+                    flex-wrap: wrap;
+                }
```
yajo: Inside this block, add this other one:

```less
.o_followers {
    order: -10;
    flex: 2 0 @chatter-side-width;
}
```

That means: *Put this element before the other ones (that default to `order: 0`); then, give him the double than the other ones when growing (because it has 2 buttons), but do not let it shrink smaller than `@chatter-side-width`*

yajo (changes requested):
With the changes I suggest below, you get this behavior, automatically, depending on user screen width:

captura de pantalla de 2018-06-20 10-07-27

captura de pantalla de 2018-06-20 10-07-36

web_responsive/static/src/less/form_view.less
Tardo: Use different grow & shrink values for me makes the view "inconsistent" ... 
2018-06-20-153006_1366x768_scrot
2018-06-20-153003_1366x768_scrot
2018-06-20-153000_1366x768_scrot
2018-06-20-152927_1366x768_scrot

web_responsive/static/src/less/form_view.less
Tardo: The element only have the class "oe_chatter" when it's empty ... perhaps best use oe_chatter for all?

web_responsive/static/src/less/form_view.less
Tardo: 2018-06-20-160030_1366x768_scrot

I have better results (if chatter have fixed size) with:
```css
.o_followers {
    order: -10;
    flex: 0 1 @chatter-side-width;
 }
```
Use all the chatter width (no more, but if needed less)
2018-06-20-155938_1366x768_scrot

web_responsive/static/src/less/form_view.less
yajo: Oh, I didn't know that. Then maybe yes, we can use `oe_chatter` for both.

web_responsive/static/src/less/form_view.less
yajo: Well, the point is that i.e. here I'm working with an ultrawide monitor. Under this circumstance, when window is full width, the chatter is extremely small. Emails sometimes have fixed widths, and with such a big monitor, it seems more comfortable to make it grow a little bit when the screen is big enough. Instead of making it feel unconsistent, here it feels more responsive.

However, if you prefer, you can combine `flex` with `min-width` and `max-width`. Play with those as you like. For instance, setting the chatter as 1/3 of the window, with `min-width: @chatter-side-width`. Do you prefer if I add the commit myself, since I have this monitor to test with?

web_responsive/static/src/less/form_view.less
yajo: OK, seems legit.

web_responsive/static/src/less/form_view.less
Tardo: Ok, i changed width to use percentage instead fixed width...
Feel free to commit changes what you think appropriate.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -9,8 +9,9 @@ odoo.define('web_responsive', function(require) {
     var SearchView = require('web.SearchView');
     var core = require('web.core');
     var config = require('web.config');
-    var FieldOne2Many = core.form_widget_registry.get('one2many');
     var ViewManager = require('web.ViewManager');
+    var Chattter = require('mail.Chatter');
```
yajo: If you do this, you must depend on mail. Check runbot, it's not loading this file.

yajo (changes requested):

web_responsive/static/src/js/web_responsive.js
```diff
@@ -291,27 +292,79 @@ odoo.define('web_responsive', function(require) {
...
+            if (self._mousePosX !== -1) {
+                $('body').css('user-select', 'initial');
+                self._mousePosX = -1;
+            }
+        },
+        _onMainMouseMove: function(ev) {
+            var self = ev.data;
+            if (self._mousePosX !== -1) {
+                var wDiff = self._mousePosX - ev.pageX;
+                self.$el.css('flex-basis', self._origWidth + wDiff + 'px');
```
yajo: I'm not sure if I prefer you to add the dependency on `mail`, or better just remove this feature. I think this feature is too complex and nobody will care about resizing that column if it has a sane and properly responsive default.

You know, the famous debate between configurability vs. sane defaults... I think this is a good case for sane defaults. If you need to see wider mails, the officially supported answer should be: "configure your chatter in the bottom; it will have full width", don't you think?

Besides, by reading the code I'd say you are not storing this anywhere... So each time I press F5 I lose my custom width? And if I resize the window up and down, do I have to be dragging it all the time? It seems one of those features that nobody will care if it's not present, but yes if it's incomplete.

Just some thoughts, please don't take this badly. :innocent:

pedrobaeza: If we have to add the dependency, I agree it's better to stick to a sane fixed width that can vary according screen size.

yajo: If you agree, since you originated the request, I'm gonna remove this. Let's see how it works after that.

pedrobaeza:
OK, I see it now perfect! @Tardo do you agree also? Should we merge?

Do you have Twitter for announcing the improvement there?

Tardo:
For me all ok to merge. Yes i have an twitter account.

pedrobaeza:
Thanks for the work. Please tell me which one for mentioning you for giving some way thanks.

Tardo:
My twitter account is: NoMoreButtons ;)

LRovira-PlanetaTIC:
The preferences option is only available for users with administration rights. It is right?

yajo:
Yikes, true, that's a bug we didn't find at time for release:

captura de pantalla de 2018-06-25 09-13-37

PR #952, opened 2018-06-11
11.0 mig web_m2x_options

ernestotejeda:
* Update README
* Change Javascript

Cc @Tecnativa


web_m2x_options/__manifest__.py
```diff
@@ -1,18 +1,23 @@
...
 {
     "name": 'web_m2x_options',
-    "version": "10.0.1.1.0",
+    "version": "11.0.1.0.0",
+    'category': 'Web',
+    "author": "ACSONE SA/NV, "
+              "0k.io, "
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    'website': 'https://github.com/oca/web',
```
yajo: OCA, uppercase

web_m2x_options/readme/USAGE.rst
```diff
@@ -0,0 +1,92 @@
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
```
yajo: :fire: Delete until here.

web_m2x_options/static/src/js/form.js
```diff
@@ -0,0 +1,393 @@
...
+    });
+
+    FormFieldMany2ManyTags.include({
+        events: _.extend({}, FormFieldMany2ManyTags.prototype.events, {
+            'click .badge': '_onOpenBadge',
+        }),
+
+        start: function() {
+            this._super.apply(this, arguments);
+            return this.get_options();
```
yajo: By doing this, you're fetching the options each time the field widget is instantiated. The options will always be the same, so this will produce many RPC calls when only the 1st one is actually needed.

You need to get the params before. For instance, when loading the addon, or when using it for the 1st time. Then cache and reuse them.

web_m2x_options/static/src/js/form.js
```diff
@@ -0,0 +1,393 @@
...
+        },
+
+        get_options: function() {
+            var self = this;
+            if (_.isUndefined(this.ir_options_loaded)) {
+                this.ir_options_loaded = $.Deferred();
+                this.ir_options = {};
+                self._rpc({
+                    model: "ir.config_parameter",
+                    method: 'search_read',
```
yajo: This model now only is readable by admins, so other users cannot get the configs. Just enter runbot and use a form as demo user.

You need to create a different method in the backend that reads the specified values with `.sudo()`. This can make the client code easier too.

yajo (changes requested):
Great JS code, congratulations! However, it's not working. https://github.com/OCA/web/pull/954 can inspire you, I'm fixing basically the same there.

yajo (approved):

yajo:
I applied some makeup to git history, to get it ready to merge, but all migration code is very clean, thanks! (And it works now)

chienandalu (changes requested):
Thanks @ernestotejeda !

Seems to work fine with m2o fields but not with m2m fields. For example:

- Set in params `web_m2x_options.create` to `False`
- In a m2o field you can't create a record from the dropdown.
- In a m2m field you have the options instead.

PR #954, opened 2018-06-12
[FIX] web_dialog_size: Fix usage for non-admins and reduce calls

yajo:
This addon had 2 problems after migrating to v11:

1. One call to backend's `ir.config_parameter.get_param` was done for
   each instantiated dialog, while the setting served for the whole
   session equally.
2. That model is now readable only by admin users, so non-admins
   couldn't use the default at all.

Fixed now.

@Tecnativa


web_dialog_size/README.rst
```diff
@@ -1,67 +0,0 @@
...
-   :alt: Odoo Community Association
-   :target: https://odoo-community.org
-
-This module is maintained by the OCA.
-
-OCA, or the Odoo Community Association, is a nonprofit organization whose
-mission is to support the collaborative development of Odoo features and
-promote its widespread use.
-
-To contribute to this module, please visit https://odoo-community.org.
```
pedrobaeza, line 67: Please don't remove totally this file, as runbot will fail. You have 2 options:

* Put a line like the template one: `**This file is going to be generated by oca-gen-addon-readme.**`
* Run oca-gen-addon-readme locally for generating the README and not letting the bot do it at night.

pedrobaeza (approved):
Little thing

web_dialog_size/README.rst
yajo: why? doesn't it create it later automatically?

pedrobaeza: Yeah, but meanwhile you will have a red build

pedrobaeza:
Travis is red

pedrobaeza:
I'm afraid this is still red

rafaelbn (approved):
:+1: all :green_apple: @pedrobaeza

chienandalu (approved):

PR #955, opened 2018-06-13
[REM] web_widget_datepicker_options: No longer needed

yajo:
This addon is no longer needed since https://github.com/odoo/odoo/pull/25044#issuecomment-396940644, so I remove it.

WIP until:

- [x] https://github.com/odoo/odoo/pull/25235

@tecnativa


pedrobaeza:
cc @moylop260

luisg123v:
@Yajo 
Just to be sure, are you sure the change in Odoo works fine for both date and datetime?

yajo:
You're right, I opened https://github.com/odoo/odoo/pull/25235 to fix that case. Thanks!

I'll set this as WIP until that's merged upstream.

tarteo (approved):
Can you also remove the pot file, please?

yajo:
Upstream was fixed in https://github.com/odoo/odoo/pull/27541 and we didn't notice it here.

This is ready to merge.

PR #963, opened 2018-06-21
[ADD] [8.0] - web_no_crawler

nedaszilinskas:


pedrobaeza:
Is this a backport of the module in 9.0? If so, you should preserve commit history for crediting the original authors. You can use https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#technical-method-to-migrate-a-module-from-80-to-90-branch, but switching the order of `8.0` / `9.0` references.

nedaszilinskas:
This is not a backport. Its a new module.

pedrobaeza:
OK, as the module already exists in next versions, but OK if you have started from scratch. Thanks for the contribution!

README.md
```diff
@@ -54,6 +54,7 @@ addon | version | summary
 [web_menu_autohide](web_menu_autohide/) | 8.0.1.0.0 | Hide top and left menu bar
 [web_menu_collapsible](web_menu_collapsible/) | 8.0.1.0.0 | Foldable second level Odoo menu
 [web_menu_navbar_needaction](web_menu_navbar_needaction/) | 8.0.1.0.0 | Show the sum of submenus' needaction counters in main menu
+[web_no_crawler](web_no_crawler/) | 8.0.1.0.0 | Implement robots.txt / disallow indexing
```
pedrobaeza: No need to change this. It's done by a bot

web_no_crawler/README.rst
```diff
@@ -0,0 +1,59 @@
...
+=============
+
+No configuration needed.
+
+Usage
+=====
+
+Once installed adds robots.txt (ex.: http://example.org/robots.txt).
+
+Known issues / Roadmap
```
pedrobaeza: Remove this section if empty because it's going to give an RST error if not

web_no_crawler/__openerp__.py
```diff
@@ -0,0 +1,18 @@
...
+    'author': 'Xpansa Group, Odoo Community Association (OCA)',
+    'website': 'http://xpansa.com',
+    'category': 'web',
+    'version': '8.0.1.0.0',
+    'depends': [
+        'base',
+    ],
+    'installable': True,
+    'application': False,
+    'images': ['static/description/icon.png'],
```
pedrobaeza: No need to add this image

web_no_crawler/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# © 2018 Xpansa Group
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Implement robots.txt / disallow indexing',
+    'author': 'Xpansa Group, Odoo Community Association (OCA)',
+    'website': 'http://xpansa.com',
```
pedrobaeza: We are now putting GitHub repository here

pedrobaeza (dismissed):
You can use the new README by fragments system for not having to maintain the full README. See the sample here: https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

nedaszilinskas:
Which module do you refer to (next version)? Maybe you refer to https://github.com/OCA/website/tree/8.0/website_no_crawler?
This module (web_no_crawler) is for those who do not want do install Website module (where the robots.txt is implemented originally).

pedrobaeza:
Ah, totally right! I have mixed both, but not having website, is it exposed to crawlers?

nedaszilinskas:
The robots.txt route is public and it disallow to index.

pedrobaeza:
With no website, there's no robots.txt nor the possibility to index AFAIK.

@Yajo @simahawk can you confirm?

nedaszilinskas:
Requested changes pushed. Please let me know if its fine.

yajo:
Indeed, all it could crawl is, basically, `/web/login`, which would never appear in a search result in any search engine IMHO.

[The `robots.txt` controller is defined in website addon](https://github.com/odoo/odoo/blob/4045e4fe66fc18f3e223a99c781a08a17ee49f60/addons/website/controllers/main.py#L85).

So this is not needed, closing.

okuryan:
@Yajo , I do not agree. Even if only opened page is web/login (also remember that there are also DB manager and etc.) I do not want them to appear in google at all.
so if I have erp.xxx.com. I do not want any entry in google to refer to erp.xxx.com

Does it make sense to you?

yajo:
OK, but you can fix that easier with a header in your proxy, isn't it?

okuryan:
Yes, it can be done. But isn't is the same comment than for website addon that is existing and approved? There are always multiple ways of doing same things.

okuryan:
@Yajo , @pedrobaeza , read here https://support.google.com/webmasters/answer/7424835?hl=en#h04
So why not having the same addon as for website, but for web? Generally all our customers do not want their Erp to appear in Google. Many of them just thinks that it is obvious not to have it in search results, so nobody asks about this feature. So why not having addon for this?

--Quote from above link----
Does my website need a robots.txt file?

No. When Googlebot visits a website, we first ask for permission to crawl by attempting to retrieve the robots.txt file. A website without a robots.txt file, robots meta tags or X-Robots-Tag HTTP headers will generally be crawled and indexed normally.

pedrobaeza:
Thanks for the explanations. For me it's enough, but I repeat I'm not an expert in this area. I reopen meanwhile.

pedrobaeza:
About website, when you install website module, you expose immediately some public pages, so the exposure is wider.

About the DB manager and login page, I have no strong opinion nor knowledge on hiding it from crawlers, but it's probably not a bad idea.

yajo (dismissed):
In this case, the addon should use the meta tags or HTTP headers:

- It would avoid conflicts with current `website` addon, which has a `robots.txt` controller.
- It would avoid `robots.txt` hacking.

okuryan:
@Yajo , but that is quotation only form google. Other search engines may use different method (only robots, for example). 
Moreover, adding it to header will also have conflicts with website related module as 2 methods (header and robots.txt) will exist in parallel. 

Maybe just add a note here that if website is used - use another addon? 

I pretty much understand your desire that all addons across all OCA will be cross-compatible. That will be also my desire. But reality is that many addons across different repos may not be working together. Where it is found - I see note in README saying not to install those addons together. If not found -well. Our customers are finding this and we have to fix for individual customers.

So that is a question regarding the approach OCA has selected to resolve such situations. What is policy about this? I guess, @pedrobaeza ,you can also be involved in this discussion.

pedrobaeza:
If there's no possibility of making both compatible, then you should put a explicit note in the installation / Known issues sections for warning about the incompatibility.

nedaszilinskas:
Added a note to readme/INSTALL.rst

nedaszilinskas:
Also to readme/ROADMAP.rst

web_no_crawler/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2018 Xpansa Group
```
simahawk: "Copyright"

web_no_crawler/__openerp__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# © 2018 Xpansa Group
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Implement robots.txt / disallow indexing',
```
simahawk: "Disallow indexing completely via robots.txt"

web_no_crawler/controllers/main.py
```diff
@@ -0,0 +1,15 @@
...
+# © 2018 Xpansa Group
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from openerp import http
+
+
+class Main(http.Controller):
+
+    @http.route('/robots.txt', type='http', auth="none")
+    def robots_txt(self):
```
simahawk: in website v8 this method is named "robots" https://github.com/odoo/odoo/blob/8.0/addons/website/controllers/main.py#L86

web_no_crawler/controllers/main.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# © 2018 Xpansa Group
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from openerp import http
+
+
+class Main(http.Controller):
```
simahawk: you should extend the same controller class as website module: 'Home'

web_no_crawler/readme/INSTALL.rst
```diff
@@ -0,0 +1,3 @@
+Installation as usual. No specific installation steps / configuration required.
+
+NOTE: this module is not to be used with `Website` module as it has a separate functionality for robots.txt.
```
simahawk: you could emphasize this w/ a warning block instead of a note

simahawk (changes requested):
as long as the possible issue is clearly stated I don't  think is that bad :)

simahawk:
check the build

nedaszilinskas:
@simahawk I have pushed the requested changes. Please let me know if its fine.

One remark - we have this new 'readme' folder for auto-generation. Should travis not require README.rst if this 'readme' folder is present?

nedaszilinskas:
Travis failed, but looks like not because of any issues with module web_no_crawler.
Do I need to do something else to proceed with this MR?

okuryan:
@pedrobaeza @simahawk who can look at what is with Travis build and help here?

simahawk:
build is fine per sé. The issue comes from [RST linting](https://travis-ci.org/OCA/web/jobs/396403583#L610) 
```
web_action_conditionable/README.rst:55: [E7901(rst-syntax-error), ]  Unexpected possible title overline or transition.|Treating it as ordinary text because it's so short.
```

okuryan:
Ok, just surprised that 8.0 branch has issues. I believe that in 8.0 tests always should be green. So what is workflow in this situation now ? I thought that every contributor is responsible only for his module. Or should we also fix other errors ?

pedrobaeza:
It's a new check that has been added because PyPi is now stricter on the RST formatting, so better to turn to red several branches here in Travis than silently on the cron that generates the wheel. I will fix it later, but it's something that any contributor can also perform and I will fast-track it if everything is green.

okuryan:
@pedrobaeza all fine. I'm not against quality code - we also have same checks in our interrnal Gitlab CI. Just question was about workflow.
@nedaszilinskas, can you fix all pylint errors? 
I see "pylint expected errors 0, found 10!"

Not much to fix there

pedrobaeza:
Better to make the linter green in other PR, please. Thanks for your comprehension.

okuryan:
@pedrobaeza seems build is green finally =)

@pedrobaeza also @nedaszilinskas  mentioned one important note:
"One remark - we have this new 'readme' folder for auto-generation. Should travis not require README.rst if this 'readme' folder is present?"

We see that https://github.com/OCA/pylint-odoo is still requires readme file. But you say that now readme will be auto-generated from here https://github.com/OCA/maintainer-tools/tree/master/template/module/readme 

So now - we have to have both:
1) README.rst - at least empty
2) new readme folder structure

It should be done so Travis will not give error/warning about missing README.rst file (because Travis still using https://github.com/OCA/pylint-odoo

Also imagine that companies like ours want to design internally modules in a way that it will be easily to contribute to OCA with new standards. So what is procedure for us than?

What set of tools we should use internally and are there any instructions on this? Because primarily we do all changes ,of course, on our own infrastructure. So we want to keep up with updated OCA guidelines.

pedrobaeza:
> "One remark - we have this new 'readme' folder for auto-generation. Should travis not require README.rst if this 'readme' folder is present?"

Yeah, the check is still present, and there's no easy balance, as it's difficult to check if all the needed fragments are present and when they generate the README, that everything is OK.

What to do? You have 2 options:

* Put a fake README.rst for avoiding the check, and after merging this PR, the OCA bot will generate the corresponding file, overwriting the previous fake one. Advantages of this method: Easier. Disadvantages: More diff (specially if migrating and erasing full README for later regenerating it - better to let it as is -), and there's a chance that the generated README is not RST compliant (this can't be checked easily as said).
* Generate the README by your own, downloading https://github.com/OCA/maintainer-tools/, installing it with `sudo python setup.py install`, and executing `oca-gen-addon-readme  --repo-name <repo> --branch <branch> --addon-dir . --no-commit` in the addons folder for including the generated README also in your PR. Advantage: everything is done in one commit and you (and Travis) see the final result. Disadvantages: a bit of initial setup + running the command.

okuryan:
@pedrobaeza , explanation is enough. I will create internal task for my DevOps to investigate possibility of doing second case for our Internal CI. 

Now about this pull request. Is it going to be accepted? Or what are next steps?

pedrobaeza (approved):
Waiting another technical review for merging

okuryan:
@pedrobaeza , can you help me to understand - for every repository we have same technical reviewers?

simahawk (approved):

pedrobaeza:
Yes, of course. Please check https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#review

okuryan:
@pedrobaeza , thanks for merging. Last question. When usually auto-generation for readme file happens? Now I see below:

screen shot 2018-06-26 at 6 40 44 pm

pedrobaeza:
OCA bot is run during nights. Tomorrow you will see the README.

PR #969, opened 2018-06-27
[11.0][FIX] web_timeline: Make colors attribute optional

tarteo:
> * colors (optional): it allows to set certain specific colors if the expressed
>   condition (JS syntax) is met.

In the current situation the colors attribute is not optional. You'll get an error when it's not provided.

image


pedrobaeza:

pedrobaeza (approved):

pedrobaeza:
We can fast-track this as obvious.

PR #970, opened 2018-06-27
[10.0] web_notify: interactive notifications

guewen:
* You can have a simple link to refresh the active view
* Or an action (client, window, ...)

I'm annoyed that I had to inherit all the chain in javascript to add the
options argument, and started to wonder if we should not create new
Notification classes and new methods for these use cases. But in any case,
let's see your ideas and interest before going further.

* [x] adapt and add tests
* [x] new Notification classes?

Screencast of the interactive notifications

### DIFF
--- a/web_notify/README.rst
+++ b/web_notify/README.rst
@@ -50,7 +50,12 @@ The action can be used using the ``action`` keyword:
         'res_id': self.id,
         'views': [(False, 'form')],
     })
-   self.env.user.notify_info('My information message', action=action)
+   self.env.user.notify_info(
+       'My information message',
+       action=action,
+       # optional
+       action_link_name='Open Sale',
+   )
 
 
 Installation
--- a/web_notify/README.rst
+++ b/web_notify/README.rst
@@ -33,12 +33,25 @@ or
 The notifications can optionally have some action buttons.
 
 * One allowing to refresh the active view
+* Another allowing to send a window / client action
 
-It is activated when sending the notification with:
+The reload button is activated when sending the notification with:
 
 .. code-block:: python
   
    self.env.user.notify_info('My information message', show_reload=True)
+
+The action can be used using the ``action`` keyword:
+
+.. code-block:: python
+
+    action = self.env.ref('sale.action_orders').read()[0]
+    action.update({
+        'res_id': self.id,
+        'views': [(False, 'form')],
+    })
+   self.env.user.notify_info('My information message', action=action)
+
 
 Installation
 ============
--- a/web_notify/README.rst
+++ b/web_notify/README.rst
@@ -30,7 +30,7 @@ or
    :scale: 80 %
    :alt: Sample notifications
 
-The notifications can optionally have some action buttons.
+The notifications can bring interactivity with some buttons.
 
 * One allowing to refresh the active view
 * Another allowing to send a window / client action
@@ -54,7 +54,7 @@ The action can be used using the ``action`` keyword:
        'My information message',
        action=action,
        # optional
-       action_link_name='Open Sale',
+       action_link_name=_('Open Sale'),
    )
 
 
@@ -92,6 +92,7 @@ Contributors
 
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Serpent Consulting Services Pvt. Ltd.<jay.vora@serpentcs.com>
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
 
 Maintainer
 ----------
--- a/web_notify/README.rst
+++ b/web_notify/README.rst
@@ -30,6 +30,16 @@ or
    :scale: 80 %
    :alt: Sample notifications
 
+The notifications can optionally have some action buttons.
+
+* One allowing to refresh the active view
+
+It is activated when sending the notification with:
+
+.. code-block:: python
+  
+   self.env.user.notify_info('My information message', show_reload=True)
+
 Installation
 ============
 
--- a/web_notify/__manifest__.py
+++ b/web_notify/__manifest__.py
@@ -18,6 +18,9 @@
     'data': [
         'views/web_notify.xml'
     ],
+    'qweb': [
+        'static/src/xml/*.xml',
+    ],
     'demo': [
     ],
     'installable': True,
--- a/web_notify/static/src/js/notification.js
+++ b/web_notify/static/src/js/notification.js
@@ -1,0 +1,38 @@
+odoo.define('web_notify.notification', function (require) {
+    "use strict";
+
+    var base_notification = require('web.notification'),
+        WebClient = require('web.WebClient'),
+        Notification = base_notification.Notification,
+        Warning = base_notification.Warning;
+
+    Notification.include({
+        events: _.extend(
+            {},
+            Notification.prototype.events,
+            {'click .o_view_reload': function(e){
+                e.preventDefault();
+                this.reload_active_view();
+            }
+            }
+        ),
+        init: function(parent, title, text, sticky, options) {
+            this._super.apply(this, arguments);
+            this.options = options || {};
+        },
+        reload_active_view: function() {
+            this.trigger_up('reload_active_view');
+        },
+    });
+
+    base_notification.NotificationManager.include({
+        notify: function(title, text, sticky, options) {
+            return this.display(new Notification(this, title, text, sticky, options));
+        },
+        warn: function(title, text, sticky, options) {
+            return this.display(new Warning(this, title, text, sticky, options));
+        },
+
+    });
+
+});
--- a/web_notify/static/src/js/notification.js
+++ b/web_notify/static/src/js/notification.js
@@ -6,7 +6,8 @@ odoo.define('web_notify.notification', function (require) {
         Notification = base_notification.Notification,
         Warning = base_notification.Warning;
 
-    Notification.include({
+    var InteractiveNotification = Notification.extend({
+        template: 'InteractiveNotification',
         events: _.extend(
             {},
             Notification.prototype.events,
@@ -21,7 +22,7 @@ odoo.define('web_notify.notification', function (require) {
             }
         ),
         init: function(parent, title, text, sticky, options) {
-            this._super.apply(this, arguments);
+            this._super.apply(this, [parent, title, text, sticky]);
             this.options = options || {};
         },
         reload_active_view: function() {
@@ -33,14 +34,23 @@ odoo.define('web_notify.notification', function (require) {
         }
     });
 
+    var InteractiveWarning = InteractiveNotification.extend({
+        template: 'InteractiveWarning',
+    });
+
     base_notification.NotificationManager.include({
-        notify: function(title, text, sticky, options) {
-            return this.display(new Notification(this, title, text, sticky, options));
+        interactive_notify(title, text, sticky, options) {
+            return this.display(new InteractiveNotification(this, title, text, sticky, options));
         },
-        warn: function(title, text, sticky, options) {
-            return this.display(new Warning(this, title, text, sticky, options));
-        },
+        interactive_warn(title, text, sticky, options) {
+            return this.display(new InteractiveWarning(this, title, text, sticky, options));
+        }
 
     });
+
+    return {
+        InteractiveNotification: InteractiveNotification,
+        InteractiveWarning: InteractiveWarning
+    };
 
 });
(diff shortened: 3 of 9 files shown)


lmignon (commented):
@guewen A great and awaited improvement! I share your concern about javascript and it could be safer to start to develop our own JS notification class if we modify the original API. (Nice to Have)

yvaucher:
Nice addition!

For the new js class, it is a good idea. IMO, the changeset is not that big, though. It would depends on what we have on the roadmap.

guewen:
I'll give a try to using a new class

guewen:
Added 2 commits using dedicated classes and functions for interactive notifications and adding variadic parameters (opening the options of the notifications for other uses cases).

guewen:
Not sure how the error is related:

```
2018-06-28 06:00:34,371 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: FAIL
2018-06-28 06:00:34,375 4376 INFO openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: ======================================================================
2018-06-28 06:00:34,375 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: FAIL: test_ui_web (odoo.addons.web_widget_slickroom.tests.test_ui.UICase)
2018-06-28 06:00:34,375 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: ` Test backend tests.
2018-06-28 06:00:34,375 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: Traceback (most recent call last):
2018-06-28 06:00:34,375 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: `   File "/home/travis/build/OCA/web/web_widget_slickroom/tests/test_ui.py", line 14, in test_ui_web
2018-06-28 06:00:34,376 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: `     login="admin",
2018-06-28 06:00:34,376 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: `   File "/home/travis/odoo-10.0/odoo/tests/common.py", line 445, in phantom_js
2018-06-28 06:00:34,376 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: `     self.phantom_run(cmd, timeout)
2018-06-28 06:00:34,376 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: `   File "/home/travis/odoo-10.0/odoo/tests/common.py", line 391, in phantom_run
2018-06-28 06:00:34,376 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: `     "PhantomJS test completed without reporting success; "
2018-06-28 06:00:34,376 4376 ERROR openerp_test odoo.addons.web_widget_slickroom.tests.test_ui: ` AssertionError: None is not true : PhantomJS test completed without reporting success; the log may contain errors or hints.
```

lmignon:
@guewen I find that using a dedicated JS class improves the readability of the code and therefore its maintenance. This opens the way to other extensions. Great work! 
For the errors, difficult to know what is happening. I admit I don't like or feel comfortable with these JS tests.

web_notify/static/src/xml/notification.xml
```diff
@@ -0,0 +1,38 @@
...
+        <a href="#" class="o_notification_reload_view">
+            <span class="fa fa-refresh"/> Reload current view
+        </a>
+    </div>
+</t>
+
+<t t-name="Notification.do_action">
+    <div class="o_notification_action" t-if="widget.options.action">
+        <a href="#" class="o_notification_do_action">
+            <span t-att-class="'fa ' + (widget.options.action_fa_icon ? widget.options.action_fa_icon : 'fa-arrow-circle-left')"/>
```
guewen, line 15: Are the `t-att-class` escaped? If not, a jsonrpc call could allow to inject something in widget.options.action_fa_icon that would directly open a notification to a user with a crafted payload, creepy.

simahawk: [`t-att`](https://github.com/odoo/odoo/blob/39f0f5db77f793c7fb996d53195524de0e035de4/odoo/addons/base/ir/ir_qweb/qweb.py#L886
) uses  [`_compile_expr`](https://github.com/odoo/odoo/blob/10.0/odoo/addons/base/ir/ir_qweb/ir_qweb.py#L293).
Seems to be safe on eval but not sure it is for escape.
Not sure how that could be dangerous on a class tho...

In any case that would come via this api so why don't you escape it in _notify_channel if present?

web_notify/README.rst
```diff
@@ -30,6 +30,34 @@ or
    :scale: 80 %
```
simahawk, line 30: switch to new readme structure? :stuck_out_tongue_winking_eye:

simahawk (approved):
LGTM

simahawk:
possible improvement: handle a list of actions instead of just one?

guewen:
> possible improvement: handle a list of actions instead of just one?

thought about that too... make sense, not sure how to keep an easy API though

guewen:
> possible improvement: handle a list of actions instead of just one?
> 
> thought about that too... make sense, not sure how to keep an easy API though

Option 1, seems a bit verbose, maybe?:
```python
self.notify_info(
    'hello world',
    actions=[
        {'action': {'type': ..., ... ,...}, 'link_name': _('Click on me!'), ...}],
        {'action': {'type': ..., ... ,...}, 'link_name': _('Do not click on me!'), ...}],
    ]
)
```
Option 2, keep the single action for 95% of the time and add a kind of builder pattern for more advanced use cases:
```python
self.notify_info(
    'hello world',
    action={'type': ..., ... ,...},
    action_link_name': _('Click on me!'),
)

# builder
builder = self._build_notify_info()
builder.add_action({'type': ..., ... ,...}, link_name=_('Click on me!'))
builder.add_action({'type': ..., ... ,...}, link_name=_('Do not click on me!'))
builder.send('hello')
```

What I like about the option 2 is: 
* we can delay the implementation of the "multiple actions" case until we really need it
* seems cleaner if you need to factorize a pre-setup notification, you can return a prepared builder from a method (well you could return a list of actions for option 1 as well of course, but I like it less ;))

simahawk:
To me opt 2 adds unnecessary complexity for such easy feature :smile: 
I mean the goal is: show more than one action.
Devs, integrators will have to **just** provide a list of actions.
We can even support backward compat by testing if given arg is a list or a dict...

simahawk:
also, can't we use action's name for link name? If not, can't we include it in the action dict?

hbrunn (approved):
it's enough to have the action, modules like https://github.com/OCA/web/tree/10.0/web_ir_actions_act_window_message add extra flexibility already, no need to cook up your own

web_notify/static/src/js/notification.js
```diff
@@ -0,0 +1,55 @@
+odoo.define('web_notify.notification', function (require) {
```
elicoidal, line 1: add copyright

web_notify/static/src/xml/notification.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elicoidal, line 1: add copyright

web_notify/tests/test_res_users.py
```diff
@@ -17,12 +18,18 @@ def test_notify_info(self):
...
-            message='message', title='title', sticky=True)
+            message='message', title='title', sticky=True,
+            show_reload=True, foo="bar"
+        )
         news = bus_bus.search(domain) - existing
         self.assertEqual(1, len(news))
-        self.assertEqual(
-            '{"message":"message","sticky":true,"title":"title"}',
-            news.message)
+        expected = ('{"message":"message","sticky":true,"title":"title",'
```
elicoidal, line 26: one attribute per line makes it easier to read

web_notify/tests/test_res_users.py
```diff
@@ -32,12 +39,46 @@ def test_notify_warning(self):
...
```
elicoidal, line 47: here too

web_notify/tests/test_res_users.py
```diff
@@ -32,12 +39,46 @@ def test_notify_warning(self):
...
+            action={
+                "type": "ir.actions.act_window",
+                "view_mode": "form",
+            },
+            action_link_name="Open"
+        )
+        news = bus_bus.search(domain) - existing
+        self.assertEqual(1, len(news))
+        # the action should be transformed by Odoo (clean_action)
+        expected = ('{"message":"message","sticky":true,"title":"title",'
```
elicoidal, line 73: and here

web_notify/__manifest__.py
```diff
@@ -18,6 +18,9 @@
     'data': [
         'views/web_notify.xml'
     ],
+    'qweb': [
+        'static/src/xml/*.xml',
+    ],
     'demo': [
```
elicoidal, line 24: remove empty keys

elicoidal (approved):
Do we need to bump version?

guewen:
> no need to cook up your own

Not sure what you are referring to?

hbrunn:
this was about supporting more than just some call to do_action - it would be reinventing the wheel to add anything specific here supporting multiple actions or the like, because we already have modules that take care of this

guewen:
> this was about supporting more than just some call to do_action - it would be reinventing the wheel to add anything specific here supporting multiple actions or the like, because we already have modules that take care of this

Got it, so the action in the notification can call one of these "multi-action" which would allow the user to choose one of the action. Well, it could feel a little bit more integrated and spare 1 click with a list of links in the notification, but at some expense of complexity, so I'm fine with your approach.

simahawk:
sorry guys but I don't get how looping on a list is going to be too heavy w/ this piece of code :smile: 
To me, using that module to achieve this simple result IS what adds complexiness in this ctx.

simahawk:
forgot: nevertheless, we can go ahead w/ current solution and improve later. Not blocking at all :wink:

guewen:
@simahawk I hate complex arguments such as `[{'action': {'type': ..., 'view_type': .....}, 'link_name': 'Open'}, {'action': {'type': ..., 'view_type': .....}, 'link_name': 'Edit'},]`. They are error-prone and hard to document. This is not complex per-se but push complexity towards the API user.

guewen:
> also, can't we use action's name for link name? If not, can't we include it in the action dict?

You'll probably want something different, as the link will probably often be a verb indicating an action "Open, View, List, ...", which will be a poor naming in the breadcrumb. Not fund of putting it in the action dict, I see it as a mix of concerns.

guewen:
In any case we all agree we can move forward with a single action at least for now :)

web_notify/tests/test_res_users.py
yvaucher: this is a dict in a string in json format for test purpose, I think it's okay this way as it is an expected output.

yvaucher (approved):
LGTM but for the missing copyright mentions in js file.

I'll add it on merge.

yvaucher:
(tests are failing on module `web_widget_slick` which seems unrelated)

yvaucher:
Added missing copyright on js file plus on res_user.py, removed key 'demo' from manifest, bumped version, squashed, rebased and merged.

merged at b4ba0eeb4a8c857eb9b488ddaf80a2af9929ccba

alexis-via:
This feature is really great... would love to use it in OCA/connector-telephony. But unfortunately, it is not available for the moment in v11 nor v12 version of web_notify.

pedrobaeza:
@alexis-via they are available already and we are using it in our PR:

https://github.com/OCA/web/tree/11.0/web_notify
https://github.com/OCA/web/tree/12.0/web_notify

alexis-via:
@pedrobaeza You're answering too fast :) As I said, this FEATURE is not available in v11 nor v12. I know that the module is available.

I tried to up-port the changes, but there are so many changes that it's not easy... so I switched to another strategy and took the current v10 version and tried to port it again to v11 by getting inspiration from the v11 migration commit https://github.com/OCA/web/pull/781/commits/63c431c5fd41e95704bde133d1644fbf69f4e13f but I didn't manage to make the JS code work on v11... I guess my JS knowledge is too limited for such a work.

If some more technical guys are ready to do it, you would be more than welcomed. In the meantime, i'll probably start to use it in v10 for OCA/connector-telephony/base_phone_popup... that would be fun!

alexis-via:
This feature is so cool that I couldn't resist porting base_phone_popup to v10 using that new feature of web_notify. We need it in v11 and v12 now :)

pedrobaeza:
Oh, got it, it's a missing feature in the migration. That's why I always asked you for forward porting changes done in lower versions ;)

PR #971, opened 2018-06-27
[IMP] web_drop_target: Upload multiple files

fuentes73:
Added the possibility to upload multiple files and improve the UX

captura de pantalla 2018-06-27 a las 16 07 25


web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -27,44 +28,67 @@ odoo.define('web_drop_target', function(require) {
...
         _on_dragenter: function(e) {
-            if(this._get_drop_item(e)) {
+            if(this._get_drop_item(e).length) {
                 e.preventDefault();
-                jQuery(e.delegateTarget).addClass(this._drag_over_class);
+                if(!this._drop_overlay){
+                    var drop_overlay_message = _t('Drop your files here');
+                    var o_content = jQuery('.o_content');
+                    var view_manager = jQuery('.o_view_manager_content');
+                    this._drop_overlay = jQuery(
```
simahawk: move this templating to an xml template or split it to separated func to ease override

simahawk: also, use `self.$(...)` to find elements inside the form only

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -27,44 +28,67 @@ odoo.define('web_drop_target', function(require) {
...
         },
 
         _on_dragleave: function(e) {
-            jQuery(e.delegateTarget).removeClass(this._drag_over_class);
+            this._drop_overlay.remove();
+            this._drop_overlay = null;
+            e.preventDefault();
         },
 
         _get_drop_item: function(e) {
```
simahawk: _get_drop_items then (+s)?

simahawk (changes requested):

simahawk:
@hbrunn ping ;)

web_drop_target/static/src/js/web_drop_target.js
fuentes73: I can't change `jQuery(...)` for `self.$` because those classes are outside the Widget, so it wouldn't find anything.
Am I wrong?

web_drop_target/__manifest__.py
```diff
@@ -3,14 +3,15 @@
...
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.1.0",
     "author": "Therp BV,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
     "license": "AGPL-3",
     "category": "Usability",
     "summary": "Allows to drag files into Odoo",
     "depends": [
         'web',
+        'document'
```
hbrunn, line 14: do we really need this?

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -5,7 +5,9 @@
 
 odoo.define('web_drop_target', function(require) {
     var Model = require('web.Model'),
-        FormView = require('web.FormView');
+        FormView = require('web.FormView'),
+        core = require('web.core');
+    var _t = core._t;
```
hbrunn: this you can just append above too

web_drop_target/static/src/js/web_drop_target.js
hbrunn: yes you are. This is a mixin class, so when this is actually instantiated, it does live within a widget: https://github.com/Studio73/web/blob/c33e6fa6745bb6fc59d2ae0f0e681c141e9bbb98/web_drop_target/static/src/js/web_drop_target.js#L138

web_drop_target/static/src/js/web_drop_target.js
hbrunn: yes please

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -27,44 +28,67 @@ odoo.define('web_drop_target', function(require) {
...
-            reader.onloadend = this.proxy(
-                _.partial(this._handle_file_drop, drop_item.getAsFile())
-            );
-            reader.readAsArrayBuffer(drop_item.getAsFile());
+            var self = this;
+            var drop_items = this._get_drop_item(e);
+            _.each(drop_items, function(drop_item) {
+                var reader = new FileReader();
+                reader.onloadend = self.proxy(
+                    _.partial(self._handle_file_drop, drop_item.getAsFile())
```
hbrunn: this is the simple way to implement this, but is it also the best way? That's a serious question, handling files separately makes sense I think, but let's think if there are cases where it makes a difference on the client side if we handle one file or 20? Can't think of one currently, but I wanted to raise the point anyways

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -27,44 +28,67 @@ odoo.define('web_drop_target', function(require) {
...
+                    o_content.append(this._drop_overlay);
+                }
                 return false;
             }
         },
 
         _on_dragleave: function(e) {
-            jQuery(e.delegateTarget).removeClass(this._drag_over_class);
+            this._drop_overlay.remove();
+            this._drop_overlay = null;
```
hbrunn: let's refactor 'do whatever we need to create/remove the d&d ui' into their own functions for simpler adaptions

hbrunn (dismissed):
thanks!

web_drop_target/__manifest__.py
fuentes73: I think yes, documents module adds the attachment button on top of the forms.

hbrunn: that's a functional thing. Is it *technically* necessary?

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Yes I know that is a mixin class and it does live within a widget. But, what I'm saying is that those classes are in other Widgets, o_content is in ActionManager and o_view_manager_content is in ViewManager. So if you use `this.$('.o_content')` the result it's empty (I've tested it).

web_drop_target/__manifest__.py
simahawk: not really AFAIS but it sucks when you add an attachment and you don't see it...
If you don't want the dependency,  maybe we can add a reminder in  the drop area saying that you won't see them unless...?

web_drop_target/static/src/js/web_drop_target.js
hbrunn: ah, this way. Well, then you break generality here. It would be better if the addon only blocks/covers the area of the widget (that's also what the drop target is)

web_drop_target/__manifest__.py
hbrunn: that's worse in my opinion. I think technically unnecessary dependencies are a bad habit because you never know what people out there do with your code, and then pulling a dependency that might or might not cause trouble can be annoying. But if you both insist, I won't block on this

fuentes73: Well, I don't think that this dependency is unnecesary, if you don't install this module you can't see the attachments. If you want I can delete it, no problem, but we must add in the readme a message warning of this.

hbrunn: just leave it as it is, there's more interesting stuff to do and to discuss here

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -6,14 +6,14 @@
 odoo.define('web_drop_target', function(require) {
     var Model = require('web.Model'),
         FormView = require('web.FormView'),
-        core = require('web.core');
-    var _t = core._t;
+        core = require('web.core'),
+        qweb = core.qweb;
```
hbrunn: sorry for this one, the linter disagrees with me: https://eslint.org/docs/rules/no-mixed-requires (TIL)

web_drop_target/static/src/js/web_drop_target.js
fuentes73: I've done this because with `position: absolute` when the user scrolls the message stays on top and this is the result:
3-absolute

One solution to this is to use `position: fixed` but, with the fixed property the element is positioned relative to the browser window
2-fixed

So if we want to cover just the area where the drag&drop works, a possible solution is something like I've done.

AFAIK ActionManager is a property of ViewManager. ViewManager is the parent view of View Widget and FormView, TreeView and KanbanView extends from View. So those css classes are always accesibles and no generality break.

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Done

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Done

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Done

web_drop_target/static/src/js/web_drop_target.js
hbrunn: concerning generality, think that tomorrow someone wants to use the mixin for enabling some specific action when dropping on some button, div, whatever on a form, but only there. Or a widget in the top bar. Or why not support drag and drop on some menu items that just create some default record based on the dropped file? That's where you break generality.

What's the problem with finding the current widget's DOM element's offset parent, and construct an overlay that covers only that?

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Yes, you're completely right. I had tested it with 3/4 pdf and worked perfectly but TBH, now I don't know a better way to do it 😅

hbrunn: let's leave this as future extension then

web_drop_target/static/src/js/web_drop_target.js
fuentes73: I've also try to do it in this way but FormView parent `.o_view_manager_content` it's positioned absolute, so `position()` returns `0` and `offset()` changes with the scroll and will became negative at the bottom of the page.  

I'm not opposing to change it, but this is the only way that I found that works.
Honestly, I'm not going to waste more time trying to explain that I already test other solutions. This cover the cases proposed at the README, Form, Tree and Kanban. If someone comes with the idea to support drag&drop in buttons or menus he will be in charge to make all compatible, not me.

web_drop_target/__manifest__.py
```diff
@@ -3,16 +3,20 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     "name": "Drop target support",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.1.0",
```
elicoidal, line 6: A question/detail: shoudnot this be 10.0.1.0.1 (do we really need to update the module?)

elicoidal (commented):
Functional Test: not sure this is linked but I cannot delete the attachments after uploading them.
Travis is complaining.
Other than that this is a great improvement!

web_drop_target/__manifest__.py
simahawk: it's an improvement :wink:

simahawk (approved):

simahawk:
@fuentes010 would you mind adding the screenshot to the README?
While you are doing this you could also cherry-pick this commit https://github.com/OCA/web/pull/974/commits/cc5cb0e5308f2c757546df0c88e1ca59a72f0b9d

simahawk:
Travis is failing here
```bash
`   File "/home/travis/build/OCA/web/web_widget_slick/tests/test_ui.py", line 18, in test_ui_web
`     login="admin",
`   File "/home/travis/odoo-10.0/odoo/tests/common.py", line 445, in phantom_js
`     self.phantom_run(cmd, timeout)
`   File "/home/travis/odoo-10.0/odoo/tests/common.py", line 391, in phantom_run
`     "PhantomJS test completed without reporting success; "
` AssertionError: None is not true : PhantomJS test completed without reporting success; the log may contain errors or hints.
```
Not sure if is related to these changes. I triggered re-build, let's see.

fuentes73:
@simahawk sure, no problem!

fuentes73:
@simahawk Done, thank's for the review!
Travis keeps complaining 😰

simahawk:
@fuentes010 try running test via /web/tests. We might get a better outcome.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -27,44 +28,45 @@ odoo.define('web_drop_target', function(require) {
...
-            }
-            jQuery(e.delegateTarget).removeClass(this._drag_over_class);
-            var reader = new FileReader();
-            reader.onloadend = this.proxy(
-                _.partial(this._handle_file_drop, drop_item.getAsFile())
-            );
-            reader.readAsArrayBuffer(drop_item.getAsFile());
+            var self = this;
+            var drop_items = this._get_drop_items(e);
+            _.each(drop_items, function(drop_item) {
```
grindtildeath: Cannot this loop be moved to _handle_file_drop (which could then be renamed _handle_drop_items or something similar) ?
The goal is to be able to process multiple file differently than creating an attachement, i.e. having one rpc call for all the items instead of one rpc call for each dropped item.
I'm actually trying to do something like that for v11 but am struggling with it.

grindtildeath: I could finally get to it, cf : https://github.com/OCA/web/pull/975/commits/87be514256265bca13a55e285233043213acae9b
Maybe there's a cleaner solution ?

simahawk: @hbrunn seems good to me, can you check it?

hbrunn: the contributor made it clear [not to want to waste time on my review](https://github.com/OCA/web/pull/971#discussion_r199557546), so I'm done with this PR

web_drop_target/static/src/js/web_drop_target.js
simahawk: as far as I see this is not breaking anything, this is ADDING a feature and JUST limiting the generalization. I wouldn't block the change just for this,

That been said, @fuentes010 I know that it can be frustrating to address changes BUT pls don't say that you are wasting time.

Explain what you have done instead (so that ppl don't have to guess what you have tried and what not) and put this clear in the roadmap/known issues session possibly including the **valueable** remark by @hbrunn .

In this way we have a trace and whoever will work again on this widget can address this issue if needed.

Guys pls, remember that we are all on the same side: trying to improve shared code. :wink:

web_drop_target/static/src/js/web_drop_target.js
simahawk: @hbrunn take a breath and see my comment there :)
In any case I'm not asking you to review again this PR but a commit on another one that can be pulled here. Thanks in advance.

web_drop_target/__manifest__.py
simahawk: @elicoidal regarding your issue, is it due to the fact that you don't see the classic dropdown?
Do you have "document" installed?

web_drop_target/static/src/js/web_drop_target.js
fuentes73: @simahawk @hbrunn first of all, thank your for your review.
I didn't mean that I'm against your review and I said cleary that i'm not opposing to change anything, all improvements are welcome, I've change all you asked for. For me address changes it's not frustrating I love to learn. I didn't try to offend to anyone with my comment, if it has been misunderstood I apologise. 

@simahawk I'll change the word waste for invest 😉

elicoidal:
@simahawk I still have the issue: I can upload but cannot delete the attachments (blocking).
Actually it is a bit disturbing not to have by default the menu attachment (while the module is installed) even if the feature for dropping works fine. I would keep the attachment menu (but not blocking for that)

screencast_2018-07-11_04 49 12 pm

fuentes73:
Hi @elicoidal,
I have reproduced your issue, trash icon doesn't appears after dropping a file in to a ir.module.module model, like in your gif. But, If you drop a file into other model like sale.order, res.partner, etc... The trash button does appear. So it might not be related to this module, maybe it's due to permissions or something else, I don't know...

@simahawk @hbrunn @grindtildeath  Any thoughts?

pedrobaeza:
Attachments get the same permissions as their "container" model, so probably you won't have permissions to unlink modules. This is not something related to this module.

elicoidal (approved):
LGTM

web_drop_target/static/src/js/web_drop_target.js
simahawk: pls, document this issue in the roadmap/known issues and add a comment in the code pointing to that item (a numbered list would be awesome :wink: )

web_drop_target/static/src/js/web_drop_target.js
simahawk: @fuentes010 would you mind to cherry-pick that commit?

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -106,7 +108,30 @@ odoo.define('web_drop_target', function(require) {
                     );
                 }
             });
-        }
+        },
+
+        _add_overlay() {
```
yajo: ES6 syntax is forbidden if you don't want to turn developers crazy, as you can see with @simahawk  in https://github.com/OCA/web/issues/980#issuecomment-405177950 :stuck_out_tongue_winking_eye: 

PhantomJS doesn't support it, and minified sources with broken syntax kill the whole JS silently.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -106,7 +108,30 @@ odoo.define('web_drop_target', function(require) {
...
+                    'top': o_content_position.top, 
+                    'left': o_content_position.left,
+                    'width': view_manager.width(),
+                    'height': view_manager.height()
+                });
+                o_content.append(this._drop_overlay);
+            }
+        },
+        
+        _remove_overlay() {
```
yajo: ES6 syntax forbidden.

yajo (changes requested):

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Upss... sorry! I didn't know about that. 😅

web_drop_target/static/src/js/web_drop_target.js
fuentes73: @grindtildeath I cherry-picked your commit and modified it to make it works at v10. But now the stored files are empty. 
Can you please, check if it's also happening at your PR? or it's just me that I've done something wrong

fuentes73: No need to check anything. I didn't see this commit [db9e332](https://github.com/OCA/web/pull/975/commits/db9e332f3b2fcd7a566a9075d5ffda0de5566302). Applied and now works properly

simahawk:
@Yajo @grindtildeath  can you update your review? :)

yajo (approved):
Code review.

grindtildeath:
@fuentes010 I did add this commit in v11 to be IE11 compliant : https://github.com/OCA/web/pull/975/commits/37202bc4b42c8ba560469c75b57c7389f7ef907f
IMO should be tested for v10. (I used this VM https://github.com/magnetikonline/linux-microsoft-ie-virtual-machines#ie11---windows-7)

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -106,7 +102,42 @@ odoo.define('web_drop_target', function(require) {
...
+
+        _handle_file_drop_attach: function(
+            item, e, res_model, res_id, extra_data
+        ) {
+            var self = this;
+            var file = item.getAsFile();
+            var reader = new FileReader();
+            reader.onloadend = self.proxy(
+                _.partial(self._create_attachment, file, reader, e, res_model, res_id, extra_data)
+            );
```
MeetKD, line 120: Can you please add FileReader error handler ?

MeetKD (changes requested):

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -27,67 +28,62 @@ odoo.define('web_drop_target', function(require) {
         },
 
         _on_drop: function(e) {
-            var drop_item = this._get_drop_item(e);
-            if(!drop_item) {
+            var drop_items = this._get_drop_items(e);
+            if(!drop_items) {
```
simahawk, line 34: pls rebase and include this fix https://github.com/OCA/web/pull/1209

web_drop_target/static/src/js/web_drop_target.js
simahawk: @fuentes010 ping

simahawk (commented):
2 comments to address then it seems good to merge, fine for every one?

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Done

web_drop_target/static/src/js/web_drop_target.js
fuentes73: Done

web_drop_target/static/src/js/web_drop_target.js
simahawk: @fuentes010 I think @MeetKD was asking for an overrideable error handler (as `self._create_attachment`).
Once this is done I think we are ready to merge.

fuentes73: @MeetKD @simahawk done.

fuentes73:
@simahawk @MeetKD changes done, do you think it's ready to merge?

Issue #540, opened 2017-01-27
[9.0] web_responsive: Some bugs + improvements


pedrobaeza:
The mobile view is still not perfect, as you can see in this screenshot:

![seleccion_026](https://cloud.githubusercontent.com/assets/7165771/22390038/0946f07c-e4e9-11e6-8def-291e41f97771.png)

Other things that can be improved:

* When clicking on a menu, hide all the rolled-up menus to the right button, as you don't want to click there again or scroll down for seeing the part you have click on.
* Hide search view by default under a glass icon or similar, and when clicked, show it. You can click again in the glass icon for hiding.
* Full statusbar widget is very big to be shown as is.

![seleccion_027](https://cloud.githubusercontent.com/assets/7165771/22390274/5955a1d4-e4ea-11e6-8a5f-1ca700d0a040.png)

  It's better to show only active state and click on it for seeing the rest of the states (and allow to click for changing state if clickable).
* The same happens with the buttons in the header. We should group them under a button called "Actions" or similar (web_enterprise called it "Action" in singular, but I prefer in plural if we choose that name).

yajo:
Can you point to how to get to screenshot 1?

pedrobaeza:
In Firefox, go to mobile view, and see a sales order in form view.

yajo:
I checked it; confirmed.

yajo:
I can confirm that problem 1 does not happen in v10.

pedrobaeza:
@lasley would you take care of the rest of the comments here?

pedrobaeza:
Closing as triage and doesn't happen in v10.

PR #973, opened 2018-06-30
[IMP][11.0][web_responsive] Sticky Statusbar & List View Header-Footer

Tardo:
sticky_statusbar

A) With Sticky Statusbar (PR)
B) Without Sticky Statusbar (Normal)

sticky_list

### DIFF
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -52,6 +52,24 @@
     }
 }
 
+// Sticky Header & Footer in List View
+.o_view_manager_content {
+    >div {
+        >.table-responsive {
+            >.o_list_view {
+                thead {
+                    position: sticky;
+                    top: 0;
+                }
+                tfoot {
+                    position: sticky;
+                    bottom: 0;
+                }
+            }
+        }
+    }
+}
+
 .o_form_view {
     // Form must fill 100% width in any size
     .o_form_sheet_bg {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Another great option! 

Maybe you can check https://github.com/OCA/web/issues/540 for other suggestions I made some time ago.

yajo (approved):
Neat! :heart:

web_responsive/static/src/less/form_view.less
```diff
@@ -52,6 +52,24 @@
     }
 }
 
+// Sticky Header & Footer in List View
+.o_view_manager_content {
+    >div {
+        >.table-responsive {
+            >.o_list_view {
+                thead {
+                    position: sticky;
```
simahawk, line 61: sounds this is not going to work for everybody https://caniuse.com/#search=sticky :)

yajo: Usually CSS would ignore unknown rules and just leave the defaults, so that shouldn't be a problem in most cases.

simahawk: sure, maybe worth to add it to known issue / roadmap?

simahawk (approved):

web_responsive/static/src/less/form_view.less
yajo: Well, it's not *known* until you open a old browser and check it's broken.
Let's just expect it works, and if it breaks somebody, and that somebody cannot stop using IE, we'll fix it then. :relaxed:

guewen (approved):
Great!

PR #974, opened 2018-07-02
[11.0][MIG] web_drop_target

grindtildeath:
TODO :
 - [ ] ~Include https://github.com/OCA/web/pull/971~ (To be done in #975 )

PR #975, opened 2018-07-02
[11.0][ADD] web_drop_target : add multi upload

grindtildeath:
Base PR : 
 - [x] https://github.com/OCA/web/pull/974

Depends on :
 - [ ] https://github.com/OCA/web/pull/971


web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -25,44 +25,67 @@ odoo.define('web_drop_target', function(require) {
...
-            }
-            jQuery(e.delegateTarget).removeClass(this._drag_over_class);
-            var reader = new FileReader();
-            reader.onloadend = this.proxy(
-                _.partial(this._handle_file_drop, drop_item.getAsFile())
-            );
-            reader.readAsArrayBuffer(drop_item.getAsFile());
+            var self = this;
+            var drop_items = this._get_drop_items(e);
+            _.each(drop_items, function(drop_item) {
```
grindtildeath: Cannot this loop be moved to _handle_file_drop (which could then be renamed _handle_drop_items or something similar) ?
The goal is to be able to process multiple file differently than creating an attachement, i.e. having one rpc call for all the items instead of one rpc call for each dropped item.
I'm actually trying to do something like that for v11 but am struggling with it.

ghost (approved):

simahawk:
@grindtildeath can you rebase on 11.0?

grindtildeath:
Superseded by #1313

Issue #976, opened 2018-07-03
[9.0]web_ir_actions_act_window_message: Uncaught TypeError: self.inner_widget.active_view.controller.recursive_reload is not a function


zakiuu:
**Use case:**
- I've a server action declared for my model:
```XML
<record id="my_model_action" model="ir.actions.server">
        <field name="name">My action</field>
        <field name="model_id" ref="model_my_model"/>
        <field name="code"><![CDATA[
action = self.my_action(cr, uid, context['active_ids'],context=context)
]]></field>
    </record>
```
- Of corse we need an `ir.value` record so the action will be added in the `actions` menu
```XML
<record model="ir.values" id="product_template_activate">
        <field name="name">My action</field>
        <field name="key2">client_action_multi</field>
        <field name="value" eval="'ir.actions.server,%d'%my_model_action"/>
        <field name="model">my.model</field>
    </record>
```
- Then inside my model, I use `my_action` method to display the window message:
```Python
    @api.multi
    def my_action(self):
        msg = _("my action message")
        return {
            'type': 'ir.actions.act_window.message',
            'title': _(u"My action title"),
            'message': msg,
            'buttons': [
                {
                    'type': 'method',
                    'name': _('My other action'),
                    'model': self._name,
                    'method': 'my_other_action',
                    'args': [self.ids],
                    'kwargs': {'context': {'active_model': self._name}},
                },
            ],
        }
```
- Then the user will select some records on the list view of my mode, then `actions` menu, then `my action`.
- the message window is well displayed, but if the user clicks on `close` button or `My other action` button, the action is executed but then I've this frontend message error:
```
Uncaught TypeError: self.inner_widget.active_view.controller.recursive_reload is not a function
http://localhost:8069/web/content/20011-fe1ba17/web.assets_backend.js:9227
Retraçage :
TypeError: self.inner_widget.active_view.controller.recursive_reload is not a function
at Object.<anonymous> (http://localhost:8069/web/content/20011-fe1ba17/web.assets_backend.js:9227:42)
at Object.<anonymous> (http://localhost:8069/web/content/19746-3b2cca7/web.assets_common.js:650:681)
at fire (http://localhost:8069/web/content/19746-3b2cca7/web.assets_common.js:644:299)
at Object.fireWith [as resolveWith] (http://localhost:8069/web/content/19746-3b2cca7/web.assets_common.js:649:198)
[...]

PR #852, opened 2018-01-31
[10.0][ADD] web_tab_focus

andreparames:
This is a proposal for a module that allows a page/tab of a notebook to be automatically focused depending on the value of a given field of the record. I suggest reading the [README](https://github.com/acsone/web/blob/85e08a7e431544ace5551eb2e80ac2123ea8e2f7/web_tab_focus/README.rst). Suggestions regarding the syntax are welcome!

PR #977, opened 2018-07-03
[9.0][REF]web_ir_actions_act_window_message: fix refresh view

zakiuu:
Take into consideration the view type when refreshing it,
the `form` view uses `recursive_reload`, and list view uses
`reload_content`

### DIFF
--- a/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
+++ b/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
@@ -25,8 +25,7 @@ odoo.define('web.web_ir_actions_act_window_message', function(require)
                     text: action.close_button_title || _t('Close'),
                     click: function() {
                         // refresh the view before closing the dialog
-                        self.inner_widget.active_view
-                            .controller.recursive_reload();
+                        self.ir_actions_act_window_message_refresh();
                         dialog.close()
                     },
                     classes: 'btn-default',
@@ -81,8 +80,7 @@ odoo.define('web.web_ir_actions_act_window_message', function(require)
                                 }
                                 // always refresh the view after the action
                                 // ex: action updates a status
-                                self.inner_widget.active_view
-                                .controller.recursive_reload();
+                                self.ir_actions_act_window_message_refresh();
                             });
                         }
                         else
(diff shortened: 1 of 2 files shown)


zakiuu:
closes https://github.com/OCA/web/issues/976

web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
```diff
@@ -25,8 +25,19 @@ odoo.define('web.web_ir_actions_act_window_message', function(require)
                     text: action.close_button_title || _t('Close'),
                     click: function() {
                         // refresh the view before closing the dialog
-                        self.inner_widget.active_view
-                            .controller.recursive_reload();
+                        var controller = self.inner_widget.active_view.controller;
```
hbrunn: that's what I meant with refactoring: Move this to its own function, something like `ir_actions_act_window_message_refresh` (to be sure we won't overwrite some action manager function) and do this there.

web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
```diff
@@ -25,8 +25,19 @@ odoo.define('web.web_ir_actions_act_window_message', function(require)
...
+                        // recursive_reload function is available
+                        // in the form view
+                        if (controller.recursive_reload)
+                        {
+                            controller.recursive_reload();
+                        }
+                        // reload_content function is available
+                        // in the list view
+                        if (controller.reload_content)
```
hbrunn: pull this up to have an `else if` branch above

web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
```diff
@@ -25,8 +25,19 @@ odoo.define('web.web_ir_actions_act_window_message', function(require)
...
+                        {
+                            controller.reload_content()
+                        }
```
hbrunn: and in an `else` branch, you can `console.log` that a refresh was requested, that's nice for the next person using this in a different sort of view

hbrunn (changes requested):

zakiuu:
@hbrunn fixed!

hbrunn (approved):

andreparames (approved):

zakiuu:
@hbrunn thanks a lot for your help 👍

ThomasBinsfeld (approved):

zakiuu:
@jbeficent  Travis seems to be red because of a `Bokeh` library, is there a solution that you can provide for that.

andreparames:
cc @simahawk 

Seems to be the same problem as in https://github.com/OCA/web/pull/852, maybe the same fix is required in v9?

andreparames:
Ah, I think the fix is this: https://github.com/OCA/web/pull/909

@zakiuu maybe you could try cherry-picking that commit here and see if it fixes Travis?

zakiuu:
@andreparames thx for solution, the `Bokeh` issue is gone
@sbidoul still some lint errors but not related to this module, so I guess we can merge

ThomasBinsfeld (approved):

benwillig (approved):

PR #978, opened 2018-07-03
[IMP][11.0][web_responsive] Suggested changes

Tardo:
Suggested changes in #540 (FieldStatus + Header Buttons)
Refactor "Chatter Position" feature (Cleaned unused/complex stuff)

screen shot 2018-07-03 at 22 44 18
screenshot_2018-07-03 so012 - odoo 1


web_responsive/static/src/js/web_responsive.js
```diff
@@ -306,11 +310,71 @@ odoo.define('web_responsive', function(require) {
...
+            this.$el.find('button.btn-default').toggleClass('hidden-xs');
+            $(e.currentTarget).blur();
+        },
+    });
+
+    // if we are in small screen create a menu for header buttons
+    FormRenderer.include({
+        _renderHeaderButton: function (node) {
+            var $button = undefined;
+            if (config.device.size_class <= config.device.SIZES.XS) {
```
yajo: Doing this through JS means that, if I resize the browser, I have to refresh to see the effects.

yajo (changes requested):
Thanks for the contribution and cleanup!

I'm seeing some weirdities:

captura de pantalla de 2018-07-04 07-57-07

TBH that was there before this PR, but now seems a good moment to fix it.

captura de pantalla de 2018-07-04 07-57-46

Maybe to reduce the duplication, you can imagine a different word for it, or you can use a proper icon.

captura de pantalla de 2018-07-04 07-57-57

I think that some of these problems come from the fact that [Bottstrap 3 expects a specific `.dropdown-menu>li>a` structure](https://getbootstrap.com/docs/3.3/components/#btn-dropdowns) to properly render a dropdown, and you're using buttons here.

So I would suggest trying this:

1. Action buttons

   - Render the structure Bootstrap expects separately from current one
   - Put all of that into a `<div class="hidden-lg hidden-md">`
   - Put upstream odoo buttons into a  `<div class="hidden-xs hidden-sm">`

2. State widget

   - This widget already supports folding. All you have to do is [override this method and set `fold: true` where needed](https://github.com/odoo/odoo/blob/daedcd92e36abdca43dd5580288311a05d87da24/addons/web/static/src/js/fields/relational_fields.js#L1996).

web_responsive/static/src/js/web_responsive.js
```diff
@@ -310,60 +310,57 @@ odoo.define('web_responsive', function(require) {
...
+            });
         },
 
-        _onClickStage: function (e) {
-            this._super(e);
-            this.$el.find('button.btn-default').toggleClass('hidden-xs');
-            $(e.currentTarget).blur();
-        },
+        _render: function () {
+            // Break inheritance!
```
yajo: Is there a chance to break it in the qweb template instead, by replacing elements?

pedrobaeza:
@Yajo can you update your review here? I have launched a rebuild on runbot.

yajo (approved):
Code OK

pedrobaeza (approved):
Thanks for the improvements!

PR #979, opened 2018-07-04
[FIX][11.0][web_tree_dynamic_colored_field] Issue #906

### LINKED ISSUE #906: [11.0] web_tree_dynamic_colored_field : my_color need to be visible for the color to be seen in the treeview options
Hello,

In web_tree_dynamic_colored_field, in the tree tag, when using the color_field property, it seems that the color field has to be visible.

ezgif com-optimize

Thank you!

Tardo:
This PR fixes #906 ... assumed that if the first record have the field all the records have the field.

### DIFF
--- a/web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
+++ b/web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
@@ -17,15 +17,12 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
                 if (colorAttr.length > 0) {
                     var colorField = colorAttr[0].split(':')[1].trim();
                     // validate the presence of that field in tree view
-                    var fieldNames = _(this.columns).map(
-                        (value) => { return value.attrs.name; }
-                    );
-                    if (fieldNames.indexOf(colorField) === -1) {
+                    if (this.state.data.length && colorField in this.state.data[0].data) {
+                        this.colorField = colorField;
+                    } else {
                         console.warn(
                             "No field named '" + colorField + "' present in view."
                         );
-                    } else {
-                        this.colorField = colorField;
                     }
                 }
             }


web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -17,15 +17,12 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
                 if (colorAttr.length > 0) {
                     var colorField = colorAttr[0].split(':')[1].trim();
                     // validate the presence of that field in tree view
-                    var fieldNames = _(this.columns).map(
-                        (value) => { return value.attrs.name; }
-                    );
-                    if (fieldNames.indexOf(colorField) === -1) {
+                    if (this.state.data.length && colorField in this.state.data[0].data) {
```
simahawk, line 20: if think `if (colorField in this.state.fields)` is enough, no?

Tardo: If others can confirm this i will change it.

Tardo: Ok... ```this.state.fields``` contains all fields of the model, ```this.state.data[0].data``` contains all the fields defined in the view for the first record.

pedrobaeza: What is the final conclusion in this?

Tardo: this.state.data[0].data -> You ensure that the value exists in the view. (In this case in the first record)

pedrobaeza: @simahawk can you update your review?

simahawk (approved):

PR #984, opened 2018-07-08
[11.0][MIG] web_advanced_search_x2x

JBF91:
First attemp on migrating web_advanced_search_x2x, It wasn't easy, but it's working. I'm sure it can be improved.

The basics of this migration:

**- Remove web_domain_v11 dependency**
**- Adapt how JS manage the fields, which has change a bit since 10.0**

I'm open to apply any change in order to improve the code, I'm sure is far to be perfect.


yajo:
It has some problems:

- Options "is equal to" and "is not equal to" do not use a many2one widget anymore.
- When clicking a field in the domain selector, the whole search dropdown is hidden.

I have WIP branch around there, still not working fine too. Let me compare both and I'll try to get best changes from both. :wink:

JBF91:
@Yajo I think my last commit fixes the dropdown hidding by prevent click event propagation.

yajo:
It's finished: https://github.com/OCA/web/pull/999

Closing in favor of that, please review there. Thanks! :raised_hands:

Issue #987, opened 2018-07-11
[11.0]web_group_expand: Expand/Collapse buttons are not shown when using a standard search (saved as favorite)


RamlaFtouh:
When we use a standard search that is saved as favorite, the Expand/Collapse buttons are not shown.

Tardo:
Can't reproduce this issue...
This module only works with "grouped" search's... Anyway I don't understand very well what "standard search" means.

RamlaFtouh:
Hello , 
I'm sorry for not being clear in my comment. 

If I have a custom filter (containing groups) that is registered , and used like a default filter ,  the Expand/Collapse options are not available (buttons not shown) when the tree view is loaded for first time
![web_group_expand_issue](https://user-images.githubusercontent.com/30926992/42740194-a5b00a46-889b-11e8-8295-52933c34463e.png)

I think the problem come from web_grroup_expand.js  -> Viewmanager ->_process_search_data function 

Regards.
.

Tardo:
Confirmed and fixed in #993 ... needs review

PR #993, opened 2018-07-16
[11.0][FIX][web_group_expand] Fix issue #987 & Improvements

Tardo:
- Fixes issue #987 
- Follow new readme structure
- Use Less
- Hide expand/collapse buttons if select a view type that isn't a 'list'


pedrobaeza:
What do you mean with `Hide expand/collapse buttons if select a view type that isn't a 'list'`?

Tardo:
@pedrobaeza I refer to this behavior:
- View Type == List
expand_s
- View Type != List
expand_n

aldo774:
@Tardo, hi dude, I made local tests, and it's worked perfectly, but an error is raised when i tried to access the action edit option in debugger options.
PS: The error isn't raised in previous implementation, at least in my local tests.

test_987

Tardo:
Thanks for the report @aldo774 ... fixed!

aldo774 (approved):

pedrobaeza (approved):
Thanks!

PR #998, opened 2018-07-18
[web_shortcut] don't destroy debug mode; allow opening shortcut in new tab

hbrunn:
without this, we don't see the interesting debug mode features like editing the current view or action


pedrobaeza (approved):
Functional review

yajo (approved):
Code OK

pedrobaeza (approved):

PR #575, opened 2017-03-08
[ADD[ web_widget_date_interval

hbrunn:
Date intervals
==============

This module was written to allow users to select some visual presentation of an interval instead of selecting a begin or end date.

Usage
=====

To use this module, use the widget ``date_interval`` on the field representing the begin of the interval. Make sure to also add the field representing the end of the interval, usually as hidden. Also add an ``options`` dictionary where you at least set the keys ``type`` and ``end_field``.

Depending on the type, you can add more configuration in the options dictionary, see below for a list.

weeknumber_iso
--------------

``hide_years``
    if this is set, don't show a year selection
``years_before``
    allow the user to move to N years before the current date (or the field's value)
``years_after``
    allow the user to move to N years after the current date (or the field's value)

Example::

    <field name="date_start" widget="date_interval" options="{'type': 'weeknumber_iso', 'end_field': 'date_end'}" />

Roadmap
=======

* add more interval types

PR #1001, opened 2018-07-23
[ADD] web_widget_date_interval

hbrunn:
This is a widget for date intervals, currently only implemented for search views (the form part will be a migration of https://github.com/OCA/web/pull/575).

For testing, open the users list on runbot and play with the two extra dropdowns allowing you to filter for some intervals on the fields `create_date` and `last_login`


ovnicraft:
@hbrunn any updates ?

hbrunn:
not at the moment, but the addon is not dead. If you prefer, I can close this until I start working on it again. It's already useful for me, so I figured it might be for others as well, that's why I kept the PR here

ovnicraft:
@hbrunn so its usable? 
We can review and merge it, so in next pr you can improve it.

hbrunn:
well, usable is a big word. For now, you can add something like
screenshot_2019-01-08_17-01-45
to a search view, which allows you to cycle through calendar (iso) weeks of whatever date{,time} field you configured quickly. But it's not feature complete, so I think it will be quite a bummer for people to see this merged but not be able to do a lot of stuff, like having a form widget for date intervals

ovnicraft:
Ok, so wait until you have time to end it 👍

web_widget_date_interval/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Date interval widget",
+    "version": "10.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
```
nikul-serpentcs, line 7: add `website` [like](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11)

nikul-serpentcs (changes requested):
Minor change

PR #1004, opened 2018-07-29
11.0 mig web listview invert selection

MeetKD:


espo-tony (commented):
Hi,
I tested this module and it works perfectly. My only concern is for the look of the "Invert selection" button, which I find a bit out of place with the rest of a listview layout. Do you think is it possible to give to the button a look which is more flat so that it can better merge with the rest of the page?
Thank you

MeetKD:
@espo-tony Agree with you. :+1:

web_listview_invert_selection/README.rst
```diff
@@ -0,0 +1,64 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
tarteo, line 1: Not important but it would be nice if you use the new partitioned readme.

web_listview_invert_selection/static/src/js/web_listview_invert_selection.js
```diff
@@ -0,0 +1,42 @@
...
+            }
+            return $header;
+        },
+
+        _renderInvertSelection: function (tag) {
+            var $icon = $('<i>', {class: 'fa fa-refresh o_invert_selection_btn', name: 'Invert Selection',
+                'aria-label': _t('Invert Selection'), 'title': _t('Invert Selection')});
+            return $('<' + tag + '>')
+                    .addClass('o_invert_selection')
+                    .append($icon);
```
tarteo, line 38: Also not blocking but I think using $.wrap() here is a bit cleaner.

web_listview_invert_selection/static/src/less/web_listview_invert_selection.less
```diff
@@ -0,0 +1,5 @@
+.o_list_view {
+    .o_invert_selection {
+        padding-left: 2px;
```
tarteo, line 3: Can you remove the padding? I think it looks better without it.

tarteo (changes requested):
Only some annoying nitpicking. :flushed:

tarteo:
@MeetKD Are you planning to finish this?

tarteo (approved):
My comments were non-blocking

PR #1005, opened 2018-07-29
[IMP] web_environment_ribbon: split renderer from rendering

blaggacao:
The motivation behind this is to make it straight-forward to separately bind
the method, so that you can bind a customized method
to a customized event, easily.

Signed-off-by: "David Arnold" <"dar@xoe.solutions">


pedrobaeza:
I think you have pinged the incorrect persons, as we are not experts on the JS side (at least me). @Yajo can you check this?

web_environment_ribbon/static/src/js/ribbon_show.js
```diff
@@ -0,0 +1,16 @@
...
+   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define('web_environment_ribbon.ribbon_show', function(require) {
+"use strict";
+
+    var core = require('web.core');
+    var showRibbon = require('web_environment_ribbon.ribbon');
+    core.bus.on('web_client_ready', null, showRibbon);
+
+}); // odoo.define
```
jabibi: \ No newline at end of file

jabibi:
Tested in runbot :+1:

web_environment_ribbon/static/src/js/ribbon_show.js
jabibi: Also you need to bump module version according to guidelines

yajo: Remove comment please.

web_environment_ribbon/data/ribbon_data.xml
```diff
@@ -6,7 +6,7 @@
 <!-- Add ribbon name default configuration parameter -->
 <record id="default_ribbon_name" model="ir.config_parameter">
     <field name="key">ribbon.name</field>
-    <field name="value"><![CDATA[TEST<br/>({db_name})]]></field>
+    <field name="value"><![CDATA[TEST<br/><sup>({db_name})</sup>]]></field>
```
yajo, line 9: Why this? :thinking:

web_environment_ribbon/static/src/js/ribbon_show.js
```diff
@@ -0,0 +1,16 @@
...
+   Copyright 2017 Xavier Jiménez <xavier.jimenez@qubiq.es>
+   Copyright 2018 David Arnold <dar@xoe.solutions>
+   License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define('web_environment_ribbon.ribbon_show', function(require) {
+"use strict";
+
+    var core = require('web.core');
+    var showRibbon = require('web_environment_ribbon.ribbon');
+    core.bus.on('web_client_ready', null, showRibbon);
```
yajo, line 14: I don't think this operation justifies the creation of a whole new module.

This patch could be way smaller by just putting this line at the end of the original JS module, and adding `return {showRibbon: showRibbon}` to the end. Could you do that please?

yajo (changes requested):
Thanks!

simahawk (changes requested):
same remarks

web_environment_ribbon/static/src/js/ribbon_show.js
blaggacao: According to guidelines, I thought,  I shouldn't touch / modify existing code.
I'm confused what rule is currently in vigor, then :wink: 
But if you confirm, I will take the opportunity to change that...

web_environment_ribbon/data/ribbon_data.xml
blaggacao: The previous one was not really readable the second line was too close to the border...
This way it looks nice and clean (see also commit message)

web_environment_ribbon/static/src/js/ribbon_show.js
blaggacao: I'm not sure if I understand how that would fit the purpose of the whole exercise.
Could you explain?
What is the goal? -> split our _rendering_ so that I can customize the rendering, not the renderer.
Or could you give an example how to achieve that more cleanly?

My line of thought: I can override the loading with an xpath and xml inheritance and therefore overload this extra module with my custom _rendering_ function, still calling the same _renderer_.

Please let me know, if this use case is intelligible? - :wink:

web_environment_ribbon/static/src/js/ribbon_show.js
blaggacao: alright, thanks.

blaggacao: I'll do that for evtl. later squashing.

blaggacao: Sorry the correct rule to cite her is: Align with style of surrounding code... :wink: 
Those rules are confusing, and interpretation is even more, some times...

blaggacao: Please, just quickly clarify, what should apply here and base on which rule to help me through this jungle...

blaggacao:
That's the goal, more or less:

web_environment_ribbon/static/src/js/ribbon_show.js
yajo: Maybe I confused you with the word "module". I was referring to a JS module, not to a Odoo module.

Summarizing, you're doing this:

1. Creating a JS module `web_environment_ribbon.ribbon` that returns the method that renders the ribbon.
2. Creating a JS module `web_environment_ribbon.ribbon_show` that binds the previous method to the `web_client_ready` event.

This adds some complexity that has actually no real benefit over just:

1. Have a single JS module `web_environment_ribbon.ribbon`, as it was before.
1. Inside that module, define a named `showRibbon` function.
1. Bind that function to the `web_client_ready` event.
1. Return the function, just in case somebody wants to extend it, by adding `return {showRibbon: showRibbon}` to the end of the module.

In its simplest form, this would take a diff of about 3 LOC.

OTOH, if your ultimate goal is to override the rendering of the ribbon, I suggest you to refactor subtly the method itself and make it render a client-side Qweb template, as normal widgets do.

web_environment_ribbon/static/src/js/ribbon_show.js
yajo: Sorry if I confused you, :sweat_smile:  please refer to https://github.com/OCA/web/pull/1005/files#r206049184, which will make this file disappear.

web_environment_ribbon/static/src/js/ribbon_show.js
blaggacao: Well, I'm still no 100% percent sure, if this is my true intention.
Let me ask: Would 3. bind the variable pointer or a copy? I don't know js enough to be able to tell that.
Or could I later just rebind the `web_client_ready` to provoke rendering a different method (which includes `showRibbon` in some way).
If this is true, then this PR is nonesense... :wink: 

I'd just need 
- return a deferred
- export the function

So that I can ultimately do: `core.bus.on('web_client_ready', null, showRibbonWithTimer);`

Is that about right?

yajo: Possibly even if you have the returned function, modifying it wouldn't mean modifying what was bound to the event.

I doubt you have more to do besides customizing the layout of the ribbon. Most useful would be then to use client-side qweb to render, and just override the qweb template in your addon.

blaggacao: I plan to wrap a java script timer around and possibly a help text pop-up.
So if you can quickly confirm, that I can rebind on the very same event, then I can achieve the same thing the way you suggested.

I'd be happy to refactor and get this trough.

yajo:
To be able to make that, I guess you need to refactor this addon, which BTW is not very well oriented.

The proper way to do this, thinking on the odoo js framework, is to declare all of this functionality as a `Widget`, and then append the widget to its parent.

The widget can then use the proper templating and event system, it can bind proxied methods to bus events, and it can be properly overriden by a submodule.

blaggacao:
@Yajo I like that idea... Would refactoring to a widget be accepted here?

There is also a new mechansim for readyness `'web.dom_ready'` haven't look into it in detail, yet. But maybe it would be more Odoonic.

yajo:
It's a good start, although I'd vote for the `web_client_ready` or directly overriding `WebClient.start()`.

I labeled as WIP, since I guess that's what this PR just became, isn't it? :stuck_out_tongue:

Issue #850, opened 2018-01-30
[10.0] web_responsive: security issue showing all menus


mohamedhagag:
Dears,
I would like to congratulate all of you on the great work and effort being done here in OCA Repos.
We found a security flaw in the latest web responsive module starting from the version providing search in the drawer.
the security issue is that all users can access all menus when searching in the drawer which exposes security.
even the user can open the menu and see all records that he should not see.

Regards,

pedrobaeza:
Are you sure about this? I think you are only able to search on allowed menus...

mohamedhagag:
One of our cutomers who is paranoid about security discovered this and we tested it and found it real to some level.
For ex. the sales user who have 'Own docs only' group permission can see and access all other users' invoices and can see the amounts and other info in the tree view, but he can't open the form view and see all other info.

I can record a video and put the url here if you want to see it in action.
Regards,

pedrobaeza:
I think that's totally unrelated to this modules, as this is controlled at another level: record rules. `web_responsive` has nothing to do with that. It only opens the action window, but if there's some record rules, they are applied at ORM level. Please try to reproduce the same on runbot.

mohamedhagag:
I tested it on OCA's /web runbot and recorded below video to show what I mean.

Runbot DB: http://3316021-10-0-929879.runbot2.odoo-community.org/web#min=1&limit=80&view_type=list&model=res.users&action=71

Video URL: https://inet.dvit.me/share/s/G_okeMrDSI2AeHYjkC4Llg

FLovison:
I have also verified the same issue ...

pedrobaeza:
@Yajo @Tardo can you confirm this issue?

Tardo:
I think that it's a permissions issue not related with "web_responsive" module... with correct access rights Odoo deny access to the views correctly...

For me the problem it's that the search feature returns entries that you really don't have access to see. (Odoo launch access rights exception correctly if you select them)

** Tested on local machine

Query for search menu entries: https://github.com/OCA/web/blob/10.0/web_responsive/static/src/js/web_responsive.js#L316
Link to launch action: https://github.com/OCA/web/blob/10.0/web_responsive/static/src/xml/app_drawer_menu_search.xml#L8

pedrobaeza:
Yeah, then it seems we have to limit to the actual menus that are visible.

PR #838, opened 2018-01-09
[IMP] web_responsive: Add search feature

lasley:
* Add feature to search menus in the app drawer

Forward port of #714

PR #1007, opened 2018-08-01
[IMP] web_responsive: Add search feature

yajo:
Indludes a squashed and slightly altered version of @lasley's #838, and supersedes that stale PR.

You should be able to search menus now.

@Tecnativa

### DIFF
--- a/web_responsive/readme/DESCRIPTION.rst
+++ b/web_responsive/readme/DESCRIPTION.rst
@@ -6,18 +6,4 @@ Features:
 * Keyboard shortcuts for easier navigation
 * Display kanban views for small screens if an action or field One2x
 * Set chatter side (Optional per user)
-* Quick search (see below)
-
-The following keyboard shortcuts are implemented:
-
-* Toggle App Drawer - `ActionKey <https://en.wikipedia.org/wiki/Access_key#Access_in_different_browsers>` + ``A``
-* Navigate Apps Drawer - Arrow Keys
-* Type to select App Links
-* ``esc`` to close App Drawer
-
-The search feature provided in the App Drawer allows you to easily
-navigate menus without a mouse.
-
-To activate the search, just begin typing while within the App Drawer.
-You can use the arrow keys or mouse to navigate and select results,
-in a similar fashion to navigating the apps in the drawer.
+* Quick search
--- a/web_responsive/readme/DESCRIPTION.rst
+++ b/web_responsive/readme/DESCRIPTION.rst
@@ -2,7 +2,22 @@ This module provides a mobile compliant interface for Odoo Community web.
 
 Features:
 
- * New navigation with an App drawer
- * Keyboard shortcuts for easier navigation
- * Display kanban views for small screens if an action or field One2x
- * Set chatter side (Optional per user)
+* New navigation with an App drawer
+* Keyboard shortcuts for easier navigation
+* Display kanban views for small screens if an action or field One2x
+* Set chatter side (Optional per user)
+* Quick search (see below)
+
+The following keyboard shortcuts are implemented:
+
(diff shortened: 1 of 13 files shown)


web_responsive/readme/DESCRIPTION.rst
```diff
@@ -2,7 +2,22 @@ This module provides a mobile compliant interface for Odoo Community web.
...
- * Keyboard shortcuts for easier navigation
- * Display kanban views for small screens if an action or field One2x
- * Set chatter side (Optional per user)
+* New navigation with an App drawer
+* Keyboard shortcuts for easier navigation
+* Display kanban views for small screens if an action or field One2x
+* Set chatter side (Optional per user)
+* Quick search (see below)
+
+The following keyboard shortcuts are implemented:
```
pedrobaeza: This should be in "Usage" section, isn't it?

pedrobaeza (dismissed):
Travis is red, and #850 is also present here. There should a method that returns the available menus instead of searching the ir.ui.menu entries.

pedrobaeza:
Please solve the conflict now that I have merged the other PR, and can you try to find a cleanest way of solving the menu issue. Standard menu and enterprise doesn't need these methods.

pedrobaeza:
@Yajo what is the status of this?

yajo:
I rebased, I wasn't aware of the git conflicts.

Current implementation is worse than enterprise because, as you said, it should use the cached menus instead of asking the server, but it's much better than it was in v10 because it defers the search until user stops writing and it aborts current search if he starts writing again, and because it fixes #850.

Implementing enterprise-like search would probably mean refactoring the addon, which is actually quite messy.

All of this is written in the known issues section. I wouldn't personally block enhancements just for not being perfect, mostly when they are much better than what was there before, but I'm not the one to decide on that. If you prefer to block it until it's perfect, just be aware it will take some time.

Besides all of that, it's still missing reviews.

pedrobaeza:
Using this PR, half of the menus disappear not in the search, but on the top part. Can you please check?

yajo:
I cannot reproduce :confused:

pedrobaeza:
Put it in our instance as a test and you'll see the problem.

Tardo:
Some error on runbot:
error_odoo

Login as "demo" user cause mistakes...
- When no app available in the menu, if use search feature and press enter you get an "undefined" url.
- When app available you can't navigate... do a click on icon doesn't launch any action.

yajo:
> Some error on runbot:

This comes from the `web_action_conditionable` addon, which asking for a dependency that doesn't exist (`web.View`). It's unrelated AFAICS.



> When app available you can't navigate... do a click on icon doesn't launch any action.

It seems to be a cache problem. Reloading without cache worked for me. Can you check please?

> When no app available in the menu, if use search feature and press enter you get an "undefined" url.

It should be fixed now.

yajo:
> half of the menus disappear not in the search, but on the top part

I finally managed to reproduce this, I'll work on the fix. Thanks!

yajo:
It should work now.

tarteo (changes requested):
Cool feature!

Is it possible to search on the display_name instead of the name e.g. 'Settings' should also find menu 'Settings/Dashboard'.

yajo:
It's not possible because it is a computed, non-stored, context-dependant field.

The only way would be doing an offline search, purely in JS, through the cached visible menus, as `web_enterprise` does, but that'd mean a refactor of this addon, which is needed anyway BTW, but honestly I'd try to tackle that when migrating to v12.

tarteo (approved):

pedrobaeza (approved):
Please squash commits before merging

pedrobaeza:
I have squashed already, so I merge.

PR #1010, opened 2018-08-02
10.0-fix-issue-727

devang-dreambits:
[FIX] fixed issue for chrome; now, for unsaved data, it displays a dialog of warning...

### DIFF
--- a/web_confirm_window_close/static/src/js/web_confirm_window_close.js
+++ b/web_confirm_window_close/static/src/js/web_confirm_window_close.js
@@ -10,7 +10,7 @@ odoo.define('web_confirm_window_close', function (require) {
             if ($('html').find('.oe_form_dirty').length) {
                 e.preventDefault();
             }
-            return 'Your have unsaved data. Are you sure you want to close window?';
+            return "";
         },
 
         start: function() {
--- a/web_confirm_window_close/static/src/js/web_confirm_window_close.js
+++ b/web_confirm_window_close/static/src/js/web_confirm_window_close.js
@@ -10,6 +10,7 @@ odoo.define('web_confirm_window_close', function (require) {
             if ($('html').find('.oe_form_dirty').length) {
                 e.preventDefault();
             }
+            return 'Your have unsaved data. Are you sure you want to close window?';
         },
 
         start: function() {
--- a/web_confirm_window_close/README.rst
+++ b/web_confirm_window_close/README.rst
@@ -42,6 +42,7 @@ Images
 Contributors
 ------------
 
+* Devang Pipaliya <devang@dreambits.in>
 * Stefan Rijnhart <stefan@opener.amsterdam>
 
 Maintainer
--- a/web_confirm_window_close/__manifest__.py
+++ b/web_confirm_window_close/__manifest__.py
@@ -4,7 +4,7 @@
 {
     'name': 'Check for unsaved data when closing browser window',
     'version': '10.0.1.0.0',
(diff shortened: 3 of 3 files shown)


devang-dreambits:
@pedrobaeza  Please help me in this PR .. Travis is failing and it doesn't look like my commit is the cause.

web_confirm_window_close/static/src/js/web_confirm_window_close.js
```diff
@@ -10,6 +10,7 @@ odoo.define('web_confirm_window_close', function (require) {
             if ($('html').find('.oe_form_dirty').length) {
                 e.preventDefault();
             }
+            return 'Your have unsaved data. Are you sure you want to close window?';
```
Tardo: Please use translations

Tardo (changes requested):

pedrobaeza:
The error is unrelated to your PR, so you can ignore it.

web_confirm_window_close/static/src/js/web_confirm_window_close.js
devang-dreambits: @Tardo  I came to know about Chrome & Firefox has removed the custom message on window close event...
Chrome
https://www.chromestatus.com/feature/5349061406228480

Firefox
https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload

So shall I add a custom message in return statement? Or returning blank string will be fine...

Tardo: https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Notes
>Starting with Firefox 44, Chrome 51, Opera 38 and Safari 9.1, a generic string not under the control of the webpage will be shown instead of the returned string.

I don't know what is the criteria to keep maintaining features that works on outdated browsers.

Any suggestion @OCA/web-maintainers ?

pedrobaeza: Well, this is a very specific case that can be determined by the maintainers of the module.

karan-dreambits: @pedrobaeza  @Tardo  @OCA/web-maintainers I checked this PR with @devang-dreambits  yesterday and what we found was that returning any string is enough to invoke the Confirm Dialog in chrome. 
In outdated browsers, both chrome and firefox did show the returned string in the dialog but that it not an requirement.
So wouldn't it be better to return an empty string as it would work well with both new and old browsers and also not get into extra effort of making it translatable as it would be ignored by browsers anyways

pedrobaeza: OK for me.

karan-dreambits: As discussed earlier and as @Tardo found here

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Notes
> Starting with Firefox 44, Chrome 51, Opera 38 and Safari 9.1, a generic string not under the control of the webpage will be shown instead of the returned string.


Any string returned would be ignored anyways and in older browser string would work just find and display their default generic dialog so better way is it to return blank string.

Please return blank string instead of this custom message

karan-dreambits (changes requested):

devang-dreambits:
@pedrobaeza  Thank you for the help.

karan-dreambits (approved):

Tardo (approved):

PR #1013, opened 2018-08-07
[11.0][FIX]web_widget_x2many_2d_matrix: Fix limit

### LINKED ISSUE #1012: [11.0][web_widget_x2many_2d_matrix] Problem with limit on matrix
The limit attribute makes that the matrix does not have more than 40 element, because default limit is 40.
Problems that appear when rows x columns > limit = 40:
1.- Undefined col
2.- If the number of columns of a row is divider of the limit, the matrix only shows the rows that have (in total) 40 elements
Steps to example:
Matrix with 5 columns and 9 or more rows, the widget only show the first 8 rows. 8*5=40, the default limit of One2many
If we have 6 columns and 7 or more rows, odoo will throw an error, because 6 dont is divider of 40.

For the moment, my solution is change the limit to Infinty on:
init_matrix: function(){
      var self = this,
          records = self.recordData[this.name].data;
      // Wipe the content if something still exists
      this.by_x_axis = {};
      this.by_y_axis = {};

      this.attrs.limit = Infinity; <---------------

Regards.

agranadosb:
Fix the problems with the limit of the matrix.

#Fixes #1012

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
@@ -67,6 +67,7 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
       // Wipe the content if something still exists
       this.by_x_axis = {};
       this.by_y_axis = {};
+      this.attrs.limit = Infinity;
       _.each(records, function(record) {
         var x = record.data[self.field_x_axis],
             y = record.data[self.field_y_axis];


JordiBForgeFlow:
Does not seem to be a proper fix. See in hr_timesheet_sheet, removing the limit=100000 I put on the matrix. It won't show the days from aug 4 to aug 14.

image

JordiBForgeFlow:
How did you test to ensure that this is fixed?

JordiBForgeFlow:
I tried to reproduce your fix with hr_timesheet_sheet in https://github.com/OCA/hr-timesheet and did not work.

As far as I can see the only way to properly display is https://github.com/OCA/hr-timesheet/blob/11.0/hr_timesheet_sheet/views/hr_timesheet_sheet_views.xml#L60

pedrobaeza:
Maybe that's the value to put as default:
```js
this.attrs.limit = 1000000;
```

JordiBForgeFlow:
No. Back when I tested this I realized that changing this.attrs.limit in
thay part of code will not have any effect.

pedrobaeza:
OK, then the patch is definitively not touching the part of code that is needed for this.

simahawk:
that's because is the view that passes the limit to `model.(re)load`.
Not sure were we can hook in current code, Have to check.

pedrobaeza:
We can simply mention this question on the README as this is a foundation module, and let each developer to add to the extension modules the `limit` option.

agranadosb:
@jbeficent I reproduce your example and i see the full timesheet

captura de pantalla de 2018-08-07 11-34-07

This example works because i created since 0. Maybe there are the problem, try to add a new timesheet. I just change that code on init, just behind of init and behind of init_matrix.

JordiBForgeFlow:
Closing in favor of https://github.com/OCA/web/pull/1014

PR #1014, opened 2018-08-07
[11.0][web_widget_x2many_matrix] fix issue with limit of records

JordiBForgeFlow:
Fixes https://github.com/OCA/web/issues/1012

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
@@ -1,0 +1,17 @@
+odoo.define( "web_widget_x2many_2d_matrix.matrix_limit_extend", function (require) {
+"use strict";
+
+        var AbstractView = require("web.AbstractView");
+
+        AbstractView.include({
+            // We extend this method so that the view is not limited to
+            // just 40 cells when the 'x2many_2d_matrix' widget is used.
+            _setSubViewLimit: function (attrs) {
+                this._super(attrs);
+                if (attrs.widget === "x2many_2d_matrix") {
+                    attrs.limit = Infinity;
+                }
+            },
+        };
+    }
+);
--- a/web_widget_x2many_2d_matrix/__manifest__.py
+++ b/web_widget_x2many_2d_matrix/__manifest__.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "2D matrix for x2many fields",
-    "version": "11.0.1.1.0",
+    "version": "11.0.1.1.1",
     "author": "Therp BV, "
               "Tecnativa, "
               "Camptocamp, "
--- a/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
@@ -319,7 +319,7 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
          * @private
          * @returns {jQueryElement} The td element with the total in it.
          */
-        _renderTotalCell: function() {
+        _renderTotalCell: function () {
             if (!this.matrix_data.show_column_totals ||
                 !this.matrix_data.show_row_totals) {
                 return;
--- a/web_widget_x2many_2d_matrix/views/assets.xml
+++ b/web_widget_x2many_2d_matrix/views/assets.xml
@@ -5,6 +5,7 @@
             <xpath expr="." position="inside">
                 <script type="text/javascript" src="/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js" />
                 <script type="text/javascript" src="/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js" />
+                <script type="text/javascript" src="/web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js" />
                 <link rel="stylesheet" href="/web_widget_x2many_2d_matrix/static/src/css/web_widget_x2many_2d_matrix.css"/>
             </xpath>
         </template>


JordiBForgeFlow:
Thanks to @tonihr for proposing the fix!

cc @agranadosb, @simahawk @pedrobaeza 

As you can see below, I can now generate huge many2many_matrix grids.

image

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,24 @@
...
+    "use strict";
+
+var AbstractView = require('web.AbstractView');
+
+var AbstractViewCustom = AbstractView.include({
+
+    _setSubViewLimit: function (attrs) {
+	this._super(attrs);
+
+    var view = attrs.views && attrs.views[attrs.mode];
```
simahawk: bad indent :)
Also, for our future selves, please add a comment explaining why we do this ;)

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,24 @@
+odoo.define('web_widget_x2many_2d_matrix.matrix_limit_extend', function (require) {
+    "use strict";
+
+var AbstractView = require('web.AbstractView');
+
+var AbstractViewCustom = AbstractView.include({
```
simahawk: should not be applied only to tree views?

simahawk (changes requested):

pedrobaeza:
Why a new file instead existing ones?

pedrobaeza:
Check syntax recommendations from https://travis-ci.org/OCA/web/jobs/413220985#L844

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
JordiBForgeFlow: See https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/views/abstract_view.js#L270
We are applying the same approach here.

JordiBForgeFlow:
@pedrobaeza We create a new file so it is easier to understand that we are changing a different class.
@simahawk I have also applied the proper indentations.

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
simahawk: I know, I was just wondering if it wouldn't be better to apply the change to ListView only (which inherits from here: AbstractView -> BasicView -> ListView)

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,22 @@
...
+    // We extend this method so that the view is not limited to
+    // just 40 cells when the 'x2many_2d_matrix' widget is used.
+    _setSubViewLimit: function (attrs) {
+	    this._super(attrs);
+        var view = attrs.views && attrs.views[attrs.mode];
+        var limit = view && view.arch.attrs.limit && parseInt(view.arch.attrs.limit);
+        if (!limit && attrs.widget === 'x2many_2d_matrix') {
+            limit = Infinity;
+            attrs.limit = limit || 40;
+            }
```
simahawk: nitpicking :nerd_face: : this parenthesis should be indented back :stuck_out_tongue:

simahawk:
I'm fine w/ the new file BTW :)

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
JordiBForgeFlow: I just tried changing to the ListView, but _setSubViewLimit is never being called, so it crashes.

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
JordiBForgeFlow: fixed!

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
simahawk: :sob:

simahawk (approved):

pedrobaeza:
Let's not introduce new JS lint warnings:
```
web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js:10: [W7903(javascript-lint), ] Unexpected tab character. [Error/no-tabs]
web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js:10: [W7903(javascript-lint), ] Mixed spaces and tabs. [Error/no-mixed-spaces-and-tabs]
web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js:12: [W7903(javascript-lint), ] Missing radix parameter. [Error/radix]
web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js:10: [W7910(wrong-tabs-instead-of-spaces), ]  Use wrong tabs indentation instead of four spaces
```

JordiBForgeFlow:
@pedrobaeza I'm not sure where you found the warnings, but I attended them.

pedrobaeza:
You have introduced even more warnings. You can see them on https://travis-ci.org/OCA/web/jobs/423957411#L2131

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,22 @@
...
+        var view = attrs.views && attrs.views[attrs.mode];
+        var limit = view && view.arch.attrs.limit && parseInt(view.arch.attrs.limit, 0);
+        if (!limit && attrs.widget === 'x2many_2d_matrix') {
+            limit = Infinity;
+            attrs.limit = limit || 40;
+        }
+    }
+});
+
+return AbstractViewCustom;
```
yajo: No need to return because you're using `.include()` above.

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,22 @@
...
+var AbstractView = require('web.AbstractView');
+
+var AbstractViewCustom = AbstractView.include({
+    // We extend this method so that the view is not limited to
+    // just 40 cells when the 'x2many_2d_matrix' widget is used.
+    _setSubViewLimit: function (attrs) {
+        this._super(attrs);
+        var view = attrs.views && attrs.views[attrs.mode];
+        var limit = view && view.arch.attrs.limit && parseInt(view.arch.attrs.limit, 0);
+        if (!limit && attrs.widget === 'x2many_2d_matrix') {
```
yajo: AFAICS [here](https://github.com/odoo/odoo/blob/15c81d74a34b57c0aa0400ec8b66bf3397904c60/addons/web/static/src/js/views/abstract_view.js#L276), this condition will never be true, because the parent method always sets the limit, so basically this method will never do nothing; isn't it? :thinking: 

I think the implementation should be just:

```js
if (attrs.widget === 'x2many_2d_matrix') {
    var view = attrs.views && attrs.views[attrs.mode];
    _.defaults(view.arch.attrs, {limit: Infinity});
}
this._super(attrs);
```

yajo (changes requested):

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,22 @@
...
+            limit = Infinity;
+            attrs.limit = limit || 40;
```
tarteo: This can be just attrs.limit = Infinity, because limit here is always truthy.

tarteo (changes requested):

simahawk (changes requested):
@jbeficent can you get back to this? :)

pedrobaeza:
@jbeficent I have faced this issue today. Are you going to finish this?

JordiBForgeFlow:
@pedrobaeza  let me finish it now

pedrobaeza:
@tarteo @Yajo @simahawk can you re-review?

tarteo (approved):

JordiBForgeFlow:
@tarteo @pedrobaeza I have no idea how to fix the lint errors. I have tried to fix, but I cannot resolve. Any idea?

tarteo:
Try it like this:

```
odoo.define(
    "web_widget_x2many_2d_matrix.matrix_limit_extend",
    function (require) {
        "use strict";

        var AbstractView = require("web.AbstractView");

        AbstractView.include({
        ...
     }
);
```

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,19 @@
+odoo.define(
```
yajo: Actually [there's an exception on the line width for the `odoo.define` lines in the linter](https://github.com/OCA/pylint-odoo/blob/eb11004b3b6a99cca23f9c096bea284e9f2d7d02/pylint_odoo/examples/.jslintrc#L200), just for the case you have at hand.

The best you can do here is to put just a single line:

```js
 odoo.define( "web_widget_x2many_2d_matrix.matrix_limit_extend", function (require) {
```

Thanks to that exception, the linter should be OK.

yajo (changes requested):

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -1,17 +1,19 @@
...
+        AbstractView.include({
+            // We extend this method so that the view is not limited to
+            // just 40 cells when the 'x2many_2d_matrix' widget is used.
+            _setSubViewLimit: function (attrs) {
+                this._super(attrs);
+                if (attrs.widget === "x2many_2d_matrix") {
+                    attrs.limit = Infinity;
+                }
+            },
+        }
```
tarteo: You're missing here `);`

JordiBForgeFlow:
We're good to go

simahawk (approved):

pedrobaeza:
There were a syntax error fixed here: 899f01f

web_widget_x2many_2d_matrix/static/src/js/abstract_view_matrix_limit_extend.js
```diff
@@ -0,0 +1,17 @@
...
+        AbstractView.include({
+            // We extend this method so that the view is not limited to
+            // just 40 cells when the 'x2many_2d_matrix' widget is used.
+            _setSubViewLimit: function (attrs) {
+                this._super(attrs);
+                if (attrs.widget === "x2many_2d_matrix") {
+                    attrs.limit = Infinity;
+                }
+            },
+        };
```
hhgabelgaard, line 15: @jbeficent In Google Chrome this fails with an Uncaught SyntaxError: missing ) after argument list in line 15
 
Changing to below seems to work:

```
                    attrs.limit = Infinity;
                }
            },
        });
    }
);

```
But I'm no javascript expert!

pedrobaeza:
@hhgabelgaard I have just written the same and fixed it on the fly. Just update...

hhgabelgaard:
@pedrobaeza Thanks!

PR #1020, opened 2018-08-17
[11.0] [ADD] web_timeline: Dependency arrows 

tarteo:
Adds a new optional attribute dependency_arrow.

If the attribute is set it will look like:
image

I created a widget TimelineCanvas which does all the svg stuff. This way the renderer will not be bloated with it.

### DIFF
--- a/web_timeline/static/src/js/timeline_controller.js
+++ b/web_timeline/static/src/js/timeline_controller.js
@@ -5,6 +5,7 @@ var AbstractController = require('web.AbstractController');
 var dialogs = require('web.view_dialogs');
 var core = require('web.core');
 var time = require('web.time');
+var Dialog = require('web.Dialog');
 
 var _t = core._t;
 
@@ -31,6 +32,9 @@ var CalendarController = AbstractController.extend({
         if (_.isEmpty(params)){
             return;
         }
+        var defaults = _.defaults({}, options, {
+            adjust_window: true
+        });
         var self = this;
         var domains = params.domain;
         var contexts = params.context;
@@ -59,7 +63,7 @@ var CalendarController = AbstractController.extend({
             },
             context: self.getSession().user_context,
         }).then(function(data) {
-            return self.renderer.on_data_loaded(data, n_group_bys);
+            return self.renderer.on_data_loaded(data, n_group_bys, defaults.adjust_window);
         });
     },
 
@@ -82,13 +86,13 @@ var CalendarController = AbstractController.extend({
         var id = item.evt.id;
         var title = item.evt.__name;
         if (this.open_popup_action) {
-            var dialog = new dialogs.FormViewDialog(this, {
+            new dialogs.FormViewDialog(this, {
                 res_model: this.model.modelName,
-                res_id: parseInt(id).toString() == id ? parseInt(id) : id,
+                res_id: parseInt(id, 10).toString() === id ? parseInt(id, 10) : id,
                 context: this.getSession().user_context,
                 title: title,
-                view_id: +this.open_popup_action,
-                on_saved: function (record) {
+                view_id: Number(this.open_popup_action),
+                on_saved: function () {
                     self.write_completed();
                 },
             }).open();
@@ -99,7 +103,7 @@ var CalendarController = AbstractController.extend({
             }
             this.trigger_up('switch_view', {
                 view_type: 'form',
-                res_id: parseInt(id).toString() == id ? parseInt(id) : id,
+                res_id: parseInt(id, 10).toString() === id ? parseInt(id, 10) : id,
                 mode: mode,
                 model: this.model.modelName,
             });
@@ -114,7 +118,7 @@ var CalendarController = AbstractController.extend({
         var event_start = item.start;
         var event_end = item.end;
         var group = false;
-        if (item.group != -1) {
+        if (item.group !== -1) {
             group = item.group;
         }
         var data = {};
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Another great addition! With this and the planning module we are doing, we finally have a superb project app!

astirpe (changes requested):
Could you increase the version in manifest? Also in https://github.com/OCA/project/pull/420 should be stated that the minimum version of the required web_timeline module is the one you are going to set.

tarteo:
I've put the arrows UML way, but I think it's more logical for a timeline which is mostly read left to right, to put the arrowhead on the depending side. What do you think?

pedrobaeza:
When I first saw it, I think there was something weird on it, but being logical when I thought about it. I'm with your that maybe we should reverse it for the shake of clarity.

astirpe:
@tarteo yes should be reversed

web_timeline/static/src/js/timeline_controller.js
```diff
@@ -226,13 +229,17 @@ var CalendarController = AbstractController.extend({
...
-    write_completed: function () {
+    write_completed: function (options) {
         var params = {
             domain: this.renderer.last_domains,
             context: this.context,
             groupBy: this.renderer.last_group_bys,
         };
-        this.update(params, null);
+
+        var defaults = _.defaults(options || {}, {
```
yajo: `_.defaults({}, options, {` please. This way we avoid modifying the `options` object in place.

yajo (changes requested):
Looks pretty nice! Just a little thing.

Also please [review the JS linter](https://travis-ci.org/OCA/web/jobs/417320520#L769-L806), which doesn't put travis in :x:  currently, but should be honored if possible (and if it has errors, it should be reported, so we polish that linter too :blush: ).

Thanks!

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -176,9 +178,41 @@ var CalendarRenderer = AbstractRenderer.extend({
...
+            });
+        });
     },
 
-    on_data_loaded: function (events, group_bys) {
+    draw_dependency: function(from, to) {
+        if (!from.displayed || !to.displayed)
+            return;
+
+        this.canvas.draw_arrow(from.dom.box, to.dom.box, 'black', 1);
```
simahawk: shouldn't "black" come from an option?

simahawk (changes requested):

tarteo:
@pedrobaeza @astirpe  I've upped the version number and reversed the arrow head. You can see the result in the PR description.
@Jayo I've applied your suggestion and cleaned up the lint (I couldn't get rid of some of the errors) I thought about using directives but I rather let it print the error instead of hiding it. It's not causing the PR to be red so it's not a problem.
@simahawk I've changed the function so the line color and line width are changeable with the options parameter.

simahawk (approved):
LG

pedrobaeza:
@Yajo waiting your blessing for merging this

pedrobaeza:
@tarteo I think the next step in the roadmap should be the possibility of defining dependencies graphically, having for example a dot in timeline object from which you can drag and drop to other element's dot for setting the dependency. How do you see it?

astirpe (approved):

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -314,16 +330,9 @@ var CalendarRenderer = AbstractRenderer.extend({
             }
         });
 
-        var content = evt.__name != undefined ? evt.__name : evt.display_name;
+        var content = typeof evt.__name === 'undefined' ? evt.display_name : evt.__name;
```
yajo: I recommend you to use `!_.isUndefined(evt)` instead.

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -342,10 +351,25 @@ var CalendarRenderer = AbstractRenderer.extend({
         return r;
     },
 
+    render_timeline_item: function(evt) {
+        if(this.qweb.has_template('timeline-item')) {
```
yajo, line 355: Since `this.qweb` is conditionally declared, this could raise the common `TypeError: undefined has no property "has_template"` error.

It would be better IMHO to always have a `this.qweb` element, that may or may not have templates.

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -342,10 +351,25 @@ var CalendarRenderer = AbstractRenderer.extend({
...
+        console.error(
+            _t('Template "timeline-item" not present in timeline view definition.')
+        );
+    },
+
     on_group_click: function (e) {
         // handle a click on a group header
-        if (e.what === 'group-label' && e.group != -1) {
-            this._trigger(e, function(){}, 'onGroupClick');
+        if (e.what === 'group-label' && e.group !== -1) {
```
yajo, line 369: You can use `~e.group` instead of `e.group !== -1` :wink:

yajo (approved):
Good!

tarteo:
@pedrobaeza Yes, for the dependencies that's exactly how I see it. E.g. It would also be nice to use parent tasks to print a kind of collapse / expand control.

web_timeline/static/src/js/timeline_renderer.js
tarteo: Made this.qweb always available even when there are no templates

PR #1024, opened 2018-08-24
[11.0][ADD] Enable multiselect and fixed time axis in timeline

tarteo:
This way you can select and move multiple items at once (using Ctrl key). This is a stock option of visjs.

http://visjs.org/docs/timeline/#Configuration_Options

image

### DIFF
--- a/web_timeline/README.rst
+++ b/web_timeline/README.rst
@@ -116,6 +116,8 @@ edit the involved record directly.
 
 Double-click on the record to edit it. Double-click in open area to create a
 new record with the group and start date linked to the area you clicked in.
+By holding the Ctrl key and dragging left to right, you can create a new record
+with the dragged start and end date.
 
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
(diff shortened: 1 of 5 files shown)


pedrobaeza:
What happens with double click then (which currently edit the selected item)?

tarteo:
@pedrobaeza If you double click with ctrl it deselects and selects it again. If you double click it (without ctrl pressed), it will deselect all the others (because ctrl is not pressed) and open the one you double clicked. Double clicking will not open all the selected onces. You can also select with shift which selects all items between 2 clicks if you double click while holding shift, it doesn't reset the selection and opens the double clicked item.

pedrobaeza:
OK, that's great then. You are improving this view a lot.

simahawk (changes requested):
LG. No mention of the feature in README?

pedrobaeza (approved):

tarteo:
@simahawk There already was a mention in the README (see line 102).

I also added / fixed the functionality to add new records by dragging (using the Ctrl key)

tarteo:
@simahawk Can you reconsider your review please? ☺️

simahawk (approved):

PR #1026, opened 2018-08-28
[10.0] web_sheet_full_width: wider field labels on form view

alexis-via:
By default on v10, in particular when you have web_sheet_full_width installed, on a form view, the field labels are pretty narrow, so they often take 2 lines, which make it less readable for the user.

I am NOT a web expert, I don't know if web_sheet_full_width is the right place to fix that... so feel free to comment and improve this PR.

BEFORE:

col_width-before

AFTER:

col_width-after

### DIFF
--- a/web_sheet_full_width/static/src/css/web_sheet_full_width.less
+++ b/web_sheet_full_width/static/src/css/web_sheet_full_width.less
@@ -7,4 +7,13 @@
     .oe_chatter{
         max-width: none;
     }
+    .o_group{
+        &.o_inner_group{
+            > tbody > tr > td{
+                &.o_td_label{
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't think this is the proper solution, specially for responsive view in mobiles.

alexis-via:
yes, feel free to suggest a better solution. But field label width in form view is really too narrow in v10, this is definitely a problem.

pedrobaeza:
You should work on `web_responsive` with an adaptive solution. This module is deprecated in favor of that full featured one. I'm not sure right now if this has been already changed. Have you tried on v11 or v10?

PR #1029, opened 2018-08-29
[FIX] flake8 errors

hbrunn:
liniting should come back green here


pedrobaeza:
PhantomJS timeout ghost problem, so I fast-track.

PR #1037, opened 2018-09-14
[10.0][IMP] Escape field values for timeline color conditions

ecino:
I had an issue where we have records that have quotes in their field values. We want to use that field in our timeline conditions to set colors, but that makes the javascript crash.

With that fix, I propose to escape the quotes that could be present in field values.

The toString() method is to convert many2one relation fields which are stored in arrays.

Should I do the same PR for v11.0 ?

### DIFF
--- a/web_timeline/static/src/js/web_timeline.js
+++ b/web_timeline/static/src/js/web_timeline.js
@@ -245,7 +245,8 @@ odoo.define('web_timeline.TimelineView', function (require) {
                 group = -1;
             }
             _.each(self.colors, function (color) {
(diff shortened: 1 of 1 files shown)


tarteo:
This doesn't really fix the issue, because value `test\\'` will still fail. A function something like: 
```
function(field, opt, color) {
   switch(opt) {
      case '=':
      ...
         break;
   }
}
```

will do the trick

tarteo (changes requested):
This doesn't really fix the issue, because value `test\\'` will still fail. A function something like: 
```
function(field, opt, color) {
   switch(opt) {
      case '=':
      ...
         break;
   }
}
```

will do the trick

PR #1040, opened 2018-09-18
[FIX] web_advanced_search: `undefined` in x2m fields 

yajo:
Before this patch, when searching with the "equals to" operator in any
x2many field, the searched parameter was always `undefined`:

captura de pantalla de 2018-09-18 09-17-46

The problem was that the underlying field manager implementation was
treating those fields as x2many, while the widget used was the `one2many`
one.

This patch simply mocks the underlying fake record to make think that
any relational field is always a `one2many`. This sets all pieces in
place and makes the field manager work as expected, and thus you can
search as expected too.

Also some linter fixes landed, just in a separate commit.

@Tecnativa

### DIFF
--- a/web_advanced_search/static/src/js/web_advanced_search.js
+++ b/web_advanced_search/static/src/js/web_advanced_search.js
@@ -39,12 +39,21 @@ odoo.define("web_advanced_search", function (require) {
      * An almost dummy search proposition, to use with domain widget
      */
     var AdvancedSearchProposition = Widget.extend({
+
+        /**
+         * @override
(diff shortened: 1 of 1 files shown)


rafaelbn:
Tested in runbot :+1: Travis is :red_circle:

yajo:
I rebuilt travis and it's :heavy_check_mark: now. It was a temporary failure as it seems.

rafaelbn (approved):

pedrobaeza (approved):

PR #1042, opened 2018-09-19
[ADD] web_fontawesome5

etobella:
This module upgrades the fontawesome font to 5.3.1.

It allows to use new icons.


simahawk:
we already have this https://github.com/OCA/server-tools/tree/10.0/base_fontawesome

Not sure is worth to leave it there tho, sure is worth to port it to v11 and upgrade FA version then use it as a base :)

etobella:
Well, It is not a direct upgrade, because some javascript is required in order to allow the usage of all icons. However, I can make the upgrade if you you think it is the right way.

simahawk:
@etobella one thing are the FA resources, one thing are your hooks for odoo UI.
FA resources should stay in the base module IMO while in web_fontawesome you can glue odoo UI.
That was my proposal ;)

web_fontawesome5/static/src/js/form_renderer.js
```diff
@@ -0,0 +1,55 @@
...
+"use strict";
+
+var FormRenderer = require('web.FormRenderer');
+
+FormRenderer.include({
+
+    _renderStatButton: function (node) {
+        var $button = this._super.apply(this, arguments);
+        if (node.attrs.brand_icon) {
+            $('<div>')
```
simahawk, line 11: just asking: wouldn't be better to override widgets' xml templates instead?

simahawk (changes requested):
also, I don't think we need minified versions as they are minified on the fly.
Or, if we want to keep the module checkout lighter then we shoud rip out the not compressed one.

web_fontawesome5/static/src/js/form_renderer.js
etobella: Stats Buttons are computed directly on Javascript on odoo11.
https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/views/form/form_renderer.js#L498
The same applies for standard buttons.

pedrobaeza:
Remove the number off the module name for avoiding migration scripts changing number

etobella:
This PR has been deprectaded, as there exists a module with the same goal on server-tools.

PR #1046, opened 2018-09-22
[FIX] Fixed inheritance bugs in web_notify. Did caused problems when …

wpichler:
…used with other modules which do inherit from WebClient (web_auto_refresh)

### DIFF
--- a/web_notify/README.rst
+++ b/web_notify/README.rst
@@ -93,6 +93,7 @@ Contributors
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Serpent Consulting Services Pvt. Ltd.<jay.vora@serpentcs.com>
 * Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Wolfgang Pichler <wpichler@callino.at>
 
 Maintainer
 ----------
--- a/web_notify/static/src/js/web_client.js
+++ b/web_notify/static/src/js/web_client.js
@@ -44,6 +44,9 @@ WebClient.include({
     ),
     init: function(parent, client_options){
         this._super(parent, client_options);
+        this.channel_warning = 'notify_warning_' + this.session.uid;
+        this.channel_info = 'notify_info_' + this.session.uid;
+
     },
     reload_active_view: function(){
         var action_manager = this.action_manager;
@@ -57,23 +60,17 @@ WebClient.include({
     },
     show_application: function() {
         var res = this._super();
-        this.start_polling();
+        base_bus.bus.start_polling();
+        base_bus.bus.add_channel(this.channel_warning);
+        base_bus.bus.add_channel(this.channel_info);
+        base_bus.bus.on('notification', this, this._web_notify_notification);
         return res
     },
     on_logout: function() {
-        var self = this;
-        base_bus.bus.off('notification', this, this.bus_notification);
+        base_bus.bus.off('notification', this, this._web_notify_notification);
         this._super();
     },
-    start_polling: function() {
-        this.channel_warning = 'notify_warning_' + this.session.uid;
-        this.channel_info = 'notify_info_' + this.session.uid;
-        base_bus.bus.add_channel(this.channel_warning);
-        base_bus.bus.add_channel(this.channel_info);
-        base_bus.bus.on('notification', this, this.bus_notification);
-        base_bus.bus.start_polling();
-    },
-    bus_notification: function(notifications) {
+    _web_notify_notification: function (notifications) {
         var self = this;
         _.each(notifications, function (notification) { 
             var channel = notification[0];


web_notify/static/src/js/web_client.js
```diff
@@ -100,5 +97,5 @@ WebClient.include({
         }
     }
 });
-
+return WebClient;
```
yajo: This is not needed.

yajo (changes requested):
AFAICS you're basically doing exactly the same that was being done, just in different methods, apart from the redundant rename of `bus_notification()` to `bus_notify_notification()`.

So, at first sight, this change would add nothing meaningful. You say it conflicts with https://github.com/OCA/web/pull/1047, but I also see no conflict points. Could you explain better why do we need to merge it? :thinking: 

Thanks!

web_notify/static/src/js/web_client.js
wpichler: Here are some things i found problematic - maybe i am wrong - but after my changes it did work as expected... ;-)

* start_polling function in web_client.js this function is ok if you use only the web_notify module. The web_auto_refresh module did used the same function (also with include on WebClient) - so you get unexpected results (last come first serve). So in my opinion it is better to not use too generic function names - or extend directly they function from where it got called.
* function bus_notification: same problem here - the name is too generic - other modules also use the name (as the docs also use this name - but for own module - not when directly included in WebClient !) - so the new name is not the best one - so maybe we could prefix such functions - lets call it _web_notify_bus_notification
* About the return WebClient - you are correct - this is not needed with include

web_notify/static/src/js/web_client.js
```diff
@@ -57,23 +60,17 @@ WebClient.include({
...
-    start_polling: function() {
-        this.channel_warning = 'notify_warning_' + this.session.uid;
-        this.channel_info = 'notify_info_' + this.session.uid;
-        base_bus.bus.add_channel(this.channel_warning);
-        base_bus.bus.add_channel(this.channel_info);
-        base_bus.bus.on('notification', this, this.bus_notification);
-        base_bus.bus.start_polling();
-    },
-    bus_notification: function(notifications) {
+    bus_notify_notification: function (notifications) {
```
yajo: I think a better name here would be `_web_notify_notification`.

web_notify/static/src/js/web_client.js
yajo: OK, I understand the problem.

Remember to follow the same logic in the other PR then, and please remove the return.

yajo (approved):

web_notify/static/src/js/web_client.js
wpichler: yep - good name

web_notify/static/src/js/web_client.js
lmignon: @Yajo Out of curiosity but why do you say that the `return` should be removed? If we want to extend the specific JS provided by this addon on the WebClient, we should return WebClient. In this way we'll be able to write
```javascript
odoo.define('my_extended_web_notify.WebClient', function (require) {
    var WebClient = require('web_notify.WebClient');
    ...
});
```

Do I miss something?

yajo: It's just not necessary. When calling `.include()`, you alter the original object's protoype, so further calls to the original object are the same as those for the returned object. If we were creating a subclass (through `.extend()`), then yes, it should be returned to be able to extend it further.

Your code and this code are exactly the same:

```js
odoo.define('my_extended_web_notify.WebClient', function (require) {
    require('web_notify.WebClient');
    var WebClient = require('web.WebClient'),
    ...
});
```

It's not actually *an error* to return that class, but it's simply not needed as you can see. And if you don't need something, it's better not to have it IMHO.

lmignon: OK thank you for the explanation. I thought that `require' was also used to ensure the loading order of the JS but on the other hand it's the extend of asset.backend  that should do it.

wpichler:
i've commited the requested changes with commit --amend and push force on my branch - i am not sure - do you get the information in this case ?

pedrobaeza:
> i've commited the requested changes with commit --amend and push force on my branch - i am not sure - do you get the information in this case ?

No, in that case, you should drop a comment saying so, as GitHub doesn't inform that change. Any way, it's always convenient to put the comment saying something like "All comments attended" or "Ready to re-review".

web_notify/static/src/js/web_client.js
yajo: require is async, so the order does not matter AFAIK.

web_notify/README.rst
```diff
@@ -1,4 +1,4 @@
-.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+q.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
yajo: typo

yajo (changes requested):
Please check Travis :x: too

web_notify/static/src/js/web_client.js
lmignon: no convinced but not an expert... :smirk:

web_notify/README.rst
wpichler: fixed

yajo (approved):
Travis not related.

PR #1054, opened 2018-10-01
[12.0][MIG] web_listview_range_select

tarteo:
I had to use window.getSelection().removeAllRanges(); to make this module work for 12.0. In 12.0 they use 'fake' checkboxes (which are actual a div). That causes shift clicking to select the content. If you have a better solution please, say :).

I updated the README using the new seperated method.


web_listview_range_select/README.rst
```diff
@@ -0,0 +1,79 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+No configuration is needed.
```
SimoRubi: might as well remove this chapter

SimoRubi (approved):
Good for me

web_listview_range_select/README.rst
tarteo: Yes :smile: removed it.

pedrobaeza (approved):
I still think it can be a good idea to allow to click on any part of the line when Shift is pressed, and also to allow to select with Ctrl in the same conditions. At least, put these ideas on the roadmap, and squash migration commits.

tarteo:
@pedrobaeza Yes, me too. I'll put them on the roadmap.
Also added:

* Shift and drag to select rows.

PR #1055, opened 2018-10-01
[10.0] Add new module system_filter

acsonefho:
- [x] New module "system_filter";
- [x] Add new field "is_system" into ir.filters (these filters are not loaded into web-interface (views));
- [x] Unit test added;

Any better ideas/improvements are welcome


system_filter/models/ir_filters.py
```diff
@@ -0,0 +1,45 @@
...
+    def _remove_system_filters(self, result):
+        """
+        Remove system filters from given result
+        :param result: list of dict
+        :return: list of dict
+        """
+        # Load every id
+        filter_ids = [r.get('id') for r in result]
+        # Get only filters where is_system is True
+        system_filter_ids = self.browse(filter_ids).filtered("is_system").ids
```
ThomasBinsfeld: @acsonefho A search with a domain filtering on ids and `is_system` would be more efficient.

ThomasBinsfeld: @acsonefho And the following lines could be replaced by: 

```python
return self.search([('id', 'in', filter_ids), ('is_system', '=', False)]).ids
```

pedrobaeza:
Hi, I would call it `base_filter_hidden`, and a best place for it it's `OCA/server-ux`.

system_filter/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
nikul-serpentcs, line 2: @acsonefho Remove Copyright from all `__init__.py` file.

system_filter/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "System Filter",
+    'description': """
+    Add possibility to have Filters used only by the system""",
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': "http://acsone.eu",
```
nikul-serpentcs, line 9: @acsonefho IMO Replace URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

nikul-serpentcs (commented):
Small Change

acsonefho:
@pedrobaeza I'll create a MR to put this module into `OCA/server-ux` (v11.0). But about the v10.0 (this MR), which repo do you suggest?
Thank you

pedrobaeza:
We can create 10.0 branch of OCA/server-ux if you want, or put it on OCA/server-tools if not.

acsonefho:
@pedrobaeza As it's a new module, I think the best/easier way is to create the 10.0 branch on OCA/server-ux

acsonefho:
When the 10.0 will be created on OCA/server-ux, I'll make a backward-port of this [MR](https://github.com/OCA/server-ux/pull/27).

pedrobaeza:
I have created 10.0 branch of OCA/server-ux.

PR #1057, opened 2018-10-02
[ADD] Added web_gesture module for swipe support on mobile devices

wpichler:


pedrobaeza:
Isn't this better to be integrated on `web_responsive` module?

PR #1059, opened 2018-10-02
Update README because of wrong license

damdam-s:
Found in this commit: https://github.com/OCA/web/commit/3965057913982b2e976d94490d027c747b2423a8


leemannd (approved):
Was wrongly setted at module migration to V11

tarteo (commented):
You can instead of fixing it in the current README, use the new splitted readme (https://github.com/OCA/maintainer-tools/tree/master/template/module) than it's also solved and the README will look alot better.

leemannd:
@tarteo He's been asked by an external guy what is the proper license, as there is another one in all other files.
IMO it should be fast-tracked to avoid any problems.

leemannd:
@gurneyalex Do you think it's mergeable?

PR #1060, opened 2018-10-02
[12.0][MIG] web_favicon

tarteo:


elicoidal (approved):
LGTM

web_favicon/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2015 Therp BV <http://therp.nl>
+# Copyright 2016 Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Custom shortcut icon",
+    "version": "12.0.1.0.0",
+    "author": "Therp BV, "
```
nikul-serpentcs, line 8: @tarteo Add website as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

nikul-serpentcs (commented):
Small Change

web_favicon/__manifest__.py
tarteo: Added it. Thanks :)

web_favicon/controllers/web_favicon.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import http
+from odoo.tools.misc import file_open
+
+
+class WebFavicon(http.Controller):
+
+    @http.route('/web_favicon/favicon', type='http', auth="none")
+    def icon(self):
+        request = http.request
+        if 'uid' in request.env.context:
```
simahawk, line 16: why not request.session.uid?

simahawk (approved):
LG just a question on controller implementation

web_favicon/controllers/web_favicon.py
nikul-serpentcs: Or request.env.uid

`If request.session.uid and request.env.uid`

web_favicon/static/src/img/master_original_favicon.svg
```diff
@@ -0,0 +1,56 @@
...
...583,0 16.2577,1.644702 23.85171,4.941729 7.5875,3.29051 14.18128,7.803012 19.76942,13.536809 5.58675,5.727838 10.02454,12.390081 13.32171,19.984093 3.29023,7.588056 4.94089,15.537841 4.94089,23.850746 -0.85928,16.90258 -6.87567,31.15898 -18.04779,42.76184 -11.17419,11.60314 -25.6454,17.97571 -43.40561,19.12329 l -0.43033,0 z M 104.43047,208 c 14.03437,0 27.28859,-2.79405 39.75226,-8.38083 12.46244,-5.58534 23.4211,-13.10687 32.87602,-22.5618 9.45355,-9.45492 16.97508,-20.48705 22.56183,-33.09085 C 205.20733,131.3558 208,117.89258 208,103.56911 208,89.239815 205.27385,75.85019 199.83679,63.387487 194.39,50.925336 187.01124,39.966662 177.70323,30.511325 168.38966,21.056678 157.431,13.602838 144.82719,8.1646355 132.21703,2.726419 118.61203,0 104.00014,0 89.671258,0 76.208041,2.5780813 63.603558,7.7347705 50.993109,12.892 39.96777,20.123798 30.512431,29.437215 21.058202,38.744259 13.605015,49.702934 8.1651761,62.313379 2.7204595,74.917448 0,88.668665 0,103.56911 c 0,14.32347 2.6462976,27.71892 7.9502436,40.1826 5.2990814,12.46381 12.6781754,23.34765 22.1335154,32.66119 9.453538,9.30801 20.553026,16.76025 33.305391,22.347 12.744188,5.58675 26.429298,8.66218 41.04132,9.2401" /></g></svg>
```
nikul-serpentcs, line 56: @tarteo add new blank line

nikul-serpentcs (commented):
Improve Code

web_favicon/static/src/img/master_original_favicon.svg
pedrobaeza: On SVG file? Any way, is this a custom icon? If not, remove standard one.

yajo (dismissed):
IMHO we should drop this module. It is supported in Odoo core now, per website:

peek 08-10-2018 13-03

tarteo:
@Yajo Without this it's only available if the website module is installed. 
Maybe we can make this only work for the backend by putting something like ```if not self.env.context.website_id:``` and by so let the website module do it for the frontend.
This way the user can change both web and website independently. What do you think?

pedrobaeza:
I think the favicon should be different on backend than on frontend. I'm not sure if this is happening right now or one of them is cached. But definitively the module is needed for cases when you don't have website installed. I'm curious what happen in the portal without website. Which favicon is used?

ghost:
in backend is the odoo-favicon used. and odoo12 has multi-website and multi-favicon.

tarteo:
Now it's only applied to the backend not the website.

pedrobaeza:
@tarteo do you know the answer for my question

> I'm curious what happen in the portal without website. Which favicon is used?

pedrobaeza:
(I'm talking with vanilla Odoo - not with this module)

tarteo:
@pedrobaeza Sorry for not answering.

Without website the default icon of Odoo:
```<link type="image/x-icon" rel="shortcut icon" t-att-href="x_icon or '/web/static/src/img/favicon.ico'"/>```

With website installed the icon of the website (/web/image/website/{id}/favicon/)

**Comment by pedrobaeza** [comment]
So maybe we should put also this icon for portal when no website? Or even having 2 selectable icons?

**Comment by tarteo** [comment]
@pedrobaeza This is already the case. If only portal and this module is installed, you can change the icon for the portal with this module:
```
            <t t-if="not website">
                <t t-set="x_icon" t-value="'/web_favicon/favicon'"/>
            </t>
```
The portal icon will be overridden when website is installed. The backend icon is still configured by this module.

pedrobaeza (approved):
I have added a little paragraph on the README about the portal thing and squash a bit the commit history. Merging when CIs finish.

ghost:
Maybe its different, if multi-website is activated?
My Odoo 12 use the standard odoo icon for portal and the website-specific icon (website-settings) for website

PR #1061, opened 2018-10-02
[12.0][MIG] web_disable_export_group

tarteo:
I also made a small change so on installation the admin and root user is automatically added to this group (the last commit).


pedrobaeza:
Seeing the name of the module, the group seems to be the contrary, but OK if it's well explained.

web_disable_export_group/templates/assets.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Onestein
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).-->
+<odoo>
+    <template id="assets_backend" name="web_disable_export_group assets" inherit_id="web.assets_backend">
+        <xpath expr="." position="inside">
+            <script type="text/javascript" src="/web_disable_export_group/static/src/js/disable_export_group.js"></script>
```
aitorbouzas, line 7: End the tag as ```<script.../>```?

aitorbouzas (commented):
LGTM

tarteo:
@pedrobaeza The name of the module is based on disabling exporting data because it is not restricted in standard Odoo. IMO users more often search (in the app store) for a module that disables the export functionality and therefore it's better to have this name. It's clearly mentioned in the README what this module does, so I don't see the problem in having this name.

web_disable_export_group/templates/assets.xml
tarteo: Old habits die hard. Is it really important? To me it isn't, because it doesn't improve the maintainability nor the functionality.

aitorbouzas: Of course it isn't. It's just more readable and a better habit. It's up to you to change it.

pedrobaeza:
I'm thinking that maybe this module should go in OCA/server-ux. What do you think?

web_disable_export_group/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+# Copyright 2018 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Web Disable Export Group',
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Onestein, '
+              'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    'website': 'http://www.github.com/OCA/web',
```
simahawk: https://github.com/OCA/web

simahawk (approved):
Just code review: LG. 
One minor non-blocking issue.
Regarding whether to move it to "server-ux", I'd say "maybe" ...if only that repo would be named "UX".
To me "server" + "UX" makes no sense at all :smile: 
But this is another story... :wink:

tarteo:
@pedrobaeza @simahawk Haha, server-ux is indeed funny. 
This module only adds a client side restriction, for this reason I think it's better here. If we're going to develop a server side restriction then it's better to be in server-ux.

astirpe (approved):
LGTM

pedrobaeza:
Squash migration commits and we are good to go

tarteo:
I'll make a simple test before merge.

tarteo:
Ok, maybe another time.

pedrobaeza:
Ah, sorry, I didn't see it... Is there any problem?

tarteo:
Just the coverage is not 100%.

pedrobaeza:
Ah, great, thanks for adding tests!

PR #1066, opened 2018-10-05
[MIG] web_responsive: Migrate to v12 and refactor

yajo:
This migration includes a full refactoring to make this module more
maintainable. Some things that have changed:

- Removed external libraries.
- Change Less for Scss.
- Reduce ES and XML to the minimal required needs.
- Implement as much features as possible with just Scss.
- Use the new hotkeys system from Odoo v12.

@Tecnativa


web_responsive/README.rst
```diff
@@ -0,0 +1,123 @@
...
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+This module is part of the `OCA/web <https://github.com/OCA/web/tree/11.0/web_responsive>`_ project on GitHub.
```
nikul-serpentcs: @Yajo IMO Change target branch `12.0`

web_responsive/README.rst
```diff
@@ -0,0 +1,123 @@
...
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Beta
+.. |badge2| image:: https://img.shields.io/badge/licence-LGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+    :alt: License: LGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fweb-lightgray.png?logo=github
+    :target: https://github.com/OCA/web/tree/11.0/web_responsive
```
nikul-serpentcs: same here

web_responsive/README.rst
```diff
@@ -0,0 +1,123 @@
...
+    :alt: OCA/web
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/web-11-0/web-11-0-web_responsive
```
nikul-serpentcs: here

web_responsive/README.rst
```diff
@@ -0,0 +1,123 @@
...
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/162/11.0
```
nikul-serpentcs: here

web_responsive/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2018  Alexandre Díaz
```
nikul-serpentcs: Remove Copyright from all `__init__` file

web_responsive/views/inherited_view_users_form_simple_modif.xml
```diff
@@ -0,0 +1,21 @@
...
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+
+<odoo>
+
+    <record id="view_users_form_simple_modif" model="ir.ui.view">
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form_simple_modif" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='email']" position="after">
```
nikul-serpentcs: IMO Directly use `field` here

nikul-serpentcs (commented):
Small Change

web_responsive/README.rst
yajo: The readme will be rebuilt automatically after merge by the bot, don't worry about these things.

FTR, if you see a `readme/` folder in an addon, that's the important part.

yajo:
@nikul-serpentcs all done :wink:

web_responsive/views/inherited_view_users_form_simple_modif.xml
yajo: I prefer to leave it as it is, since it works fine and it is perfectly OK. I changed enough things, no need to change everything :stuck_out_tongue_winking_eye:

nikul-serpentcs: @Yajo Yes, but if you Get `field` directly then use `<field>` first, if not get then it's Ok.

here, First Priorty use to direct `field` and if not get then use`xpath`.

chienandalu (changes requested):
Very cool!

I found a small issue in Chrome:

- When you open the apps box it's not possible to navigate with the keys. No response from the arrows and the tab kicks you out of the menu.

In the aesthetics side, wouldn't it be nice to have a colored background in the apps box?

yajo:
> When you open the apps box it's not possible to navigate with the keys. No response from the arrows and the tab kicks you out of the menu.

App navigation is not supported right now. Only navigating through results when you search. I'll add it to roadmap.

> In the aesthetics side, wouldn't it be nice to have a colored background in the apps box?

I suppose, but I wanted to focus on getting it functional. It can be beautier later :wink:. Adding to roadmap as well.

yajo:
Roadmap updated.

chienandalu:
@Yajo I mean the search results. It's not possible to select them with the keys in Chrome. It works perfectly in Firefox though.

yajo:
Oh sorry! Let me check that.

yajo:
It should work now. Problem: https://stackoverflow.com/a/11031754/1468388

chienandalu (approved):
Great work! :+1: :+1: :+1:

web_responsive/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import inherited_res_users
```
aitorbouzas: The file should me called just res_users as the [convention](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming) says.

aitorbouzas (commented):
Love to see this module migrated! Nice work, thanks for your effort.
Just a small note.

web_responsive/models/__init__.py
yajo: Done.

aitorbouzas: Nice, thank you!

web_responsive/static/src/js/web_responsive.js
```diff
@@ -0,0 +1,238 @@
...
+    var core = require("web.core");
+    var Menu = require("web.Menu");
+
+    /**
+     * Recursively search for menus with their full names
+     *
+     * @param {Object} menu
+     * @returns {Array} It contains many sublevel; you should flatten it
+     */
+    function findNames (menu) {
```
tarteo: Can you put this into the class or return it in web_responsive to expose the function.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -0,0 +1,238 @@
...
+
+    var AppsMenu = require("web.AppsMenu");
+    var config = require("web.config");
+    var core = require("web.core");
+    var Menu = require("web.Menu");
+
+    /**
+     * Recursively search for menus with their full names
+     *
+     * @param {Object} menu
```
tarteo, line 44: Can you put a description of what you're expecting, is it a record, or something from another source?

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -0,0 +1,280 @@
...
+            display: none;
+        }
+
+        // Fix toggler button padding
+        .o-menu-toggle {
+            cursor: pointer;
+            padding: 0 $o-horizontal-padding;
+        }
+
+        // Cusom fullscreen layout when showing submenus
```
tarteo: Custom* :grimacing:

tarteo (changes requested):
I love the search function because it searches on display_name.
I've also did a functional test. It seems some features are lost because there's no keyboard navigation to change the app selection anymore. I don't know if other features are lost too.

pedrobaeza (changes requested):
Functional review. Great improvement on the search!

Several points to discuss/solve:

- Arrows don't work on app switcher for selecting directly the app and press enter.
  There are 2 solutions for this:
  * Present the app icon also on the list of candidates, and remove all of them when you have typed anything
  * when you press the arrows on the text box, you navigate on the apps icons.
- Apps icons are moved down and top each time you type something on the search box. The first solution for the previous point will solve this also.
- On Firefox, switching to mobile view, it's not possible to search, as I can't type in the textbox. It's curious, as I have been able to do it on a real mobile, so this is not critical.
- As the search happens in another "screen", can you hide the searchbox on mobile view and let only the glass icon? This way, we gain a bit of space.
- We have lost the improvements made on v11 for collapsing statusbar and buttons. Please restore it:

  seleccion_005

  About the name of the button, I'm still not very convinced for calling it "Tasks". Enterprise responsive interface calls this "Action", but hides "Print" and  "Action" dropdown that are when full. I consider this a loss, as you can't for example remove a record from mobiles. Is it feasible to join the 3 of them (*Print* dropdown, *Action* dropdown and buttons) in the same mobile dropdown, split by a separator? Enterprise also increases the space for tapping each option and put text in upper for better recognition.
- On a notebook with pages that doesn't fit current width, full page appears side scrollable, which breaks mobile purpose. Enterprise makes scrollable in this case only the pages part instead of everything. Other technique can be to collapse extra pages that doesn't fit on mobile width (or put all of them as collapsable?)

  seleccion_006

  EDIT: Looking to other form views, it's an incorrect size of the notebook container, as here we don't have too many pages and they fit on the width:

  seleccion_008

- On mobile view, shadow and padding is preserved on forms with `sheet` attribute. We should gain this space:

  seleccion_007

@Tardo can you review this also?

Tardo (changes requested):
As said by others... good work with search feature!

Bad news:
- Can't use shortcuts, browser interpret it as "accelerator key" (ALT+<key>)
- Settings pages doesn't display properly
- Sticky feature has been lost
- You can move out of the "search list bounds" when use arrow keys
- Search-box dispatch events when press modifier keys
- I see some app names "cutted":
app_title

web_responsive/models/res_users.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    chatter_position = fields.Selection([
+        ('normal', 'Normal'),
+        ('sided', 'Sided'),
+    ], string="Chatter Position", default='normal')
+
```
bealdav, line 14: Nice to see this settings.

But don't you think it can have side effect with some module like 

https://github.com/OCA/web/tree/10.0/web_sheet_full_width_selective

Do you think it possible whatever user setting here, the chatter can go normal position if not enough width space ?

May be it works as it ?

Thanks s lot

pedrobaeza:
Another pending thing is the inclusion of web_searchbar_full_width module in this one, which is a simply CSS rule.

osoul-dev:
I found an issue with RTL direction where the app box is not visible. This is due to hardcoding the left position to be zero at the element level. If it was in css, then I believe rtlcss library that odoo is using will take care of flipping the layout.

This is happening in both desktop and mobile screens. 
screen shot 2018-10-27 at 1 08 22 pm

pedrobaeza:
@osoul this doesn't belong here, but on Odoo itself.

osoul-dev:
@pedrobaeza the RTL issue is related to this pull request. The issue is with web_responsive app box as odoo's app dropdown menu shows fine in RTL

screen shot 2018-10-27 at 1 44 36 pm

pedrobaeza:
You are mixing things, as the screenshot doesn't match with colors and app switcher that this pull request is adding.

rafaelbn:
Maybe we can discard this module as has a high maintenance an start using

- https://github.com/Openworx/backend_theme/issues/22

- https://github.com/Openworx/backend_theme/tree/12.0/openworx_backend_theme

It is already done and supported by Openworx

pedrobaeza:
AFAIK that one is not responsive. And IMO is ugly. It doesn't include menu search nor app switcher (only a variant of the navbar). And finally, it's not OCA.

mgielissen:
> 
> 
> AFAIK that one is not responsive. And IMO is ugly. It doesn't include menu search nor app switcher (only a variant of the navbar). And finally, it's not OCA.

The first version only has the app sidebar and more enterprise look'n'feel. The second version will using web_responsive when it's available.

web_responsive/static/src/js/web_responsive.js
yajo: It is not needed because this is just a helper to generate `this._searchableMenus`, which is the same data that we already had, but in a format usable by `fuzzy.js`. I'll fix docstrings to make this clearer.

If you need to override something, it should be usually `AppsMenu.init()`, where this is used.

web_responsive/models/res_users.py
yajo: That's how it works actually.

yajo:
> there's no keyboard navigation to change the app selection anymore

(and other similar concerns).

This is on purpose. The code to achieve that was awful, and to maintain that is all but easy.

One of my goals in this refactoring is to make this sensible and smooth again. The fact is that you don't need that, because you can simply write the app name, and the new fuzzy searcher will present that as the first candidate. You end up being faster this way, as you can use 10 fingers instead of one :stuck_out_tongue_winking_eye: and it's way easier to maintain.

yajo:
> Apps icons are moved down and top each time you type something on the search box

This is fixed now. App icons disappear when searching, and appear again when not, so we avoid confusion.

Still WIP in other stuff.

pedrobaeza:
> This is fixed now. App icons disappear when searching, and appear again when not, so we avoid confusion.

And if you look for the app name the icon appears the first the same as in enterprise? That's important to be preserved, and more if you have removed the possibility of navigating with the arrows.

yajo:
Not sure if I understand correctly, but the idea I'm implementing is that you can choose that item with its name, not the icon:

peek 12-11-2018 09-16

When you start a search, icons disappear. You can still choose the "Discuss" app, but through its listed name. The result and speed for the user is the same, or more, and the code is much simpler.

It still has bugs though.

Were you talking about this?

pedrobaeza:
The icon is a way to highlight that the entry is a main app, so it's better to have it. I think it's not too much complicated, as you have an attribute for the icon linked to the menu, so you only have to adapt CSS and show conditionally that icon if exists.

pedrobaeza:
Another important thing is that you can look for the complete menu path (I don't know if you have taken into account this already).

yajo:
Yes, that's the way it's searched.

I think current code with last commit is simple and still have the app icon indicator:

peek 12-11-2018 12-55

BTW you can notice that scrolling is synced and looped fine. :blush:

pedrobaeza:
:heart_eyes:

yajo:
Buttons folded now, still WIP on other stuff:

peek 13-11-2018 11-16

yajo:
> I see some app names "cutted":

@Tardo this is Odoo problem. Fix: https://github.com/odoo/odoo/pull/28628

yajo:
This is ready to review now.

Features for all devices:

* New navigation with an app drawer
peek 13-11-2018 13-21

* Quick menu search from the app drawer
peek 13-11-2018 13-23


Features for mobile:

* App-specific submenus are shown on full screen when toggling them from the
  "hamburger" menu 
peek 13-11-2018 13-36

* View type picker dropdown displays confortably
peek 13-11-2018 13-37

* Top app bar is always visible
* Top control panel is hidden when scrolling down
top bar visible

* Form status bar action buttons are collapsed in a dropdown
captura de pantalla de 2018-11-13 13-48-00

* Form status bar record state hides all but current state
statusbar


Features for computers:

* Keyboard shortcuts for easier navigation, visible when pressing `alt`
captura de pantalla de 2018-11-13 13-49-24

* Autofocus on search menu box when opening the drawer
* Set chatter on the side of the screen, optional per user
peek 13-11-2018 13-34

* Full width form sheets
captura de pantalla de 2018-11-13 13-33-17

eLBati:
@Yajo thanks for this great work.

A note about the `discuss` page: messages within a channel

image

disappear in small screens

image

pedrobaeza:
@eLBati you have missed to reload page. Take into account that part of the responsive part is attached to JS, so when you have resized your window, you have to reload to let JS to detect that part (and that can't be done in a resize event). On a mobile, you won't have that problem as it starts small from the beginning :wink:

pedrobaeza (changes requested):
Excellent work you are making with this. I think you can put in the README the list of the features you have collected in your last comment, even with the GIFs for enriching the description.

Several things still remains that are important IMO to be improved for a full responsive experience:

1. The search box on mobile should be hidden in favor of only a glass icon, as any way no active filters are shown directly on the view, and when you click on the icon it goes to another screen:

   seleccion_005

2. Create button and pager controls are not on the same row, wasting some precious vertical space:

   seleccion_006

3. When we click through records, breadcrumb bar keeps growing, filling more and more vertical space:

   seleccion_007

   What enterprise has made is to only show latest record name_get and put a left arrow for going back. Another option is to put breadcrumb in only one line and make it side scrollable.

4. On form view with print and action dropdown, the vertical space used is still large. I would like to include print and action options inside the same "Quick actions" dropdown, putting a little separator with the names of each section: Print, Action, Buttons. And then you can put the pager in the same line as "Create" button:

   seleccion_008

   In this case, we can rename again the dropdown to simply "Actions" for avoiding very large translations that, joined with a large statusbar value, makes both to not fit in a line.

eLBati:
@pedrobaeza thanks!

yajo:
Thanks for the reviews.

The points you say, @pedrobaeza, while true, are not so simple to fix as it might seem at first sight. I'll try though.

Point 4, however, is almost impossible. Those are completely separated components and views, it's gonna be a lot of code to put it all together in the same place.

WIP on the others.

pedrobaeza:
Do you see any other option for point 4? The goal is to reduce the header height.

pedrobaeza:
Maybe replace texts by icons (the printer icon is easy, and the other one can be the magic wand), and reduce the margin between create/edit button and these buttons, so everything fits in one line.

yajo:
I can try, maybe replacing texts for icons is the easiest way, but still the CSS is gonna be hacky.

mohshedid:
Thanks for your effort 
I note the following issue at sub menu 

image

yajo:
What's the problem, @mohshedid? :confused:

mohshedid:
> What's the problem, @mohshedid? 😕

it's display as a  popup , below there is a  screen shot for expected behavior 

whatsapp image 2018-11-20 at 11 20 18 am

yajo:
You probably misunderstood this. I'm not trying to clone `web_enterprise`; I'm just trying to make our own version, that works fine and makes it usable in phones.

Being a popup is still quite usable, so that's not a problem. Or did you find a usability issue?

mohshedid:
> You probably misunderstood this. I'm not trying to clone `web_enterprise`; I'm just trying to make our own version, that works fine and makes it usable in phones.
> 
> Being a popup is still quite usable, so that's not a problem. Or did you find a usability issue?

I got your point , sorry for misunderstand

pedrobaeza:
@Yajo what is the status of this?

yajo:
Under a pile of work. :roll_eyes:

pedrobaeza:
You might need to modify something when https://github.com/odoo/odoo/pull/29994 is merged.

yajo:
I fixed all concerns. Some new screenshots/gifs:

- View type switcher on mobiles: 
peek 10-01-2019 10-58

- The control panel is hidden when scrolling, to save screen space, but the main top bar is always visible: 
peek 10-01-2019 11-02

- Form status bar action buttons are collapsed in a dropdown.
  Other control panel buttons use icons to save space.
peek 10-01-2019 11-21


- Breadcrumbs navigation is collapsed with a "back arrow" button. 
peek 10-01-2019 11-15

All these gifs appear in the updated README now.

This is ready to review.

pedrobaeza (approved):
All the comments have been attended, so I'm going to merge this after a long incubation phase.

@Yajo full width searchbar is still missing, but we can add it later on a new PR.

steync027:
Great work!

mgielissen:
When will it be published in the Odoo Apstore?

PR #1071, opened 2018-10-10
[12.0][MIG] web_notify

aitorbouzas:
Migrated and tested. Minor changes in JS.

Messed up the commit history in #1070 
CC for help with JS/code review: @lmignon @JayVora-SerpentCS


lmignon (approved):
LGTM (Code review only)

pedrobaeza:
Maybe we can add as demo data a button (or 2, one for each kind of notification) on user form for launching a default notification, which can be done using the message as keyword argument, and this way, we can self-test the module. What do you think?

aitorbouzas:
@pedrobaeza Awesome. Maybe we can show the buttons only when debug mode is active?

pedrobaeza:
Well, if we load them as demo, that extra layer maybe is not needed, but it's up to you to decide which one is best.

web_notify/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/web',
+    'depends': [
+        'web',
+        'bus',
+    ],
+    'data': [
+        'views/web_notify.xml'
+    ],
+    'demo': [
```
nikul-serpentcs, line 23: @aitorbouzas IMO Remove `demo`, If not added any `value`

web_notify/views/web_notify.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
```
nikul-serpentcs: remove  ` <data>`, It's not need.

nikul-serpentcs (commented):
Minor Change

web_notify/__manifest__.py
aitorbouzas: Added some demo now, so this stays :)

aitorbouzas:
@pedrobaeza Added the buttons and updated USAGE.rst to show how to test it.
There's only one little problem, the ```message``` keyword argument shows the context when using the buttons, why is the context parsed as message?

pedrobaeza:
Uhm, maybe there's an argument mix, but I'm testing on runbot and can't get the notification. I think it's better to put the buttons on user form, not user preferences. Being a popup can also waste a bit.

aitorbouzas:
I'm getting the notifications correctly in my local environment... That's strange, can I check runbot's Odoo log?
Let's see if it works now that I moved the buttons to users form.

aitorbouzas:
Ok, I think I know the reason why it fails, could it be due to have more than 1 worker in runbot?

web_notify/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
JayVora-SerpentCS, line 3: good to add:
# Copyright 2016 ACSONE SA/NV

JayVora-SerpentCS (changes requested):

web_notify/__init__.py
aitorbouzas: Thanks Jay, Copyright should not be filled in ```__init__.py``` [(CHECK HERE)](https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py) but I'll add the license

web_notify/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+                and any(user.id != self.env.uid for user in self)):
+            raise exceptions.UserError(
+                _('Sending a notification to another user is forbidden.')
+            )
+        bus_message = {
+            'message': message,
+            'title': title,
+            'sticky': sticky
+        }
+        notifications = [(getattr(record, channel_name_field), bus_message)
```
simahawk: don't use `getattr`, you can access fields in dict-like notation: `(record[channel_name_field], bus_message)`

web_notify/models/res_users.py
```diff
@@ -0,0 +1,46 @@
...
+            'notify_info_channel_name', message, title, sticky)
+
+    def notify_warning(self, message="Default message",
+                       title=None, sticky=False):
+        title = title or _('Warning')
+        self._notify_channel(
+            'notify_warning_channel_name', message, title, sticky)
+
+    def _notify_channel(self, channel_name_field, message, title, sticky):
+        if (self.env.uid != SUPERUSER_ID
```
simahawk: I know this a recent patch but: wouldn't be better to check for a group or to use `self.env.user._is_admin()`?
/cc @hbrunn

simahawk (changes requested):
LG overall, small remarks

web_notify/models/res_users.py
hbrunn: `_is_admin`

web_notify/models/res_users.py
aitorbouzas: Thanks! It could be also a security problem. Bad habit to use ```getattr```...

web_notify/models/res_users.py
aitorbouzas: :+1:

simahawk:
@aitorbouzas seems good: pls, squash your commits :)

aitorbouzas:
@simahawk Squashed!

simahawk:
@aitorbouzas do you really want to keep "Add buttons to users form"? If yes, you must split it as it contains things not related to it.
NOTE: next time when you use "--fixup" you can pass the commit to fixup like:
```bash
git commit --fixup 1234
git log
1234 A
4323 B
2343 fixup! A
```
then when you rebase w/ auto-squash (ie: `git rebase -i --autosquash HEAD~3`) you'll get them sorted properly.
In your case I'd expect to have a log like:
```bash
xxx web_notify: migrate to v11
yyy web_notify: add buttons to users form
zzz web_notify: improve security check

Do not rely on SUPERUSER_ID:
delegate higher permisions check to `user._is_admin()`.

Improves https://github.com/OCA/web/pull/1071/commits/ae8e4ec59deea6063bd912d6be0f62e4e5a8bc3b
```
kind of... :wink: 
NOTE2: you can reword commits in the rebase interactive session

aitorbouzas:
@simahawk resquashed commits and reworded some. How's it now?

simahawk (approved):

simahawk:
@JayVora-SerpentCS ok for you?

simahawk:
well, seem so, his comment is attended anyway.
@aitorbouzas thanks for your work!

aitorbouzas:
@simahawk Thank you for the reviews!

JayVora-SerpentCS:
Does it also solve the known issue of it with workers ?

aitorbouzas:
@JayVora-SerpentCS it doesn't, anyway I don't know if workers are the reason of the problem. It was just a supposition. Has this happened to you too?

JayVora-SerpentCS:
Happened once. I have @serpentcs-dev1 check it once.

PR #1079, opened 2018-10-15
[FIX] web_widget_char_switchcase: check input value

SimoRubi:
Before this PR:
When field is empty (false), the following error happens: 
`TypeError: val.toUpperCase is not a function`

### DIFF
--- a/web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
+++ b/web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
@@ -7,7 +7,7 @@ odoo.define('web_widget_char_switchcase', function (require) {
 
     form_widgets.FieldChar.include({
         events: _.extend({}, form_widgets.FieldChar.prototype.events, {
-            'keyup': 'key_up_handler',
+            'keyup': '_onKeyUp',
         }),
         transformations: ['default', 'upper', 'lower', 'title', 'sentence', 'camel', 'pascal', 'snake'],
         init: function (field_manager, node) {
@@ -20,11 +20,10 @@ odoo.define('web_widget_char_switchcase', function (require) {
             if (!_.contains(this.transformations, this.current_transformation))
                 console.error(this.current_transformation + ' case unknown');
         },
-        key_up_handler: function ($event) {
-            var $input = $($event.target);
-            var old_val = $input.val();
+        _onKeyUp: function (event) {
+            var old_val = this.$input.val();
             var new_val = this.current_transformation_handler(old_val);
-            $input.val(new_val);
+            this.$input.val(new_val);
         },
         parse_value: function (val, def) {
             return this._super(this.current_transformation_handler(val), def);
@@ -32,22 +31,25 @@ odoo.define('web_widget_char_switchcase', function (require) {
         format_value: function (val, def) {
             return this._super(this.current_transformation_handler(val), def);
         },
+        check_val: function (val) {
+            return val && val.trim();
+        },
         get_transformation_handler: function () {
             switch (this.current_transformation) {
                 case 'upper':
                     // HELLO WORLD!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         return val.toUpperCase();};
                 case 'lower':
                     // hello world!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         return val.toLowerCase();};
                 case 'title':
                     // Hello World!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         return val.replace(
                             /\w\S*/g,
                             function(txt) {
@@ -58,7 +60,7 @@ odoo.define('web_widget_char_switchcase', function (require) {
                 case 'sentence':
                     // Hello world!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         var first = true;
                         return val.replace(
                             /\w\S*/g,
@@ -75,7 +77,7 @@ odoo.define('web_widget_char_switchcase', function (require) {
                 case 'camel':
                     // helloWorld!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         var first = true;
                         return val.replace(
                             /\w\S*/g,
@@ -92,7 +94,7 @@ odoo.define('web_widget_char_switchcase', function (require) {
                 case 'pascal':
                     // HelloWorld!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         return val.replace(
                             /\w\S*/g,
                             function(txt) {
@@ -103,7 +105,7 @@ odoo.define('web_widget_char_switchcase', function (require) {
                 case 'snake':
                     // hello_world!
                     return function (val) {
-                        if (!val) return val;
+                        if (!this.check_val(val)) return val;
                         return val.toLowerCase().replace(' ', '_');
                         };
                 case 'default':
--- a/web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
+++ b/web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
@@ -6,6 +6,9 @@ odoo.define('web_widget_char_switchcase', function (require) {
     var form_widgets = require('web.form_widgets');
 
     form_widgets.FieldChar.include({
+        events: _.extend({}, form_widgets.FieldChar.prototype.events, {
+            'keyup': 'key_up_handler',
+        }),
         transformations: ['default', 'upper', 'lower', 'title', 'sentence', 'camel', 'pascal', 'snake'],
         init: function (field_manager, node) {
             this._super(field_manager, node);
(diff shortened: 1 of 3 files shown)


astirpe (approved):
Thanks!

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
```diff
@@ -41,13 +41,18 @@ odoo.define('web_widget_char_switchcase', function (require) {
             switch (this.current_transformation) {
                 case 'upper':
                     // HELLO WORLD!
-                    return function (val) { return val.toUpperCase(); };
+                    return function (val) {
```
simahawk, line 41: @SimoRubi In my example in former PR I've put this before the call to the transformer, exactly for this reason -> do not run the transform if there's nothing to transform.... I'd move this check before the call and leave this func as it is

simahawk: if you want to keep it here: move it before the switch

simahawk (changes requested):

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
SimoRubi: @simahawk the check is there, inside the *keyup_handler*. Where would you put that too? I the `init` function?

simahawk: then how do you hit the case of calling the transformer w/out a value? :)

SimoRubi: @simahawk I don't know if in the `init` I already have a value in the input field, actually I think it is not there yet, that's why I didn't check it in there.

In the `initialize_content` I might have the value already (after the `super` call), but the transformation_handler is already assigned and I think that's correct because the transformation_handler should not depend on the content but only on the option.

I could put the check in the `parse_value` and `format_value`, but I thought that checking the value can be something that the transformation_handler should manage by itself.


Then I've run out of functions! That's why the only solution I came up with is putting the check in each transformation_handler.
Where do you suggest to put the check?

simahawk: then how don't get how the module is supposed to work properly IF [this keyup event](https://github.com/OCA/web/blob/50e7cf588ea6bf8676dd0a0fc0b644f54d52477c/web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js#L24) is not triggered EVERY time you type.
If this is triggered properly then the handler should never be called. :thinking:

Ahhhhh now I see: you are binding the event only if you have a value at initialize content (and we did not catch that in reviews :cry:).
I think you should delegate the binding to `events` mapping. Ideally I'd say you should add your own event to internal events (by extension) but it seems [there's already one that should do what we need](https://github.com/odoo/odoo/blob/1868713dbd07e0b518f91dffe73e62d85e6ab9a6/addons/web/static/src/js/fields/basic_fields.js#L157).
As you see, it says:
```js
/**
     * Called when the user is typing text -> By default this only calls a
     * debounced method to notify the outside world of the changes.
     * @see _doDebouncedAction
     *
     * @private
     */
    _onInput: function () {
        this.isDirty = true;
        this._doDebouncedAction();
    },
```
Hence, I'd say you should override it, do your stuff (what you do now inside the keyup callback) and call super.
**UPDATE: this will work in v11 only, see below!**

simahawk: dammit, this would work in v11 (I was grepping in the wrong branch :stuck_out_tongue:). Then, the solution is:
```js
    events: _.extend({}, form_widgets.FieldChar.prototype.events, {
        'keyup': '_onKeyUp',
    },
    [...]
    _onKeyUp: function () {  ...your callback code here... }
```
This should do the trick.

SimoRubi: @simahawk thanks for your thorough analysis (in v11 too :) )!
Even if seems to me that the keyup event is already triggered everytime I type in the input field, your `event` extension is much more clean than my binding so I'll try to use that :)

SimoRubi: @simahawk used the new event binding in https://github.com/OCA/web/pull/1079/commits/9602babdb9809b0d68f2dc9877d7ea04e15079fe

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
```diff
@@ -6,6 +6,9 @@ odoo.define('web_widget_char_switchcase', function (require) {
     var form_widgets = require('web.form_widgets');
 
     form_widgets.FieldChar.include({
+        events: _.extend({}, form_widgets.FieldChar.prototype.events, {
+            'keyup': 'key_up_handler',
```
simahawk: good convention: `_on$EVENT_NAME` (w/ or w/out `_`)

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
```diff
@@ -17,19 +20,11 @@ odoo.define('web_widget_char_switchcase', function (require) {
...
-                    var old_val = self.$input.val();
-                    if (!old_val)
-                        return;
-                    var new_val = self.current_transformation_handler(old_val);
-                    self.$input.val(new_val);
-                });
-            }
-            return res;
+        key_up_handler: function ($event) {
+            var $input = $($event.target);
```
simahawk: don't you have `this.$input.val()`?
And you should do the check here on `if this.$input.val().trim()` and call the transform only if satisfied.
also, is not `$event` but `event` as is not a jq object

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
```diff
@@ -41,13 +36,18 @@ odoo.define('web_widget_char_switchcase', function (require) {
...
                     // HELLO WORLD!
-                    return function (val) { return val.toUpperCase(); };
+                    return function (val) {
+                        if (!val) return val;
+                        return val.toUpperCase();};
                 case 'lower':
                     // hello world!
-                    return function (val) { return val.toLowerCase(); };
+                    return function (val) {
+                        if (!val) return val;
```
simahawk: you remove all these checks now

web_widget_char_switchcase/static/test/web_widget_char_switchcase.js
```diff
@@ -3,88 +3,87 @@ odoo.define_section('web_widget_char_switchcase', ['web.form_common', 'web.core'
...
     test('Default does nothing', function(assert, form_common, core, web_form_widgets) {
-        this.field = createField(form_common, web_form_widgets, {'attrs': {}});
+        var fieldWidget = createField(form_common, web_form_widgets, {'attrs': {}});
 
         var orig_val = 'Hello World!';
-        this.field.$input.val(orig_val);
-        this.field.$input.keyup();
-        assert.strictEqual(this.field.$input.val(), orig_val);
+        fieldWidget.$input.val(orig_val);
+        fieldWidget.$input.keyup();
```
simahawk: fieldWidget.$input.trigger('keyup')

simahawk (changes requested):

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
SimoRubi: @simahawk If I remove these checks, `parse_value` and `format_value` might fail, and I think the *emptiness* check should be performed by the handlers.
I know repeated code is a bad thing, but maybe I can create a method 
```
check_val: function (val) {
    return val; // Check that it is not empty
}
```
even though it is so simple that turns out to be almost silly.
Then I'd use `check_val` in all the possible handlers

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
SimoRubi: Done in https://github.com/OCA/web/pull/1079/commits/63e2623b69d22a27b224fe80586e2fef4a9ac99e

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
SimoRubi: Yes, I realized that I had the correct `this` after writing that.
Done in https://github.com/OCA/web/pull/1079/commits/63e2623b69d22a27b224fe80586e2fef4a9ac99e

web_widget_char_switchcase/static/test/web_widget_char_switchcase.js
SimoRubi: Done in https://github.com/OCA/web/pull/1079/commits/63e2623b69d22a27b224fe80586e2fef4a9ac99e

web_widget_char_switchcase/static/src/js/web_widget_char_switchcase.js
SimoRubi: Anyway, added the `check_val` in https://github.com/OCA/web/pull/1079/commits/63e2623b69d22a27b224fe80586e2fef4a9ac99e

simahawk (approved):

PR #1080, opened 2018-10-15
[FIX] web_notify README

aitorbouzas:
@simahawk The readme contained some old instructions


web_notify/README.rst
```diff
@@ -64,7 +64,7 @@ or
    :alt: Sample notifications
 
 You can test the behaviour of the notifications by installing this module in a demo database.
-   Click in your user logo in the top right corner of the screen then click "Preferences".
```
simahawk, line 67: this is duplicated below, keep instructions in USAGE only

simahawk (changes requested):
also, pls, try to include module name in PR title and commit message ;)

web_notify/README.rst
aitorbouzas: What's duplicated? README.rst is the result of the new generator that takes content from ```readme/``` folder...

simahawk: hmmm sorry, I mis-read the path above the content! :stuck_out_tongue_closed_eyes:

simahawk (approved):

pedrobaeza (approved):

PR #1082, opened 2018-10-17
[11.0] web_responsive : Support sticky table headers on chrome

grindtildeath:
FF 62.0.3 (64-bit) on ubuntu 16.04 before :
ff_before

FF 62.0.3 (64-bit) on ubuntu 16.04 after :
ff_after

Chrome 70.0.3538.67 (64-bit) on ubuntu 16.04 before :
chrome_before

Chrome 70.0.3538.67 (64-bit) on ubuntu 16.04 after :
chrome_after

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -8,7 +8,7 @@
                "web",
     "version": "11.0.2.0.1",
     "category": "Website",
-    "website": "https://laslabs.com/",
+    "website": "https://github.com/OCA/web",
     "author": "LasLabs, Tecnativa, Alexandre Díaz, "
               "Odoo Community Association (OCA)",
     "license": "LGPL-3",
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -6,7 +6,7 @@
     "name": "Web Responsive",
     "summary": "It provides a mobile compliant interface for Odoo Community "
                "web",
-    "version": "11.0.2.0.0",
+    "version": "11.0.2.0.1",
     "category": "Website",
     "website": "https://laslabs.com/",
     "author": "LasLabs, Tecnativa, Alexandre Díaz, "
--- a/web_responsive/static/src/less/form_view.less
+++ b/web_responsive/static/src/less/form_view.less
@@ -57,13 +57,16 @@
     >div {
         >.table-responsive {
             >.o_list_view {
-                thead {
+                thead tr:nth-child(1) th {
                     position: sticky;
                     top: 0;
+                    background-color: #EEEEEE;
+                    border: none;
                 }
-                tfoot {
+                tfoot tr:nth-child(1) td {
                     position: sticky;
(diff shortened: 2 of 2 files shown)


simahawk:
tested on latest chrome on windows too ;)
@grindtildeath FTR you could mention here caniuse.com link and stackoverflow workaround

web_responsive/static/src/less/form_view.less
```diff
@@ -57,13 +57,16 @@
     >div {
         >.table-responsive {
             >.o_list_view {
-                thead {
+                thead tr:nth-child(1) th {
                     position: sticky;
                     top: 0;
+                    background-color: #EEEEEE;
```
simahawk, line 63: isn't there any variable for this color?

web_responsive/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     "name": "Web Responsive",
     "summary": "It provides a mobile compliant interface for Odoo Community "
                "web",
-    "version": "11.0.2.0.0",
+    "version": "11.0.2.0.1",
     "category": "Website",
     "website": "https://laslabs.com/",
```
simahawk: not related to this change, still: https://github.com/OCA/web

simahawk (changes requested):

web_responsive/static/src/less/form_view.less
grindtildeath: What do you mean ?

simahawk: `@odoo-color-for-thead` or similar -> likely there's a variable to reuse and you don't hardcode this color

grindtildeath: https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/less/list_view_extra.less#L4 :man_shrugging:

simahawk: amazing :rofl:

simahawk (approved):

tarteo (changes requested):
Is it possible to keep the border on the table header? Now it's all one color gray.

grindtildeath:
@tarteo I tried, but it unfortunately seems not possible and that's why I had to set it to 0.
The border would appear as long as you didn't scroll down yet, but when scrolling the border will disappear and there will be a 1px transparent border between the sticky header cells.
From what I tested, to have the border, we would have to switch `border-collapse` property which is set to `collapse` for `table` on bootstrap (https://github.com/odoo/odoo/blob/11.0/addons/web/static/lib/bootstrap/less/normalize.less#L417), but I couldn't get a good rendering.
Therefore, I dropped the border as IMO usability (not having to scroll up in a big list to know what you're looking at) is more important than esthetics.
Maybe there's another solution to keep it, but I couldn't find any.

tarteo (approved):

pedrobaeza:
@Yajo check this for the v12 migration

PR #1041, opened 2018-09-19
[IMP] Change display of m2o link

apollet:
Change display of m2o link in order to not have a full clickable link but just an icon next to it.
The icon shows up only when cursor is on the value.

It prevents from opening the many2one form by accident.
m2o_clickable

PR #1088, opened 2018-10-22
[10.0] [IMP] web_tree_many2one_clickable. Default false.

NL66278:
Having this module not set all Many2one fields clickable by default, greatly increases the
possibility to reuse the functionality in other OCA modules.

For instance I would like to make partner_multi_relation more user friendly, with the possibility to click on the partner fields in a relation list view to go directly to the partner. The widget is ideal for this, but I can not/should not force all users of partner_multi_relation to have all of their Many2one fields clickable.

I think the same would go for other OCA modules that would like to use these widgets.

### DIFF
--- a/web_tree_many2one_clickable/data/ir_config_parameter.xml
+++ b/web_tree_many2one_clickable/data/ir_config_parameter.xml
@@ -6,7 +6,7 @@
 
 <record id="default" model="ir.config_parameter">
     <field name="key">web_tree_many2one_clickable.default</field>
-    <field name="value">true</field>
+    <field name="value">false</field>
 </record>
 
 </odoo>


pedrobaeza:
I prefer to have it to True, as I install it for having that behavior. What we should do is to be less invasive with something like #1041

NL66278:
@pedrobaeza I agree that the less invasive method would certainly help, but this PR still has issues and is not merged yet.

My point is that if you specifically (and presumably manually) install this module, it is easy to set the default to True. But if the module is installed as a prerequisite for other modules, in order to use the many2one_clickable widget where appropriate, many users might be very surprised that all their views suddenly changed their appearance and workings.

pedrobaeza:
Well, I prefer to stay this way, as it's the same to you to disable after installing whatever, but let's see what others say.

tarteo:
> My point is that if you specifically (and presumably manually) install this module, it is easy to set the default to True.

It's also easy to set it to False after installation. I expect when I install this module it to be active by default.

simahawk:
me too :)
FTR this setting is totally ignored in v11

NL66278:
It is not about the manual installation of the module. Of course I understand it is as easy to set the value to False as it is to set it to True. But you can NOT use the module AS IS as a prerequisite for other modules, to use the many2one_clickable widget, because that would change all views in the whole system AS A SIDE EFFECT.

So if the maintainers of this module do not want to change the default - and as maintainers of course it is up to them - I will propose  to split the functionality of this module in two parts. A base web_tree_many2one_clickable_widget module to contain the widget and the parameter, and the original module, that will depend on the base module but only contains code to set the parameter to True (on installation).

Would this be an acceptable solution?

simahawk:
@NL66278 if the issue is to disable/enable this on demand per model or per context I'd suggest you to propose an improvement in this direction. Make the widget read a param (from ctx or config or model attr) to turn on/off the feature.

PR #873, opened 2018-02-21
[FIX] web_responsive : fixed the issue of fullscreen

JayVora-SerpentCS:
Fixes #859

PR #1089, opened 2018-10-23
[11] Backport web_responsive fix #873 to v11

Ark74:
Adding patch of #873 to v11
Cheers!

### DIFF
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -46,23 +46,15 @@ main {
 }
 
 .o_web_client {
-        &.o_fullscreen {
-        #oe_main_menu_navbar {
-            display: none;
-        }
-        .o_main .o_sub_menu {
-            display: none;
-         }
-        .navbar {
-            display: none;
-        }
-        .navbar-default main-nav{
-            display: none
-        }
+    &.o_fullscreen {
+        #oe_main_menu_navbar,
+        .o_main .o_sub_menu,
+        .navbar,
+        .navbar-default .main-nav,
         .main-nav{
             display: none
-        }
-      }
+         }
+    }
 
     >.o_main {
         overflow: auto;
--- a/web_responsive/static/src/less/main.less
+++ b/web_responsive/static/src/less/main.less
@@ -46,6 +46,24 @@ main {
 }
 
 .o_web_client {
+        &.o_fullscreen {
+        #oe_main_menu_navbar {
+            display: none;
+        }
+        .o_main .o_sub_menu {
+            display: none;
+         }
+        .navbar {
+            display: none;
+        }
+        .navbar-default main-nav{
+            display: none
+        }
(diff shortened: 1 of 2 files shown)


web_responsive/static/src/less/main.less
```diff
@@ -46,6 +46,24 @@ main {
 }
 
 .o_web_client {
+        &.o_fullscreen {
```
tarteo: Can you fit the identation, please?

web_responsive/static/src/less/main.less
```diff
@@ -46,6 +46,24 @@ main {
...
+        #oe_main_menu_navbar {
+            display: none;
+        }
+        .o_main .o_sub_menu {
+            display: none;
+         }
+        .navbar {
+            display: none;
+        }
+        .navbar-default main-nav{
```
tarteo: This doesn't seem right. ```.navbar-default .main-nav```?

web_responsive/static/src/less/main.less
```diff
@@ -46,6 +46,24 @@ main {
...
+            display: none
+        }
+        .main-nav{
```
tarteo, line 53: Why not use multiple selectors once, like this:
```
#oe_main_menu_navbar,
.o_main .o_sub_menu,
.navbar,
.main-nav {
     display: none;
}
```

tarteo (changes requested):
Thanks!

web_responsive/static/src/less/main.less
Ark74: Sure,let me check that.

web_responsive/static/src/less/main.less
Ark74: Ohh! You're right.

web_responsive/static/src/less/main.less
Ark74: Ohh, nice, haven't thought of it.
I'm not too familiar with less, I think of it like css maybe there lays the issue.

Ark74: Should `.navbar-default .main-nav` be removed?

tarteo: It did nothing before but we can keep it so the intention of the original developer is kept. :P

tarteo (approved):

pedrobaeza (approved):
@Yajo please review and put it on v12

web_responsive/static/src/less/main.less
yajo: Remove it please. We don't need uneeded code :wink:

yajo (changes requested):

web_responsive/static/src/less/main.less
Ark74: Done

yajo (approved):

PR #1091, opened 2018-10-24
Add new module web_tree_background_color

LeartS:


web_tree_background_color/static/src/js/list_view.js
```diff
@@ -0,0 +1,53 @@
+/* Copyright 2018 Leonardo Donelli (donelli@webmonks.it)
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+odoo.define('web_tree_background_color.stuff', function (require) {
```
simahawk, line 4: s/stuff/list_view

web_tree_background_color/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module allows you to dynamically set the **background** color
+of list view rows, according to python expressions; similarly to the
+`standard odoo feature to set the text color`_.
+
+.. _standard odoo feature to set the text color: https://www.odoo.com/documentation/10.0/reference/views.html#lists
```
simahawk, line 5: worth to mention that the module simply adds extra classes, so if you want to use custom colors you can define your own class + LESS file to style it.

web_tree_background_color/static/src/js/list_view.js
```diff
@@ -0,0 +1,53 @@
...
+            var res = this._super.apply(this, arguments);
+            this.bg_decoration = null;
+            return res;
+        },
+
+        willStart: function() {
+            this.bg_decoration = _.pick(this.fields_view.arch.attrs, function(value, key) {
+                return key.lastIndexOf('decoration-bg-') === 0;
+            });
+            this.bg_decoration = _.mapObject(this.bg_decoration, function(value) {
```
simahawk, line 22: you can chain this calls, see example from v11 https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/views/list/list_renderer.js#L54

web_tree_background_color/static/src/js/list_view.js
```diff
@@ -0,0 +1,53 @@
...
+            }
+            return res;
+        },
+
+        compute_decoration_classnames: function(record) {
+            var classnames = this._super.apply(this, arguments);
+            var context = _.extend({}, record.attributes, {
+                uid: session.uid,
+                current_date: moment().format('YYYY-MM-DD')
+                // TODO: time, datetime, relativedelta
```
simahawk, line 41: willing to do it now? Otherwise name this into ROADMAP

web_tree_background_color/static/src/js/list_view.js
```diff
@@ -0,0 +1,53 @@
...
+    "use strict";
+
+    var ListView = require('web.ListView');
+    var session = require('web.session');
+
+    ListView.include({
+
+        init: function() {
+            var res = this._super.apply(this, arguments);
+            this.bg_decoration = null;
```
simahawk, line 14: mention what this var is going to contain (eg: list of something, mapping expr: rule, or similar) in a comment above it

simahawk (changes requested):
LG, not tested. Minor glitches + a screenshot included in README would be nice.

pedrobaeza:
This one is very similar to https://github.com/OCA/web/tree/8.0/web_tree_dynamic_colored_field.

simahawk:
@pedrobaeza indeed, but that one allows to define custom styles.

I think we could have one single module `web_tree_color` and make them work seemingly on both rows and fields. Sounds sub-optimal to have 2 modules that do almost the same thing.

At the same time I think we should wait for v11 where `web_tree_dynamic_colored_field` has been refactored and start from there w/ the merge. Or even in v12.

What do you think?

pedrobaeza:
OK, but remember to remove plurals from module names.

simahawk:
> OK, but remember to remove plurals from module names.

:roll_eyes:  ...fixed :rofl:

web_tree_background_color/readme/DESCRIPTION.rst
LeartS: I didn't even think of that, that's great to do some advanced stuff. Will surely add to the README.

LeartS:
@pedrobaeza yeah, I saw the other module after I made the pull request. Still, I think that setting the entire row background instead of single fields it's a much needed feature. It would be definitely better to add the "entire row" functionality to the `web_tree_dynamic_colored_field` or (`web_tree_color` as @simahawk suggests), but in the meantime I think it's better to have the two modules (as soon as I address @simahawk remarks) rather than not having the "entire row background" feature.

simahawk:
@LeartS cool, tnx. One more thing: please mention this option about future merge into `web_tree_color` (even  maybe `web_view_color` if there are bits reusable in other view types) in the ROADMAP.

web_tree_background_color/static/src/js/list_view.js
steync027: The TODO is from Odoo: https://github.com/odoo/odoo/blob/10.0/addons/web/static/src/js/views/list_view.js#L209

PR #1092, opened 2018-10-24
[FIX] depend on inexisting module make openupgrade failing if module is installed in v 8.0

legalsylvain:
the commit remove the dependency on a unexisting module. 
Note that the web_graph_improved module is **NOT** installable on the version 9.0

summary of the problem : 

- use a V8 database
- installed web_graph_improved 
- migrate on v9, making an update all 

odoo will scan all the module in v9 version to create dependencies tree, **including uninstallable modules**, so will fail because the module web_graph_improved depends on web_graph that doesn't exist in V9.

CC : @OCA/openupgrade-maintainers


web_graph_improved/__openerp__.py
```diff
@@ -27,7 +27,7 @@
     'author': "Leonardo Donelli,Odoo Community Association (OCA)",
     'license': 'AGPL-3',
     'website': 'http://learts.glaucus.in',
-    'depends': ['web_graph'],
+    'depends': ['web'],
     'data': ['view/graph_improved_view.xml'],
     'installable': False,
```
MiquelRForgeFlow: 🤔

legalsylvain: Hi @mreficent, your review is quite cryptic !
could you be more explicit ? 
regards.

MiquelRForgeFlow: It's a bit odd doing changes in a module that is not installable 😏

legalsylvain: yes ! but even if it's not installable, it is read by odoo, and as it depends on a non existing module, it make the upgrade failing. 
The alternative was to delete fully this module, that has nothing to do in V9 serie. I thought it was less brutal, just to remove the undesired dependency.

MiquelRForgeFlow: I think the best solution is just delete this deprecated module from the repo! :)

MiquelRForgeFlow: Please, delete it. And also, you could delete it from v10 too :)

legalsylvain: > Please, delete it.

OK done.

> And also, you could delete it from v10 too :)

I'll do it when I'll work on 9.0 -> 10.0 migration.

MiquelRForgeFlow (approved):

PR #1093, opened 2018-10-25
fixup! [IMP] web_widget_x2many_2d_matrix: New option field_att_<name>

hbrunn:
phantomjs doesn't support `startsWith`, so without this, we can't use the widget in phantomjs tests


pedrobaeza (approved):
I think we can fast-track this.

PR #1094, opened 2018-10-31
[11.0] web_advanced_search: fix initialization of 1st domain node + add changelog

simahawk:
Sometime the dialog is not ready yet, like on EE version.
Hence when you inject the 1st domain node
the dialog must be already opened.

Example on EE before:

screenshot from 2018-10-31 10-45-38

after:

screenshot from 2018-10-31 10-44-40

**NOTE: I added a changelog too.**

Looking at git log it seems that some version upgrades were missing. Namely this fix from @Yajo https://github.com/OCA/web/commit/7cc961fb7b1ba02b53a335b5164b7f96ee8ebb2e

Hence I've include it in the changelog and upgraded module version by 0.0.2.

@pedrobaeza @sbidoul @Yajo I'd like to hear your voice on how to handle such situations.

Also, what about former odoo versions changelogs? Do we want to segregate them into each branch?

Somewhat related to https://github.com/OCA/maintainer-tools/issues/381

### DIFF
--- a/web_advanced_search/__manifest__.py
+++ b/web_advanced_search/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "Advanced search",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.2",
     "author": "Therp BV, "
               "Tecnativa, "
               "Odoo Community Association (OCA)",
--- a/web_advanced_search/README.rst
+++ b/web_advanced_search/README.rst
@@ -75,6 +75,50 @@ Improvements to the search view in this addon:
 * Beautiful, human-readable, domain representation when adding an
   advanced filter
 
+Changelog
+=========
+
+11.0.1.0.2 (2018-10-31)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* Fix initialization of 1st domain node
+
+  Sometime the dialog is not ready yet, like on EE version.
+  Hence when you inject the 1st domain node
+  the dialog must be already opened.
+
+  [simahawk]
+
+
+11.0.1.0.1 (2018-09-18)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* Fix `undefined` in x2m fields
+
+  Before this patch, when searching with the "equals to" operator in any
+  x2many field, the searched parameter was always `undefined`.
+
+  The problem was that the underlying field manager implementation was
+  treating those fields as x2many, while the widget used was the `one2many`
+  one.
+
+  This patch simply mocks the underlying fake record to make think that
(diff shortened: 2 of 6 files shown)


yajo (approved):
Thanks! Bug & fix confirmed.

web_advanced_search/__manifest__.py
```diff
@@ -5,9 +5,10 @@
 
 {
     "name": "Advanced search",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.2",
     "author": "Therp BV, "
               "Tecnativa, "
+              "Camptocamp, "
```
pedrobaeza: I think these changes don't promote Camptocamp to co-author.

pedrobaeza (changes requested):
About the changelogs, I'm against them as it requires more time to author to polite it correctly. As now, it's very technical and doesn't add information to users. If you want to see a log, use `git log` :wink: 

About patches in different versions, this module is totally different from v10 to v11, and it's a coincidence that the same patch applies to both versions.

web_advanced_search/__manifest__.py
simahawk: fixed

simahawk:
@pedrobaeza changelogs != git log. https://keepachangelog.com/en/1.0.0/

pedrobaeza:
> changelogs != git log

I know, but right now this changelog is the same, and if not, it requires another effort more for polishing it as said.

simahawk:
@pedrobaeza the changelog is the same because the log in the last 2 commits was written correctly :sunglasses: 

If we'd have always well-written commits we could generate the changelog automatically.
Problem is that many times you see really bad commits and we cannot always bug ppl (especially new comers) to rewrite their commits (sometimes I do it actually).

I think we can/should ask to keep the changelog up to date (or the maintainer does it).
In any case, if you write your commits properly the changelog won't give you much burden.

I'll quote keepachangelog.com:

> Don’t let your friends dump git logs into changelogs.
>
> Why keep a changelog?
> To make it easier for users and contributors to see precisely what notable changes have been made between each release (or version) of the project.
> 
> Who needs a changelog?
> People do. Whether consumers or developers, the end users of software are human beings who care about what's in the software. When the software changes, people want to know why and how.

The above statement is particularly true when you publish a module on PyPI (as we are already doing since quite some time tnx to @sbidoul).

Plus, since it lands in the README users like PMs, non-dev collaborators, etc will be able to check by their own if an issue is gone or a new feature has been introduced, etc.

So, all in all, is a matter of QUALITY. IMO :)

gurneyalex (approved):
Thanks for caring :smile_cat:

pedrobaeza (approved):

PR #1108, opened 2018-11-20
[FIX] show hidden loading indicator with web responsive theme

Guenzn:
The " Loading " indicator is hidden by the new navbar.


pedrobaeza:
Can you please include screenshots of before and after for illustrating the change?

Guenzn:
See images before...
before_fix
...  and after fix!
after_fix

....

PR #1121, opened 2018-12-04
[8.0][ADD] web_action_conditionable: Expression based on group xml ids

tarteo:
With this addition you can use `_group_refs` to make a condition based on the user's groups:

`<tree delete="'base.group_user' in _group_refs">`

This way you don't have to make multiple views for every group.

cc: @pankk


luc-demeyer (approved):
Nice contribution.
A small remark/suggestion: it would be nice to also have the context available

ovnicraft:
@tarteo great contribution ! could be done for next major versions too ?

About this LGTM

tarteo:
@ovnicraft Yes, I think not much has to be done to migrate to later versions.

pankk:
> 
> 
> Nice contribution.
> A small remark/suggestion: it would be nice to also have the context available

Done.


One thing to discuss is the fact that (with the changes I added) we get an error when a requested key on the context does not exist. So for example if I've defined `_context.state == 'draft'` and state is not in the context, we get an error. The benefit is that this error indicates that the use case has not been taken into account when adding the condition, and should be re-evaluated.

pankk (approved):
Technical review and tests in multiple production environments by multiple end users.

Issue #1084, opened 2018-10-18
[11.0][web_widget_x2many_2d_matrix] cannot store fields values not included in the matrix


simahawk:
Origin: https://github.com/OCA/product-variant/pull/109#discussion_r225880090

Fields not included in the matrix are not considered on save.

@pedrobaeza @Yajo is that all? Have you identified specific flaws or possible solutions?

pedrobaeza:
Thanks for continuing this. You can check on the module `sale_order_variant_mgmt` in version 10, that the model shown on the matrix contains an extra field:

https://github.com/OCA/product-variant/blob/249bea07fcf1b202254c953d09c8b5d740f06ad6/sale_order_variant_mgmt/wizard/sale_manage_variant.py#L96-L105

which is `product_id`. That field was used for storing a temp value for finding the line.

Now on v11, extra values are not stored, giving an error with the same code (on v11 there's a workaround in this case).

GSLabIt:
@simahawk do u know if it is possible to show fields not included in matrix (before/after it)? Thx

PR #1124, opened 2018-12-06
[FIX] web_widget_x2many_2d_matrix: _renderBodyCell

alexey-pelykh:


alexey-pelykh:
@simahawk @Yajo @pedrobaeza @tarteo please review this fix, and sorry for introducing an issue

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -288,11 +288,24 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
...
+            var name = node.attrs.name;
+            var field = this.state.fields[name];
+            var value = record.data[name];
+            var formattedValue = field_utils.format[field.type](value, field, {
+                data: record.data,
+                escape: true,
+                isPassword: 'password' in node.attrs,
+            });
+            this._handleAttributes($td, node);
+            return $td.html(formattedValue);(widget.$el);
```
tarteo: `return $td.html(formattedValue);`**(widget.$el);** Please, remove the bold part.

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -288,11 +288,24 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
...
+            }
+            if (node.attrs.widget || (options && options.renderWidgets)) {
+                var $el = this._renderFieldWidget(node, record, _.pick(options, 'mode'));
+                this._handleAttributes($el, node);
+                return $td.append($el);
+            }
+            var name = node.attrs.name;
+            var field = this.state.fields[name];
+            var value = record.data[name];
+            var formattedValue = field_utils.format[field.type](value, field, {
```
tarteo, line 314: field.type must be node.attrs.widget, right? So float_time format etc can be used.

tarteo (changes requested):
Looks good. For the options parameter in the function I would rather use _.defaults once instead of checking every time (also on line 286) it `options` is something.

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: Oh, sorry, being in a rush is not helping me :)

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: I would be glad to answer, yet I didn't even get what you've asked about - yet code seems ok, taking into account that list_renderer was always a base for 2d_matrix_renderer

JordiBForgeFlow:
@bjeficent @mreficent can you review plz?

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: https://github.com/odoo/odoo/blob/2a4c44bb4f0ef0d02c62991f588c20c0741616e2/addons/web/static/src/js/views/list/list_renderer.js#L301

alexey-pelykh: @tarteo ^

hbrunn:
this is an example why open source is so super cool: I introduced this widget a while ago, others picked it up, extended, and now I'm at a state where I don't understand the diff without reading all of what people did during the v11 and v12 migration. Looks great! But I'm still reading, so can't put a review currently

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -288,11 +288,24 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
             }
             // Enforce mode of the parent
             options.mode = this.getParent().mode;
-            var widget = this._renderFieldWidget(
-                node, record, _.pick(options, 'mode')
-            );
-            this._handleAttributes(widget.$el, node);
-            return $td.append(widget.$el);
+            if (node.tag === 'widget') {
```
simahawk, line 303: TBH I don't remember why I discarded this stuff during refactoring :thinking: 
Fact is: it looks good and it's the way the std list renderer uses it's :ok_hand:

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
simahawk: @tarteo nope, the format utils works at field type level (m2o, char, etc) so: this sounds correct :+1: 

https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/js/fields/field_utils.js#L651

simahawk (changes requested):
@alexey-pelykh LG. I'd like to see a version upgrade and a line in the changelog. Can you do that?

simahawk:
@tarteo checking that format thingy and how it changed... on abstract field they do this:

https://github.com/odoo/odoo/blob/1868713dbd07e0b518f91dffe73e62d85e6ab9a6/addons/web/static/src/js/fields/abstract_field.js#L164-L173

```js
        // formatType is used to determine which format (and parse) functions
        // to call to format the field's value to insert into the DOM (typically
        // put into a span or an input), and to parse the value from the input
        // to send it to the server. These functions are chosen according to
        // the 'widget' attrs if is is given, and if it is a valid key, with a
        // fallback on the field type, ensuring that the value is formatted and
        // displayed according to the choosen widget, if any.
        this.formatType = this.attrs.widget in field_utils.format ?
                            this.attrs.widget :
this.field.type;
```
so, yes, there could be also a specific formatting by widget if found in the registry,

However, in the list renderer is ignored and I think it's because you already try to render via widget (which is going to handle it using `formatType`) and it just fallbacks when no option is left to the mere field type.

Seems legit, no?

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: Who knows :) Also, button seems to have no meaning fir this renderer, right?

alexey-pelykh:
@simahawk done, also please check:
- removed button-related code (is it even possible to have one in 2d matrix?)
- added few comments

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
tarteo: @simahawk There's also float_time in format. https://github.com/odoo/odoo/blob/d40956c7909bffa338fd0138565044571e852da3/addons/web/static/src/js/fields/field_utils.js#L660

tarteo:
@simahawk Ah yes, thanks. @alexey-pelykh Can you add some comments in the code on what's happening.

web_widget_x2many_2d_matrix/__manifest__.py
```diff
@@ -4,11 +4,12 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     'name': '2D matrix for x2many fields',
-    'version': '12.0.1.0.0',
+    'version': '12.0.2.0.0',
```
simahawk: this is a bugfix not a major release. Increase only 0.0.1

web_widget_x2many_2d_matrix/readme/HISTORY.rst
```diff
@@ -0,0 +1,5 @@
+12.0.2.0.0 (2018-12-07)
```
simahawk: would be nice to catch the commit date (or merge date) of the v12 migration to add the 1st version before

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -51,23 +59,30 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
          * @returns {Deferred} this deferred is resolved immediately
          */
         _renderView: function () {
+            // ListRenderer._renderView {
```
simahawk: there's no need to keep all these comments. This is not the list renderer nor it inherits from it.
Saying that we were inspired by it is enough.

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
simahawk: not sure we need buttons here but leaving it won't hurt as long as it does not fail.

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -401,6 +441,32 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
             return this._super.apply(this, arguments);
         },
 
+        /**
+         * Manages the keyboard events on the list. If the list is not editable, when the user navigates to
```
simahawk: this is an addition, should be treated as such: split the commit, add a new line in the changelog

web_widget_x2many_2d_matrix/__manifest__.py
simahawk: as you added the key nav feature you can raise by 0.1.0

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: I think it would be beneficial for next generation of contributors to give hints what was replaced or sourced as-is from ListRenderer, for that only purpose

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: It won't render anyway, thus no need of keeping this unless adding support for buttons

web_widget_x2many_2d_matrix/README.rst
```diff
@@ -42,7 +42,7 @@ An example use case would be: Select some projects and some employees so that
 a manager can easily fill in the planned_hours for one task per employee. The
 result could look like this:
 
-.. image:: https://raw.githubusercontent.com/web_widget_x2many_2d_matrix/static/description/screenshot.png
+.. image:: /web_widget_x2many_2d_matrix/static/description/screenshot.png
```
yajo: Please let it download from GH. It makes the link work, no matter where the README is rendered.

The link was wrong, however. Use https://raw.githubusercontent.com/OCA/web/12.0/web_widget_x2many_2d_matrix/static/description/screenshot.png

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
yajo: I agree with @simahawk. We have git history to know where code comes from, so if you wish, you can add notes in the commit, but don't pollute all of this please. :blush:

yajo (changes requested):
Yikes, this is really hard to review. Please, could you add some details on what are you fixing and how, and remove all the clutter introduced with all that commented code, to keep a smaller diff, easier to review?

Please abstain from opening big PRs without description. Thanks!

web_widget_x2many_2d_matrix/README.rst
pedrobaeza: @Yajo, the bot now do this work automatically thanks to https://github.com/OCA/maintainer-tools/pull/397

pedrobaeza: But yes, I think it has to be a bit adjusted for working right now with this automation.

alexey-pelykh: Done

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: Well, I don't think git history will help here - taking into account how much time I spent doing 3-way diff 11.0 odoo:list_renderer vs 12.0 odoo:list_renderer vs this module

yajo: OK, then just remove the comments please.

AaronHForgeFlow (commented):
I will only say that this makes https://github.com/OCA/hr-timesheet/pull/140 to work with hr_timesheet_sheet. So It'd be great if all comments here are attended and we can move on with that.

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: Removed, although I'm still sure that by doing this we've added extra couple of fun hours for future migrations

alexey-pelykh:
All changes made as of now

simahawk (approved):

yajo (approved):

yajo:
@tarteo can you please update your review?

alexey-pelykh:
Significant note: for some reason, editable mode is not working.
```
options.mode = this.getParent().mode;
```
worth checking

alexey-pelykh:
`readonly` vs `edit` are getting propagated properly, that's all I know right now

alexey-pelykh:
/me dumb - now fields can be editable properly

pedrobaeza:
Several questions about this (some might be out of the scope of this PR):

* Can you use a many2one for values of the matrix?
* Can you specify editable by cell possibilities through `attrs`?
* Is #1084 fixed here?

alexey-pelykh:
@pedrobaeza unfortunately I'm unable to address any of these questions properly, this PR is related to migration, so I don't think it would fix anything by magic. Yet, it may have introduced support for widgets (question 1.), and right now the editable state is propagated from parent (question 2.)

pedrobaeza:
Well, maybe 3 might be introduced in your migration, so that's why I asked... @tarteo please confirm the review for merging this.

tarteo (approved):
Thanks!

PR #1125, opened 2018-12-06
[8.0][IMP][web_m2x_options] Do not limit the items quantity in the M2O search pop up

### LINKED ISSUE #1123: [8.0][web_m2x_options] m2o pop up selection hide result > 80
Step to reproduce : 
- go to runbot OCA/web/8.0
- install point_of_sale (to have data populated)
- go to invoicing, and create a new invoice
- select a product and click on "search more"
- the pop up screen mentions that there are only 80 elements

image

- uninstall the module web_m2x_options. 
- refresh the cache
-> the behaviour of the pop up is now ok 

image

version 8 : :-1: impacted
version 9 : :question: not tested
version 10.0 : :question:  not tested
version 11.0 : :+1: OK 

CC : authors : @hbrunn, @njeudy, @yvaucher, @Yajo, @chienandalu ....

legalsylvain:
Fix #1123. 

- The current module web_m2x_options hardcodes 80 items limit.

- The Odoo web modules hardcodes 160 items limit.
https://github.com/odoo/odoo/commit/8c3cdce539d87775b59b3f2d5ceb433f995821bf

This patch remove this limits and make the search faster. (no call to useless name_search anymore)

Note that maybe it can be slow in very huge database. (tested on DB with 50k products without problems)

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -127,6 +127,11 @@ Known issues
 Double check that you have no inherited view that remove ``options`` you set on a field ! 
 If nothing works, add a debugger in the first line of ``get_search_result method`` and enable debug mode in Odoo. When you write something in a many2one field, javascript debugger should pause. If not verify your installation.
 
+This module fixes an Odoo bug in the pop up of search that limit the items quantity
+to 160 https://github.com/odoo/odoo/commit/8c3cdce539d87775b59b3f2d5ceb433f995821bf
+For that purpose, it makes previously a light ``search`` call, to get all the ids.
+This process could make the display of the pop up slow, in case of huge databases.
+
 Roadmap
 =======
 
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -167,14 +167,20 @@ openerp.web_m2x_options = function (instance) {
                     values.push({
                         label: _t("Search More..."),
                         action: function () {
-                            // limit = 80 for improving performance, similar
-                            // to Odoo implementation here:
+                            // Odoo JS framework is quite bad designed
+                            // _search_create_popup requires to receices the ids
+                            // even if it will do again a search when displayed
+                            // By default, Odoo display hard code a domain of 160 ids
+                            // preventing to make a search if your datalist has more that 160 items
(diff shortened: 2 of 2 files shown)


tarteo (commented):
Why not cap it at 160 like Odoo? I can't test if this gets unbearably slow with many records, so I have to trust you on that. :smile:

legalsylvain:
> Why not cap it at 160 like Odoo?

Well, the problem is that odoo doesn't works correctly. if you have 200 products, odoo will show you a page 1-80 and another 81-160. impossible to see the rest of the result, and it's quite disturbing for the user to see a wrong total number. (160 instead of 200) 

Yes, another approach is just to replace in the current module 80 by 160. With that, the OCA module will works "bad as odoo".

regards.

tarteo:
@legalsylvain I think for most databases this change will make it better.

tarteo (approved):

web_m2x_options/README.rst
```diff
@@ -127,6 +127,11 @@ Known issues
 Double check that you have no inherited view that remove ``options`` you set on a field ! 
 If nothing works, add a debugger in the first line of ``get_search_result method`` and enable debug mode in Odoo. When you write something in a many2one field, javascript debugger should pause. If not verify your installation.
 
+This module fixes an Odoo bug in the pop up of search that limit the items quantity
```
yajo, line 130: Just let me remind you that addons is not the place to fix bugs. If you really consider this a bug, it should be fixed where it is: OCB.

yajo (approved):

web_m2x_options/README.rst
legalsylvain: Indeed. but it should be done twice. If we fix Odoo, (via OCB) but install this module, this bug come back -;)

yajo: Yes, it's just that adding this comment in the README is a little bit misleading. If there's a bug, then it shouldn't be documented in some random addon readme; it should be just fixed.

Also, if this addon avoids to introduce a bug, that's just a bug fix. No need to document that, unless you just want to add a note in the release notes.

I don't know if I explain myself :sweat_smile: 

Well in any case you got my approval here :blush:

Issue #1095, opened 2018-11-01
[11.0][web_listview_range_select] Can not delete multiple records selected by holding Shift


novawish:
# Current situation
When select multiple records on list view (with Shift) and click _Action > Delete_, the system will only delete the first and the last record, records within the range are not affected.
Same issue happens with _Export_ and _Invoice Orders_ on Sale Order list view.

# Expected behavior
Should be able to do all the options inside `Action` with list view range selection

pedrobaeza:
Version?

novawish:
I tested on odoo version 11.0 but i think it would be the same for 10.0 and 12.0

tarteo:
I tested it it's only for 11.0. I've to see what changed in the migration to 11.0 https://github.com/OCA/web/pull/910

novawish:
Hello @tarteo , thanks for your reply, do you have any news on this?

tarteo:
@novawish Hey, I created a fix: #1127

metalmonkey69:
@tarteo I've tested your fix. If i apply it for range export, the last record will be exported twice. I guess it works for range delete because you cannot delete a record twice. I am using Odoo11 latest version.

tarteo:
@peterhoth Thanks, I also fixed this in the pull request. Can you check and leave a review. #1127

PR #1127, opened 2018-12-07
[11.0][FIX] web_listview_range_select: Internal selection

tarteo:
Fixes the internal selection of the list view when selecting a range of records.

https://github.com/OCA/web/issues/1095

### DIFF
--- a/web_listview_range_select/__manifest__.py
+++ b/web_listview_range_select/__manifest__.py
@@ -6,7 +6,7 @@
     'summary': """
         Enables selecting a range of records using the shift key
     """,
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'category': 'Web',
     'author': 'Onestein,Odoo Community Association (OCA)',
     'website': 'https://github.com/oca/web',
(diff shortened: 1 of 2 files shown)


novawish (approved):
Tested on runbot and it works

metalmonkey69:
@tarteo range export works. Only small thing i need to pick is the records are not in the correct order as listed on screen. i.e. First 2 records will be the start record and end record followed by the rest of the selected records. 
e.g.
Screen order,   Exported order
001,                 001
002,                 004
003,                 002
004,                 003

It would be perfect and users will not be confused if the export/delete order is the same as the selection shown.

tarteo:
@peterhoth I understand, but this is unfortunately not possible. This is out of control of the module.

metalmonkey69:
@tarteo thanks anyway !

pedrobaeza:
@peterhoth please give your specific approval in this PR.

tarteo:
@peterhoth Can you approve the changes? You can do this in the tab 'Files changed' and then the button 'Review changes'.

metalmonkey69 (approved):
Tested range export and delete and it is working.

PR #1128, opened 2018-12-09
[12.0][MIG] web_ir_actions_act_multi

alexey-pelykh:


alexey-pelykh:
@pnajman-modoolar @simahawk @lreficent since you were involved in 11.0, please review this migration for 12.0

rodrigets (approved):

tarteo (changes requested):
Can you change the name of the module to web_ir_action_act_multi (singular) and take a look at the jslint issues?

alexey-pelykh:
@tarteo re: renaming yet there are other `web_ir_actions_*` modules - won't this break consistency? Maybe we should postpone it until 13.0 migration?

hbrunn:
I know the conventions say no plurals, but I think we should apply this rule with some context. I started naming those modules like that years ago because that's how the models in Odoo are called. In my opninion, we shouldn't change this.

pedrobaeza:
Yes, on guidelines it's said that names that are plurals in Odoo, should be preserved: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#11modules, so it's OK to leave it as is.

tarteo (approved):
:+1: Thanks. code looks good. I couldn't test the functionality though. Maybe you can add demo action so it's easier to test the functionality.

alexey-pelykh:
@tarteo thanks for a hint, I'll consider adding example as a separate module. Not sure if this module is used a lot elsewhere, yet at least it's being used in this one https://github.com/OCA/hr-timesheet/pull/139

hbrunn:
you can just add some example action as demo data, much less work and we don't clutter out repos with a lot of _demo modules.

alexey-pelykh:
@hbrunn a snippet or a link to example would be much appreciated, since I haven't yet worked with demo data.

web_ir_actions_act_multi/static/src/js/web_ir_actions_act_multi.js
```diff
@@ -1,29 +1,51 @@
...
-"use strict";
+    "use strict";
 
     var ActionManager = require('web.ActionManager');
 
     ActionManager.include({
 
-        execute_ir_actions_act_multi: function(actions, options, index){
+        /**
+         * Intersept action handling to detect extra action type
```
hbrunn: interCept

hbrunn (approved):

web_ir_actions_act_multi/static/src/js/web_ir_actions_act_multi.js
alexey-pelykh: Thanks!

hbrunn:
check out https://github.com/OCA/web/blob/10.0/web_ir_actions_act_window_page/__manifest__.py#L18

hbrunn:
whatever you add to the `demo` key in the manifest is only loaded when the demo flag for the module is set, which is never the case on production servers (if not somebody messed up majorly).

Meant to add data needed for tests, but we can use it perfectly fine to, well, demo a module's functionality.

alexey-pelykh:
@hbrunn thought about examples - this is so purely developer-related module, not even sure it's worth adding an example besides describing what this module does. If you have any particular ideas for demos - it would be great to hear them!

hbrunn:
hmm, I just answered your question on how to do @tarteo's proposal, don't have much of an opinion about this. As everyone approved, I'll just merge.

alexey-pelykh:
@hbrunn true, sorry for mis-tagging!

PR #1130, opened 2018-12-11
[10.0] web_widget_color allow empty values

Martronic-SA:
Allow for empty values and let Odoo framwork to decide it the field is required or not.

### DIFF
--- a/web_widget_color/__manifest__.py
+++ b/web_widget_color/__manifest__.py
@@ -8,7 +8,7 @@
 {
     'name': "Web Widget Color",
     'category': "web",
-    'version': "10.0.1.0.0",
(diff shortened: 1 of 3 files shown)


tarteo (changes requested):
Can you increase the version number?

Martronic-SA:
Hi @tarteo I'm sorry but I haven't made any testings on the v11/12 versions.
It's probably working but it would require some checks first.

tarteo:
@Martronic-SA I mean in the manifest file. 😄

tarteo (approved):
Thanks!

pedrobaeza (approved):
This should be forward-ported to v11/v12

PR #1134, opened 2018-12-18
[web_timeline][IMP] Add stack support

cvinh:


web_timeline/static/src/js/timeline_view.js
```diff
@@ -111,12 +111,18 @@ odoo.define('web_timeline.TimelineView', function (require) {
             if (!isNullOrUndef(attrs.quick_create_instance)) {
                 self.quick_create_instance = 'instance.' + attrs.quick_create_instance;
             }
+            if (isNullOrUndef(attrs.stack) || !_.str.toBoolElse(attrs.stack, "true")) {
+                this.stack = false;
```
tarteo, line 116: By default it should stack IMO.

tarteo (changes requested):

web_timeline/static/src/js/timeline_view.js
tarteo: The toBoolElse function is used the wrong way here (by reading the definition). 
Also you can remove the else by doing it like this:

```
            this.stack = true;
            if (!isNullOrUndef(attrs.stack) && !_.str.toBoolElse(attrs.stack, true)) {
                this.stack = false;
            }
```

tarteo:
Can you increase the version number in the manifest file.

tarteo (approved):
Thanks!

PR #1136, opened 2018-12-21
[10.0][ADD] web_sheet_widescreen

markrobinson85:
Forked web_sheet_full_width to instead use a media query to increase the width of sheets for widescreens, while maintaining the comfort of using a sheet with negative space around it. Can be easily extended to support more display types/sizes.


web_sheet_widescreen/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Web - Sheets for Widescreen",
+    "version": "10.0.1.0.1",
+    "author": "Mark Robinson",
```
nikul-serpentcs: @markrobinson85 author like [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L12)

Add 'website' key like [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L11)

web_sheet_widescreen/view/qweb.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
```
nikul-serpentcs: @markrobinson85 use `<odoo>` and remove `<data>`

nikul-serpentcs (commented):
Improve Code

hbrunn (changes requested):
I think you shouldn't add a new module for this, but change the existing one: Add an extra group, add a config wizard item to enable this group for all employees, and add a template that sets some class on the `o_form_sheet` element. Use this class to select the element then, the net effect is that we can turn this behavior on and off with the config wizard and you barely have to write code for this.
Then I think your CSS is very specific to your setting, aren't we better off with some `calc` expression combined with media queries?

web_sheet_widescreen/__manifest__.py
hbrunn: and OCA as author

PR #1142, opened 2018-12-25
[10.0] web_advanced_search_wildcard: adds 'starts with' and 'ends with'

alexis-via:
We keep the historic feature *matches* provided by web_advanced_search_wildcard, but we add *starts with* and *ends with* with are super-easy and intuitive to use and cover the majority of use cases.

### DIFF
--- a/web_advanced_search_wildcard/README.rst
+++ b/web_advanced_search_wildcard/README.rst
@@ -6,11 +6,16 @@
 Wildcard in advanced search
 ============================
 
-Allows =ilike ('matches') operator to advanced search option.
+This module adds 3 options to advanced search of char, many2one,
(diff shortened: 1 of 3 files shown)


alexis-via:
@tafaRU @lfreeke What do you think about this improvement ?

alexis-via:
I found out that this PR breaks Odoo web interface with older versions of firefox such as Firefox 45 ESR.

alexis-via:
Now working fine with old Firefox 45 AND newer Firefox. The joys of JS !

MOSAMAD (approved):
Tests only, no code review

tarteo (approved):

PR #1150, opened 2019-01-08
[12.0] [MIG] web_no_bubble

anandkansagra:
Issue [Reference](https://github.com/OCA/web/issues/1048)


web_no_bubble/views/web_no_bubble.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="assets_backend" name="web_no_bubble_assets" inherit_id="web.assets_backend">
+        <xpath expr="." position="inside">
+            <link rel="stylesheet" href="/web_no_bubble/static/src/css/web_no_bubble.scss"/>
```
nikul-serpentcs: @anandkansagra change `css` directory to `scss`
Ex. `/web_no_bubble/static/src/css/web_no_bubble.scss"/` to `/web_no_bubble/static/src/scss/web_no_bubble.scss"/`

anandkansagra: Thanks @nikul-serpentcs ,
Could you check it now?

pedrobaeza:
Is this still needed? I see no bubbles in runbot for v12.

anandkansagra:
> Is this still needed? I see no bubbles in runbot for v12.

@pedrobaeza , I have found a class in ODOO base so this module will work.
https://github.com/odoo/odoo/blob/12.0/addons/web_tour/static/src/scss/tip.scss#L71

pedrobaeza:
Class can be there, but if it's not used, this module is unneeded.

anandkansagra:
> Class can be there, but if it's not used, this module is unneeded.

@pedrobaeza , what if somebody use it in custom module?

pedrobaeza:
I don't think so, but any way, if not used in Odoo, it doesn't matter. As said, for me this module should disappear.

nikul-serpentcs:
> I don't think so, but any way, if not used in Odoo, it doesn't matter. As said, for me this module should disappear.

Yes, @pedrobaeza In v12 I think not used bubbles, I checked in runbot but not find it. 

@anandkansagra Thanks for contribution

pedrobaeza:
I have included the part for migration path to not report this module as absent in v12: https://github.com/OCA/OpenUpgrade/commit/662858b3effaf1c63078346ce83c21fdc42031b4, so closing this. Thanks any way for the effort.

PR #1154, opened 2019-01-15
[FIX] web_decimal_numpad_dot: Fix inconsistent behavior

yajo:
When user was replacing some specific sections or being too fast, the module wasn't behaving fine. Also, it was always displaying the original `.` slightly before doing the replacement.

This is before the fix:
bad

Now, after the fix:
good

@Tecnativa

### DIFF
--- a/web_decimal_numpad_dot/static/src/js/numpad_dot.js
+++ b/web_decimal_numpad_dot/static/src/js/numpad_dot.js
@@ -9,7 +9,7 @@ odoo.define("web_decimal_numpad_dot.FieldFloat", function (require) {
     var NumpadDotReplaceMixin = {
         init: function () {
             this.events = $.extend({}, this.events, {
-                "keyup": "numpad_dot_replace",
+                "keydown": "numpad_dot_replace",
             });
             return this._super.apply(this, arguments);
         },
@@ -19,14 +19,13 @@ odoo.define("web_decimal_numpad_dot.FieldFloat", function (require) {
                 ? ":" : translation._t.database.parameters.decimal_point;
         },
 
-        _replaceAt: function (cur_val, index, replacement) {
-            return cur_val.substr(0, index) + replacement +
-                   cur_val.substr(index + replacement.length);
+        _replaceAt: function (cur_val, from, to, replacement) {
+            return cur_val.substring(0, from) + replacement +
+                   cur_val.substring(to);
         },
 
         numpad_dot_replace: function (event) {
             // Only act on numpad dot key
-            event.stopPropagation();
             if (event.keyCode !== 110) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


web_decimal_numpad_dot/static/src/js/numpad_dot.js
```diff
@@ -19,14 +19,13 @@ odoo.define("web_decimal_numpad_dot.FieldFloat", function (require) {
...
-            return cur_val.substr(0, index) + replacement +
-                   cur_val.substr(index + replacement.length);
+        _replaceAt: function (cur_val, from, to, replacement) {
+            return cur_val.substring(0, from) + replacement +
+                   cur_val.substring(to);
         },
 
         numpad_dot_replace: function (event) {
             // Only act on numpad dot key
-            event.stopPropagation();
```
pedrobaeza, line 29: why removing this?

web_decimal_numpad_dot/static/src/js/numpad_dot.js
```diff
@@ -9,7 +9,7 @@ odoo.define("web_decimal_numpad_dot.FieldFloat", function (require) {
     var NumpadDotReplaceMixin = {
         init: function () {
             this.events = $.extend({}, this.events, {
-                "keyup": "numpad_dot_replace",
+                "keydown": "numpad_dot_replace",
```
pedrobaeza, line 12: I remember that this was changed to `keyup` on purpose due to other reason. Can you check? What happened when you keep the key pressed?

web_decimal_numpad_dot/static/src/js/numpad_dot.js
yajo: Because it didn't make sense in the first place.

web_decimal_numpad_dot/static/src/js/numpad_dot.js
yajo: Yes, I checked that. When you type fast, sometimes you hit another number before releasing the dot key, so we actually need it to be keydown. When you keep the key pressed, multiple keydown are sent, while only one keyup is sent at the end when you release it.

web_decimal_numpad_dot/static/src/js/numpad_dot.js
yajo: To explain a little better, if any handler reacts to a keypress, it would be inhibited as long as focus was in a float field. That didn't make sense, we don't care about event propagation in this case, the important part is to stop and modify the native behavior in case it's a numpad dot, and that's done below.

pedrobaeza (approved):

tarteo (approved):

pedrobaeza:
@Yajo, please cherry-pick it on v12

yajo:
Done at 75ac50be53113289396f2f722e78c77eb156d37d

pedrobaeza:
Thanks

PR #1155, opened 2019-01-16
[FIX] web_m2x_options: Fix error in tree views with many2many_tags

yajo:
The method `_onOpenColorPicker` is defined on `FormFieldMany2ManyTags`, not in `FieldMany2ManyTags`, so it can only be called on its descendants.

Otherwise, clicking on a `many2many_tags` tag when found in a tree view produces an error such as:

    TypeError: self._onOpenColorPicker is not a function

    http://localhost/web_m2x_options/static/src/js/form.js:354
    Rastreo de error:
    _onOpenBadge@http://localhost/web_m2x_options/static/src/js/form.js:354:17
    proxy/<@http://localhost/web/static/src/js/core/mixins.js:279:20
    dispatch@http://localhost/web/static/lib/jquery/jquery.js:4640:50
    add/elemData.handle@http://localhost/web/static/lib/jquery/jquery.js:4309:41

@Tecnativa

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -9,10 +9,12 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
         Dialog = require('web.Dialog'),
         view_dialogs = require('web.view_dialogs'),
         relational_fields = require('web.relational_fields'),
-        FieldMany2One = relational_fields.FieldMany2One,
+        rpc = require('web.rpc');
+
+    var _t = core._t,
         FieldMany2ManyTags = relational_fields.FieldMany2ManyTags,
-        rpc = require('web.rpc'),
-        _t = core._t;
+        FieldMany2One = relational_fields.FieldMany2One,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
With this, the tag will be clickable for selecting color? I think this is not the default behavior, and for this module, I would say that it's something not wanted.

yajo:
This patch only fixes the bug, but the logic beneath remains the same as it was.

pedrobaeza:
Yeah, but I'm asking what is the logic, and if it's what I'm saying, we should inhibit it.

yajo:
If the widget is supposed to open the record when clicking on the tag, or if it is marked as `no_color_picker: true`, then the color picker will not display.

Otherwise it will.

yajo:
I fixed a little bug, this should be ready to merge.

pedrobaeza (approved):
I'm curious why do you need separated var definitions

rafaelbn (approved):

Issue #1138, opened 2018-12-22
Error while changing view from timeline to another and backward  


tglukhov:
Hello everyone, 

There is a problem with the view timeline in odoo V12, timeline V12

steps to reproduce : 
- open timeline view
- change view to any different
(- you can switch to any other view except timeline as many times as you want)
- back to view timeline 
the everything is frozen, you need to reload page 

message error : 
```
Error:
Uncaught TypeError: Cannot read property 'then' of undefined

http://localhost:8069/web/static/src/js/chrome/action_manager_act_window.js:650
Traceback:
TypeError: Cannot read property 'then' of undefined
at http://localhost:8069/web/static/src/js/chrome/action_manager_act_window.js:650:69
at http://localhost:8069/web/static/lib/jquery/jquery.js:3276:89
at fire (http://localhost:8069/web/static/lib/jquery/jquery.js:3119:58)
at Object.add [as done] (http://localhost:8069/web/static/lib/jquery/jquery.js:3165:49)
at Array. (http://localhost:8069/web/static/lib/jquery/jquery.js:3275:77)
at Function.each (http://localhost:8069/web/static/lib/jquery/jquery.js:383:58)
at Object. (http://localhost:8069/web/static/lib/jquery/jquery.js:3272:56)
at Function.Deferred (http://localhost:8069/web/static/lib/jquery/jquery.js:3332:30)
at Object.then (http://localhost:8069/web/static/lib/jquery/jquery.js:3271:55)
at Object. (http://localhost:8069/web/static/src/js/chrome/action_manager_act_window.js:649:68)

```

cvinh:
Please give us which version of web_timeline you are using

pedrobaeza:
And Odoo version.

pedrobaeza:
Also the traceback doesn't serve for anything if you are not in debug with assets mode.

tglukhov:
oh yeah, sorry, 
the odoo version is v12 and web_timeline is a OCA module, also version 12. 

here is the error message in debug=assets mode 
Error:
Uncaught TypeError: Cannot read property 'then' of undefined

http://localhost:8069/web/static/src/js/chrome/action_manager_act_window.js:650
Traceback:
TypeError: Cannot read property 'then' of undefined
    at http://localhost:8069/web/static/src/js/chrome/action_manager_act_window.js:650:69
    at http://localhost:8069/web/static/lib/jquery/jquery.js:3276:89
[...]

PR #1156, opened 2019-01-16
[12.0][FIX] web_timeline: Reloading view

tarteo:
Fixes: https://github.com/OCA/web/issues/1138

### DIFF
--- a/web_timeline/static/src/js/timeline_controller.js
+++ b/web_timeline/static/src/js/timeline_controller.js
@@ -37,9 +37,9 @@ odoo.define('web_timeline.TimelineController', function (require) {
          * @override
          */
         update: function (params, options) {
-            this._super.apply(this, arguments);
+            var res = this._super.apply(this, arguments);
             if (_.isEmpty(params)){
-                return;
+                return res;
             }
             var defaults = _.defaults({}, options, {
                 adjust_window: true
@@ -63,17 +63,20 @@ odoo.define('web_timeline.TimelineController', function (require) {
 
             var fields = this.renderer.fieldNames;
             fields = _.uniq(fields.concat(n_group_bys));
-            self._rpc({
-                model: self.model.modelName,
-                method: 'search_read',
-                kwargs: {
-                    fields: fields,
-                    domain: domains,
-                },
-                context: self.getSession().user_context,
-            }).then(function(data) {
-                return self.renderer.on_data_loaded(data, n_group_bys, defaults.adjust_window);
-            });
+            return $.when(
+                res,
+                self._rpc({
+                    model: self.model.modelName,
+                    method: 'search_read',
+                    kwargs: {
+                        fields: fields,
+                        domain: domains,
+                    },
+                    context: self.getSession().user_context,
+                }).then(function (data) {
+                    return self.renderer.on_data_loaded(data, n_group_bys, defaults.adjust_window);
+                })
+            );
         },
 
         /**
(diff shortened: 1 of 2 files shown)


web_timeline/static/src/js/timeline_controller.js
```diff
@@ -39,7 +39,7 @@ odoo.define('web_timeline.TimelineController', function (require) {
         update: function (params, options) {
             this._super.apply(this, arguments);
             if (_.isEmpty(params)){
-                return;
+                return $.Deferred().resolve();
```
yajo: You can do `return $.when()`, which is the same. Possibly not needed anyways due to what is explained above.

web_timeline/static/src/js/timeline_controller.js
```diff
@@ -39,7 +39,7 @@ odoo.define('web_timeline.TimelineController', function (require) {
         update: function (params, options) {
             this._super.apply(this, arguments);
```
yajo: This is probably creating a race condition: `this._super` returns a deferred, which is being ignored and you're creating your own deferred.

Please could you check if this is what you really want here?:

```js
return this._super.apply(this, arguments).then(function () {
   // Your code here
}); //Maybe you need to chain several .then()
```

If your code execution doesn't depend on super's resolution, then you should probably change it for this:

```js
return $.when(this._super.apply(this, arguments), function () {
  // Your code here
});
```

yajo (changes requested):

web_timeline/static/src/js/timeline_controller.js
tarteo: I now just return the deferred from the super.

yajo: OK, just to be sure: The operation you're performing here depends somehow in the result from super?

tarteo: No. The update function is now done when the super and the reloading of records are done.

tglukhov:
Your fix is working great, thank you a lot !

web_timeline/static/src/js/timeline_controller.js
```diff
@@ -63,17 +63,20 @@ odoo.define('web_timeline.TimelineController', function (require) {
...
-                    fields: fields,
-                    domain: domains,
-                },
-                context: self.getSession().user_context,
-            }).then(function(data) {
-                return self.renderer.on_data_loaded(data, n_group_bys, defaults.adjust_window);
-            });
+            return $.when(
+                res,
+                self._rpc({
```
simahawk, line 68: Not blocking: wouldn't make sense to move the RPC call to a separated func in case a custom behavior is needed?
BTW you could also wrap all the calls into `$.when`, like `$.when(res, self._call_rpc, self._update_rendering)` and `_update_rendering` (random names here) would accept both results of previous calls and then call `on_data_loaded`. The whole chain in this case would become modular.

yajo (approved):

web_timeline/static/src/js/timeline_controller.js
tarteo: Do you mean `$.when(res, this._loadRecords).then(this._updateRenderer)` where _updateRenderer will accept the results of res, and this._loadRecords?

If this the case this it can be done but then I've to refactor many things, because they're dependent on variable n_group_bys, defaults, and domain.
I think it's better to do it in a different PR along with other refactorings.

With this PR I just want to fix the issue.

simahawk (approved):

Issue #1149, opened 2019-01-07
[10.0] web_widget_image_webcam Error getting stream


ngrandjean:
Hello,

Module web_widget_image_webcam worked just fine until very recently. Now I have en error everytime I try to connect to the webcam : 
![image](https://user-images.githubusercontent.com/25527410/50765032-f0626f80-1274-11e9-9227-4ab0fce7f80a.png)

I've tried with Chrome and Firefox.

pedrobaeza:
Have you installed any additional extension? Or maybe have you updated the browser?

ngrandjean:
I guess the browser has been updated yes. After some researches it seems that the method createObjectURL() is deprecated for MediaStream objects (see https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL).

I did a quick fix for my system because it was very blocking but I didn't fully tested it for previous Chrome/Firefox version or for other browsers. Would it still help if I did a PR with it ?

pedrobaeza:
Yes, please! but the method should be compatible for both old and new browsers.

aitlamsali:
Hello any update about this, still waiting, thank you very much!

PR #964, opened 2018-06-21
[11.0][MIG] web widget image webcam

bud-e:

PR #1159, opened 2019-01-18
[10.0][FIX] web_widget_webcam: fix compatibility for video source

ngrandjean:
Fixes issue #1149

### DIFF
--- a/web_widget_image_webcam/static/src/js/webcam.js
+++ b/web_widget_image_webcam/static/src/js/webcam.js
@@ -139,7 +139,11 @@ var Webcam = {
 			})
 			.then( function(stream) {
 				// got access, attach stream to video
-				video.src = window.URL.createObjectURL( stream ) || stream;
+				try {
+				    video.srcObject = stream;
+                } catch (error) {
+                    video.src = window.URL.createObjectURL(stream) || stream;
+                }
 				self.stream = stream;
 				self.loaded = true;
 				self.live = true;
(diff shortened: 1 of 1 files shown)


web_widget_image_webcam/static/src/js/webcam.js
```diff
@@ -139,7 +139,11 @@ var Webcam = {
 			})
 			.then( function(stream) {
 				// got access, attach stream to video
-				video.src = window.URL.createObjectURL( stream ) || stream;
+				try {
+				    video.srcObject = stream;
```
tarteo, line 143: Does this line throw an error in older browsers?

tarteo (changes requested):
Can you look at the indentation, please?

web_widget_image_webcam/static/src/js/webcam.js
ngrandjean: Apparently it does. Honestly I couldn't reproduce a case when there's an error so I just followed the recommended fallback of the MDN doc, just in case (it's here if you want to check https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/srcObject)

tsabi:
This fix works for me! Tested with Firefox and Chrome (without this fix none of these browsers worked).

Please merge it as soon as you can.

web_widget_image_webcam/static/src/js/webcam.js
bud-e: :+1: 
As this code deprecated, So this change is necessary. As this change is also done by the maintainer of this webcam library in the latest version.
https://github.com/OCA/web/pull/964/files#diff-c740e52ad4ba0136715ea03661ca21f6R319

Thereby, the library version is also updated for v11.0 module.
https://github.com/OCA/web/pull/964#issuecomment-473201816

PR #1162, opened 2019-01-20
[12.0] [MIG] web_tree_many2one_clickable

anandkansagra:
Issue [Reference](https://github.com/OCA/web/issues/1048)


pedrobaeza (changes requested):
I'm seeing that this module has lost the possibility of defining an option as field level for indicating if clickable or not, as well as the system parameter that defines the behavior of making clickable all many2one. I think this has happened on v11

@simahawk I think you were the porter. Why removing that features?

bealdav:
@pedro I think lose the feature you said is not problematic if we can get this feature https://github.com/OCA/web/pull/1041 (not a lot of code).
May be in v12 no more fail with big web 10.0 branch.
What do you think ?

pedrobaeza:
Yes, if #1041 is implemented here, then it's enough IMO.

pedrobaeza:
Superseeded by #1290

PR #1164, opened 2019-01-21
[11.0] web_export_view: add security group

chienandalu:
- Add a security group to disallow users to export data from the view

cc @Tecnativa

### DIFF
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -16,8 +16,12 @@ odoo.define('web_export_view', function (require) {
             var self = this;
             this._super.apply(this, arguments);
             if (self.getParent().renderer.viewType == 'list') {
-                self.$el.find('.o_dropdown').last().append(QWeb.render('WebExportTreeViewXls', {widget: self}));
-                self.$el.find('.export_treeview_xls').on('click', self.on_sidebar_export_treeview_xls);
+                session.user_has_group('web_export_view.group_disallow_export_view_data_excel').then(function (has_group) {
+                    if (!has_group) {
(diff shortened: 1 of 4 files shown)


tarteo:
Is this https://github.com/OCA/web/tree/11.0/web_disable_export_group not the same?

pedrobaeza:
No, that one is a complement. This one is restricting the new button at list level that this module adds that allows to export directly "what you see".

web_export_view/static/src/js/web_export_view.js
```diff
@@ -16,8 +16,12 @@ odoo.define('web_export_view', function (require) {
             var self = this;
             this._super.apply(this, arguments);
             if (self.getParent().renderer.viewType == 'list') {
-                self.$el.find('.o_dropdown').last().append(QWeb.render('WebExportTreeViewXls', {widget: self}));
-                self.$el.find('.export_treeview_xls').on('click', self.on_sidebar_export_treeview_xls);
+                session.user_has_group('web_export_view.group_disallow_export_view_data_excel').then(function (has_group) {
```
pedrobaeza, line 19: Why calling again `has_group` if you have already called `user_has_group`?

chienandalu: No, `user_has_group` returns a deferred object that we need to resolve in the proper method. `has_group` is the boolean value we're evaluating after all.

pedrobaeza: Sorry for the noise. My lack of JS knowledge...

pedrobaeza (approved):
Add yourself to contributors in the README.

tarteo (approved):

PR #1165, opened 2019-01-22
[ADD] web_widget_iframe: new module to display an iframe in a form view

benwillig:
This modules simply adds an form widget which display an iframe.


web_widget_iframe/README.rst
```diff
@@ -0,0 +1,74 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/web-10-0/web-10-0-web_widget_iframe
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/162/10.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module simply adds a new widget which display an Iframe for a given
```
tarteo, line 28: Can you add an usage section

web_widget_iframe/static/src/js/web_widget_iframe.js
```diff
@@ -0,0 +1,25 @@
+odoo.define('web_widget_iframe.web_widget_iframe', function(require) {
```
tarteo: Can you rename the namespace to web_widget_iframe.FieldUrlIframe

web_widget_iframe/static/src/js/web_widget_iframe.js
```diff
@@ -0,0 +1,25 @@
...
+
+    var core = require('web.core');
+
+    var FieldChar = core.form_widget_registry.get('char');
+
+    var FieldUrlIframe = FieldChar.extend({
+        template: 'FieldIframe',
+
+        render_value: function() {
+            var self = this;
```
tarteo, line 12: There's not really a reason for this variable you can just use `this`.

web_widget_iframe/static/src/js/web_widget_iframe.js
```diff
@@ -0,0 +1,25 @@
+odoo.define('web_widget_iframe.web_widget_iframe', function(require) {
+    "use strict";
+
+    var core = require('web.core');
+
+    var FieldChar = core.form_widget_registry.get('char');
```
tarteo, line 6: You can also use FieldUrl which maybe automatically adds the protocol etc (don't know for sure :P).

tarteo (changes requested):
Very nice! Can you also look at the lint errors?

pedrobaeza:
Isn't this a bit of security whole allowing any external object to catch data from backend?

hbrunn:
you should set `sandbox=""`, cf https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe to have this locked down by default and possibly add an option to the widget to selectively add permissions

benwillig:
@hbrunn Thanks for your feedback. I added the empty sandbox attribute by default and it can be overriden by specifying it on the field tag.

@tarteo Fixed the code based on your notes.

@pedrobaeza What do you mean?

pedrobaeza:
@benwillig the sandbox mode suggested by @hbrunn answers my concern.

web_widget_iframe/README.rst
pedrobaeza: Yes, please, with the `sandbox` part included.

tarteo (approved):

PR #1166, opened 2019-01-23
[ADD] web_widget_embed: new module to display an embedded document in a form view

benwillig:


pedrobaeza:
Shouldn't it be `web_widget_embedded`?

benwillig:
The HTML tag is named ```embed```.

pedrobaeza:
OK then

sudhir-erpharbor:
It would be better if you add sample code (example) to use the widget.
Would be best if you add the screenshots as well in the README file.

bosd:
@benwillig Can you force push to create a new runboat?

PR #1169, opened 2019-02-02
[11.0][FIX] web_export_view: empty column w/ group by #1099

### LINKED ISSUE #1099: [11.0][BUG] web_export_view: empty column w/ group by on
In any list view with group, if you export view, the excell generated add a void field on header.

contacts odoo

export

Maybe the problem is related with javascript that get header names, and don't take care agout grouping add a void field on header.

https://github.com/OCA/web/blob/11.0/web_export_view/static/src/js/web_export_view.js#L53

Tardo:
Possibly fix #1099

### DIFF
--- a/web_export_view/readme/ROADMAP.rst
+++ b/web_export_view/readme/ROADMAP.rst
@@ -1,0 +1,9 @@
+Pedro M. Baeza (pedro.baeza@tecnativa.com):
+When you have groups, they are not exported to Excel. It would be desirable to have this option.
+One of the problems with this module is that you can't export data from a view with mode="tree".
+Changing the approach to have the button always visible (we should relocate it also to another place,
+as the current location is not visible for these views), and digging correctly in the DOM elements
+for this view (very similar to the normal tree view one) will do the trick. This will also help users
+to locate the feature, as it's hidden now by default and users don't think about selecting records.
+The behavior will be: nothing selected > you export all (including groups).
+Something or all selected: export the selection.
--- a/web_export_view/readme/USAGE.rst
+++ b/web_export_view/readme/USAGE.rst
@@ -1,0 +1,3 @@
+After you installed it, you'll find an additional link 'Export current view'
+right on the sidebar. By clicking on it you'll get a XLS file contains
+the same data of the tree view you are looking at, headers included.
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -1,5 +1,5 @@
 odoo.define('web_export_view', function (require) {
-"use strict";
+    "use strict";
 
     var core = require('web.core');
     var Sidebar = require('web.Sidebar');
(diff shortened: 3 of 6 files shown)


pedrobaeza:
I think this is happening also in v10, isn't it? Maybe you can provide also the fix there.

Tardo:
@pedrobaeza 10.0 appears to work fine. In 11.0 doesn't work in the same way, can't use "string" property because, for some reason, are not set on all columns (on my tests only in the first) :/

pedrobaeza:
Then is this completed?

Tardo:
For me, yes.

web_export_view/static/src/js/web_export_view.js
```diff
@@ -50,14 +50,16 @@ odoo.define('web_export_view', function (require) {
             var export_columns_names = [];
             var column_index = 0;
```
yajo, line 55: Wouldn't it be the same to just put a `1` here? :thinking:

Tardo: Nope :/ The problem is "export_columns_keys"... if you set to 1 in non-grouped views you get an exception... and in grouped you see good headers but bad column content.
This fix really is a "workaround". For me a best solution would be to improve "ListRenderer" and store header strings, avoiding css query selector usage.

Tardo:
Launching flare... What does this PR need to close?

pedrobaeza:
Well, reviewers. I can approve it blindly trusting you, but I prefer @Yajo "JS master" 's opinion

web_export_view/static/src/js/web_export_view.js
yajo: This will turn it `1` when it is grouped, and `0` otherwise:

```suggestion
             // Skip 1st empty header cell if list is grouped
             var column_index = view.renderer.state.groupedBy.length && 1;
```

web_export_view/static/src/js/web_export_view.js
```diff
@@ -50,14 +50,16 @@ odoo.define('web_export_view', function (require) {
...
             var column_header_selector;
+            var isGrouped = view.renderer.state.groupedBy.length > 0;
```
yajo, line 57: ... then you can remove this change.

web_export_view/static/src/js/web_export_view.js
```diff
@@ -50,14 +50,16 @@ odoo.define('web_export_view', function (require) {
...
             $.each(view.renderer.columns, function () {
                 if (this.tag == 'field' && (this.attrs.widget === undefined || this.attrs.widget != 'handle')) {
                     // non-fields like `_group` or buttons
                     export_columns_keys.push(column_index);
-                    column_header_selector = '.o_list_view > thead > tr> th:not([class*="o_list_record_selector"]):eq('+column_index+')';
+                    var css_selector_index = isGrouped?column_index+1:column_index;
+                    column_header_selector = '.o_list_view > thead > tr> th:not([class*="o_list_record_selector"]):eq('+css_selector_index+')';
```
yajo: ... and undo this change

web_export_view/static/src/js/web_export_view.js
```diff
@@ -50,14 +50,16 @@ odoo.define('web_export_view', function (require) {
...
                     export_columns_names.push(view.$el.find(column_header_selector)[0].textContent);
                 }
-                column_index ++;
+                ++column_index;
```
yajo, line 71: ... and this one too.

yajo (changes requested):
Nothing special, just minifying the diff :blush:

simahawk (approved):
Approving @Yajo 's version :wink: 
I mean: fix it as he proposed and will be fine.

simahawk:
also, when you fixup pls squash and amend commit msg to contain module's name + upgrade version by 0.0.1. Thanks :)

Tardo:
@Yajo @simahawk the proposed improvement doesn't work properly. Explained in my previous message ;)
All other changes done!

web_export_view/readme/ROADMAP.rst
```diff
@@ -0,0 +1,12 @@
+Known Issues
+============
```
yajo: This header will fail. Remove it.

web_export_view/readme/ROADMAP.rst
```diff
@@ -0,0 +1,12 @@
...
+
+Pedro M. Baeza (pedro.baeza@gmail.com):
```
yajo: wrong email, is `@tecnativa.com`

web_export_view/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+After you installed it, you'll find an additional link 'Export current view'
+right on the sidebar. By clicking on it you'll get a XLS file contains
+the same data of the tree view you are looking at, headers included.
+
+Configuration
+=============
```
yajo: This header will fail. Use `CONFIGURATION.rst` file instead.

yajo (changes requested):
Yikes, thanks for the README and linter update, although its hard to differentiate now between the real fix to the problem and the cosmetic update. Could you at least split that in 2 commits?

Also, I cannot find the comment where you explain why the suggestion doesn't work. :roll_eyes:

Tardo:
@Yajo done!
Here 🙈: https://github.com/OCA/web/pull/1169#discussion_r253539462

yajo (approved):

yajo (changes requested):
OK, sorry I had to re-read that answer more slowly; I understand now. Just one more little thing to merge:

web_export_view/README.rst
```diff
@@ -1,81 +1 @@
...
-.. image:: http://odoo-community.org/logo.png
-   :alt: Odoo Community Association
-   :target: http://odoo-community.org
-
-This module is maintained by the OCA.
-
-OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
-
-To contribute to this module, please visit http://odoo-community.org.
+This file will be modified by OCA gen-addon-readme script.
```
yajo: You have 2 options here:

1. Do not touch this file. The bot will generate it after merging.
2. Generate the readme manually and commit it.

But please don't do this. :sweat_smile:

yajo (changes requested):
I had to read more slowly your answer, you're right.

This is ready now, just one minor nitpick. :blush:

web_export_view/README.rst
Tardo: Reverted README modifications ;)

yajo: Lol, you added a commit to clean up the diff :laughing: 

NP, I fixed it for you. :wink:

yajo (approved):

PR #1172, opened 2019-02-04
[ADD] make the module play nice with web_shortcut

hbrunn:
without this, either `web_shortcut` comes first and then we don't have a full width searchbar, or `web_searchbar_full_width` comes first and then we have the start on top of the breadcrumb, not next to it.


simahawk:
@hbrunn a screenshot of before and after helps in such cases :)

tbaden (approved):

PR #1178, opened 2019-02-06
[11.0] [web_company_background_color]

JordiBForgeFlow:
This module allows you to change the color of the navigation bar
depending on the company you are logged into.

Configuration
=============

Go to *Settings / Users & Companies* and select the company.
In the field 'Background color' select the background of your choice.

image


JordiBForgeFlow:
@ageficent @tarteo @Yajo what do you think about this proposal? I'm trying to make it compatible with Odoo CE, OCA web responsive and Odoo EE

pedrobaeza:
Isn't better to add this to `web_responsive` module?

JordiBForgeFlow:
@pedrobaeza not everyone uses the `web_responsive` module. This module would be intended to be used independently.

JordiBForgeFlow:
@pedrobaeza But perhaps you're right. I am not sure

JordiBForgeFlow:
cc @hveficent

pedrobaeza:
> not everyone uses the `web_responsive` module

Really? I don't see why don't you want that module in any case.

JordiBForgeFlow:
If someone uses EE, for example?

pedrobaeza:
But this doesn't serve for them either... as there's no such class there.

tarteo:
I would prefer something more generic like in the website module where you can install a different theme for each company. Then it also doesn't matter if web_responsive / EE is used. And it's more flexible so e.g. the text color can also be altered etc. This way a programmer is needed to make the themes, though.

HviorForgeFlow:
It's fine as a quick feature to ease multi company handle.

As you only change color from background header buttons can be hard to read.

A good approach could be modify dynamically less variables like ```@odoo-view-background-color``` and the module could charge some color templates optimizing user readability.

I'm not sure if it can be done something like http://lesscss.org/usage/#using-less-in-the-browser-modify-variables

Definitely I don't know if it make sense.

simahawk:
> A good approach could be modify dynamically less variables like `@odoo-view-background-color` and the module could charge some color templates optimizing user readability.

hmm but then you have to re-trigger the computation of all LESS files AFAIS. `**`

I agree on having a small module that works for both CE and EE.

I would not tie it to background naming only tho.

We could:

1. use a sparse field and proxy the values that we want customers to edit (like bg in this case).
2. register a CSS like `/custom-styles/company/$ID?ts=$WRITE_TIMESTAMP.css` (to be excluded from cached assets)
3. this path matches a specific route that catches the company, read the config, output a valid css content  and proper headers. Up to you if you want to use LESS o SCSS or pure CSS and a template for it to produce it.

`**` NOTE that if you don't care about re-computing them you can still use the above solution to output `variable.less` file to override the default one.

JordiBForgeFlow:
@simahawk ugh that's something beyond my skills.....

simahawk:
@jbeficent I'm not asking you to do it, I'm trying to get to a versatile and shared solution ;)

yajo:
Odoo has a theme that includes this handy customizer:

captura de pantalla de 2019-02-07 09-47-45

Customizations have to be made via Less, that's for sure. You shoud take a look on how does it work for the frontend and replicate that for the backend (not the GUI, but the style overriding method).

OTOH, doing an addon to customize this seems to me like it is very complex and it can quickly become a feature creep. Usually every customer has some `mycompany_custom` addon, and for most of them this is not important, but if any of them considers it very important, dropping a Less file there might be the easiest way to handle this.

pedrobaeza:
@Yajo we are talking the background in backend theme, not in website.

yajo:
I know, but the assets handling system is the same.

Tardo (changes requested):
Wrong site comment :d

simahawk:
FTR there's something nice happening here https://github.com/OCA/web/pull/1253
@jbeficent I'd keep this closed or in stand by until that one is done and then you can update this or create a new one for v11. 
What do you think?
BTW @Yajo @tarteo your reviews on the other side would be nice :wink:

HviorForgeFlow:
@lreficent can we close this? I think another module was proposed on OCA for this feature.

LoisRForgeFlow:
@hveficent Yes, in v12 this need is covered with web_company_color but we are actually using this module in some installation so I would keep it here. Then, we'll take care when migrating to v12 to move to the formerly mentioned.

PR #1179, opened 2019-02-06
web_environment_ribbon: allow to set ribbon in configuration file

isabellerichard:
Allow to override ribbon settings inside configuration file.

This option will be useful in the case of a restoration of a production database on a staging environment, where we want to see a different ribbon (ie. you want to see "staging" instead of "prod").

### DIFF
--- a/web_environment_ribbon/__manifest__.py
+++ b/web_environment_ribbon/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': "Web Environment Ribbon",
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
     'category': 'Web',
     'author': 'Francesco OpenCode Apruzzese, '
               'Tecnativa, '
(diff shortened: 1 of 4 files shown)


isabellerichard:
Hi, 

What should I do to ensure that codecov checks will be sucessful ?

Regards,

Isabelle

simahawk:
@isabellerichard there are no tests in this module: you can add them ;)

sbidoul:
Have you considered using `server_environment_ir_config_parameter` to achieve this? There is a v12 migration PR: https://github.com/OCA/server-env/pull/15

simahawk:
@sbidoul I was looking for the link... thanks :smile: 
Actually I don't think we need this improvement as we are already living fine w/ server-env.

@isabellerichard I close this, feel free to reopen if needed.
In any case, tests are always welcome, especially if we have none :wink:

pedrobaeza:
Hey, server_environment is a solution that doesn't fit all tastes, so can be the case of @isabellerichard.

isabellerichard:
Hi, thanks for the suggestion, it's ok for me!

PR #1180, opened 2019-02-07
[ADD] web_database_rollback: allow rollback database changes

mcojocaru:


pedrobaeza:
Hi, thanks for contributing, and I don't want to demotivate you, but I don't think this specific module is a good idea. You have other mechanism for doing the same, performing duplications of the DB. And the limitation of workers notes that this is not intended for real usages.

tarteo:
@pedrobaeza It can be nice for developers (that are too lazy to make unit tests :smile:) to test some code. I think it's supposed to be for local development environment only and should never be installed in production.

I think this belongs in server-ux though.

mcojocaru:
@tarteo , this module is indeed more useful for developers, to be used for local dev environment. I will close this PR and open it in the server-ux repo.
 
@pedrobaeza, working with large databases (10GB+) can take minutes to create a backup, and if you just want to check some odoo settings, this module would help to see quickly the results and if they are not what you expect, just rollback and have the initial db state. and try again with other odoo settings.

PR #1183, opened 2019-02-08
[12.0][ADD] web_edit_user_filter: Easy adjust user filters and facets

tarteo:
In standard Odoo you can edit user filters via the debug module.
Normal users often don't have access to this menu therefore can't adjust filters once they're saved.
This module makes this feature available for normal users with a user friendly interface.
It also adds the ability to adjust facets (a single part of the filter).

Usage
=====

Edit a favourite filter:

1. Go to a list or kanban view;
2. open the advanced search options;
3. open the 'Favorites' menu;
4. click on the pencil icon to start editing the filter.

Edit a facet:

1. Click on the facet;
2. a menu is now shown which allows you to remove values from the facet;
3. to cancel removal you can click on the button 'Done'.

edit_facet


pedrobaeza:
This sounds great! Maybe using the widget domain/web_advanced_search is the ideal to complete this.

tarteo:
@pedrobaeza I'm not sure how to integrate the DomainSelectorDialog into this module. Maybe when on a saved filter is clicked there's an extra button to edit it in the dialog. But i think it would be pretty hard to do. I can add it to the roadmap if you want.

pedrobaeza:
Well, maybe @Yajo has some ideas.

web_edit_user_filter/models/ir_filters.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class IrFilters(models.Model):
+    _inherit = 'ir.filters'
+
+    facet = fields.Text()
+
+    @api.model
+    def get_filters(self, model, action_id=None):
+        res = super().get_filters(model, action_id)
```
yajo, line 14: :thinking: I don't understand what this whole file is doing. I feel like all of it could be done via JS, which is where all filters are constructed and displayed. Can't this be done without this backend code?

Otherwise, in any case it would be helpful some comments explaining this code.

web_edit_user_filter/static/src/xml/backend.xml
```diff
@@ -0,0 +1,34 @@
...
+                    </span>
+                    <button class="btn btn-sm btn-link pull-right">
+                        <span class="fa fa-close" />
+                    </button>
+                </a>
+            </t>
+        </div>
+
+        <button type="button" class="btn btn-primary btn-block mt-2 btn-done">
+            Done
```
yajo: It seems to me like this button is not needed. Just clicking outside the bubble should close it.

yajo (approved):
It's great, thanks!

There are just a couple details missing:

1. When user hits the pencil button and edits the filter, the only way to update it later is deleting the previous one and saving it again with the same name. Some "update" button, or the ability to save it with the same name again and trigger that as an update, would be great.
1. When the facet is editable, the cursor should become a pointer over it.

No blocking comments though

web_edit_user_filter/models/ir_filters.py
tarteo: It can be done with an extra call but this seemed the easiest and fastest solution. I'll add some comments to the method.

web_edit_user_filter/static/src/xml/backend.xml
tarteo: I removed it and made a simple mechanism to hide the popovers.

tarteo:
> 
> 1. When user hits the pencil button and edits the filter, the only way to update it later is deleting the previous one and saving it again with the same name. Some "update" button, or the ability to save it with the same name again and trigger that as an update, would be great.
> 2. When the facet is editable, the cursor should become a pointer over it.
> 

I've added 1 to the roadmap and 2 is done. :+1: Thanks for the review.

yajo:
I think opening one bubble should close the others:

peek 13-02-2019 13-06

Also, if there's only 1 element in the facet, maybe it doesn't make much sense to add this feature to it. The user can just delete it with the button Odoo has.

tarteo:
@Yajo Oh yes done, thanks. I keep the popover when there's only 1 value so when values become editable nothing has to be changed.

foutoucour (approved):

web_edit_user_filter/readme/ROADMAP.rst
```diff
@@ -0,0 +1 @@
+* Make individual values in facets editable.
```
yajo, line 1: > > 1. When user hits the pencil button and edits the filter, the only way to update it later is deleting the previous one and saving it again with the same name. Some "update" button, or the ability to save it with the same name again and trigger that as an update, would be great.
> 
> I've added 1 to the roadmap

I think you didn't get the point :thinking:. I was referring to the fact that you cannot update a favorite filter; you have to delete it and save it again instead:


peek 14-02-2019 08-40

This fact is a little bit confusing for a new user, so it should be probably documented here.

Making individual values in facets editable would be good, but the fact that you can only remove them and you can add a new one using normal odoo search is, instead, very straightforward and quick to understand with the UI you designed. :blush:

yajo (approved):
It works fine now, thanks!
peek 14-02-2019 08-41

web_edit_user_filter/readme/ROADMAP.rst
tarteo: I understand now :smile:

I have some ideas to do this:
1. The pencil icon changes into a floppy disk icon (or can be added next to it) for the filter that is being edited. If it's pressed the filter will be overwritten.
2. or with a checkbox 'Overwrite' in "Save current search" which overwrites the filter if the name already exists.

tarteo: I like 1 more, what do you think?

PR #1185, opened 2019-02-09
[11.0][MIG] web_widget_text_markdown

Tardo:
Changes from 10.0:
- Removed CSS
- Removed Templates
- Removed 'highlight' (JS library)
- Updated Readme
- Updated JS dependencies
- Now it's based on ```FieldText``` instead of ```AbstractField```
- Roadmap: Now supports drag&drop images (ir.attachment)
- Roadmap: Enabled dropzone (to support drag&drop images)


Tardo:
@pedrobaeza @tarteo @Yajo @simahawk This need tests?

pedrobaeza:
Well, QUnit test would be awesome for UI, and for backend part (the upload method), a regular test is also welcome, but it's not strictly required.

web_widget_text_markdown/views/main.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
simahawk: rename this file to `templates/assets.xml` pls

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
```diff
@@ -0,0 +1,101 @@
...
+            LIBS_PATH + '/bootstrap-markdown.min.css',
+        ],
+
+        _getValue: function () {
+            return this.$markdown.getContent();
+        },
+
+        _prepareInput: function () {
+            var $result = this._super.apply(this, arguments);
+            _.defer(function ($elm) {
```
simahawk, line 37: not a blocker but some thing might be split out to ease override (eg: dropzone params).

web_widget_text_markdown/controllers/web_main.py
```diff
@@ -0,0 +1,44 @@
...
+                # fail
+                filename = unicodedata.normalize('NFD', ufile.filename)
+
+            try:
+                attachment = Model.create({
+                    'name': filename,
+                    'datas': base64.encodestring(ufile.read()),
+                    'datas_fname': filename,
+                    'res_model': model,
+                    'res_id': int(id)
```
simahawk: pls rename `id` to `res_id`.
Also, I wonder: when you are creating a record you won't have a real id here... hence you'll lose the reference.

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
```diff
@@ -0,0 +1,101 @@
...
+                    dropZoneOptions: {
+                        paramName: 'ufile',
+                        url: '/web_widget_text_markdown/upload_attachment',
+                        acceptedFiles: 'image/*',
+                        params: {
+                            csrf_token: core.csrf_token,
+                            session_id: this.getSession().override_session,
+                            callback: '',
+                            model: this.model,
+                            id: this.res_id,
```
simahawk: watch out, this is not a real ID on record create. Either we handle it on final create/write or we state clearly in the readme that we cannot use image upload on create.
Am I wrong?

simahawk (changes requested):

Tardo:
@simahawk Done, thanks!
@pedrobaeza I added some tests but i don't know why travis doesn't like them... and i don't know how to run test for submodules. "fields > FieldTextMarkdown"

pedrobaeza:
It seems the tour is not correctly configured for reading data. Isn't QUnit tests already in v11?

web_widget_text_markdown/controllers/main.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class MarkdownController(http.Controller):
+    # This is a modified method from web>controllers>main 'upload_attachment'
+    @http.route(
+        '/web_widget_text_markdown/upload_attachment',
+        type='http', auth="user")
+    def upload_markdown_attachment(self, model, res_id, ufile):
+        attachment_obj = request.env['ir.attachment']
+        filename = unicodedata.normalize('NFD', ufile.filename) \
```
simahawk: would be more readable like:

```suggestion
        filename = ufile.filename
        if request.httprequest.user_agent.browser == 'safari':
            filename = unicodedata.normalize('NFD', filename)
```

web_widget_text_markdown/controllers/main.py
```diff
@@ -0,0 +1,32 @@
...
+
+        try:
+            attachment = attachment_obj.create({
+                'name': filename,
+                'datas': base64.encodestring(ufile.read()),
+                'datas_fname': filename,
+                'res_model': model,
+                'res_id': int(res_id)
+            })
+        except Exception:
```
simahawk: can we track any specific exception?

web_widget_text_markdown/controllers/main.py
```diff
@@ -0,0 +1,32 @@
...
+            _logger.exception(
+                "Fail to upload attachment %s" % ufile.filename)
+        else:
+            return "/web/image/%d" % attachment.id
```
simahawk: I'd return always a dict cointaining: 
* success: true|false
* url: (if success)
* msg: (handle custom messasge but also the error message in case success = false (you can catch it in the exception)

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+
+            return markdownOpts;
+        },
+
+        _markdownUploadSuccess: function () {
+            this.isDirty = true;
+            this._doDebouncedAction();
+            $(".dz-success-mark svg").css({
```
simahawk: can we have SCSS file for this?
This way you can use standard odoo colors (or theme colors) and ease customization.
Not mandatory.

simahawk (commented):
Thanks for the fixups :)
Sorry, I have a few more :stuck_out_tongue_winking_eye:

web_widget_text_markdown/controllers/main.py
Tardo: Removed method... now uses the Odoo 'original' (looks a bit hackish, but not repeated code for the same purpose)

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
Tardo: Ok! but v11 uses only less no?

Tardo:
@pedrobaeza Other modules uses it, for example: https://github.com/OCA/OCB/blob/11.0/addons/calendar/static/tests/calendar_tests.js#L9
Perhaps this is not really necessary? https://github.com/OCA/web/pull/1185/files#diff-8c77323714c54ade2edecc55bf665a33R10

pedrobaeza:
AFAIK, tours are intended for website tests, but QUnit are for backend ones, but I'm not sure if you can do the latter with tours, as I'm not an expert on it.

web_widget_text_markdown/static/src/less/web_widget_text_markdown.less
```diff
@@ -0,0 +1,18 @@
+.dz-preview {
```
simahawk: would be better to isolate the styles by wrapping them in the widget's class

simahawk (changes requested):
LG but tests are red 

`2019-03-21 17:32:52,866 6338 ERROR openerp_test odoo.addons.web_widget_text_markdown.tests.test_ui: ` AssertionError: Ran out of data to read`

web_widget_text_markdown/static/src/less/web_widget_text_markdown.less
Tardo: bootstrap markdown doesn't have an option to add a custom class :/
These CSS rules are generic, other more specific can overwrite them.

Tardo:
@simahawk I don't know what is wrong with the tests... i need help to solve this :/ on local environment works fine.

web_widget_text_markdown/static/src/less/web_widget_text_markdown.less
simahawk: I'm talking about the odoo widget class: you should have a specific class for it and the dropzone elements should stay into the widget element, so that you could wrap styles into `.o_widget_text_markdown *` or similar.

simahawk (commented):
@Tardo  are you sure you are running tests properly?
It seems the build goes time out because you are running too many things.

web_widget_text_markdown/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module adds a new widget for text field in form view on Odoo:
+
+- In readonly mode, it uses text contents to parse and render them to html markdown syntax.
+- In write mode, use [bootstrap-markdown][1]
+
+[1]: https://github.com/toopay/bootstrap-markdown    "bootstrap-markdown"
```
elicoidal:
```suggestion
[1]: `bootstrap-markdown <https://github.com/toopay/bootstrap-markdown>`_
```

web_widget_text_markdown/readme/ROADMAP.rst
```diff
@@ -0,0 +1,5 @@
+* Improve user experience with Odoo specific syntax
+* Improve user experience with Github specific syntax
+
+**KNOWN ISSUES**
+* Can't create attachments on virtual records
```
elicoidal: add empty line

web_widget_text_markdown/views/assets.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- * Copyright 2014 Sudokeys <http://www.sudokeys.com>
+ * Copyright 2017 Komit - <http:///komit-consulting.com>
+ * Copyright 2019 Alexandre Díaz - <dev@redneboa.es>
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).-->
```
elicoidal, line 5: Rest of the module is LGPL: can you recheck this one?

web_widget_text_markdown/static/src/lib/bootstrap-markdown.js
```diff
@@ -0,0 +1,1575 @@
+/* ===================================================
```
elicoidal, line 1: Isnot there a possibility not to embed the external library in the module and use better an external dependency?

elicoidal (commented):
some nitpicking

web_widget_text_markdown/readme/DESCRIPTION.rst
Tardo: This is a markdown syntax and i don't get the same result with this changes :/

web_widget_text_markdown/views/assets.xml
Tardo: Rest of the module is AGPL... i will change all to AGPL ok? (other modules uses AGPL too)

web_widget_text_markdown/static/src/lib/bootstrap-markdown.js
Tardo: You mean use a CDN? Odoo can work on 'offline' environments (doesn't need an internet connection to run).
If you have security concerns can use tools like 'sha1sum' to check files integrity.

Tardo:
@simahawk I think so, yeah.
Only JS, using QUnit view:
http://localhost:11069/web/tests?failfast&moduleId=636ea481  (i don't know if the "moduleId" is only valid for me)
PhantomJS (unittest):
https://github.com/Tardo/oca_contrib/blob/master/oca_contrib.sh#L142

Using "PhantomJS (unittest)" way i get fails but related with 'fields > relational_fields > FieldMany2One'

web_widget_text_markdown/readme/DESCRIPTION.rst
elicoidal: The README file is rst format not md: http://docutils.sourceforge.net/docs/user/rst/quickref.html#hyperlink-targets

web_widget_text_markdown/views/assets.xml
elicoidal: this is indeed consistent with 10

web_widget_text_markdown/readme/DESCRIPTION.rst
Tardo: Oops! done, thx! me confused with available markup languages :S

Tardo:
Launching flare... I can't resolve travis problems. I use the same tests configuration on other module and all works fine (local & travis).

web_widget_text_markdown/tests/test_ui.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2019 Alexandre Díaz <dev@redneboa.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.tests import HttpCase
+
+
+class TestUi(HttpCase):
+
+    post_install = True
+    at_install = False
```
yajo: You shouldn't need these.

web_widget_text_markdown/tests/test_ui.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def test_ui_web_field_text_markdown(self):
+        self.phantom_js(
+            "/web/tests?module=fields&failfast",
```
yajo: This is weird. Maybe you need this?:

```suggestion
            "/web/tests?module=web_widget_text_markdown&failfast",
```

web_widget_text_markdown/static/tests/js/web_widget_text_markdown.js
```diff
@@ -0,0 +1,72 @@
...
+odoo.define('web_widget_text_markdown.test', function (require) {
+    'use strict';
+
+    var FormView = require('web.FormView');
+    var testUtils = require('web.test_utils');
+
+    var createAsyncView = testUtils.createAsyncView;
+
+
+    QUnit.module('fields', {
```
yajo: Shouldn't you declare a new namespace for your tests?

```suggestion
    QUnit.module('web_widget_text_markdown', {
```

yajo (commented):
I'm just giving you some clues on where could the problem be, although I'm not actually sure because [the travis error](https://travis-ci.org/OCA/web/jobs/513893096#L1129-L1131) is weird. In fact, this kind of errors make me think that we have an ES6 problem. Loading ES6 code, in a minimized file (as Odoo does), makes an old engine (like PhantomJS) to fail loading the whole file. Phantom will not tell you about the syntax error and fail silently. I can't seem to find that code anywhere, I don't know if it could be in a lib.

Do some research on that and suggestions below; let's see if we find it. :wink:

Tardo:
@simahawk Travis green.

Can be tested on runbot in 'res.partner > Internal Notes' (if no one rebuild)

Masthercell-IT:
Tested on runbot, It's ok

Good job

simahawk (approved):

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
```diff
@@ -0,0 +1,159 @@
...
+        _markdownDropZoneInit: function (markdown) {
+            var self = this;
+            var caretPos = 0;
+            var $textarea = null;
+            markdown.on('drop', function (e) {
+                $textarea = $(e.target);
+                caretPos = $textarea.prop('selectionStart');
+            });
+            markdown.on('success', function (file, response) {
+                var text = $textarea.val();
```
daramousk, line 121: When trying to upload a file I get:
`TypeError: Cannot read property 'val' of null`

Also, it seems that not all the file types can be uploaded ex. use drag-drop on a pdf file and an error is raised `You can't upload files of this type.`. Since this an unhandled exception, the web client renders it as if a crash happened. Perhaps we can catch that and raise it diferently

web_widget_text_markdown/README.rst
```diff
@@ -0,0 +1,105 @@
...
+    ...
+    <field name="field_name" widget="bootstrap_markdown"/>
+    ...
+
+Known issues / Roadmap
+======================
+
+* Can't create attachments on virtual records
+
+* Enable dropzone support (Disabled for support IE11)
```
daramousk: Perhaps now is the time to enable dropzone support? Since the latest dropzone version seems to support IE11 https://www.dropzonejs.com/#compatibility

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
Tardo: Thanks, changed to console warning... user can know that is a failed upload with the icons.

web_widget_text_markdown/README.rst
Tardo: Not yet... http://www.odoo.com/documentation/11.0/setup/deploy.html#supported-browsers

Tardo: Yes, sorry... :P dropzone is used to allow attach images ;)

yajo (approved):

Issue #1177, opened 2019-02-05
[12.0] web_responsive: Improvement - Make chatter top bar sticky


pedrobaeza:
I think that to make this sticky can be a good idea, but only when chatter is on right side, and including the attachment widget:

![seleccion_006](https://user-images.githubusercontent.com/7165771/52294732-5593a880-297a-11e9-944d-c75e2a58193e.png)

@Tecnativa

pedrobaeza:
What do you think, @Tardo @Yajo ?

yajo:
I like it.

Tardo:
Looks good! i like it too

pedrobaeza:
Then which volunteer to do it? :angel:

Tardo:
If @Yajo doesn't mind, I can try.

yajo:
Yes, I don't mind at all. Thanks! :wink:

PR #1187, opened 2019-02-12
[12.0][IMP] web_responsive: Sticky chatter topbar

Tardo:
Improve chatter topbar as suggested in https://github.com/OCA/web/issues/1177

Sets sticky at top 0 because topbar uses margin, if preserve these margin, when sticky, conversation can be seen in these space because is an a transparent zone.

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -420,6 +420,22 @@ html .o_web_client .o_main .o_main_content {
                     max-width: 50%;
                     min-width: 30%;
                     overflow: auto;
+
+                    .o_chatter_topbar,
+                    .o_thread_composer {
+                        position: sticky;
+                    }
+
+                    .o_chatter_topbar {
+                        top: 0;
+                        z-index: 2;
+                        background-color: $o-view-background-color;
+                    }
+
+                    .o_thread_composer {
+                        top: $o-statusbar-height;
+                        z-index: 1;
+                    }
                 }
             }
         }
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Web Responsive",
     "summary": "Responsive web client, community-supported",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.1.0",
     "category": "Website",
     "website": "https://github.com/OCA/web",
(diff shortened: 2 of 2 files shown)


tarteo (approved):

tarteo (changes requested):
Please increase version number

Tardo:
@tarteo updated version number to v1.1.0

tarteo (approved):
Thanks!

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -420,6 +420,22 @@ html .o_web_client .o_main .o_main_content {
...
+
+                    .o_chatter_topbar,
+                    .o_thread_composer {
+                        position: sticky;
+                    }
+
+                    .o_chatter_topbar {
+                        top: 0;
+                        z-index: 2;
+                        background-color: white;
```
yajo: It should be `$o-webclient-background-color` instead. Just to make theming easier. :wink:

yajo (changes requested):
Almost perfect! Just one minor requirement.

Apart from the comment below, I've been testing and it seems a little bit confusing that the attachments panel isn't also sticky. This is not a requirement, but I think that if it is easy to fix, it'd be good to fix it too. What do you think?
peek 14-02-2019 08-31

Tardo:
@Yajo idk how to stack sticky elements dynamically only with CSS... the problem it's that the composer and attachments can be show at the same time (or not).

yajo (approved):
Hmm I have been playing around and it seems like the only sane way is to put `.o_chatter_topbar`, `.o_thread_composer` and `.o_mail_chatter_attachments` inside the same div, which should be the sticky one:

peek 15-02-2019 10-53

(Look at the new div highlighted in blue in the dom panel, and its css properties highlighted below in green).

However, after thinking more about it, maybe a record with too many attachments would lose its ability to scroll over them, so better leave as it is.

pedrobaeza:
Thanks to all.

Tardo:
@Yajo Oh! xD good solution :P and can be useful to preserve margin top. For the scroll issue something like this can work:
```
.o_attachments_previews {
  overflow: auto;
  max-height: $o-mail-attachment-image-size * 2.5 * 3.0;
}
```
I try it?

pedrobaeza:
Let's try to see how it's seen!

yajo:
Another solution would be to make the right panel scroll up if the user clicks on clippy.

PR #1194, opened 2019-02-15
[12.0][IMP] web_widget_x2many_2d_matrix: allow HTML usage

alexey-pelykh:
This would give ability to format row/col headers

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
@@ -146,7 +146,7 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
             if (_.isNull(description)) {
                 description = node.attrs.string || field.string;
             }
-            $th.text(description).data('name', name);
+            $th.html(description).data('name', name);
 
             if (
                 field.type === 'float' || field.type === 'integer' ||
@@ -225,7 +225,7 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
                 value = value.data.display_name;
             }
             // Get 1st column filled w/ Y label
-            $td.text(value);
+            $td.html(value);
             return $td;
         },
 
--- a/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
@@ -153,6 +153,8 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
                 field.type === 'monetary'
             ) {
                 $th.addClass('text-right');
+            } else {
+                $th.addClass('text-center');
             }
 
             if (config.debug) {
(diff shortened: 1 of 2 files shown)


alexey-pelykh:
Specifically, to be able to add `&nbsp;` as-is

pedrobaeza:
Why not using `\t` instead?

alexey-pelykh:
`\t` as tab? It's not the same as non-breakable-space, quite opposite

pedrobaeza:
Well, it acts similar in some occasions, but yes, not the same. Your idea is to keep the header as one line as I see. But isn't going to be each column very wide? And more in mobile devices?

alexey-pelykh:
`Mon, Feb&nbsp;15` would break exactly on non-nbsp space

pedrobaeza:
Very fine-tuned, hehe, but OK, understood.

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
```diff
@@ -146,13 +146,15 @@ odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer', function (requ
             if (_.isNull(description)) {
                 description = node.attrs.string || field.string;
             }
-            $th.text(description).data('name', name);
+            $th.html(description).data('name', name);
```
yajo, line 149: By doing this, modules using this module should escape HTML manually, and a simple translation could inject a XSS. I think it's a bad idea in general terms. Have you tried writing i.e. your description as `Mon, Feb 15` (that's a real nbsp there between Feb and 15).

yajo (changes requested):

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
alexey-pelykh: The security concern is valid, yet by not-trusting a module or it's translation (which is a resource for system) within the system is kind-a odd

pedrobaeza: @alexey-pelykh I'm afraid you can't trust anything...

yajo: Anyways the problem of somebody having a header such as `Count < total` becoming suddenly broken unles you change it to `Count &lt; total` is still valid. IMHO we gain much more by autoescaping than by allowing to use nbsp.

alexey-pelykh: I guess @Yajo is right, we'd gain more issues than benefits for it. I'll try out the unicode nbsp symbol w/o html and see if it will work, otherwise I think this PR would get transformed to `html(escape(input))` in order to translate unicode nbsp to `&nbsp;` somehow

yajo: That seems a better idea, yes. In any case, before continuing I think the best is that you test if just pasting a raw nbsp character works for you, and in such case this can be closed.

alexey-pelykh:
@Yajo you were right, thanks for the hint!

Issue #1189, opened 2019-02-14
[12.0] web_responsive: Shrinking window with sided chat hides some buttons


yajo:
Just see those buttons at the right, for followers, attachments and notifications:

![peek 14-02-2019 08-34](https://user-images.githubusercontent.com/973709/52773696-796a8480-3033-11e9-9a07-57854ad0b0e9.gif)

I should fix this, but in case someone wants to volunteer, this one should be easy.

gaikaz:
https://github.com/Tardo/web/commit/ac656d7499207a76db33f10bbe486ace91b12f91 fixed the sheet going under the sideded chatter, but chatter toolbars items getting misplaced (the described issue) problem is still there. Should this be re-opened as a new issue?

yajo:
The fix is in a009528. If you see further problems please publish a new issue. Thanks!

gaikaz:
Sorry to bother, but I double checked now on the runbot (and locally). The exact problem is still there.
Fresh screenshot from runbot:
![3367603-12-0-501adc runbot3 odoo-community org_web_debug=1](https://user-images.githubusercontent.com/17097810/57128357-d1e03100-6d9b-11e9-961b-2c819f5e24c9.png)

As this is not a different problem, I think it's best discussed here. If not, I can publish a new issue and could even try to fix it myself. Already have been fiddling with CSS here and might have a solution.

yajo:
OK, yes it seems the same. Then may you open the PR?

gaikaz:
This can probably be closed now as it's fixed by #1305

pedrobaeza:
Yes, thanks for the reminder.

PR #1198, opened 2019-02-19
[12.0][IMP] Chatter Topbar

Tardo:
- Preserve margin top
- Sticky attachments (thx to @Yajo) **[$o-mail-attachment-image-size depends on 'web' module]**
- Toggle composer: Now click on tab toggle composer visibility. **[Chatter modifications depends on 'web' module]**
- Fix issue #1189

Best not use $o-mail-attachment-image-size and don't touch Chatter?

Continue https://github.com/OCA/web/pull/1187


web_responsive/static/src/xml/form_view.xml
```diff
@@ -138,4 +138,14 @@
         </t>
     </t>
 
+    <t t-extend="mail.Chatter">
+        <t t-jquery=".o_chatter_topbar" t-operation="replace">
+            <div class="o_chatter_header_container">
+                <div class="o_chatter_topbar">
```
yajo, line 144: Use `<t t-raw="0"/>` instead. It's the system they added to wrap elements.

yajo (approved):
Awesome.

I'm [rebuilding 12.0 branch](https://travis-ci.org/OCA/web/builds/493719384) to see if travis failure is related to this PR or to some random change in odoo itself.

These weeks it seems like they've decided to destroy all possible tests out there... :disappointed:

pedrobaeza:
Travis is red, and we can add this feature in the README with the same fancy GIF animation style

web_responsive/static/src/xml/form_view.xml
Tardo: How? i don't know how use raw="0" in this context :S

Tardo:
How i can record gifs with the quality and size of used in the README?
I'm triying using gimp, ffmpeg, gifscile, gifski and convert ... but i can't get the same results xD

pedrobaeza:
I use peek https://github.com/phw/peek. It works very well.

Tardo:
I can't use it... Openbox interpret all the window as "not event zone" and i can't resize, move, click on all the window :(

pedrobaeza:
Can you switch to other window manager? @Yajo maybe you can provide it.

yajo:
> I can't use it... Openbox interpret all the window as "not event zone" and i can't resize, move, click on all the window :(

Are you using wayland? On Xorg it will work probably.

Rebuilding runbot, just in case.

web_responsive/static/src/xml/form_view.xml
yajo: It allows to move the target element that is being replaced to a different place. In your case, it would be:

```xml
<t t-jquery=".o_chatter_topbar" t-operation="replace">
    <div class="o_chatter_header_container">
        <t t-raw="0"/>
    </div>
</t>
```

Tardo:
Yep, use x11 but doesn't work properly (Manjaro Openbox).
Anyway I hit the key to generate good quality gifs and weight :)
https://gist.github.com/Tardo/b20c2f0c7908d2208a9e3075245ff6af

web_responsive/static/src/xml/form_view.xml
Tardo: Ummm doesn't work... I understand the "t-raw='0'" on a "t-call" context... but here i can't see it :/

This change, without something like "t-raw='0'", is incompatible with other modules that modify the chatter, right? It will depend on the order of the dependencies/installations...

yajo: Yikes! Sorry I was mixing 2 concepts. :confounded: 

You're right. I was meaning to use [`position="move"`](https://www.odoo.com/documentation/12.0/reference/views.html#inheritance-specs), but after a little research, it seems that's only available for backend Qweb, so there's nothing that can be improved here.

yajo:
I have triggered [a rebuild for the 12.0 branch](https://travis-ci.org/OCA/web/builds/496966442) to see if the error is from this PR or not.

Tardo:
Auto-Closed :scream:

pedrobaeza:
Yes, GitHub does that when merging the linked PR

Tardo:
Can re-open? Has closed it without merging the changes proposed in this pr :/

yajo:
I think you have to rebase.

pedrobaeza:
This is the culprit for being auto-closed:
seleccion_024
Please use other text next time

Tardo:
Launching flare... What does this PR need to close?

pedrobaeza:
Isn't there any possibility of avoiding `mail` dependency? I know "everyone" install this module, but for trying to decouple the maximum.

Tardo:
@pedrobaeza yes, it's possible avoid mail dependency (Odoo works normally if can't find the template to extend)...
 
The main problem is that the "replace" operation makes the module "incompatible" with other modules that changes mail.Chatter template.
https://github.com/OCA/web/pull/1198/files#diff-1d4576177cd6d759ee794bcba371ea3bR141

pedrobaeza:
> The main problem is that the "replace" operation makes the module "incompatible" with other modules that changes mail.Chatter template.

I don't get that part... The incompatibility is going to be the same due to the `replace` operation, nevertheless the dependency IMO. What you can do is to put that on a template with a very high priority number for being always processed the latest.

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -55,3 +55,7 @@ Features for computers:
 * Full width form sheets
 
   .. image:: https://user-images.githubusercontent.com/973709/48417428-ac5a5f00-e74a-11e8-8839-5bc538c54c1d.png
+
+* Sticky chatter topbar
+
+  .. image:: ./images/chatter_topbar.gif
```
yajo: May I suggest you to link the gif from Github's CDN instead? It will be rendered everywhere and reduce repo size. (Sometimes relative links are interpreted differently on odoo itself, on the apps store, on github, etc.)

Also, remove the gif file from the PR.

```suggestion
  .. image:: https://user-images.githubusercontent.com/973709/55539179-f8854c00-56b7-11e9-9dc5-1a060123edf3.gif
```

It's using the same URL Github assigns when drag & drop this image here:

chatter_topbar

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -421,20 +425,21 @@ html .o_web_client .o_main .o_main_content {
...
                     overflow: auto;
 
-                    .o_chatter_topbar,
-                    .o_thread_composer {
-                        position: sticky;
-                    }
-
-                    .o_chatter_topbar {
+                    .o_chatter_header_container {
+                        padding-top: $grid-gutter-width*0.5;
```
yajo: Add beautiful spacing please :relaxed: 

```suggestion
                        padding-top: $grid-gutter-width * 0.5;
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -421,20 +425,21 @@ html .o_web_client .o_main .o_main_content {
...
-                        top: $o-statusbar-height;
                         z-index: 1;
+
+                        .o_chatter_topbar {
+                            margin-top: 0;
+                        }
+
+                        .o_attachments_previews {
+                            overflow: auto;
+                            max-height: $o-mail-attachment-image-size * 2 * 3;
```
yajo: :thinking: Why not just this?

```suggestion
                            max-height: $o-mail-attachment-image-size * 6;
```

yajo (changes requested):
That will not work, @pedrobaeza, JS Qweb doesn't have that feature. And in any case, lowering priority just defers the error, doesn't disable it.

We have these options currently:

1. Add dependency, as done here.
1. Use a glue, auto-installable addon. Not sure if it's worth.
1. Separate all mail-dependant code in other JS/XML files. These files will fail to load due to `require()` not resolving. It will probably mean errors in CI.

I've been thinking about this and I think the best solution is the one presented here (number 1). We were already styling mail stuff without adding the dependency because the CSS code doesn't have such requirement, but actually styling things that can't exist without your dependency chain is not much elegant. At the end of the day, everybody installs `mail` because almost every addon depends on it, so this shouldn't be much of a problem IMHO. The other solutions are too bloated for a case that will almost never exist in practical terms.

pedrobaeza:
OK about the dependency then. About the image, that's not the best way, @Yajo, but use README generator relative URLs resolver + adding the image in the repo. See https://github.com/OCA/maintainer-tools/pull/397/commits/77ed2b229ee9e63c9a0ef52b71a9705c48617636 for details of the implementation.

The aim of this is to point to the proper version image. If you did that with current ones, @Yajo , you should change them, or in v13, it will be very difficult to update them to new version.

web_responsive/static/src/css/web_responsive.scss
Tardo: An inaccurate attempt to make changes easier... xD

yajo: Answering to the previous version of your answer, best is always to make it easier for humans as long as it doesn't affect performance.

I'd appreciate a comment then explaining this. Not a blocker though. :blush:

yajo (approved):

yajo:
Ready to merge, @Tardo?

Tardo:
@Yajo for me yes, but i'm a bit lost with image resources. xD

pedrobaeza (approved):
@Tardo you did it the change in the README well!

I have pre-generated the README for you to see how it's changed, and also for putting the image in /static/img for following the guidelines folder structure. I have done also a bit of squashing.

Merging this when green.

Thanks again for your contributions.

PR #1200, opened 2019-02-21
[ADD] web_widget_domain_editor_dialog: New module

chienandalu:
Since v11 introduced the new domain editor widget it's not possible to edit
the selected records from the current domain.

This module reintroduces that dialog to complement the current widget with the
powerful search engine of Odoo.

cc @Tecnativa


pedrobaeza:
Some screenshots or GIFs would be very interesting to have in the README + generate it.

web_widget_domain_editor_dialog/static/src/js/widget_domain_editor_dialog.js
```diff
@@ -0,0 +1,98 @@
+/* Copyright 2019 Tecnativa - David Vidal
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define("web_widget_domain_editor_dialog.widget_domain_editor_dialog", function (require) {
```
yajo: I'd call this module `web_wiget_domain_editor_dialog.DomainEditorDialog` and return that object directly.

Another option would be to have a single `web_widget_domain_editor_dialog` js module that performs all actions and returns the dict as it is doing now here.

yajo (approved):
I'm becoming used to attach gifs to Github and then link them directly from there. This way we don't need to keep those gifs in git, and they are visible anywhere also.

All things are suggestions, so OK anyways. Code & functional review.

pedrobaeza:
@Yajo @chienandalu about the GIFs/images, the README generator includes code for unraveling its path and put an absolute one for not having problem with relative paths. Are you aware of that?

yajo:
I'm not aware, but do you think it's better than using a free CDN such as github's?

hbrunn:
using CDNs is a problem for deployments that can't access the internet

pedrobaeza:
It uses GitHub as well, through raw.githubusercontent.com. See the initial commit for this:

https://github.com/OCA/maintainer-tools/pull/397/commits/77ed2b229ee9e63c9a0ef52b71a9705c48617636

pedrobaeza:
@hbrunn too late...

hbrunn:
ah, this is just about the readme. There I think it's fine, but all resources the module needs for functioning it should bring with it

pedrobaeza:
@hbrunn yes, of course libs and so on are still bundled with the module itself. That hasn't changed and we also have that clear.

web_widget_domain_editor_dialog/static/src/js/widget_domain_editor_dialog.js
chienandalu: There you are :slightly_smiling_face:

chienandalu:
All comments attended, thanks for the hints :smile:

chienandalu:
@pedrobaeza I must be doing something wrong or do the image get uploaded to the CDN by the bot once the module is merged

pedrobaeza:
@chienandalu the absolute path is intended for being used when the module is merged. Here in the PR, they are broken links.

chienandalu:
Perfect then :smile:

yajo (approved):

pedrobaeza (approved):

PR #1205, opened 2019-02-27
 [ADD][web_action_conditionable] support forms

hbrunn:
I added a few other improvements in separate commits, you'll see that in the messages

### DIFF
--- a/web_action_conditionable/__openerp__.py
+++ b/web_action_conditionable/__openerp__.py
@@ -11,6 +11,9 @@
         'web',
     ],
     "license": "AGPL-3",
+    "demo": [
+        "demo/ir_ui_view.xml",
+    ],
     'data': [
         'views/view.xml'
     ],
--- a/web_action_conditionable/__openerp__.py
+++ b/web_action_conditionable/__openerp__.py
@@ -2,7 +2,7 @@
 
 {
     "name": 'web_action_conditionable',
-    "version": "8.0.1.1.0",
+    "version": "8.0.1.1.1",
     "category": "Web",
     "website": "https://github.com/OCA/web",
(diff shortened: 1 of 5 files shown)


tarteo:
@pankk

tarteo (changes requested):
`<tree create="'some_group' in _group_refs">` in tree views doesn't hide the 'Create' button anymore.

hbrunn:
@tarteo thanks, didn't consider this use case. Fixed and updated the readme accordingly

tarteo:
The same case is for trees in form views.

hbrunn:
I can't reproduce this locally, can you add an example on runbot?

tarteo:
http://3360805-1205-62dbc4.runbot3.odoo-community.org/web?debug=1#id=3&view_type=form&model=res.groups&action=75

I've changed 

`                                <tree string="Access Rules" editable="top">
`

to

`                                <tree string="Access Rules" editable="top" create="'some_group' in _group_refs">
`

but I can still add items.

hbrunn:
aha! you try this in a many2many field, I believe this never has worked: https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_form.js#L5024

But simple to fix, so I just added that even though not in the scope of this PR.

tarteo (approved):
The previous worked with one2many and many2many fields on my local machine. But this is more clear. Thanks!

Can you increase the version number in the manifest.

tbaden (approved):
code review

pedrobaeza:
@hbrunn can you squash commits at your taste?

hbrunn:
@pedrobaeza done

PR #1209, opened 2019-03-08
[FIX] web_drop_target

daramousk:
Make sure that when a File is dropped into a form the event is propagated and the action only takes place when the element dropped is actually a Blob object.

cc: @hbrunn


hbrunn (approved):
build error unrelated and fixed in https://github.com/OCA/web/pull/1213

simahawk:
@daramousk pls increase version number by 0.0.1

daramousk:
@simahawk it should be fine now, green as well.

simahawk (approved):

PR #1225, opened 2019-03-25
[10.0] web_ckeditor4: fix no re-init on form discard

simahawk:
When you hit "cancel" button or when you navigate away from the form,
for instance by clicking on the breadcrumb or on  "edit translations",
we have to remove the CKEditor widget.

BUT then if you hit "create" Odoo's form machinery is not initializing
the widget anymore, which really sounds inconsistent.
If the widget is not initialized again it means that if CKEditor
got destroyed there's no way to re-init again.

Here we make sure that on create (no id on datarecord)
if the editor is not initialized yet we force it.

Collateral change: moved default configuration to a separated file to make the main one less messy.


Tonow-c2c (approved):
But I'm not JS expert

mpanarin (approved):

PR #1229, opened 2019-03-27
[ADD] web_qr

etobella:
This module allows to print QR in better structure than the standard odoo.

The original image looks like the following
old_qr

With the new generator, it looks like:
new_qr

As you can see, some lines are added on the original QR. As the QR gains complexity, the lines are worse.

With this new method, now you can generate a QR without this kind of lines


etobella:
On a printed report the difference look like this:
image

image

As you can see, the Odoo QR is not readable by a scanner.

tarteo:
Looks good, but why make this a module? IMO it should be improved in https://github.com/odoo/odoo.

etobella:
The change is too big to apply on a stable release for odoo.
Also, it changes the behaviour of QR calls.
However, it would be great to include this directly to odoo.

tarteo:
Ok maybe later then, I think this module is more suitable in https://github.com/oca/server-tools

etobella:
Well, no problem to change for me, but I will wait for another comment about the change before making it (just to have a confirmation) :wink:

web_qr/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
```
JordiBForgeFlow: Add here Enric Tobella

web_qr/controllers/qr.py
```diff
@@ -0,0 +1,17 @@
+import werkzeug
+from odoo import http
+from odoo.http import request
+
+
+class Home(http.Controller):
+
+    @http.route('/report/qr', type='http', auth="public")
+    def report_qr(self, value, box_size=3, border=3, **kwargs):
```
JordiBForgeFlow, line 9: According to https://github.com/lincolnloop/python-qrcode/blob/master/README.rst the minimum border should be 3.

web_qr/README.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+When creating new reports, you should use a path like the following::
+
+    <img t-att-src="'/report/qr/?value=%s&amp;error_correction=%s' % ('HELLO WORLD!', 3)" style="width:100;height:100"/>
```
JordiBForgeFlow, line 52: I would add some explanation on the possible parameters, as explained here: https://github.com/lincolnloop/python-qrcode/blob/master/README.rst

The `error_correction` parameter controls the error correction used for the QR Code. The following four constants are made available:
error_correction = 1: About 7% or less errors can be corrected.
error_correction = 0: About 15% or less errors can be corrected.
error_correction = 3: About 25% or less errors can be corrected.
error_correction = 2: About 30% or less errors can be corrected.

The `box_size` parameter controls how many pixels each "box" of the QR code is. The default is 3.
The `border` parameter controls how many boxes thick the border should be (the default is 4, which is the minimum according to the specs).

The `version` parameter is an integer from 1 to 40 that controls the size of the QR Code (the smallest, version 1, is a 21x21 matrix). Set to None and use the fit parameter when making the code to determine this automatically.

`fill_color` and `back_color` can change the background and the painting color of the QR, when using the default image factory.

JordiBForgeFlow (approved):
:+1: functional test and technical review. Minor comments.

In my tests I added the following line to the invoice report.

image

Then the invoice looks like:
image

Then I scan it with my mobile. to obtain:
image

JordiBForgeFlow:
For me I don't see how moving to server-tools would make a difference. Is it because server-tools is more generic?

@hveficent @bjeficent  can you review this one?

pedrobaeza:
This repository is for web client modifications, and this is not that. You can put it in OCA/reporting-engine and call it `report_qr` for example.

HviorForgeFlow (commented):
Functional Review: LGTM :+1: 
It makes sense to move it to reporting engine.
Waiting next PR

etobella:
Replaced by OCA/reporting-engine#288

PR #1230, opened 2019-03-27
 [11.0][ADD] Datatable Widget

eilst:
A module to add datatable https://datatables.net/ as a widget


pedrobaeza:
Interesting. What are the intended uses? Why not putting `field_name` in `my_table` instead?

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "web_widget_table",
```
nikul-serpentcs: @eilst Name Should be `Web Datatable Widget`

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "cateogry": "Hidden",
```
nikul-serpentcs: web

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": "This widget allows to display tables using Dtatable jQeury library.",
+    "author": "Savoir-faire Linux",
```
nikul-serpentcs: added one more author: ' Odoo Community Association (OCA)'

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "maintainer": "Savoir-faire Linux",
+    "website": "http://www.savoirfairelinux.com",
```
nikul-serpentcs: URL : `https://github.com/OCA/web`

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "license": "LGPL-3",
```
nikul-serpentcs: AGPL-3.0

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "version": "0.1",
```
nikul-serpentcs: version : `11.0.1.0.0`

nikul-serpentcs (changes requested):
netpicking

sudhir-erpharbor:
@eilst 
Travis is red. Please give some love to it.

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "web_widget_table",
+    "cateogry": "Hidden",
+    "summary": "This widget allows to display tables using Dtatable jQeury library.",
```
sudhir-erpharbor: Datatables

web_widget_table/__manifest__.py
eilst: Thanks!

web_widget_table/__manifest__.py
eilst: Thanks!

eilst: I will do those corrections now

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,56 @@
...
+            this.$el.empty(); 
+            var field = this.nodeOptions['field_name'],
+                fieldModel = this.record.data[field].model,
+                fieldRecordIds = this.record.data[field].res_ids,
+                table = `<table><thead><tr>`,
+                table_fields = this.nodeOptions['fields'],
+                headers = this.nodeOptions['headers'];
+                datatable_params = this.nodeOptions['datatable_params'];
+
+            headers.forEach(element => {
```
Tardo: Odoo 11.0 doesn't support ES6, more info: https://github.com/OCA/odoo-community.org/issues/37

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,56 @@
...
+                table += '<th>' + element + '</th>';
+            });
+            table+= `</tr></thead><tbody>`;
+            this._rpc({
+                model: fieldModel,
+                method: 'read',
+                args: [fieldRecordIds],
+                })
+                .then(function(result){
+                    Array.from(result).forEach(record => {
```
Tardo: Odoo 11.0 doesn't support ES6

web_widget_table/views/assets.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="assets_backend" inherit_id="web.assets_backend">
+        <xpath expr=".">
+            <link href="/web_widget_table/static/css/web_widget.css"
+                  rel="stylesheet"/>
+            <link href="https://cdn.datatables.net/v/dt/dt-1.10.18/datatables.min.css"
```
Tardo: I don't know what is the official posture in OCA for the CDN usage. But the modules that i can see don't use CDN.
Odoo can run on environments without internet.

Please, don't use minified files... Odoo does it automatically ;) and is usefull for debugging.

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,56 @@
...
+                        table+='<tr>'
```
Tardo: Semicolon missed.
To prevent these type of "mistakes" you can use eslint with the OCA rules: https://raw.githubusercontent.com/OCA/pylint-odoo/master/pylint_odoo/examples/.jslintrc

Here can found a guide to setup Atom editor: https://gist.github.com/Tardo/52a08cb7f98ae3cbfafdf47203c29a4b

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,56 @@
...
+                        table_fields.forEach(function(key){
+                            table += (record[key] ? '<td>' + record[key] + '</td>': '<td></td>');      
+                        });
+                        table+='</tr>'
```
Tardo: Semicolon missed.

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,56 @@
...
+                        })
+                    );
+                    jQuery('#datatable').DataTable(datatable_params);
+                }.bind(this));             
+        }
+    });
+    registry.add('widget_table', FieldDataTable);
+
+    return {
+        FieldDataTable: FieldDataTable,
```
Tardo: You can return "FieldDataTable" directly...

Tardo (changes requested):

web_widget_table/README.rst
```diff
@@ -0,0 +1,88 @@
...
+   :linenos:
+<field name="my_table" widget="widget_table" options="{
+                            'field_name': 'child_ids',
+                            'datatable_params':{'searching': False,
+                            'paging': False,
+                            'info': False},
+                            'fields': ['name','date_start', 'date_end', 'real_date_start', 'real_date_end', 'total_time'],
+                            'headers': ['Name','Date Start', 'Date End', 'Real Date Start', 'Real Date End', 'Total Time']}"/>
+
+"my_table" field must be of type char, it is only used to display the table. For example:
```
tarteo, line 50: Is it possible to make the widget work directly on one2many fields, without adding a new 'empty' field?

tarteo (commented):
Can you add some demo data / view to make it easier to test the functionality?

web_widget_table/__manifest__.py
eilst: ok thanks

web_widget_table/__manifest__.py
eilst: ok

web_widget_table/__manifest__.py
eilst: ok

web_widget_table/__manifest__.py
eilst: ok

web_widget_table/__manifest__.py
eilst: ok

web_widget_table/__manifest__.py
eilst: ok, changes done

web_widget_table/static/js/web_widget_table.js
eilst: I did the changes, thank you

web_widget_table/views/assets.xml
eilst: Thanks for the comment, I included all the external references directly to my application context so there are no external dependencies. However, these are big files, I'm wondering that if this widget was bigger and used many packages, would I need to include all the packages here? Do you know if we can use a package manager?

Tardo: @pedrobaeza @tarteo @simahawk @yajo can help you better.

yajo: :zipper_mouth_face: Don't ask about JS package managers! This could become a emacs vs vim situation! :bomb: 

:laughing: Jokes apart, it's a pending subject where there's no consensus here at @OCA, so we've just continued Odoo's methodology of just include the external libraries, unminified, under `/static/lib/` and assets from there. Odoo itself will take care of minification.

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Web Datatable Widget",
+    "cateogry": "web",
```
max3903:
```suggestion
    "category": "web",
```

web_widget_table/README.rst
```diff
@@ -0,0 +1,88 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+===============
```
max3903:
```suggestion
================
```

web_widget_table/README.rst
```diff
@@ -0,0 +1,88 @@
...
+Web Widget Table
+===============
```
max3903:
```suggestion
================
```

web_widget_table/README.rst
```diff
@@ -0,0 +1,88 @@
...
+
+A module to add datatable https://datatables.net/ as a widget
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=============
```
max3903:
```suggestion
=====
```

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": "This widget allows to display tables using Datatable jQeury library.",
```
max3903:
```suggestion
    "summary": """
        This widget allows to display tables using Datatable jQeury library.
    """,
```

web_widget_table/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "author": "Savoir-faire Linux, Odoo Community Association (OCA)",
+    "maintainer": "Savoir-faire Linux",
+    "website": "https://github.com/OCA/web",
+    "license": "ALGPL-3.0",
+    "version": "11.0.1.0.0",
+    "depends": ["web"],
+    "data": [
+        "views/assets.xml",
+    ],
```
max3903, line 17:
```suggestion
    ],
    "development_status": "Beta",
    "maintainers": ["eilst"],
```

web_widget_table/README.rst
eilst: Hey, thanks it makes sense what you are saying, I did it like this because the use case that I thought was to add any field of the model to the table, not only the fields of many2many record records. I created an issue for this comment, I think it would be more intuitive to if a many2many field is referenced to make it as you say, so I created an issue for this: https://github.com/OCA/web/issues/1273

max3903 (approved):
I am not an expert in JS, but this looks good to me.

nikul-serpentcs (approved):
Code Review LGTM :+1: 

`<link href="/web_widget_table/static/css/web_widget.css" rel="stylesheet"/>`
change to `<link href="/web_widget_table/static/src/css/web_widget.css" rel="stylesheet"/>`
Add `src` directory after `css`, same as JS also

web_widget_table/static/css/web_widget.css
```diff
@@ -0,0 +1,6 @@
+.oe_form_field_table
+{
+    margin: 0em .2em .2em 0em;
+    display: inline-block !important;
```
yajo, line 4:
```suggestion
    display: inline-block;
```

Please check if there's any way to remove the `!important` from here. Maybe you can just add a more specific selector so normal CSS priority rules apply.

It's a good practice to avoid these, because later is too hard to override in a theme or submodule.

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,55 @@
...
+
+odoo.define('web_widget_table', function (require) {
+    "use strict";
+
+    var registry = require('web.field_registry');
+    var AbstractField = require('web.AbstractField');
+    var FieldDataTable = AbstractField.extend({
+        className: 'oe_form_field_table',
+        supportedFieldTypes: ['char'],
+        init: function init () {
```
yajo, line 11: :fire: Remove this method, it's doing nothing.

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,55 @@
...
+                fieldModel = this.record.data[field].model,
+                fieldRecordIds = this.record.data[field].res_ids,
+                table = "<table><thead><tr>",
+                table_fields = this.nodeOptions.fields,
+                headers = this.nodeOptions.headers,
+                datatable_params = this.nodeOptions.datatable_params;
+            headers.forEach(function (element) {
+                table += '<th>' + element + '</th>';
+            });
+            table += "</tr></thead><tbody>";
```
yajo, line 26: IMHO this would be better done in a client qweb template.

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,55 @@
...
+                        table += record[key] ? '<td>' + record[key] +
+                                 '</td>' : '<td></td>';
+                    });
+                    table += '</tr>';
+                });
+                table += "</tbody></table>";
+                this.$el.append(jQuery(table).attr({
+                    'id': 'datatable',
+                    'class': 'display',
+                }));
```
yajo, line 45: All of this would be better done in the qweb template too.

web_widget_table/static/js/web_widget_table.js
```diff
@@ -0,0 +1,55 @@
...
+                table = "<table><thead><tr>",
+                table_fields = this.nodeOptions.fields,
+                headers = this.nodeOptions.headers,
+                datatable_params = this.nodeOptions.datatable_params;
+            headers.forEach(function (element) {
+                table += '<th>' + element + '</th>';
+            });
+            table += "</tr></thead><tbody>";
+
+            this._rpc({
```
yajo, line 28: You need to:

```suggestion
            return this._rpc({
```

web_widget_table/views/assets.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="assets_backend" inherit_id="web.assets_backend">
+        <xpath expr=".">
+            <link href="/web_widget_table/static/css/web_widget.css"
+                  rel="stylesheet"/>
+            <link href="/web_widget_table/static/css/dt-1.10.18-datatables.css"
```
yajo: Put all lib files in a separate folder:

```suggestion
            <link href="/web_widget_table/static/lib/dt-1.10.18-datatables.css"
```

web_widget_table/views/assets.xml
```diff
@@ -0,0 +1,13 @@
...
+                  rel="stylesheet"/>
+            <script type="text/javascript" src="/web_widget_table/static/lib/jsdt/jsdt-1.10.18-datatables.js"></script>
```
yajo: Possibly that extra folder is not needed.

```suggestion
            <script type="text/javascript" src="/web_widget_table/static/lib/jsdt-1.10.18-datatables.js"></script>
```

web_widget_table/README.rst
```diff
@@ -0,0 +1,89 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+================
+Web Widget Table
+================
+
+A module to add datatable https://datatables.net/ as a widget
```
yajo, line 28: You better use the split readme system.

yajo (changes requested):
Code review done.

My main question is: why not just use odoo list views instead?

eilst:
Hey @Yajo 
> Code review done.
> 
> My main question is: why not just use odoo list views instead?


The main use case for this widget is:
Given a [view](https://github.com/savoirfairelinux/project-addons/blob/11.0/project_event/views/project_task_view.xml) with two list views;
one in a page ["Tasks"](https://github.com/savoirfairelinux/project-addons/blob/4f184de98098b2d036e7c246d810462031346d39/project_event/views/project_task_view.xml#L62) 
Screenshot from 2019-05-07 16-53-02

and the other in another page ["Technical Sheet"](https://github.com/savoirfairelinux/project-addons/blob/4f184de98098b2d036e7c246d810462031346d39/project_event/views/project_task_view.xml#L91)
Screenshot from 2019-05-07 16-53-11
I couldn't use a referenced list view to display different list views of the same model in the same form view, so I created this widget.

tarteo (changes requested):
Can you add an use case in de README. And please use Qweb template to render the widget.

yajo:
:roll_eyes: I'm not sure this is a good design for that...

You could just add the fields to the 1st view and have them all at a glance, or you could add a smart button that redirects the user to a different view of the same things. It could even be shown in a modal.

It's not like the module is wrong per se, but if you can avoid having to support ~15K LOC, it's generally a good idea. :sweat_smile:

web_widget_table/static/js/web_widget_table.js
eilst: Done

web_widget_table/README.rst
eilst: Right, done

web_widget_table/readme/USAGE.rst
```diff
@@ -0,0 +1,37 @@
...
+                                'paging': False,
+                                'info': False},
+                                'fields': ['name','date_start', 'date_end', 'real_date_start', 'real_date_end', 'total_time'],
+                                'headers': ['Name','Date Start', 'Date End', 'Real Date Start', 'Real Date End', 'Total Time']}"/>
+
+"my_table" field must be of type char, it is only used to display the table. For example:
+
+.. code-block:: python
+
+    my_table = fields.Char(
```
simahawk, line 17: do you really need this? FYI you can use `<widget />` tag to add a widget w/out link to a field. Then you handle the link (as you already seem to do) w/ params.

web_widget_table/readme/USAGE.rst
```diff
@@ -0,0 +1,37 @@
...
+
+.. code-block:: xml
+
+    <field name="my_table" widget="widget_table" options="{
+                                'field_name': 'child_ids',
+                                'datatable_params':{'searching': False,
+                                'paging': False,
+                                'info': False},
+                                'fields': ['name','date_start', 'date_end', 'real_date_start', 'real_date_end', 'total_time'],
+                                'headers': ['Name','Date Start', 'Date End', 'Real Date Start', 'Real Date End', 'Total Time']}"/>
```
simahawk, line 11: headers should be optional and come by default from field names, especially because here you cannot handle translations.

web_widget_table/static/js/web_widget_table.js
simahawk, line 28: as far as I understand, you already have the field in the view, no?
If true, can't you simply read data from current record instead of issuing another request?

simahawk (changes requested):
:+1:  for specific comments by @Yajo 
Regarding the size of the module... if you split out `datatable` source code size is not that big :wink: 
Regarding the use case: sometimes it happened to me to address the same issue and I went w/ the solution Jairo suggested. BUT is definitely an hack for not having a good way to display the info you want on the same form view.
That been said there are some design flaws that should be addressed.

moylop260:
> @Yajo 
but if you can avoid having to support ~15K LOC, it's generally a good idea.

@eilst 
You can use a hook in order to download the data table library.
You can find other examples using:
 - `rgrep "hook" . --include=__manifest__.py`

Example:
 - https://github.com/odoo/odoo/blob/c493aa0e9d29a4b101a8efc9150f16ea10df4967/addons/stock/__init__.py#L14-L19

But using OCA structure guideline:
 - https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#installation-hooks
 - https://github.com/OCA/maintainer-tools/blob/master/template/module/hooks.py

daramousk (changes requested):
Great stuff, thank you for this contribution.

I went on to see what `datatables` is all about, so from my understanding `datatables` is a basically an attempt to present tabular data in a more powerfull way. The library does that by utilizing `<table>` and working around it adding more functionality and operations on it.

Now the library has two modes, read and edit. If we say we are going to provide an integration with this library we should at least over the read part.

Now going through your implementation I see the following:

1) You extend `AbstractField`
2) You override `_render` and do the following:
a) assemble the headers
b) make an rpc to fetch the data
c) do some attribute assignment
d) closing with some standard operations (adding the widget to the appropriate registry and exporting)

Now, I see that `datatable` declares that there is only a simple call needed for their library to work.

So I went to see what kind of requirements do we need to satisfy before this call can be successful, it turns out that the only thing needed is a `<table>`. That's great because we have a table when `ListRenderer._renderView` does its magic

Now, going back to your operations we see the following:
a) assemble the headers -> you can let `_renderHeader` do this and if you want something fancy override it.
b) make an rpc to fetch the data -> Great news, ListRenderer has the data in `this.state.data`
c) and d) can be done when the `_renderView` is done

Now, keep in mind that some of the operations that `datatables` defines are already defined by odoo (sorting, paging etc.) and we probably need to override these to delegete them from odoo to datatables (otherwise just having render of the datatables table would not be of much use to us).

Either than the last paragraph, everything else should be straightfowardly done in `ListRenderer`, and if you do not have time to implement a edit function, perhaps put it on the `ROADMAP.rst` so that someone else can pick it up

bosd:
What is the status of this pr?

PR #1231, opened 2019-03-29
[12.0] web_notify: improve popup UI

shepilov-vladislav:
Added all available bootstrap-like notifications (success/danger/warning/info/default).

Screen Shot 2019-03-31 at 5 37 38 PM


tarteo (approved):
Cool!

web_notify/models/res_users.py
```diff
@@ -1,46 +1,92 @@
...
 
-from odoo import api, exceptions, fields, models, _
+from enum import Enum
+
+from odoo import _, api, exceptions, fields, models
+
+DEFAULT_MESSAGE = "Default message"
+
+
+class NotificationType(Enum):
```
simahawk: I don't think you need this, you can just define N costants or simply use namedtuple since there's actually no enumeration here

web_notify/static/src/js/widgets/notification.js
```diff
@@ -0,0 +1,27 @@
...
+
+    var Notification = require('web.Notification');
+
+    Notification.include({
+        init: function () {
+            this._super.apply(this, arguments);
+            // Delete default classes
+            this.className = this.className.replace(' o_error', '');
+            var addClass = 'o_' + this.type;
+            if (this.type === 'success') {
```
simahawk: pls define a mapping to be included in this class (`this.icon_mapping[this.type]` is going to be easier to customize and makes the code more clean

simahawk (changes requested):

shepilov-vladislav:
@simahawk Did everything as you requested ;)

simahawk:
@shepilov-vladislav thanks! One thins tho: seems demo is not working on runbot, if I hit the buttons nothing happens (beside refreshing the form)

shepilov-vladislav:
> @shepilov-vladislav thanks! One thins tho: seems demo is not working on runbot, if I hit the buttons nothing happens (beside refreshing the form)

It's required longpoling. Longpoling works on runbot?

shepilov-vladislav:
> @shepilov-vladislav thanks! One thins tho: seems demo is not working on runbot, if I hit the buttons nothing happens (beside refreshing the form)

@simahawk Can you double-check? I made some edits.

simahawk:
@shepilov-vladislav on runbot it does not work but I think you are right about longpolling.
I trust you it works locally because I don't have time to test it.

simahawk (approved):

PR #1232, opened 2019-03-29
[11.0][MIG] web_access_rule_buttons

espo-tony:
Standard migration


web_access_rule_buttons/static/src/js/form_controller.js
```diff
@@ -0,0 +1,35 @@
...
+odoo.define("web_access_rule_buttons.main", function(require) {
+	"use strict";
+	var FormController = require("web.FormController");
+	FormController.include({
+
+        _update: function () {
+            var self = this;
+            var ret = this._super.apply(this, arguments);
+            self.show_hide_buttons();
+            return ret;
```
simahawk: I think you should chain to it:

```js
_update: function (state) {
    var self = this;
    return this._super(state).then(self.proxy('show_hide_buttons'));
},
```

web_access_rule_buttons/static/src/js/form_controller.js
```diff
@@ -0,0 +1,35 @@
...
+        },
+		show_hide_buttons : function() {
```
simahawk: bad indent here

simahawk (changes requested):

espo-tony:
Hi @simahawk ,
Thanks for your help

web_access_rule_buttons/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
tarteo: Can you remove this.

web_access_rule_buttons/static/src/js/form_controller.js
```diff
@@ -0,0 +1,33 @@
...
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+odoo.define("web_access_rule_buttons.main", function (require) {
+    "use strict";
+    var FormController = require("web.FormController");
+    FormController.include({
+
+        _update: function (state) {
+            var self = this;
+            return this._super(state).then(self.proxy('show_hide_buttons'));
```
tarteo: you can use this here and remove `var self = this;`

tarteo (changes requested):
Looks good! Thanks.

tarteo (approved):

simahawk:
hmm I've just realized there's something missing: HISTORY :)
@espo-tony while you rebase, pls squash your commits, tnx

espo-tony:
Yes, you're right.
Do you know of any quick way to fix this, or I need to close the current PR and to open a different one?

sersanchus:
I'm having problems with this code because initialstate.data.id doesn't change when a different item on list view is selected. Al alternative solution must be used like storing a copy to the value of this.state.res_id in init().

espo-tony:
Hi, @sersanchus 
You're right: initialstate.data.id keeps only track of the state of the 1st element which is opened in a form view. I'm working on finding a better way to do this. 
In additions, as the PR was opened without history, I'm going to close it and to open a new one following the proper procedure. 

Thanks to all

espo-tony:
I opened a new PR #1243 importing history and hopefully fixing the bug @sersanchus  reported.
Please, give it a look.

PR #1239, opened 2019-04-07
web_timeline: fix group orders

jcdrubay:
Without this change, the groups are ordered kind of randomly because the javascript operation `-` does not work between 2 strings.


jcdrubay:
The fix seems to work with Firefox but not with Chrome :/

tarteo (commented):
Is `grp1.content` and `grp2.content` always a string? If this is the case you can use https://www.w3schools.com/jsref/jsref_localecompare.asp

jcdrubay:
Sorry for the mess in the history, I created this new PR with a clean history: https://github.com/OCA/web/pull/1283

PR #1247, opened 2019-04-11
[12.0][ADD] web_tree_duplicate: Duplicate records from the tree view

tarteo:
Simple module which lets you duplicate records directly from the tree view.

image


pedrobaeza:
Does this present the duplicated records once they are done? If not, you are not going to distinguish which are the duplicated and which are the originals (not all models add the ` (copy)` suffix or similar).

This can be a confusing and error prone option also, so I would put this as an opt-in option for the models (through ir.config_parameter or similar).

tarteo:
No, didn't thought of it ;) I'll try to make a solution for it.

tarteo:
@pedrobaeza I implemented the opt-in with an attribute, what do you think?

web_tree_duplicate/README.rst
```diff
@@ -0,0 +1,116 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+The duplicate option is disabled by default.
+To enable it you have to add attribute `duplicate` to the tree view.
```
pedrobaeza, line 39: Shouldn't this be called `multi_duplicate` for avoiding misunderstandings?

pedrobaeza:
What about the option of showing after the multi duplication the resulted records filtered in the same view or a similar technique?

web_tree_duplicate/README.rst
tarteo: Hmm yes, that's better. 👍  I'll also try to make it filter on the new copied records.

daramousk:
A question on this, wouldn't adding a server action that executes the code:
```
for rec in self:
   rec.copy()
```
on the model we want achieve the same thing? When I tested this on runbot, the treeview is actually refreshed and showing the new records.

web_tree_duplicate/README.rst
tarteo: Tested it. multi_duplicate gives an error:
`Invalid view account.analytic.line.tree.hr_timesheet definition in False`
The attribute `duplicate` is also used in the form view, so I'm keeping it `duplicate` it's more consistent and no extra python code is needed.

pedrobaeza: That's because you have to change the RNG that indicates the allowed structure:

https://github.com/odoo/odoo/blob/c9d3653f486f6c22b47c91f9dc7583219fffc47b/odoo/addons/base/rng/tree_view.rng#L17

I'm not sure if we can inherit that file. But if we reuse the existing attribute, we are overwriting other possible functionality and this might fail in the future.

tarteo: It's impossible to inherit :( But I checked the web.ListView, and web.ListRenderer and it doesn't use the duplicate attribute. For now it's ok to use it IMO. Indeed, in another version of Odoo this module probably gets deprecated.

pedrobaeza: Is it or is not possible? (your sad face confused me), but if possible, I prefer to do it at once than to be exposed in later versions or in this one with an overlap.

tarteo: not possible :smile:

pedrobaeza: OK, that's something limiting indeed, so another option is to use `options` attribute. I let in your hands the path to take for not being more picky.

tarteo:
@daramousk Yes, but with this module it takes less time to achieve the same thing. Also this can be applied per tree view e.g. I can enable it on the timesheet tree view but keep it disabled on the other account.analytic.line tree views.

Tardo (approved):

simahawk (approved):
LG, functional test: ok.
Nice feature mate :)
I love the automatic filter!

PR #1248, opened 2019-04-12
[ADD] web_edit_mode addon

mharenz:
Add new module with that the user can change the read mode to edit mode and do not have to click the edit button every time he wants to change something in a form view.

### DIFF
--- a/web_edit_mode/models/res_users.py
+++ b/web_edit_mode/models/res_users.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    def __init__(self, pool, cr):
+        """ Override of __init__ to add access rights.
+        Access rights are disabled by default, but allowed on some specific
+        fields defined in self.SELF_{READ/WRITE}ABLE_FIELDS.
+        """
+        super().__init__(pool, cr)
+        # duplicate list to avoid modifying the original reference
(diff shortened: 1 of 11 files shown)


web_edit_mode/models/res_users.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
yweng8111, line 1: You have forgot the licence and copyright here.

Tardo:
On runbot with this module can't edit any record... all time opens in "only-read mode" (except some type of fields like m2o).

pedrobaeza:
For me this concept is very dangerous, as the edit mode is there for a reason (not a technical one, but readability + security one).

mharenz:
@Tardo 
> 
> On runbot with this module can't edit any record... all time opens in "only-read mode" (except some type of fields like m2o).

You have to change it for your user as in configuration (README.rst) described:
At user settings the user can change the default behavior from read mode
to edit mode.

I will soon make the description clearer.

@pedrobaeza 
> 
> For me this concept is very dangerous, as the edit mode is there for a reason (not a technical one, but readability + security one).

We are running odoo (version 8) in a manufacturing company with 120 users since 4 years. This was one of the first improvements that our users wanted. Some users did change data by mistake, but this cases were only at beginning.

pedrobaeza:
OK, this can host any opinion on this matter, so go ahead, but I don't share your PoV on the safety part and prefer to leave dual mode, plus extra of readability of the data on view mode, that I appreciate.

Tardo:
@mharenz I see it, but some views doesn't work properly. Try it on invoice model in form view.

PR #1252, opened 2019-04-15
[FIX] web_responsive: Allow searching apps in mobile Chrome

### LINKED ISSUE #1251: [12.0] web_responsive: App drawer search menus not working properly from mobile phone.
Tested in runbot, after last commits search menu doesn't work with virtual keyboard on mobile. 

Whatever key is pressed, search input loose focus and doesn't show search results.

Have a look at video: 
20190413114412-4b38295101 gif-2-mp4 com

Can someone double check and confirm? 

Thx

yajo:
This patch fixes #1251.

The keydown event is faked in mobile chrome because it's not something reliable due to today's virtual keyboards prediction features. Typing a <kbd>C</kbd> character doesn't mean it will be there when user uses autocorrection/autoguessing.

Instead of that, we use now the `keydown` event exclusively for movement features, and `input` (TIL) event exclusively for content changes in the search input element.

Apart from fixing search in mobile Chrome, it also makes code more readable.

The search input also disables autocompletion since it would be useless and annoying here.

@Tecnativa

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -72,6 +72,7 @@ odoo.define('web_responsive', function (require) {
     AppsMenu.include({
         events: _.extend({
             "keydown .search-input input": "_searchResultsNavigate",
+            "input .search-input input": "_searchMenusSchedule",
             "click .o-menu-search-result": "_searchResultChosen",
             "shown.bs.dropdown": "_searchFocus",
             "hidden.bs.dropdown": "_searchReset",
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

GSLabIt (approved):
LGTM

Tardo (changes requested):
Mobile works fine, but now crash when press TAB key...

GSLabIt:
@Tardo tested on runbot, can't replicate the issue

Tardo:
@GSLabIt umm runbot too.. tested on Firefox and Chromium... Here a video to clarify the steps:
bug

GSLabIt:
@Tardo you're right, reproduced the issue; press Tab with empty input search

yajo:
It should be fixed now.

GSLabIt (approved):
Looks like issue is fixed

PR #1253, opened 2019-04-15
[12.0][ADD] web_company_color

Tardo:
This module changes navbar colors automatically using company logo colors. Can be changed by user.
Supports multi-company.

It's too similar to https://github.com/OCA/web/pull/1178


Tardo:
Ok, ready for reviews.

LoisRForgeFlow (approved):
Cool, I like it :+1:

tarteo:
Can you move the functionality to this module? https://github.com/OCA/web/pull/1178 The chameleon functionality will be the default behavior (if no color is manually selected).

simahawk:
@Tardo @tarteo I think it makes more sense to keep this open and close the other.
Also, I would change the name to something more wide `web_company_[colors|styles|...]`.

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+from odoo import models, fields, api
+
+URL_BASE = '/web_navbar_chameleon/static/src/scss/'
+URL_SCSS_GEN = URL_BASE + 'web_navbar_chameleon.%d.gen.scss'
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    navbar_color_background = fields.Char('Navbar Background Color')
```
simahawk: I'd make this fields sparse:
```
company_colors = fields.Serialized()
f1 = fields.Char(sparse='company_colors')
f2 = fields.Char(sparse='company_colors')
```

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+    navbar_color_background_hover = fields.Char(
+        'Navbar Background Color Hover')
+    navbar_color_text = fields.Char('Navbar Text Color')
+
+    @api.model
+    def create(self, vals):
+        company_id = super(ResCompany, self).create(vals)
```
simahawk: s/company_id/rec

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @api.multi
+    def unlink(self):
+        IrAttachmentObj = self.env["ir.attachment"]
+        for record in self:
+            custom_url = URL_SCSS_GEN % record.id
+            IrAttachmentObj.search([
+                ("url", '=', custom_url),
+            ]).unlink()
+        return super(ResCompany, self).unlink()
```
simahawk: I'd call super first and I think we should use sudo to delete the attachment

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @api.onchange('logo')
+    def change_logo(self):
+        def n_rgb2hex(_r, _g, _b):
```
simahawk: this func can be global

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+            return '#%02x%02x%02x' % (int(255*_r), int(255*_g), int(255*_b))
+
+        if self.logo:
+            _r, _g, _b = self._get_image_mix_color_rgb(self.logo)
```
simahawk: this can be moved to a "gimme_some_colors_whatever_better_name" method :)

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+            self.navbar_color_text = '#000' if _a < 0.5 else '#fff'
+        else:
+            self.navbar_color_background = ''
+            self.navbar_color_background_hover = ''
+            self.navbar_color_text = ''
+
+    @api.onchange('navbar_color_background', 'navbar_color_background_hover',
+                  'navbar_color_text')
+    def change_navbar_color(self):
+        if self._origin.id:
```
simahawk: why do you handle this in an onchange?
I'd rather to this on write.

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @api.multi
+    def update_navbar_chameleon_scss(self):
+        for record in self:
+            datas = base64.b64encode(
+                record._generate_navbar_chameleon_scss().encode("utf-8"))
+            self._create_or_update_navbar_chameleon_scss_attachment(record.id,
+                                                                    datas)
+
+    def _get_image_mix_color_rgb(self, img64):
```
simahawk: same here: I'd move it to a global func. 
You can collect them all in a `utils.py` file.

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+            datas = base64.b64encode(
+                self._generate_navbar_chameleon_scss().encode("utf-8"))
+            self._create_or_update_navbar_chameleon_scss_attachment(
+                self._origin.id, datas)
+
+    @api.multi
+    def _generate_navbar_chameleon_scss(self):
+        self.ensure_one()
+
+        return """
```
simahawk: I'd move this to `self.SCSS_TEMPLATE`

web_navbar_chameleon/models/res_company.py
```diff
@@ -0,0 +1,164 @@
...
+                    color: %(text_color)s !important;
+
+                    &:hover, &:focus, &:active, &:focus:active {
+                      background-color: %(bghover_color)s !important;
+                    }
+                  }
+                }
+              }
+            }
+        """ % {
```
simahawk: here I'd move all defaults to each respective field definition and do `self.SCSS_TEMPLATE.format(**self.company_colors)`

simahawk (commented):
thanks for implementing my idea :smile: :raised_hands:

web_navbar_chameleon/view/assets.xml
```diff
@@ -0,0 +1,17 @@
...
+    Copyright 2019 Alexandre Díaz
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="assets_backend" name="Web Navbar Chameleon Assets"
+              inherit_id="web.assets_backend">
+        <xpath expr=".">
+            <link rel="stylesheet" type="text/scss"
+                  t-attf-href="/web_navbar_chameleon/static/src/scss/web_navbar_chameleon.#{request.env.user and request.env.user.company_id.id or '1'}.gen.scss"
```
simahawk: to prevent caching we should store a timestamp when styles are updated and use it in the URL

yajo:
You probably know that, when enabling the default theme in a website, you get these features for free:

Captura de pantalla de 2019-04-22 08-42-49

Behind the scenes, it's creating a custom SCSS file for the website.

I'm not sure if you can reuse any of those because they might be too related to website stuff and you are theming the backend here AFAICS. However, it uses [the method `save_scss` from `web_editor`](https://github.com/odoo/odoo/blob/326a44c3f7f754160c78649f7c6c69dc0c030d92/addons/web_editor/controllers/main.py#L397-L401). I'm pretty sure you can use it to store customizations and reduce this module a lot.

web_company_color/view/assets.xml
```diff
@@ -0,0 +1,18 @@
...
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="assets_backend" name="web_company_color assets"
+              inherit_id="web.assets_backend">
+        <xpath expr=".">
+            <t t-set="company_id" t-value="request.env.user and request.env.user.company_id" />
+            <link t-if="company_id" rel="stylesheet" type="text/scss"
+                  t-attf-href="/web_company_color/static/src/scss/web_navbar_chameleon.#{company_id.id}.#{company_id.scss_modif_hash}.gen.scss"
```
tarteo: web_navbar_chameleon > web_company_color

tarteo (changes requested):

Tardo:
Thx all for the help!
@Yajo yeah, i saw it in the eficent pr ;) but i don't know how to call controller method from model.
I added @lreficent @jbeficent (Original Idea) @simahawk (Dev. Help) and @Yajo (Doc. Help) to contributors file ... if i forgotten someone or you don't want to appear in "contributors", please, tell me.

web_company_color/view/assets.xml
Tardo: thx, done!

tarteo (approved):
Works perfectly now, thanks!

JordiBForgeFlow:
@Tardo It's great! Small comment. I can see that the company fields are not required, but you're always getting a default #FFFFFF for all fields initially. Would it make sense to allow to set the fields blank, and it that case the system would use the system defaults?

yajo:
> i don't know how to call controller method from model

Possibly it should be done through JS.

simahawk:
> but i don't know how to call controller method from model.

```py
from web_editor.controllers.main import Web_Editor
[...]
ctrl = Web_Editor()
ctrl.call_what_you_want()
```
if you have all the info to pass you should be able to do it since the request is used only for the env and the params are passed straight.

This is how they call it from the color picker https://github.com/odoo/odoo/blob/8f4b98e67fe49516c0c830524de351db67c58357/addons/website/static/src/js/widgets/theme.js#L532

Tardo:
@jbeficent 'web_widget_color' doesn't allow non-required fields.... needs be improved.

@Yajo You mean use bus notifications? I'm not able to visualize the "flow"...

@simahawk Something missed... or i'm doing something wrong :/... The method "stops" when try do an database operation.

yajo:
@Tardo I mean that when you save the form, it should save some custom CSS file, so it would be nice if you are able to reuse [Odoo's method for such things](https://github.com/odoo/odoo/blob/8f4b98e67fe49516c0c830524de351db67c58357/addons/web_editor/controllers/main.py#L389-L442). It'd save you quite a few code.

---

Another option I'm thinking now would be to add a controller called `/web_company_color/custom.css` that returns the CSS with customizations, and then add it to backend with something like:

```xml
<data>
  <template id="company_css" inherit_id="web.webclient_bootstrap">
    <xpath expr="//t[@t-set='head_web']">
      <link type="text/css" rel="stylesheet" href="/web_company_color/custom.css"/>
```

You cannot include it in `assets_backend` because it's not a file but a controller. Doing it like this would include that HTML raw, and then would make a call to such controller to obtain the customized CSS in a separate file.

Tardo:
@Yajo yeah, i understand these part... the thing that i don't know is how to "reuse" the controller method xD... i try @simahawk way but doesn't work for me :/ and i don't know (and don't see clear) how to use javascript here...
The other solution you're proposing sounds great! but... you don't like ir.attachment usage for some reason?

P.S: Travis error is an outdated test.

yajo (approved):
> you don't like ir.attachment usage for some reason?

Well, I was thinking more in a way to simplify code, but as you say my idea wouldn't be "simple" also to implement, so it seems this is simple enough after all! :wink:

web_company_color/models/res_company.py
```diff
@@ -0,0 +1,165 @@
...
+            if any([field in values for field in fields_to_check]):
+                self.create_or_update_scss_attachment()
+        else:
+            result = super(ResCompany, self).write(values)
+        return result
+
+    @api.multi
+    def _sanitize_scss_values(self, values):
+        self.ensure_one()
+        values = dict(self.company_colors)
```
simahawk, line 109: do you need to call `dict` here? I think you get an empty one anyway

web_company_color/models/res_company.py
```diff
@@ -0,0 +1,165 @@
...
+                    })
+                else:
+                    values.update(self.default_get(fields_to_check))
+
+            result = super(ResCompany, self).write(values)
+
+            if any([field in values for field in fields_to_check]):
+                self.create_or_update_scss_attachment()
+        else:
+            result = super(ResCompany, self).write(values)
```
simahawk: you can use super()

web_company_color/models/res_company.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @api.multi
+    def get_url_scss(self):
+        self.ensure_one()
+        return URL_SCSS_GEN_TEMPLATE % (self.id, self.scss_modif_hash)
+
+    @api.multi
+    def create_or_update_scss_attachment(self):
+        IrAttachmentObj = self.env['ir.attachment']
+        self.write({'scss_modif_hash': str(uuid.uuid4())[:6]})
```
simahawk: why not using a simple timestamp?
Also, wouldn't be better to write at the end?

web_company_color/models/res_company.py
```diff
@@ -0,0 +1,165 @@
...
+        NTEMPLATE = '.'.join(URL_SCSS_GEN_TEMPLATE.split('.')[:2])
+        return NTEMPLATE % self.id
+
+    @api.multi
+    def get_url_scss(self):
+        self.ensure_one()
+        return URL_SCSS_GEN_TEMPLATE % (self.id, self.scss_modif_hash)
+
+    @api.multi
+    def create_or_update_scss_attachment(self):
```
simahawk: matter of taste probably: I'd rename all the methods and the class/instance vars to be prefixed by `scss_` or `_scss`.
This way you categorize all of them automatically.
I do this when I want to "tag" and "isolate" new set of features... but feel free to ignore my comment ;)

web_company_color/view/res_company.xml
```diff
@@ -0,0 +1,27 @@
...
+-->
+
+<odoo>
+
+    <record id="view_company_form" model="ir.ui.view">
+        <field name="model">res.company</field>
+        <field name="inherit_id" ref="base.view_company_form" />
+        <field name="arch" type="xml">
+            <notebook position="inside">
+                <page string="Navbar" name="navbar" groups="base.group_system">
```
simahawk: pls rename the tab to "Company styles" or "Backend styles" or similar.
We might add some other styles not related to navbar only.

web_company_color/view/res_company.xml
```diff
@@ -0,0 +1,27 @@
...
+                    <group string="Navbar Colors" name="navbar_colors">
+                        <field name="color_navbar_bg" widget="color" />
+                        <field name="color_navbar_bg_hover" widget="color" />
+                        <field name="color_navbar_text" widget="color" />
+                    </group>
```
simahawk, line 22: add a `div class="alert alert-info"` block at the end to remind to refresh the page to see the changes

simahawk (changes requested):
@Tardo thanks for the amazing work :)
Few details to change, few cosmetic details to be addressed only if you prefer to.

Tested on runbot: works smoothly upon refresh. :tada:

web_company_color/models/res_company.py
Tardo: Fixed, thanks

web_company_color/models/res_company.py
Tardo: Improved, thanks

web_company_color/models/res_company.py
Tardo: Looks cool with uuid... xD changed to use timestamp integer format

uuid offer a better random value... the current timestamp implementation can get the same value if two users modify the scss at the same time (in seconds).

web_company_color/models/res_company.py
Tardo: Done

web_company_color/view/res_company.xml
Tardo: Done

web_company_color/view/res_company.xml
Tardo: Done

simahawk (approved):

simahawk:
@Tardo thanks for your work! (note: I squashed on merge to rewrite the commit msg that kept the old module name :wink: )

PR #1256, opened 2019-04-19
[12.0][FIX] web_notify: prevent notify non-type message

anhvu-sg:
There is a popup without any message appear unexpectedly and continuously.

This issue comes from module web_notify and user have more than 1 direct message

image


simahawk (changes requested):
Please increase version number by 0.0.1

simahawk:
BTW Do you know if there's any other case where we miss the type?

anhvu-sg:
Hi @simahawk, it had been fixed in #1249. I will close this PR

PR #1263, opened 2019-04-26
[10][IMP] switch company warning : Add detection of switching database

hparfr:
You have multiple databases on the same host (like in a dev env) and you have multiple tabs open.
When you switch from a base to another in one tab, thanks to this PR, you get a warning message.


tarteo (approved):
Tested locally

PierrickBrun (approved):

PR #1266, opened 2019-04-30
12.0 timeline view for calendar (Waiting feedback)

apineux:
Contains improvement for timeline view:
- Allow x2x field as group
- Fix group display for all type of fields

Contains timeline view for calendar


pedrobaeza:
Hi, Arnaud, thanks for this. Can you please split the new module into a new PR and let this one only for the improvement over `web_timeline`. This way, it's easier to review and we can speed up the merge.

PR #1269, opened 2019-05-03
12.0 mig web_export_view

ernestotejeda:
* Adapt JS

Cc @Tecnativa


web_export_view/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
Tardo: Python3 doesn't need declare file coding (utf-8): PEP 3120

Tardo (approved):

web_export_view/__init__.py
ernestotejeda: Done

pedrobaeza (approved):

PR #1274, opened 2019-05-04
[FIX] web_widget_timepicker: fix show current time when choose 0:00

JonathanOsAlc:
Now when we use the `web_widget_timepicker` and choose a `0:00` it will show that expected value.

- Before this MR:
before_test

- After this MR:
after_test


moylop260 (approved):
@oscarolar Could you review it, please?

pedrobaeza:
There's already #1272 fixing this.

PR #1275, opened 2019-05-04
[12.0] [ADD] web_widget_text_markdown: adds markdown editor widget for text fields.

tomeyro:
This module adds a `markdown` widget for the text fields.

It uses [`bootstrap-markdown`](https://github.com/toopay/bootstrap-markdown) to create an editor on edit mode, and [`showdown`](https://github.com/showdownjs/showdown) to convert the markdown into HTML while on preview or readonly mode.

Just add the markdown widget attribute to any text field like so:

```xml
<field name="textfield" widget="markdown"/>
```

Supports translatable fields, and all views (form, tree, kanban...).

demo editor
demo preview

This module was inspared by original module at version 10.0 located at:

https://github.com/OCA/web/tree/10.0/web_widget_text_markdown#credits

\#OCADaysManzanillo \#OCADays2019


moylop260:
if you are migrating the module you need get commit history.
Could you review the following script, please?

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

janverb:
Thank you for this, it works well.

The only problem I had was that it didn't work in Firefox. That turned out to be because of showdownjs/showdown#363. Replacing `showdown.min.js` by the non-minified version fixes it, as in 4e1576181.

jpoa:
Please apply the no stale label.

pedrobaeza:
Review the PR instead of asking that:

https://odoo-community.org/page/review

jpoa:
Greetings,

More than happy to do that, and Pedro I appreciate all the effort you put into OCA but my feeling and also of other colleagues is that getting a PR is tantamount to climbing a very steep mountain due to unclear processes, but perhaps it’s just my point of view.

I’m trying to say this in a constructive way, many times I hold back from contributing due to the overhead, perhaps the attrition can be lessened to some extent without compromising quality. I don’t know how but am willing to participate in discussing that.

Again, I am trying to deliver constructive criticism please ignore if it is coming across otherwise, that is not the intention at all.

That said thanks for the feedback and all your efforts and much appreciated contributions, I will do as you have mentioned.

Best regards and thanks in advance

--

pedrobaeza:
The question is that people thinks that contributing is only to put a PR, but it's just one step of several. Reviewing, or managing what is open (attending to comments) is the same important as the first step.

Everyone is always struggled with overheads, but that's because doing these tasks instead of "customer tasks" are considered less important than the other, and that's the main problem here. I know that part of this problem is is you invoice per hour, and this is not an invoiceable task. That's what you (your company) have to get over it: make it invoiceable, or change your approach to have R&D budget.

PR #1277, opened 2019-05-05
add encoding line to web_listview_range_select/__init__.py

KareemAbuzaid:
Add encoding line as the © is causing an issue.


pedrobaeza:
This is not needed since Python 3, and on contrary, we encourage to remove all of them.

PR #1279, opened 2019-05-08
[12.0][ADD] web_widget_uom

AdriaGForgeFlow:
This module aims to allow the user to decide how many decimal places should be displayed in the Odoo User Interface for each Unit of Measure.

image

image


web_widget_uom/static/src/js/widget.js
```diff
@@ -0,0 +1,64 @@
...
+    var FieldUoM = basic_fields.FieldFloat.extend({
+        /**
+         * Float fields have an additional precision parameter that is read from
+         * either the field node in the view or the field python definition itself.
+         *
+         * @override
+         */
+        init: function () {
+            this._super.apply(this, arguments);
+            var decimal_places = 4;
```
Tardo, line 19: Never used

Tardo (changes requested):

web_widget_uom/static/src/js/widget.js
```diff
@@ -0,0 +1,64 @@
...
+         * Also, the user experience will be better if your rendering function is
+         * synchronous.
+         *
+         * @private
+         * @returns {Deferred|undefined}
+         */
+        _render: function () {
+            var self = this;
+            var parent = this._super;
+            this._getDecimalPlaces().then(function (){
```
Tardo, line 35: I think that 'always' would be more appropriate.

web_widget_uom/static/src/js/widget.js
```diff
@@ -0,0 +1,64 @@
...
+        },
+
+        _getDecimalPlaces: function () {
+            var self = this;
+            var uomField = this.nodeOptions.uom_field || this.field.uom_field || 'uom_id';
+            var uomID = this.record.data[uomField] && this.record.data[uomField].res_id;
+
+            return this._rpc({
+                model: 'uom.uom',
+                method: 'get_decimal_places',
```
daramousk, line 47: Why not just get the data here and avoid defining that `get_decimal_places` completely?

web_widget_uom/static/src/js/widget.js
```diff
@@ -0,0 +1,64 @@
...
+    var rpc = require('web.rpc');
+
+    var FieldUoM = basic_fields.FieldFloat.extend({
+        /**
+         * Float fields have an additional precision parameter that is read from
+         * either the field node in the view or the field python definition itself.
+         *
+         * @override
+         */
+        init: function () {
```
daramousk, line 17: This init does not seem to be doing anything, the docstrings are pretty much copies of their super definitions so I believe you could remove these as well.

daramousk (commented):
Thanks for this, I wonder if a purely backend solution would be feasible and/or more appropriate here, something like this:

1) For all the fields that are foreign keys to `uom.uom`
2) inject the digits attribute in the appropriate place in the initialization phase of the `class Float`

So now you could have a piece of code that runs once and avoids all the js.

What do you think?

PR #1280, opened 2019-05-10
[12.0][IMP] Timeline View - Allow group by all type of fields (including m2m & o2m)

apineux:
Originally the timeline view only allows to visualize data grouped by simple field such as Char, Text, M2O. 
The timeline didn't allow to group by multiple field such as M2M or O2M because it was not possible to display more than one time the element.

This PR has for purpose to allow to group by M2M and O2M.
To do so, the idea is to create multiple element (that are the same) on the timeline view. We will create as many element as there are relation on it.
For example: a calendar.event having the name 'Meeting' and with 2 attendees 'You and Me' will be displayed twice. One on the group "You" and one on the group "Me".
To goup by a field M2M or O2M, it must be done on the default_group_by param specified on the timeline view definition.

The multiple elements are created and managed on the Javascript part (This allows us to keep the standard behaviour on Python side).

If we take the previous example with "You and Me". Let's assume that you want to filter on the "Me" so that I only display the events link to "Me".
By doing so on the search view, you will received a list of event where "Me" is part of the attendee. This is the normal behaviour of Odoo. As those events can be linked to other attendees too, the Timeline view will display not only the group "Me" but also the groups corresponding to attendees of the events displayed. To help the user to visualize only the group "Me", an addition filter has been added on the top left of the timeline view.
Without changing the filter on the search view, you can choose to display or hide specific groups.


sbejaoui (approved):
Good improvement. Thanks.

pedrobaeza:
@sbidoul should I add Arnaud to Acsone ECLA?

sbidoul:
@apineux is covered by the ACSONE ECLA. I just updated the odoo-community.org database.

pedrobaeza:
OK

apineux:
@pedrobaeza @sbidoul Is this PR accepted? If yes can you merge it? Thanks

pedrobaeza:
@apineux you must split each of the changes in one PR, as there are too many of them, and some of them can be conflicting / to be discussed, so having them in each PR, they can follow their path. For example, I don't see how can you group by a one2many / many2many.

pedrobaeza:
@Tardo don't review this until the PR split. I will ping you then.

sbidoul:
@apineux could you update the description with a summary of the features of this PR, and an explanation of what grouping on a m2m means in this case? Also, do you think it's feasible to split the PR as @pedrobaeza suggests?

apineux:
@sbidoul @pedrobaeza 
I will do the slip of feature asap and also include an explanation.
Regards,

apineux:
@sbidoul @pedrobaeza 

I split the feature. In this PR there is the possibility to group by M2M and O2M on the timeline view.
On the PR: https://github.com/OCA/web/pull/1347 there is the feature to define a field for the color.

A simple description of the feature is written on the first comment of this discussion.

Regards,

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -416,14 +582,16 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                if (group instanceof Array) {
+                    group = _.first(group);
+                }
             } else {
                 group = -1;
             }
-            _.each(self.colors, function (color) {
-                if (eval("'" + evt[color.field] + "' " + color.opt + " '" + color.value + "'")) {
-                    self.color = color.color;
+            _.each(self.colors, function (col) {
```
pedrobaeza: Why are you changing this that is touched in the other PR?

web_timeline/static/src/css/web_timeline.css
```diff
@@ -9,7 +9,7 @@
     background: #DCDCDC;
 }
 
-.vis-item .vis-item-overflow {
+.vis-item .vis-item-overflow:hover {
```
pedrobaeza, line 12: Why this change?

web_timeline/static/src/css/web_timeline.css
```diff
@@ -29,3 +29,7 @@
     left: 0px;
     top: 0px;
 }
+
+.selected-groups {
+    display: inline-flex;
+}
```
pedrobaeza: Add always a final empty line

pedrobaeza (dismissed):
I still don't see the application of the grouping on one2many fields, as they are going to be always unique.

Also adding the bootstrap multiselect library seems overkill for this feature. Can't you do it other way? I suppose this is only for helping in visuals, but having a composed text for x2x is enough in most cases.

I'm still not seeing any description in README, and adding screenshots would definitively help to understand how this works.

web_timeline/static/src/js/timeline_renderer.js
apineux: My bad, I thought I reverted it correctly, it has been adapted and I squashed the commits.

sbidoul:
@pedrobaeza at first grouping on a o2m was not intuitive for me also. And indeed in a pivot table it does not make sense. But here in a timeline view it does. Imagine for instance a view that is used to plan field service interventions where each intervention can be done possibly by more than one technician. In that case it is useful to see the intervention in the timeline of each technician.

So +1 for the feature, I think it is generic enough to land in the base module.

Putting the use case in the README is indeed a must :)

apineux:
@pedrobaeza 
First of all, thank you for your reviews :+1: 

The multiselect library allows to filter the display instead of filtering on the data (this is the purpose of the search view). It is useful because this library make the selection of group really user friendly. This is a small library so I don't think that it is an issue.

I will do the small changes (as you suggested) and I will do my best to include a better and complete description of the feature (with screenshot).

Regards,

pedrobaeza:
Thanks for the extra effort!

web_timeline/static/src/css/web_timeline.css
apineux: There is too much text overflow in the timeline view. This allows to keep the text inside the "element" of the timeline view, except when you put the mouse over. This is a usability improvement.

apineux:
The readme is updated. Here is an image to understand better the concepts.

x2x_description

Regards

pedrobaeza:
Thanks for the screenshot. Then I see that you are adding 2 features:

* A filter for the group by with the bootstrap library.
* The possibility of grouping x2x.

About one2many, then I understand that you are grouping by a subfield inside the one2many, isn't it?

apineux:
There are two features indeed, but there are strongly linked together. 
For the 02M, it is the same as M2M. I get the values from the field (so it will be a list of ids) then I make a name_get on those ids to the right model (I get the model from field information). The name became the group name and the id its content.
There are a lot of code that updates the groups when we update an existing record or create a new one, so that we can directly see them and filter them in the "multi select".

Regards,

pedrobaeza:
I would approve the filter right now, as it's useful for any possible grouping if you have too much possible groups, but the one2many feature is still debatable, so if you want to speed up that part, split it also in another PR.

For o2m thing, it's not the same as m2m, as with m2m, you can potentially have several records for the same group. For o2m, they aren't groups at all, as each one will have only one record. An added value will be to group by a subfield of that o2m (for example, the product of a sale.order.line, not each sale.order.line as group entity).

Another thing that concerns me doing x2x grouping (both m2m or o2m) is that you may see the same record under several groups, changing the cardinality of things and confusing users about the number of records shown. One potential problem is the update of one record not refreshing details on the rest of instances of that record.

apineux:
@pedrobaeza 
I agree that grouping by the o2m isn't that great because the group will have only one record. But I disagree in the fact that we have to group by sub field. It will make the thing more complexe in Javascript knowing that you can do exactly the same by creating you own M2M field that get the subfield of your O2M (computed field).
Example: In the calendar.event, you have a O2M relationship called attendee_ids. If you group by that, you will have multiple times the same name "Arnaud Pineux" (for example) because many attendee can be created on the same res.partner.
BUT If you use the partner_ids field on calendar.event (that gets the sub field partner_id of the attendee_ids), the problem is fixed.

If you modify or create any event, it will update the linked events. In the screenshot, if I modify one of the "Test 00001", all of them will be updated. In my opinion it is not confusing as it is the purpose of that feature.

I hope it clarify what I did.

Regards,

pedrobaeza:
OK, thanks for all the explanations. It's good enough for me right now. I discard my review and waiting for technical review by my colleague @Tardo

web_timeline/static/src/js/timeline_controller.js
```diff
@@ -192,7 +195,7 @@ odoo.define('web_timeline.TimelineController', function (require) {
                     model: self.model.modelName,
                     method: 'write',
                     args: [
-                        [item.event.data.item.id],
+                        [self.renderer.x2x ? item.event.data.item.id.split('_')[0] : item.event.data.item.id],
```
Tardo: Can be good gets the real id from a function... something like getRecordId(item.event.data.item.id)

web_timeline/static/src/js/timeline_controller.js
```diff
@@ -314,9 +317,15 @@ odoo.define('web_timeline.TimelineController', function (require) {
                 context: this.context,
             })
             .then(function (records) {
+                if (self.renderer.x2x) {
+                    records[0].id = records[0].id + "_" + records[0][self.renderer.grouped_by][0];
```
Tardo, line 318:
```suggestion
                    records[0].id = _.uniqueId(records[0].id + "_");
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -63,6 +63,7 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
                 throw new Error(_t("Timeline view has not defined 'date_start' attribute."));
             }
             this._super.apply(this, self);
+
```
Tardo: :fire:

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
         on_data_loaded: function (events, group_bys, adjust_window) {
             var self = this;
             var ids = _.pluck(events, "id");
-            return this._rpc({
-                model: this.modelName,
+            var group_by_field = self.view.fields[group_bys[0]];
+            if (group_by_field.type == 'one2many' | group_by_field.type == 'many2many') {
```
Tardo:
```suggestion
            if (group_by_field.type === 'one2many' | group_by_field.type === 'many2many') {
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                self.x2x = true;
+                return self.split_groups_x2x(events, group_bys).then(function (groups) {
+                    self.groups = groups.sort((a, b) => (a.content > b.content) ? 1 : -1);
```
Tardo: Please, don't use ES6 'arrows'

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                        model: self.modelName,
+                        method: 'name_get',
+                        args: [
+                            ids,
+                        ],
+                        context: self.getSession().user_context,
+                    }).then(function(names) {
+                        var nevents = _.map(events, function (event) {
+                            return _.extend({
+                                __name: _.detect(names, function (name) {
```
Tardo, line 325: ```_.find and _.filter are now the preferred names for _.detect and _.select.```
```suggestion
                                __name: _.find(names, function (name) {
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+        updateGroups: function(events) {
+            var self = this;
+            var group_ids = [];
+            // Get the missing groups from new events
+            _.each(events, function (event) {
+                var group = event[_.first(self.last_group_bys)];
+                if (group) {
+                    _.each(group, function (gr) {
+                        var found = _.find(self.groups, function(existing_group) {
+                            return existing_group.id == gr
```
Tardo, line 363:
```suggestion
                            return existing_group.id === gr;
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                        });
+                        if (found === undefined) {
```
Tardo, line 365:
```suggestion
                        if (_.isUndefined(found)) {
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
-                        __name: _.detect(names, function (name) {
-                            return name[0] === event.id;
-                        })[1]
-                    }, event);
+                _.each(names, function (gr_name) {
+                    self.groups.push({
+                        id: gr_name[0],
+                        content: gr_name[1],
+                    })
+                    self.$select_groups.append($("<option value='" + gr_name[0] + "' selected='selected'>" + gr_name[1] + "</option>"));
```
Tardo, line 385:
```suggestion
                    self.$select_groups.append($(_.str.sprintf("<option value='%s' selected='selected'>%s</option>", gr_name[0], gr_name[1])));
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                    id: elmt.value,
+                    content: elmt.text,
+                })
+            });
+            self.timeline.setGroups(self.selected_groups);
+        },
+        createSelectGroups: function () {
+            var self = this;
+            var select_groups = $(core.qweb.render('TimelineSelectGroups', {'groups': self.groups}));
+            self.$el.find('.selected-groups').html(select_groups);
```
Tardo:
```suggestion
            self.$('.selected-groups').html(select_groups);
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
             var groups = [];
             this.grouped_by = group_bys;
             _.each(events, function (event) {
                 if (event[self.date_start]) {
-                    data.push(self.event_data_transform(event));
+                    if (x2x) {
+                        _.each(event[group_bys], function (gr) {
+                            var x2x_object = jQuery.extend({}, event);
+                            x2x_object[group_bys] = [gr];
+                            x2x_object.id = event.id + "_" + gr;
```
Tardo, line 452:
```suggestion
                            x2x_object.id = _.uniqueId(event.id + "_");
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                    }
                 }
             });
-            groups = this.split_groups(events, group_bys);
+            if (x2x) {
+                groups = this.groups;
+                this.selected_groups = this.groups;
+                this.createSelectGroups();
+            } else {
+                if (this.$select_groups !== undefined) {
```
Tardo:
```suggestion
                if (!_.isUndefined(this.$select_groups)) {
```

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -308,41 +308,164 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
+                    this.$el.find('.selected-groups').html('');
```
Tardo:
```suggestion
                    this.$('.selected-groups').html('');
```

Tardo (changes requested):

web_timeline/static/src/js/timeline_controller.js
apineux: @Tardo 
It could if there were multiple use of that function but in this case It isn't.

web_timeline/static/src/js/timeline_controller.js
apineux: @Tardo 
I don't see any good reason to use this method knowing that I will not have duplicate id by combining record.id and group (the relationship between the record and the group is always uniq for one2many and many2many)

web_timeline/static/src/js/timeline_renderer.js
apineux: agreed

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo I'm not sure that it is necessary to check that the type are matching too when the value is the same in this case. Do you have a good reason to do that? thanks

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo This is nice that you provide review and feedback but please give the reason of your suggestion. Why I shouldn't use the arrow function?

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo sprintf is usefull if you want to format the data too but in this case the operator + is faster. Why should we change? thanks

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo I don't see any good reason to change that.
The method find is more efficient than $(...) because $(...) is the method (jQuery) and does almost everything. This find is limited by the scope of $el (which is already shorter than the whole page).

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo Same comment as previously. The id I generate is alredy unique (due to the uniq relationship between record in the one2many and many2many relationship).
In addition to that it will be much more easier to debug when you know that your id is id + _ + group than something completely random

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo from what I read, there is no good reason to use the underscore lib here instead of !== . Do you have an explanation? thanks

web_timeline/static/src/js/timeline_renderer.js
apineux: @Tardo As previously, $(...) is slower then the specific find. I don't see any good reason to change.

web_timeline/static/src/js/timeline_renderer.js
Tardo: Because can't use ES6 string templates I think is best for readability but if you think that it's not useful in this case... OK for me.

web_timeline/static/src/js/timeline_renderer.js
Tardo: Odoo widgets have the "$" function defined that call to "this.$el.find()".
https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/js/core/widget.js#L299

`this.$() != $()`

web_timeline/static/src/js/timeline_renderer.js
Tardo: Ok, can you add a comment to know fast that it's a unique id and that  in "[id]_[group]", the group part, is only used for debugging purposes?

web_timeline/static/src/js/timeline_renderer.js
Tardo: You can see: https://github.com/OCA/odoo-community.org/issues/37#issuecomment-470686449

ES6 will be supported in Odoo v14.0 (I think :/ ...)

web_timeline/static/src/js/timeline_renderer.js
Tardo: == and != are called the "evil twins" ... in this case you are right, doesn't need strict equality comparison. But it's a good practice to avoid possible type coercion.

web_timeline/static/src/js/timeline_renderer.js
Tardo: I think that in IE11 'undefined' can be shadowed... and in modern browsers can be shadowed by a local variable.
If you don't want to use underscore best use ```typeof this.$select_groups !== 'undefined'```

web_timeline/static/src/js/timeline_controller.js
Tardo: Here are other places where you use '.split('_')[0]'

https://github.com/OCA/web/pull/1280/files#diff-5c962e3fc7da04ac9df86aee662e205fR111
https://github.com/OCA/web/pull/1280/files#diff-5c962e3fc7da04ac9df86aee662e205fR228

Perhaps something like this...
```javascript
_getDatabaseId: function (id) {
    return this.renderer.x2x ? +id.split('_')[0] : id;
}
```

web_timeline/static/src/js/timeline_renderer.js
apineux: For the moment there is no use of the "gr" in the id_gr. It is only for debugging purpose but I also think that it is the most precise id for this situation. this element represent the record with id = id in the group with id = gr.

apineux:
@Tardo The changes from your review are done. Thanks for your feedback

lmignon (approved):

web_timeline/static/src/js/timeline_renderer.js
Tardo: What happens with this?

web_timeline/static/src/js/timeline_renderer.js
Tardo: What happens with this?

web_timeline/static/src/js/timeline_renderer.js
Tardo: What happens with this?

Tardo (changes requested):
Thanks for the work! Only a few unattended suggestions...
** See it on 'Files changed' tab :/

Martronic-SA:
What is the status here? Almost two years since last review?

PR #1286, opened 2019-05-24
[11.0][MIG] bokeh library to version 1.1.0

AdriaGForgeFlow:
Migration of bokeh library to its newest version 1.1.0 in module web_widget_bokeh_chart (v11.0)

See: https://bokeh.pydata.org/en/1.1.0/docs/releases.html

cc ~ @lreficent @Eficent


LoisRForgeFlow (approved):
Travis error unrelated :+1:

web_widget_bokeh_chart/README.rst
```diff
@@ -27,7 +27,7 @@ Installation
 
 You need to install the python bokeh library::
```
daramousk, line 53: The example function refers to `p.html.data`. This does not seem to work, perhaps readme should be updated here as well so that we know how to fetch the data from the graph

LoisRForgeFlow: v12 is up-to-date: https://github.com/OCA/web/tree/12.0/web_widget_bokeh_chart. With that syntax it should work

PR #1294, opened 2019-05-27
[12.0][IMP] web_tree_many2one_clickable: Readme - USAGE

Tardo:
This adds changes suggested in the original PR related with the readme


AdriaGForgeFlow (approved):
:+1:

web_tree_many2one_clickable/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+Put the mouse pointer over a many2one cell and click the button.
+.. image:: ./static/img/clickable.gif
```
pedrobaeza: Let an empty line in between for proper RST formatting and use `..` instead of `.` for being working in this file and the general one.

pedrobaeza (changes requested):

web_tree_many2one_clickable/readme/USAGE.rst
Tardo: Done, thanks!

pedrobaeza (approved):

PR #1296, opened 2019-05-29
[12.0] web_timeline: small fixups

anishagg17:
1)  A more better colour that will suit the display of various devices is chosen recent colour blows into white for some devices due to resolution limits.
2) padding of 18px will suit more appropriately
3)  when left ,top were set to 0px  each it may cause sometimes crashing issues.

### DIFF
--- a/web_timeline/static/src/css/web_timeline.css
+++ b/web_timeline/static/src/css/web_timeline.css
@@ -6,7 +6,7 @@
 /* very light gray background in weekends */
 .vis-timeline .vis-grid.vis-saturday,
 .vis-timeline .vis-grid.vis-sunday {
-    background: #DCDCDC;
+    background: #e7e7e7;
 }
 
 .vis-item .vis-item-overflow {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please sign the CLA, and put screenshots of before and after for each feature to compare.

Tardo:
ping @anishagg17

simahawk:
also pls rewrite the commit msg (or let's rewrite it w/ squash on merge) to something more meaningful as "web_timeline: styles fixups", tnx :)

daramousk (commented):
Hello there and thank you for this contribution, some points which I believe should be addressed

> A more better colour that will suit the display of various devices is chosen recent colour blows into white for some devices due to resolution limits.

You need to provide screenshots for this so that we can see exactly what the problem is, what if the colour you have selected now causes the same issue on other agents? Let's see.

> padding of 18px will suit more appropriately

Maybe, but that sounds opinionated and perhaps we should stick to what we already have, because people could tell us that 17px will suit more appropriately.

> when left ,top were set to 0px each it may cause sometimes crashing issues.

If indeed there is a crash, please provide the stack trace so that we can see what the crash is abount.

Good first PR, keep it up.

PR #1298, opened 2019-06-03
[11.0] web_widget_x2many_2d_matrix enhancements

vincent-hatakeyama:


pedrobaeza:
The module is already migrated: https://github.com/OCA/web/tree/11.0/web_widget_x2many_2d_matrix

What do you want to achieve?

vincent-hatakeyama:
I wanted to propose our enhancements.
I don’t get why most of the changes are displayed as new as they were already merged in. Unless several people made the migration in parallel.

pedrobaeza:
You have to cherry-pick only the commits with the enhancements then.

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -63,6 +63,8 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                 this.parse_boolean(node.show_row_totals || '1');
             this.show_column_totals =
                 this.parse_boolean(node.show_column_totals || '1');
+            this.label_cell_max_char = this.nodeOptions.label_cell_max_char || -1;
+            this.header_max_char = this.nodeOptions.header_max_char || -1;
```
daramousk, line 67: I am not sure how this can work here, `this.nodeOptions` contains the value of the `options` attribute, why don't you `node.label_cell_max_char` and `node.header_max_char` respectively so that you can be consistent with the way the rest of the attributes are set?

daramousk: Also, why should the default be `-1`? It could just be `undefined`

daramousk (commented):
Hello there and thank you for this contribution, I have made some comments on this:

1) You have set these attributes as keys of the options attribute. Wouldn't it be better if you just follow the convention and have them as attributes of the element?
2) The only usage I see for this addition is to deal with overflowing text, for that we have `text-overflow` https://drafts.csswg.org/css-ui-3/#propdef-text-overflow. Would this be a better alternative?

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
vincent-hatakeyama: If I remember correctly, the CSS would not apply and the column would get quite large or split on many lines.

It will not get fixed so closing this.

PR #1300, opened 2019-06-04
[10.0][FIX][web_widget_image_webcam] Fix swfURL property set

MicheleZak:
Actually set the swfURL property on Webcam object.

This property needs to be set with the proper setter method 'setSWFLocation'
since the Webcam object does not look at the params property when searching for it.

### DIFF
--- a/web_widget_image_webcam/static/src/js/webcam_widget.js
+++ b/web_widget_image_webcam/static/src/js/webcam_widget.js
@@ -31,9 +31,9 @@ odoo.define('web_widget_image_webcam.webcam_widget', function(require) {
                 image_format: 'jpeg',
                 jpeg_quality: 90,
                 force_flash: false,
-                fps: 45,
-                swfURL: '/web_widget_image_webcam/static/src/js/webcam.swf',
+                fps: 45
             });
+            Webcam.setSWFLocation('/web_widget_image_webcam/static/src/js/webcam.swf');
 
             self.$el.find('.o_form_binary_file_web_cam').removeClass('col-md-offset-5');
 


web_widget_image_webcam/static/src/js/webcam_widget.js
```diff
@@ -31,9 +31,9 @@ odoo.define('web_widget_image_webcam.webcam_widget', function(require) {
                 image_format: 'jpeg',
                 jpeg_quality: 90,
                 force_flash: false,
-                fps: 45,
-                swfURL: '/web_widget_image_webcam/static/src/js/webcam.swf',
+                fps: 45
```
Tardo, line 34: :fire: 
'fps' doesn't seem to be a valid parameter
https://pixlcore.com/read/WebcamJS#configuration

Tardo (changes requested):

web_widget_image_webcam/static/src/js/webcam_widget.js
MicheleZak: I don't know about that, I changed only the 'swfURL' property related stuff. I'll check it.

daramousk (commented):
Hello there and thank you for this contribution.

Now, I went to test this and bumped into an error which is not relevant with your changes rather with the changes of the web api's.

> Webcam.js Error: Could not access webcam: TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.

With that being said, that `createObjectURL` seems to have been deprecated, and updating the `webcam.js` library actually resolves the issue. So it would probably make more sense to just backport this

https://github.com/OCA/web/pull/964

What do you think?

MicheleZak:
Sure, it seems the best thing to do.

Thank you
Have a good week end 
:)

PR #1288, opened 2019-05-24
[11.0][MIG] web_switch_context_warning

AdriaGForgeFlow:
Standard migration of module web_switch_context_warning to v11 from #1287 

cc ~ @Eficent

PR #1302, opened 2019-06-07
Add module web_switch_company_warning

Linekio:
Add a v11 of https://github.com/OCA/web/tree/10.0/web_switch_company_warning


pedrobaeza:
Isn't this the same as #1288 ?

Linekio:
Seems you're right

LoisRForgeFlow:
This can be closed I think.

PR #1289, opened 2019-05-24
[12.0][MIG] web_switch_context_warning

AdriaGForgeFlow:
Standard migration of module web_switch_company_warning to v12 from #1288 

cc ~ @Eficent

PR #1303, opened 2019-06-07
Add module web_switch_company_warning

Linekio:
Add a v12 of https://github.com/OCA/web/tree/10.0/web_switch_company_warning


pedrobaeza:
There's already #1289 with commit history, so review that one and it's better to look before.

Issue #1157, opened 2019-01-17
New Widget: "Like Github Contributions Widget" for Partner


agent-z28:
the description of your feature : I suggest a widget like the GitHub Contributions Widget.
the Odoo version you expect it to run on: v11+
mockups and visual materials
![bildschirmfoto vom 2019-01-17 23-06-11](https://user-images.githubusercontent.com/7473428/51351950-8768cc00-1aac-11e9-97ba-f64958ffd040.png)

e.g. in client view, it could be used to visualize all order dates. in some industries its from importance to see order behaviour.

pedrobaeza:
This is a kind of heatmap. Are you going to implement it?

agent-z28:
Maybe yes, if nobody did it already. My first attempt was to bring the idea in:)

pedrobaeza:
Let's see if somebody has made something...

hbrunn:
gather the data in some model, override the matrix widget to show blocks instead of inputs, should be quite straight forward

pedrobaeza:
Well, the design with that will be very ugly, but it can be a rudimentary solution for someone in a hurry

hbrunn:
it takes only some css to make this beautiful in my opinion

agent-z28:
Nice Idea!

nikul-serpentcs:
Good Idea! :bulb:

tarteo:
Very cool. Maybe this will help: https://github.com/bachvtuan/Github-Contribution-Graph

Looking forward to try it out.

agent-z28:
JS Base : https://cal-heatmap.com/
We Sponsored IT PROJECTS LLC to realize a first Version.
It looks as they did a pretty Good job
Pull Request is almost ready.
Now we need to get it oca compliant

agent-z28:
This is the repository https://github.com/it-projects-llc/addons-dev/pull/642

pedrobaeza:
Are they going to do the pull request here?

agent-z28:
Thats the idea, i come back later with an answer. But at the end we will have a pull request here.

pedrobaeza:
Great

GabbasovDinar:
Hello @pedrobaeza @agenterp 

Today I will send the PR with this module, I will be glad to any feedback and suggestions

pedrobaeza:
Great, please include a sample.

PR #1304, opened 2019-06-10
[NEW] HeatMap Widget for One2many and Many2many records

GabbasovDinar:
This module allows you to display the One2many and Many2many records as a calendar of heatmaps.

Example::

    <record id="view_partner_form" model="ir.ui.view">
        <field name="name">res.partner.view.buttons</field>
        <field name="model">res.partner</field>
        <field name="inherit_id" ref="base.view_partner_form" />
        <field name="priority" eval="20"/>
        <field name="groups_id" eval="[(4, ref('sales_team.group_sale_salesman'))]"/>
            <page name="sales_purchases" position="after">
                <page name='sale_orders' string="Orders">
                    <field name="sale_order_ids" widget="heatmap"/>
                </page>
            </page>
        </field>
    </record>

Result:

screenshot-misc odoo11 local-2019 06 10-17-53-40

Discussion: https://github.com/OCA/web/issues/1157

CC: @pedrobaeza @agenterp 
@it-projects-llc


pedrobaeza:
I'm wondering if this is worth more as view than as widget, as loading it as widget overloads the destination record a lot.

agent-z28:
The initial idea is to show directly how to the order behaviour is. 
From performance site we'll have to check.

pedrobaeza:
I agree on the feature, but Odoo paradigm is to put an smart-button on the partner (like the existing one with the sales count), and when clicking navigate to the other view (right now, to the sales order list, but you can add the same smart-button for navigating to the heat-map). Take into account than adding this heatmap at partner level, overloads the server having to load all sales order each time you access the partner, even for modifying other things like the name, query the email, etc.

ilmir-k:
@pedrobaeza Thanks for your feedback. Actually we thought about making it as a view rather than widget but stayed on the latter at this moment. We'll take it into account.
cc @GabbasovDinar

GabbasovDinar:
What is the difference between displaying records as a tree view or using this method? Both methods overload the server to load the entire sales order.

pedrobaeza:
Exactly, and that's the reason why Odoo doesn't do neither of both inside res.partner. It always relies in an smart-button that navigates to the object.

agent-z28:
Github is able to load the whole! page within 1 second, and shows the widget in the mainscreen.
I assume they generate the widget html in the background, like a depends method on commit. 
What about a similar solution?

pedrobaeza:
Uhm, any cached solution always have the problem of out of sync data. Any way, the problem here is not the time it takes to render (which can be more or less depending on the volume data, the number of customizations in the target model, etc), the question is that we are going to render it uselessly a lot of times when accessing the partner for other reasons.

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+            '/web_widget_heatmap/static/lib/js/cal-heatmap.js'
+        ],
+
+        /**
+         * Calculate and get the maximum date (up to which date the grid will be built).
+         * Display of control buttons depends on the maximum date.
+         * @param {Object} options
+         * @returns {Date} max_date
+         */
+        get_max_date: function(options) {
```
tarteo, line 44: please use dromedary case.

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+
+    var AbstractField = require('web.AbstractField');
+    var registry = require('web.field_registry');
+
+
+    var HeatMapWidget = AbstractField.extend({
+
+        fieldsToFetch: {
+            display_name: {type: 'char'},
+            create_date: {type: 'datetime'},
```
tarteo, line 17: Can you make the date field configurable?

web_widget_heatmap/static/src/css/heatmap_widget.css
```diff
@@ -0,0 +1,12 @@
+/* Copyright 2019 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+.heatmap_widget {
+	width: 100%;
```
tarteo, line 4: Can you use spaces here instead of tabs

tarteo (changes requested):
I like this widget.
The loading problem can also be solved by making the widget work for Serialized fields. The programmer can make a computed and stored Serialized field which stores the records as JSON. This way every time the partner record opens, not all the sale orders are fetched instead just a text field essentially.

web_widget_heatmap/__manifest__.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2019 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Web Widget HeatMap",
+    "summary": "Displaying your records in calendar HeatMap",
+    "version": "11.0.1.0.0",
+    "development_status": "Stable",
+    "category": "Extra Tools",
+    "website": "https://github.com/OCA/web/tree/11.0/web_widget_heatmap",
```
Tardo:
```suggestion
    "website": "https://github.com/OCA/web",
```

web_widget_heatmap/static/src/css/heatmap_widget.css
```diff
@@ -0,0 +1,12 @@
+/* Copyright 2019 Dinar Gabbasov <https://it-projects.info/team/GabbasovDinar>
```
Tardo, line 1: Please, use ~~'less'~~ 'Sass'

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+        /**
+         * Find the records that match the selected date and show in the tree view
+         * @param {Date} date
+         * @returns {jQuery.Deferred} Action loaded
+         */
+        onClickHeatMap: function(date) {
+            date = moment(date);
+            var options = this.heatmap_options;
+            var elements = this.value ? _.pluck(this.value.data, 'data') : [];
+            if (elements && elements.length) {
```
Tardo:
```suggestion
            if (_.some(elements)) {
```

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+                // TODO: make faster
+                var current_elements_ids = elements.filter(function(el) {
+                    var duration = moment.duration(date.diff(el.create_date));
+                    if (options.domain === 'hour') {
+                        return duration.years() === 0 && duration.months() === 0 && duration.weeks() === 0 && duration.days() === 0 && duration.hours() === 0;
```
Tardo: Can be better for not equal cases change the order:
```suggestion
                        return duration.hours() === 0 && duration.days() === 0 && duration.weeks() === 0 && duration.months() === 0 && duration.years() === 0;
```

Tardo (changes requested):

web_widget_heatmap/static/src/css/heatmap_widget.css
pedrobaeza: Better to use `sass` as `less` is deprecated

Tardo: Is 11.0

pedrobaeza: But sass is also present in 11.0. In 12.0, you have scss, with retro-compatibility with sass.

Tardo: Oh! thanks.

web_widget_heatmap/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+        "web",
+    ],
+    "data": [
+        "views/web_widget_heatmap_template.xml",
+    ],
+    "demo": [],
+    "qweb": [
+        "static/src/xml/base.xml"
+    ]
+}
```
MeetKD, line 30: You can remove empty keys and may b change category to web?

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+                        return duration.years() === 0 && duration.months() === 0;
+                    } else if (options.domain === 'year') {
+                        return duration.years() === 0;
+                    }
+                    return false;
+                }).map(function(el) {
+                    return el.id;
+                });
+                var action = {
+                    name: date.format('YYYY-MM-DD HH:mm'),
```
MeetKD, line 178: Can we take dateformat from l10n itself?

MeetKD (changes requested):

web_widget_heatmap/static/src/css/heatmap_widget.css
agent-z28: Thats 

> I like this widget.
> The loading problem can also be solved by making the widget work for Serialized fields. The programmer can make a computed and stored Serialized field which stores the records as JSON. This way every time the partner record opens, not all the sale orders are fetched instead just a text field essentially.

Thats the perfect solution, with a depends on the interesting fields

web_widget_heatmap/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+    "development_status": "Stable",
+    "category": "Extra Tools",
+    "website": "https://github.com/OCA/web",
+    "author": "Dinar Gabbasov, Odoo Community Association (OCA)",
+    "maintainers": ["GabbasovDinar"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "preloadable": True,
+    "external_dependencies": {
```
daramousk, line 16: Remove mappings with no values.

web_widget_heatmap/static/lib/js/cal-heatmap.js
```diff
@@ -0,0 +1,3487 @@
+/*! cal-heatmap v3.6.2 (Mon Oct 10 2016 01:36:20)
```
daramousk, line 1: Use `cal-heatmap.min.js` instead.

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+*/
+
+odoo.define('web_widget_heatmap.widget', function (require) {
+    "use strict";
+
+    var AbstractField = require('web.AbstractField');
+    var registry = require('web.field_registry');
+
+
+    var HeatMapWidget = AbstractField.extend({
```
daramousk, line 13: If this widget can work with x2many fields, why not inherit from `FieldX2Many` instead?

web_widget_heatmap/static/src/js/heatmap_widget.js
```diff
@@ -0,0 +1,194 @@
...
+        previousHeatMap: function() {
+            this.heatmap.previous();
+        },
+
+        /**
+         * Find the records that match the selected date and show in the tree view
+         * @param {Date} date
+         * @returns {jQuery.Deferred} Action loaded
+         */
+        onClickHeatMap: function(date) {
```
daramousk, line 154: I am not sure I understand this function, and the references to dates.

A heatmap's elements as I understand can represent a lot of things, not just dates/time. This heatmap could have as elements the records of the XMany field and depending on a field on those records proper opacity can be generated. 
I thin

daramousk (commented):
Great contribution, thanks.

I was under the impression that something like this existed, and after looking a little bit around it looks that the module `web_graph` existed on 8.0 that did something like this, but it seems it has been removed, anyway I have left some comment in which I would like your opinion on.

Thanks

web_widget_heatmap/static/lib/js/cal-heatmap.js
GabbasovDinar: Why? Odoo automatically minimizes js files. This code format allows you to conveniently debug code in debug=assets mode

GabbasovDinar: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#4javascript

legalsylvain: OK to keep non minimizes files.

PR #1305, opened 2019-06-11
[12.0][IMP] web_responsive

### LINKED ISSUE #1189: [12.0] web_responsive: Shrinking window with sided chat hides some buttons

### LINKED ISSUE #1228: [12.0] web_responsive section change delay
In the web_responsive module once you want to change sections as shown in the gif below, the interface will dismiss the Apps menu and will remain for some time in the previous section while it charges and goes to the next one. 
GIFrecord_2019-03-27_111936
This can be a little bit confusing for the user experience when the charging takes more than a few seconds.

I'm working on finding a clean fix to this trying to make the dropdown menu to be dismissed once the new section has been fully charged instead of when the user has clicked the app icon.
Any help is welcomed.

### LINKED ISSUE #1309: [12.0] web_responsive: The icons on the dashboard are displayed disproportionately (In Google Chrome only)
The icons on the dashboard are displayed disproportionately (In Google Chrome only)
IconDasboard

Tardo:
- Fix long title in forms:
title_ellipsis

- Added Sticky style in tree view and form again (like in v11.0):
sticky

- Should fix #1189 issue:
resize

- Should fix #1228 issue:
appmenu

- Should fix #1309 issue


pedrobaeza:
For "Fix long title in forms:" in read mode, I would prefer text-overflow capability with tooltip with full text.

Tardo:
@pedrobaeza To do this need:
- Javascript to set the tooltip in .oe_title element
- More javascript for mobile solution (perhaps click and expand?)

Other problem is that the tooltip with the full text will overwrite the "help" tooltip and the debug tooltip in developer mode.

pedrobaeza:
And what about only text overflow?

Tardo:
Title with ellipsis, it expands when it is active (left click is maintained). It works also in mobile. What do you think?
title_ellipsis

** Made with only CSS

pedrobaeza:
Wonderful!

pedrobaeza:
#1228 feature deserves an entry in the README, although the GIF you put is not very explanatory as there's no delay. Maybe you can simulate the delay forcing in developer tools of the browser a very slow connection.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -1,6 +1,29 @@
...
+  75% {
+    transform: translateY(-10px);
+  }
+  100% {
+    transform: translateY(0);
+    transform-origin: bottom center;
+  }
+}
+.o-app-icon-waiting {
+  animation: 1.3s infinite cubic-bezier(.65,.05,.36,1) o-app-icon-bounce-animation;
```
yajo: I'd add here also a `cursor: progress`. If possible, in the whole screen; otherwise just in the icon (you are supposed to have just clicked there).

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -31,8 +54,7 @@
             }
 
             .o_menu_sections,
-            .o_menu_systray,
-            {
+            .o_menu_systray {
```
yajo: I'd leave this as it was, it looks good for when there are multiple selectors IMHO.

The same applies for similar diffs.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -324,19 +344,67 @@ html .o_web_client .o_main .o_main_content {
 .o_content, .modal-content {
     max-width: 100%;
 
+    // Sticky Header & Footer in List View
+    >div {
```
yajo: You're inside `.modal-content`. Does that make sense?

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -324,19 +344,67 @@ html .o_web_client .o_main .o_main_content {
...
+        >.table-responsive {
```
yajo: Can't we just apply it to all `.table-responsive` found? If possible it's good to `avoid > too > strict > structures`. It's usually better to be less strict and just apply to classes inside classes or such. In a future version they could add/remove an element in the chain and nobody would notice this section is not working. Being less strict, that's less possible. It's also a new approach in BS4.

Of course, it's not always possible; I'm not sure if it's the case here...

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -324,19 +344,67 @@ html .o_web_client .o_main .o_main_content {
...
+            overflow-x: initial;
+
+            >.o_list_view {
```
yajo: same about hierarchy strictness.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -324,19 +344,67 @@ html .o_web_client .o_main .o_main_content {
...
+                // th & td are here for compatibility with chrome
+                thead, thead tr:nth-child(1) th {
```
yajo: So if a `thead` has multiple rows, why only the 1st one is sticky? :thinking: 

Also, if `thead th` works for other browsers than Chrome too, then can we just leave that?

(the same questions apply to other similar selectors below)

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -324,19 +344,67 @@ html .o_web_client .o_main .o_main_content {
...
+            width: initial;
+
+            >h1>span {
+                max-width: 100%;
+                text-overflow: ellipsis;
+                white-space: nowrap;
+                overflow: hidden;
+                width: initial;
+            }
+            >h1>span:active {
```
yajo: Here and above, the same about tight structure. Can't we be less strict?

Also a suggestion about readability (this applies to your whole patch, BTW):

```suggestion
            > h1 > span:active {
```

web_responsive/static/src/js/web_responsive.js
```diff
@@ -264,6 +265,14 @@ odoo.define('web_responsive', function (require) {
                 },
             });
         },
+
+        // Load view indicator
+        _onAppsMenuItemClicked: function (ev) {
+            ev.preventDefault();
+            ev.stopPropagation();
```
yajo: Isn't this done in `this._super()`?

web_responsive/static/src/js/web_responsive.js
```diff
@@ -264,6 +265,14 @@ odoo.define('web_responsive', function (require) {
...
+            $(ev.currentTarget).find('img').addClass('o-app-icon-waiting');
```
yajo: I'd change it like this:

```suggestion
            $(ev.currentTarget).addClass('o-app-icon-waiting');
```

Then the CSS above can just apply to `.o-app-icon-waiting img`. It gives you a little bit more flexibility in case you want i.e. add more styling to the whole block.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -358,7 +367,24 @@ odoo.define('web_responsive', function (require) {
...
+    });
+
+    // Hide AppMenu & remove waiting anim when loaded action
+    ActionManager.include({
+        doAction: function () {
+            return this._super.apply(this, arguments).then(function () {
+                var $app_menu = $('.o_menu_apps .dropdown');
+                if ($app_menu.length) {
+                    if ($app_menu.hasClass('show')) {
+                        $app_menu.dropdown('toggle');
```
yajo: I think you can just do `$app_menu.dropdown("hide")`, and avoid the `if` here.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -358,7 +367,24 @@ odoo.define('web_responsive', function (require) {
...
+                    }
+
+                    $app_menu.find('img.o-app-icon-waiting').removeClass(
```
yajo: As always, the less strict you are, the better :blush::

```suggestion
                    $app_menu.find('.o-app-icon-waiting').removeClass(
```

yajo (changes requested):
A great patch. Some code comments. Thanks!

web_responsive/static/src/css/web_responsive.scss
Tardo: I changed it because the linter doesn't like the last comma, and change the Allman brace style to K&R style to be like as the rest of the file:
screenshot_20190612-120115_1920x1080

web_responsive/static/src/js/web_responsive.js
Tardo: no, the click propagates and causes the menu to close

web_responsive/static/src/js/web_responsive.js
Tardo: Only appears to be implemented the "toggle" method... :suspect: use "hide" or "show" prints: ```TypeError: No method named "hide"```

Tardo:
Changes done... waiting for @Yajo responses.
Changed the "waiting anim" to be less "intrusive":
appmenu

web_responsive/static/src/css/web_responsive.scss
yajo: Hmm but it's valid syntax AFAIK...

Honestly we don't have a linter for *CSS right now at @OCA. It was just a matter of taste. As we usually leave the trailing comma wherever possible to reduce future diffs, I was doing the same here.

I wouldn't introduce these kind of diffs until at least we have a linter here, but I'm open. What do you think @pedrobaeza @moylop260?

web_responsive/static/src/js/web_responsive.js
yajo: Yikes, that's because Odoo 12 uses Bootstrap 4.1, and the `hide` method was implemented in 4.2.

OK, then could you add a TODO comment here indicating that, please?

web_responsive/static/src/css/web_responsive.scss
pedrobaeza: Let's minimize now the diff and open a discussion in MQT about the CSS/SCSS linter.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -270,7 +270,12 @@ odoo.define('web_responsive', function (require) {
         _onAppsMenuItemClicked: function (ev) {
             ev.preventDefault();
             ev.stopPropagation();
-            $(ev.currentTarget).find('img').addClass('o-app-icon-waiting');
+            // Prevents anim more app-icon if user click other before action
+            // is fully loaded
+            this.$el.find('.o-app-waiting').removeClass(
+                'o-app-waiting');
+            $(ev.currentTarget).addClass('o-app-waiting');
+            document.body.style.cursor = 'progress';
```
yajo: Yikes, I have to admit I don't like this approach... A random bug in the JS call stack could leave the cursor in progress mode and confuse the user... As a general rule of thumb, one also try to avoid inline styles as well as `!important` flags in modules. :thinking: 

Yikes, how many rules of thumb I have nowadays! :sweat_smile: 

How about this? You can add a class to the whole app drawer (`this.$el.addClass(".o-app-waiting")`). Then apply the CSS like:

```scss
.o-app-waiting {
    cursor: progress;
    .active img {
        // apply animation here
    }
}
```

Another option would be to just use the waiting overlay that is shipped by odoo  (the famous one of "go grab a coffee"). This would seem like a more integrated UX and probably would mean less code to maintain. Did you consider this option?

web_responsive/static/src/css/web_responsive.scss
pedrobaeza: Any way, @Tardo, any lint fixing is convenient to be put in a separate commit for isolate this noise from the rest of the changes.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -372,16 +377,23 @@ odoo.define('web_responsive', function (require) {
...
+                if ($app_menu.length && options &&
+                        'action_menu_id' in options) {
+                    var $app = $app_menu.find('.o_app.active');
+                    var menu_id = $app.data('menuId');
+                    if (menu_id === options.action_menu_id) {
+                        if ($app_menu.hasClass('show')) {
+                            $app_menu.dropdown('toggle');
+                        }
+
+                        document.body.style.cursor = 'default';
```
yajo: This should disappear too after the above suggestion is applied.

yajo (changes requested):

web_responsive/static/src/js/web_responsive.js
Tardo: I don't like the Odoo overlay because I associate it to "long loading" period... see it all the time can give the feeling of a slow system, when it is not.
To avoid apply styles directly i change it to add a class instead: ```$(document.body).addClass('o-cursor-waiting');``` Is applied to whole page because the user can hide the AppDrawer before action is completed.
Now the check to remove the 'waiting class' are done "always" (sucess and fail).

Tardo:
@pedrobaeza Added your suggestion:
menu

web_responsive/static/src/js/web_responsive.js
```diff
@@ -358,7 +405,61 @@ odoo.define('web_responsive', function (require) {
...
+
+                        $menu.find('li.show').each(
+                            function (i, el) {
+                                if ($.contains(el, $item[0])) {
+                                    $(el).find('.dropdown-toggle')
+                                        .dropdown('toggle');
+                                }
+                            });
+
+                        $(document.body).removeClass('o-cursor-waiting');
```
yajo: Both of these calls are inside `if` blocks.

If none of them evaluate to true, nothing would remove this class. Is that the expected behavior, or should this call be moved to the end, outside of both `if`s?

I cannot test because there's no runbot... :confused:

yajo (approved):
It looks much better now, thanks! :blush:

web_responsive/static/src/js/web_responsive.js
pedrobaeza: @Tardo please confirm this

GSLabIt:
Seems there is a small issue on click search field in app drawer, it close the drawer instead of allowing to search.

tested on mobile Samsung internet browser and Chrome 74.

step to reproduce:

1. login to runbot
2. click to open app drawer
3. click on settings (or other app)
4. click again to open app drawer
5. click inside search input

the app drawer will be closed instead of selecting search input field.

pedrobaeza:
@GSLabIt and that wasn't happening before this PR?

GSLabIt:
@pedrobaeza nope

web_responsive/static/src/js/web_responsive.js
Tardo: Added a "fallback": waiting cursor are removed after ~3 seconds (is the time it takes to appear the BlockUI)

Tardo:
@GSLabIt fixed, thanks!

GSLabIt:
Tested on runbot, it seems to be fixed

yajo:
I hope this helps: https://github.com/odoo/odoo/pull/34147

Tardo:
Code refactored, thanks to @Yajo for the help

web_responsive/readme/DESCRIPTION.rst
```diff
@@ -59,3 +59,7 @@ Features for computers:
 * Sticky chatter topbar
 
   .. image:: ./static/img/chatter_topbar.gif
+
+* AppMenu waits for action finished to show the view
+
+  .. image:: ./static/img/appmenu.gif
```
pedrobaeza: Please put `..` instead of `.` (and you can do the same change in the rest).  With this, you can see from both the generated README and this one properly the images.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -330,6 +331,29 @@ html .o_web_client .o_main .o_main_content {
             max-width: calc(100% - 32px);
         }
 
+        // Sticky statusbar
```
pedrobaeza, line 335: This features is not mentioned in the README, isn't it?

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -446,3 +482,32 @@ html .o_web_client .o_main .o_main_content {
         }
     }
 }
+
+// Sticky Header & Footer in List View
```
pedrobaeza, line 521: This features is not mentioned in the README, isn't it?

gaikaz:
Nice work, but tested the fix for #1189 issue and it's better, but not fully fixed.
The *Send message* and *Log note* tabs open up message box that messes things up a little. Look here:
image
This happens on screens between 768px and 1128px wide.
Personally, I really don't see the floating message box being very useful when chatter is on the side on screens smaller than 1200px. It's not built to be responsive to such small spaces (the avatar takes too much space; the margins/paddings are too big)

Maybe just make it move down sooner with media queries?

Also, I've seen third party modules add additional buttons and icons near the attachments icon. In those cases, the list is longer and could drop to second row again (like in the issue #1189).

pedrobaeza:
Yes, the solution should be to put on that window sizes the chatter again on bottom (or technically, I would say not to move it to the side on that resolutions).

Tardo:
@pedrobaeza @gaikaz @Yajo Changed to use sided chatter on large screens or higher.

pedrobaeza:
I still have some problems on mobile view. Simulating a Samsuns S9, I find that the mobile view has side scrollbar due to the length of the chatter options:

Selección_037

and other weird thing is that clicking on the scroll bar (with left click) and dragging it, a right click is performed (due to extra JS code).

Tardo:
@pedrobaeza try resizing and F5 to refresh the view (updates the layout, fields are displayed using two rows)

pedrobaeza:
I have done it on Firefox. Are you using Chrome maybe?

Tardo:
Firefox too :/
screenshot_20190618-012743_1920x1080


The right-click issue it's really weird :/ appears to be an firefox issue... it happens on all the pages.

pedrobaeza:
Let's check tomorrow about the width issue.

Tardo:
Ok! ah, I know what's happening ... Firefox is making use of mobile gestures. Long press = Show contextual menu

pedrobaeza:
So it's going to happen in mobiles or dragging there doesn't trigger this?

pedrobaeza:
And putting this in Known issues will save us further investigation in the future.

Tardo:
On mobile works as expected... no menu shown.

yajo:
Thanks to https://github.com/OCA/web/pull/1305#issuecomment-502796749 I just noticed this, that you might want to fix too:

Captura de pantalla de 2019-06-18 08-33-57

Captura de pantalla de 2019-06-18 08-38-27



Captura de pantalla de 2019-06-18 08-34-55


Captura de pantalla de 2019-06-18 08-35-07



When re-enabling the followers widget and adding these rules to small screens:

```css
.o_chatter .o_chatter_topbar {
    height: auto;
    flex-wrap: wrap-reverse;
}

.o_chatter .o_chatter_topbar > .o_topbar_right_area {
    flex: 1 0 auto;
    flex-wrap: wrap;
    max-width: 100%;
}
```

It looks like this:

Peek 18-06-2019 08-49

:point_up: You see the followers button is still glitchy in veeery small sizes because it overlaps with the attachments button, but I'm pretty sure you can fix that in no time. :wink:

Tardo:
Now the followers section and send button is displayed on mobile. The avatar is hidden to gain space
chatter

AdriaGForgeFlow (changes requested):
The following error happened to me sometime ago when I was working with the web_responsive module. It appears that the action used to hide the apps menu is not triggered when you have already visit the page as if the page was somehow stored in cache, I'm not entirely sure. Maybe you can find out what's going on.

Peek 2019-06-20 15-14


**Update**: It only happens whit the Discuss app. Whenever you click on it the second time it stops working. Works fine with the other apps though.

Tardo:
@ageficent Thanks for report. Hope now it's fixed.

AdriaGForgeFlow (approved):
Works like a charm now!

Thanks for these improvements, great job :+1:

GSLabIt (approved):
Functional tests LGTM

PR #1306, opened 2019-06-13
[10.0][ADD] Add web_selection_autocomplete module

MicheleZak:
Add _selection_autocomplete_ widget, which allow to use the jquery-ui _autocomplete_ widget on selection fields.


pedrobaeza:
What are the advantages of using this widget?

MicheleZak:
It allows to type inside the input element to filter the selection values.

PR #1308, opened 2019-06-18
[12.0][ADD] web_selection_autocomplete

MicheleZak:
Add jquery-ui autocomplete widget on selection fields, allowing to
type inside the input to filter the values.


pedrobaeza:
What is the advantage of using this?

MicheleZak:
Same as the previous pr, but for version 12 of odoo.
Make selection field ui more similar to the many2one's, allowing to filter the selection values.

The most advantageous case is when:
- there are a lot of selection values
- new model not needed but still want the many2one field ui

pedrobaeza:
Well, I'm not able to see the advantages over other UI confusions, but it's only my opinion. While you specify everything well in the README (please add explanations like the one you have given to me) and comply with OCA guidelines, this can continue.

web_selection_autocomplete/.eslintrc.json
```diff
@@ -0,0 +1,291 @@
+{
```
daramousk, line 1: I think this needs to be removed.

web_selection_autocomplete/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2019 LevelPrime srl
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': "Web Selection Autocomplete",
+    'summary': """
```
daramousk, line 6: This has a max length of 256 chars, reduce this description to less than 256 characters because it will error out on postgres.

web_selection_autocomplete/static/src/scss/web_selection_autocomplete.scss
```diff
@@ -0,0 +1,9 @@
+.o_selection_autocomplete_invalid {
+    &::before {
+        color: #DF382C !important;
+        content: 'Wrong value. Insert one of the available options.' !important;
```
daramousk, line 4: Is there a need for this? Can we use the already existing mechanism (ServiceProviderMixin.do_warn) to inform the user that their input is invalid?

web_selection_autocomplete/static/src/xml/template.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates id="template" xml:space="preserve">
+
+<t t-name="SelectionAutocomplete">
+    <span t-if="widget.mode === 'readonly'"/>
```
daramousk, line 5: Why not `t-if t-else`?

daramousk (commented):
Ok I can see why you need this, it does work ok in listview and form view but I understand that some modifications could be made.

web_selection_autocomplete/__manifest__.py
NL66278, line 6: A summary should be a single line. The longer description is in the README.

web_selection_autocomplete/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+            - there are a lot of selection values
+            - new model not needed but still want the many2one field ui
+    """,
+    'author': "LevelPrime srl, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/web",
+    'license': 'AGPL-3',
+    'development_status': 'Beta',
+    'category': 'Web',
+    'version': '12.0.1.0.0',
+    'depends': ['base', 'web'],
```
NL66278, line 20: Na need for depends on base.

web_selection_autocomplete/static/src/xml/template.xml
```diff
@@ -0,0 +1,20 @@
...
+
+    <input t-if="widget.mode !== 'readonly'"
+        class="o_input"
+        type="text"
+        t-att-name="widget.name"
+        t-att-tabindex="widget.attrs.tabindex"
+        t-att-autofocus="widget.attrs.autofocus"
+        t-att-id="widget.idForLabel"/>
+</t>
+
```
NL66278, line 15: Why the empty lines?

NL66278 (changes requested):
Please take up comments and make sure travis is green.

PR #1311, opened 2019-06-23
[FIX] web_responsive: AppDrawer Discard

keshrath:
ezgif com-video-to-gif (19)

### DIFF
--- a/web_responsive/README.rst
+++ b/web_responsive/README.rst
@@ -156,6 +156,7 @@ Contributors
 * Dennis Sluijk <d.sluijk@onestein.nl>
 * Sergio Teruel <sergio.teruel@tecnativa.com>
 * Alexandre Díaz <dev@redneboa.es>
+* Mathias Markl <mathias.markl@mukit.at>
 
 Maintainers
 ~~~~~~~~~~~
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -7,6 +7,7 @@ odoo.define('web_responsive', function (require) {
     var ActionManager = require('web.ActionManager');
     var AbstractWebClient = require("web.AbstractWebClient");
     var AppsMenu = require("web.AppsMenu");
+    var BasicController = require('web.BasicController');
     var config = require("web.config");
     var core = require("web.core");
     var FormRenderer = require('web.FormRenderer');
@@ -110,6 +111,14 @@ odoo.define('web_responsive', function (require) {
             this.$search_input = this.$(".search-input input");
             this.$search_results = this.$(".search-results");
             return this._super.apply(this, arguments);
+        },
+
+        /**
+         * @override
+         */
+        _onAppsMenuItemClicked: function (ev) {
+            this._super.apply(this, arguments);
+            ev.preventDefault();
         },
 
         /**
@@ -276,10 +285,21 @@ odoo.define('web_responsive', function (require) {
         },
     });
 
+    BasicController.include({
+        
+        /**
+         * @override
+         */
+        canBeDiscarded: function (recordID) {
+            if (this.model.isDirty(recordID || this.handle)) {
+                $('.o_menu_apps .dropdown:has(.dropdown-menu.show) > a').dropdown('toggle');
+            }
+            return this._super.apply(this, arguments);
+        },
+    });
+
     Menu.include({
         events: _.extend({
-            // Clicking on apps menu
-    	    "click .o_menu_apps a[data-toggle=dropdown]": "_onAppsMenuClick",
             // Clicking a hamburger menu item should close the hamburger
             "click .o_menu_sections [role=menuitem]": "_hideMobileSubmenus",
             // Opening any dropdown in the navbar should hide the hamburger
@@ -324,22 +344,6 @@ odoo.define('web_responsive', function (require) {
         _updateMenuBrand: function () {
             if (!config.device.isMobile) {
                 return this._super.apply(this, arguments);
-            }
-        },
-        
-        /**
-         * Check if Controller can be removed
-         */
-        _onAppsMenuClick: function(event, checkedCanBeRemoved) {
-            var action_manager = this.getParent().action_manager;
-            var controller = action_manager.getCurrentController();
-            if (controller && !checkedCanBeRemoved) {
-                controller.widget.canBeRemoved().done(function () {
-                    $(event.currentTarget).trigger('click', [true]);
-                    $(event.currentTarget).off('.bs.dropdown');
-                });
-                event.stopPropagation();
-                event.preventDefault();
             }
         },
     });
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -114,6 +114,8 @@ odoo.define('web_responsive', function (require) {
         },
 
         /**
+         * Prevent the menu from being opened twice
+         * 
          * @override
          */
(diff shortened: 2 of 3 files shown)


keshrath:
@Yajo @Tardo can you please take a look at this PR.

pedrobaeza:
@Yajo is currently on holidays. I suppose that the problem is that previously to this PR the message doesn't appear, isn't it?

keshrath:
No, the gif shows the current situation. So technically there is no error, but I found it iritating that the warning pops up twice on top of the menu. The PR checks if a save is necessary before opening the drawer. The behaviour after the PR is comparable to the EE version.

ezgif com-video-to-gif (20)

pedrobaeza:
Well, then I don't see any problem. Clicking (or pushing the combination key shortcut) doesn't mean that you finally go away from current form. You just need to press Escape or click again on the app switcher to come back to current form. Only when clicking actually on a menu is when you go away from current form, and thus showing that message.

So for me, this is :-1: I will wait any way the opinion of others.

Tardo:
Thanks for report. The problem is that the load is executed two times (one by AppMenu and other time because 'a' change the url). The fix is remove the template that adds "href" to '.o_app' elements.
https://github.com/OCA/web/blob/12.0/web_responsive/static/src/xml/apps.xml#L7-L10

pedrobaeza:
@Tardo that is put for a reason: to have the possibility to right click (or Ctrl-Click) on the app for opening it in another tab. Any way, that's not exactly what we are discussing here IMO.

Tardo:
I think like Pedro, going to the menu does not have to indicate that I want to leave the view.

I think that this issue need fix two problems:
1. Fix app click:
On 'AppsMenu':
```javascript
_onAppsMenuItemClicked: function (ev) {
    ev.preventDefault();
},
```

2. Close AppsMenu if have changes to discard:
```javascript
BasicController.include({
    canBeDiscarded: function (recordID) {
        if (this.model.isDirty(recordID || this.handle)) 
            $('.o_menu_apps .dropdown:has(.dropdown-menu.show) > a')
                .dropdown('toggle');
        }
    },
});
```
Best wait for @Yajo that is the principal maintainer of 'web_responsive' module and have a better idea about how fix this.

keshrath:
I see your point about opening the menu without directly assuming that it will be changed. I have implemented the suggestion of @Tardo, which solves the problem without having to check it when opening the menu.

Result:

ezgif com-video-to-gif (21)

pedrobaeza:
And what happens if you click in OK? Does it go to the clicked App? It has been implemented recently that the app switcher or the menu keeps opened while data are not totally received, so I think this goes against this pattern.

keshrath:
In this case it switches to the menu that was clicked before. In my opinion this is the effect I would expect.

1. Readonly mode -> menu click -> stay in the menu until the action is loaded.
2. Edit mode (no change) -> menu click -> stay in the menu until the action is loaded.
3. Edit mode (change) -> menu click -> show the warning on the record, because it belongs there contextually -> click ok -> switch to the selected menu
4. Edit mode (change) -> menu click -> display the warning on the record, because it belongs there contextually -> click cancel -> stay on the record

ezgif com-video-to-gif (22)

pedrobaeza:
Thanks for your PoV, @keshrath, and thanks also for involving here :smiley: It's good to see you around in OCA.

@Tardo what's your opinion? I also want to wait for @Yajo's opinion when he comes back from holidays.

Tardo:
For me, it's ok :+1: but can be better if can "fix" this:
menu_f

keshrath:
I think that should solve it:

ezgif com-video-to-gif (23)

web_responsive/static/src/js/web_responsive.js
```diff
@@ -276,6 +285,20 @@ odoo.define('web_responsive', function (require) {
...
 
+    BasicController.include({
+        
+        /**
+         * @override
+         */
+        canBeDiscarded: function (recordID) {
+            if (this.model.isDirty(recordID || this.handle)) {
+                $('.o_menu_apps .dropdown:has(.dropdown-menu.show) > a').dropdown('toggle');
+                $('.o_menu_sections li.show:has(.dropdown-menu.show) > a').dropdown('toggle');
```
Tardo: I know that is confuse.. but best "attack" the element that have "toggle='dropdown'" defined in the dataset.

``` $('.o_menu_sections li.show .dropdown-toggle').dropdown('toggle');```

**Review by Tardo — CHANGES_REQUESTED** [review]

File: web_responsive/static/src/js/web_responsive.js (hunk 1 above)
Comment (keshrath) [reply web_responsive/static/src/js/web_responsive.js]: https://github.com/OCA/web/compare/044ccb31de238ccb90e302bc0fca9376f3bd777f..156d3bb5d69e5397223c6f63680716634ae00f40

**Review by Tardo — APPROVED** [review]
LGTM :+1:

**Comment by pedrobaeza** [comment]
As said, I'll wait @Yajo's opinion on this when he comes back from holidays if you don't mind.

**Comment by keshrath** [comment]
No of course not, but this way he can immediately have a look at the best possible version of the PR 😃

File: web_responsive/static/src/js/web_responsive.js [hunk 2]
```diff
@@ -112,6 +113,14 @@ odoo.define('web_responsive', function (require) {
             return this._super.apply(this, arguments);
         },
 
+        /**
+         * @override
+         */
+        _onAppsMenuItemClicked: function (ev) {
+            this._super.apply(this, arguments);
+            ev.preventDefault();
```
yajo, line 123: Add a comment explaining why you do this.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -276,6 +285,20 @@ odoo.define('web_responsive', function (require) {
         },
     });
 
+    BasicController.include({
+        
+        /**
+         * @override
+         */
+        canBeDiscarded: function (recordID) {
+            if (this.model.isDirty(recordID || this.handle)) {
```
yajo, line 298: Also here. Code is not obvious.

yajo (changes requested):

web_responsive/static/src/js/web_responsive.js
keshrath: https://github.com/OCA/web/commit/ce0391a0a0a21de4cc3caffa6829a99437e3db0e

web_responsive/static/src/js/web_responsive.js
keshrath: https://github.com/OCA/web/commit/ce0391a0a0a21de4cc3caffa6829a99437e3db0e

yajo (approved):
It seems good to me.

web_responsive/README.rst
```diff
@@ -156,6 +156,7 @@ Contributors
 * Dennis Sluijk <d.sluijk@onestein.nl>
 * Sergio Teruel <sergio.teruel@tecnativa.com>
 * Alexandre Díaz <dev@redneboa.es>
+* Mathias Markl <mathias.markl@mukit.at>
```
pedrobaeza, line 159: Unless you put this on `readme/CONTRIBUTORS.rst`, it will be lost on README regeneration.

pedrobaeza (changes requested):

web_responsive/README.rst
keshrath: https://github.com/OCA/web/commit/2e911c6b028ecbd700ae041e7f426e4ca30703c1

pedrobaeza (approved):

PR #1320, opened 2019-07-03
[FIX] web_export_view: float time format broken

syci:
In views with float_widget fields when we use the export, truncate the data

### DIFF
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -69,7 +69,7 @@ odoo.define('web_export_view', function (require) {
                             if (text.search(":")) { 
                                 export_row.push(text);
                             }
-                            if ($cell.hasClass("o_list_number")) {
+			    else if ($cell.hasClass("o_list_number")) {
                                 export_row.push(parseFloat(
                                     text
                                     // Remove thousands separator
--- a/web_export_view/static/src/js/web_export_view.js
+++ b/web_export_view/static/src/js/web_export_view.js
@@ -65,6 +65,10 @@ odoo.define('web_export_view', function (require) {
                         }
                         else {
                             var text = $cell.text().trim();
+                            // Search : for hour value
+                            if (text.search(":")) { 
+                                export_row.push(text);
+                            }
                             if ($cell.hasClass("o_list_number")) {
                                 export_row.push(parseFloat(
                                     text


pedrobaeza:
Please document in the commit message and here the error you get.

syci:
Sorry, new commit on file https://github.com/syci/web/commit/8ff068506a0e4a145147126f02d242bf4f0e0862

web_export_view/static/src/js/web_export_view.js
```diff
@@ -65,7 +65,11 @@ odoo.define('web_export_view', function (require) {
                         }
                         else {
                             var text = $cell.text().trim();
-                            if ($cell.hasClass("o_list_number")) {
+                            // Search : for hour value
+                            if (text.search(":")) {
```
simahawk, line 69: As far as I remember you still get the class for a number but you get `:` into it, right?
If so, this should be an exception for the other IF block (w/ proper comment):
```js
if ($cell.hasClass("o_list_number") && !text.search(":")):
    // here we have a number w/out time formatting (eg: 3:00)
```

simahawk (changes requested):

web_export_view/static/src/js/web_export_view.js
alexey-pelykh: Maybe test `parseFloat()` results for NaN and in that case write text as-is? I don't think there's any benefit of testing for `:` presence only

alexey-pelykh (changes requested):

web_export_view/static/src/js/web_export_view.js
simahawk:
```js
parseFloat('03:00')
3
```
I  wish you were right :smile:

alexey-pelykh: 🙈 javascript... maybe `parseFloat('03:00').toString() === '03:00'`? Yet this may not be super ok

simahawk:
@syci can you check my comment here https://github.com/OCA/web/pull/1320/files#r300233688 ?

web_export_view/static/src/js/web_export_view.js
alexey-pelykh: I meant 
```
if (parseFloat(text).toString() !== text) {
  // totally not a float
}
```

Tardo: Perhaps can use ```_.isNaN(+text)``` (ensure that text have `.` as decimal separator)

daramousk (commented):
Hi there, thanks for the fix.

I went on to reproduce and I see what you mean, I am rather curious as to why the authors decided to manipulate the `o_list_number` elements in the first place though. If this module is supposed to export
> the tree view they are looking to.

then why manipulate the numbers in the first place?
1) the thousands seperator is different between countries, others use comma others dot, it should probably left as it it
2) same for the decimal separator
3) Non numeric characters are removed. Why?

Going through the commit messages it seems that there was a problem with monetary widgets and that's why this happened, but this handling does not seem appropriate to me, because it can arbitrarily mess with numeric output and cause situations like this.

So perhaps this PR could be in reverting that commit and fixing the issue with the monetary field?

web_export_view/static/src/js/web_export_view.js
scigghia: > As far as I remember you still get the class for a number but you get `:` into it, right?
> If so, this should be an exception for the other IF block (w/ proper comment):
> 
> ```js
> if ($cell.hasClass("o_list_number") && !text.search(":")):
>     // here we have a number w/out time formatting (eg: 3:00)
> ```

it works for me

PR #1326, opened 2019-07-08
 [12.0][MIG] - Wildcard in advanced migration to 12.0

sbejaoui:
In addition to the migration this PR add two new options to web_advanced_search_wildcard: 
- doesn't start with
- doesn't end with


sbejaoui:
@gva-acsone, can you review this please.

gva-acsone:
@sbejaoui : 2 observations:

1. When I apply the 'Doesn't start with' filter, records that contain the value mentioned (for instance, at the end of the value) don't appear in the results.

Example: I make the following research: ZIP doesn't start with 74201. 
Hence, I expect all records that have a zip code that starts differently than with 74201 to appear in the results.
However, some records, that contain the value '74201', but not at the beginning of the zip code, don't appear in the results of the research. For instance, a record with the following zip code: 12374201.
This should be changed.

2. Empty values should appear in the 'Doesn't start with XXXX' filter. However, this is currently not the case.


Can you apply those 2 changes? I will then proceed to another round of tests.

gva-acsone (approved):
+1 Functional test

pedrobaeza:
Shouldn't this be renamed to accommodate new features? And also as the search feature is now called "Custom filter", this can be confused with existing OCA module that adds a new feature `web_advanced_search`. I recommend it to call it `web_custom_filter_operator_extra` or similar.

sbejaoui:
> Shouldn't this be renamed to accommodate new features? And also as the search feature is now called "Custom filter", this can be confused with existing OCA module that adds a new feature `web_advanced_search`. I recommend it to call it `web_custom_filter_operator_extra` or similar.

+1

web_advanced_search_wildcard/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014 initOS GmbH & Co. KG (<http://www.initos.com>).
+# Copyright 2016 Therp BV <http://therp.nl>.
+# Copyright 2017 Alex Comba - Agile Business Group
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Wildcard in advanced search",
+    "summary": "Webmodule to add wildcard operators in advanced search field",
+    "version": "12.0.1.0.0",
```
Tardo, line 8: Please, add the 'website' key

web_advanced_search_wildcard/static/src/js/search.js
```diff
@@ -0,0 +1,29 @@
...
+        {value: "not_startswith", text: _lt("doesn't start with")},
+        {value: "endswith", text: _lt("ends with")},
+        {value: "not_endswith", text: _lt("doesn't end with")},
+        {value: '=ilike', text: _lt("matches")}
+    );
+
+    Char.include({
+        get_domain: function (field, operator) {
+            switch (operator.value) {
+            case '∃': return [[field.name, '!=', false]];
```
Tardo: Undocumented?

Tardo (changes requested):

web_advanced_search_wildcard/static/src/js/search.js
sbejaoui: it was redundant code with the inherited method. I used super instead.

alexey-pelykh (approved):

Tardo (approved):
LGTM :+1:, code review

web_advanced_search_wildcard/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
daramousk: Remove encodings from all *.py files.

web_advanced_search_wildcard/static/src/js/search.js
```diff
@@ -0,0 +1,28 @@
...
+        {value: "endswith", text: _lt("ends with")},
+        {value: "not_endswith", text: _lt("doesn't end with")},
+        {value: '=ilike', text: _lt("matches")}
+    );
+
+    Char.include({
+        get_domain: function (field, operator) {
+            var res = this._super(field, operator);
+            switch (operator.value) {
+            case 'startswith': return [[field.name, '=ilike', this.get_value() + '%']];
```
daramousk: Please fix the switch formatting here if it's possible so that it is more readable.

web_advanced_search_wildcard/static/src/js/search.js
```diff
@@ -0,0 +1,28 @@
...
+            case 'not_startswith': return ['!', [field.name, '=ilike', this.get_value() + '%']];
+            case 'endswith': return [[field.name, '=ilike', '%' + this.get_value()]];
+            case 'not_endswith': return ['!', [field.name, '=ilike', '%' + this.get_value()]];
+            default: return res;
```
daramousk: You could move the super call here:
```
default:
    return this._super(field, operator);

```

so that you can avoid unessesary calls to super since you are not using `res` in any way. So first call your stuff and leave super call for last.

daramousk (commented):
Works great, some recommendations.

andreampiovesana (approved):
go

Abranes (approved):

andreampiovesana:
merge!

legalsylvain:
@sbejaoui : could you take a look on the @daramousk remarks ? 

If you have no time, I propose to merge this PR, and wait for another one regarding @daramousk remarks that seems to be cosmectics / non blocking remarks.

thanks.

sbejaoui:
> @sbejaoui : could you take a look on the @daramousk remarks ?

Done

fuentes73 (approved):

ValentinVinagre (approved):

PR #1327, opened 2019-07-09
[MIG] web_editor_background_color: Migrate to v12

yajo:
- Remove bootstrap-colorpicker library, which is deprecated.
- Use Odoo's internal color picker dialog.

This is how it looks now, pretty similar to https://github.com/odoo/odoo/pull/34687:

Peek 09-07-2019 12-04

@Tecnativa


pedrobaeza (approved):
Tried on runbot and working on both backend (mass mailing) and website.

Is there any change of getting live preview when selecting the color the same you have when you pass the mouse over the rest of the template colors?

yajo:
I thought about that, but I researched and we'd need  the color picker dialog to emit specific events when the user plays with the colors (to preview) and when the dialog is cancelled without saving (to restore). However, none of these exist in current implementation AFAICS, so I'd have to override quite a bit of that class behavior to get that... and after all the dialog is modal and grays out the background, so I'm not very sure this would be worth the effort...

Maybe we could handle that for v13 if they merge upstream, since controlling upstream code all of this is much easier.

Also, I wouldn't like to add a feature that might disappear in v13.

pedrobaeza:
OK, please annotate this on ROADMAP then.

yajo:
Done

web_editor_background_color/static/src/css/background_color.scss
```diff
@@ -0,0 +1,14 @@
+/* Copyright 2017-2019 Jairo Llopis <jairo.llopis@tecnativa.com>
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+.snippet-option-colorpicker {
+    .o_colorpicker_section[data-name="custom"]>.text-muted {
```
Tardo:
```suggestion
    .o_colorpicker_section[data-name="custom"] > .text-muted {
```

Tardo (approved):
Only some observations that aren't related with this PR directly... but Color Picker Dialog can restore all values except "Opacity" if you input a wrong value :/ and hex color field doesn't support abbreviated format: ```#CCC -> #CCCCCC```

PR #1330, opened 2019-07-10
[FIX][12.0] web_timeline

hpatelserpentcs:
Users are having trouble scrolling timeline view.

### DIFF
--- a/web_timeline/README.rst
+++ b/web_timeline/README.rst
@@ -183,6 +183,7 @@ Contributors
 * Leonardo Donelli <donelli@webmonks.it>
 * Adrien Didenot <adrien.didenot@horanet.com>
 * Dennis Sluijk <d.sluijk@onestein.nl>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
 
 Other credits
 ~~~~~~~~~~~~~
--- a/web_timeline/readme/CONTRIBUTORS.rst
+++ b/web_timeline/readme/CONTRIBUTORS.rst
@@ -4,3 +4,4 @@
 * Leonardo Donelli <donelli@webmonks.it>
 * Adrien Didenot <adrien.didenot@horanet.com>
 * Dennis Sluijk <d.sluijk@onestein.nl>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
--- a/web_timeline/static/src/css/web_timeline.css
+++ b/web_timeline/static/src/css/web_timeline.css
@@ -29,3 +29,35 @@
     left: 0px;
     top: 0px;
 }
+
+.vis-labelset {
+    position: relative;
+    width: 100%;
+    height: 100%;
+    overflow: visible !important;
+}
+
+.vis-labelset .vis-label:first-child {
+    background: lightgrey;
+    position: sticky;
+    top: 0px !important;
+    z-index: 999;
+    max-height: 100px;
+    height: 100px !important;
+    display: -webkit-flex;
+}
+
+.vis-foreground .vis-group:first-child {
+    background: lightgrey;
+    position: sticky;
+    top: 0px !important;
+    z-index: 999;
+    max-height: 100px;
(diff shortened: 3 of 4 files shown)


hpatelserpentcs:
@max3903 Can you please review this PR.

Cc @wolfhall

pedrobaeza:
Please show before and after screenshots

wolfhall (approved):

smangukiya (approved):

max3903 (approved):

pedrobaeza:
I have asked for screenshots of the changes because I'm not sure about these changes. I will check and if not agree, I will revert.

wolfhall:
Pedro, 
The current time view has some serious limitation: We have unassigned work that needs to be dragged to a worker (see screenshot)  we have maybe 50 items scheduled for the same time and 100 people to do the work.  then the items need to be dragged and dropped onto the workers below.  the current time view does not support that because you cannot drop an item on workers down in the list. See attached screen for the current implementation.
FSM-0002
The proposed fix creates separate sections that are individually scrollable. 
If this is too disruptive for other users who have fewer items to work with, we could make these views turn on or off via configurations.

hpatelserpentcs:
@pedrobaeza As per above @whafhall comment I will add .gif file of it.

**Before Scroll configuration :** 
demo1


**After Scroll Configuration:**

demo2

Cc @wolfhall @max3903

pedrobaeza:
Thanks for the screenshots. That's exactly what I needed before merging so fast for discussing this.

Seeing the GIFs, can you tell me what is the grey area?

hpatelserpentcs:
@pedrobaeza 

In grey area(First row of timeline view) displaying all orders list which is not assigned to partner.

pedrobaeza:
This is what I supposed, and this is why I don't like this change: doing sticky something that way matches your use case for field-service, but not the general one. Sticky group must be optional to be selected through the timeline definition. I would say that it would be even desirable to have the possibility of setting which value of the group must be the sticky one. Imagine you group by state, and you want the stick one to be the state "Draft".

hpatelserpentcs:
> This is what I supposed, and this is why I don't like this change: doing sticky something that way matches your use case for field-service, but not the general one. Sticky group must be optional to be selected through the timeline definition. I would say that it would be even desirable to have the possibility of setting which value of the group must be the sticky one. Imagine you group by state, and you want the stick one to be the state "Draft".

@max3903 @wolfhall

wolfhall:
@hpatelserpentcs  ok. we can make it configurable.  Hemangini, please make the change to allow the scrollable function to be turned on and off.  If it is not much effort  (one hour) , selecting the field to see what is in the upper area is a plus.  Others can make that improvement otherwise.

gaikaz:
This seems to be breaking some uses of the timeline module. We (company I'm from) are developing a custom app for restaurant menu tracking and after this PR got merged, these problems rose:
1. It doesn't work well with `web_responsive`, as the new sticky part of the timeline is on top of app menu (see **pic. 1**)
2. Some odd spacings appear above and inside the sticky part. (see **pic. 2**)
3. ~~If the active model of items has the Y axis field set to required, we get an error (see **error** below) if the item is put inside the sticky part. _I guess the solution here would be, that the sticky part should not exist, if Y axis field is a required field._~~ _Edit: This was a non-related issue, just much less noticeable before this PR. Might need a separate issue created_

~~**error about the required field**:~~
```
The operation cannot be completed:
- Create/update: a mandatory field is not set.
- Delete: another model requires the record being deleted. If possible, archive it instead.

Model:  ...
```

**pic 1**:
pic1

**pic 2**:
pic2

pedrobaeza:
Reverting the PR then, as confirmed it can't be done this wild mergings without more agreement

pedrobaeza:
Reversed on 79df8aadbe538b83ec2ad620291bbd2007b42478.

Please do another patch and take into account all the remarks, and checking that you don't break things.

PR #1234, opened 2019-04-01
[ADD] Added new module web_widget_prefixed_url

Swapnesh-SerpentCS:
A simple module for adding 'callto:' widget to your odoo fields

`<field name="skype_name" widget="callto"/>`

Ps: I will update PR with README.rst and other fixes If this module is usefull for Community and will be accepted here.

PR #1337, opened 2019-07-20
[12.0][MIG] web_widget_url_advanced

alexey-pelykh:


web_widget_url_advanced/static/src/js/url_widget.js
```diff
@@ -0,0 +1,29 @@
...
+
+    basic_fields.UrlWidget.include({
+        /**
+         * @override
+         */
+        init: function () {
+            this._super.apply(this, arguments);
+            // retrieve customized `<a />` text from a field
+            // via `text_field` attribute or `options.text_field`
+            // NOTE: `<field text_field="foo"..` is not supported in tree views.
```
Tardo: Please, add this in ROADMAP.

Tardo (changes requested):

web_widget_url_advanced/static/src/js/url_widget.js
alexey-pelykh: Thanks, done!

simahawk (approved):

Swapnesh-SerpentCS:
We can add https://github.com/OCA/web/pull/1234#issuecomment-511708931 here too so we don't need an additional module.

simahawk:
@Swapnesh-SerpentCS can be done in second step after this PR is merged. Please review it to speed it up ;)
@Tardo good for you?

Tardo (changes requested):
I see that when you get the "text_field" empty... the field with the url widget doesn't update the text until the page is refreshed.

url_widget

Please, fix or add it to roadmap.

alexey-pelykh:
@simahawk should that be changed? I think it's due to `if (field_value)` check?

alexey-pelykh:
@Tardo I've adjusted the implementation to update in any case as long as field is set

Tardo (approved):
LGTM :+1:, great!

simahawk:
@alexey-pelykh @Tardo sorry guys, I missed it :stuck_out_tongue: You managed it so: let's merge it :)

PR #1338, opened 2019-07-23
[FIX] [no task] web_listview_custom_column Dont crash on non existstant `type_id`

daramousk:
cc: @hbrunn

### DIFF
--- a/web_listview_custom_column/migrations/8.0.1.0.1/post-migrate.py
+++ b/web_listview_custom_column/migrations/8.0.1.0.1/post-migrate.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import re
+from openerp import api, SUPERUSER_ID
+
+
+def migrate(cr, version):
+    """
+    Find all the custom views which are supposed to apply to everyone and have
+    a priority of 10000 and change it to 10001.
+    """
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    model_data = env['ir.model.data']
+    model_view = env['ir.ui.view']
+    data = model_data.search([
+        ('module', '=', 'web_listview_custom_column'),
+        ('model', '=', 'ir.ui.view'),
+        ])
+    # find the views applying to everyone
+    views = model_view.browse(
+        data.filtered(
+            lambda rec: rec.name.endswith('_all')).mapped(
+                'res_id'),
+    )
+    views.filtered(lambda rec: rec.priority == 10000).write({
+        'priority': 10001,
+    })
+    # find the views applying to individuals
+    views = model_view.browse(
+        data.filtered(
+            lambda rec: re.search(
+                '^custom_view_.*_user_.*$', rec.name)).mapped(
+                    'res_id'),
+    )
+    views.filtered(lambda rec: rec.priority == 10001).write({
+        'priority': 10000,
+    })
--- a/web_listview_custom_column/models/ir_ui_view.py
+++ b/web_listview_custom_column/models/ir_ui_view.py
@@ -2,7 +2,7 @@
 # © 2017 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from lxml import etree
-from openerp import _, api, models, tools
+from openerp import _, api, models, tools, exceptions
 
 
 class IrUiView(models.Model):
@@ -39,6 +39,21 @@ class IrUiView(models.Model):
                     element.getnext().addnext(element)
                     break
         elif diff['operation'] == 'reset':
+            # if empty, then search if a global view has been set on this
+            # and prompt user who tries to delete to talk to user who
+            # created those custom views.
+            if not customized_view:
+                views = self.search([
+                    ('model', '=', self.model),
+                    ('mode', '=', 'extension'),
+                    ('user_ids', '=', False),
+                    ('create_uid', '!=', self.env.user.id),
+                    ])
+                raise exceptions.ValidationError(_(
+                    'Users %s have created views that apply to everyone. \n '
+                    'If you want to change this view you need to notify \n'
+                    'them to reset the views they have set by clicking this \n'
+                    'button') % ', '.join(views.mapped('create_uid.name')))
             customized_view.unlink()
             return []
         elif diff['operation'] == 'to_user':
--- a/web_listview_custom_column/models/ir_ui_view.py
+++ b/web_listview_custom_column/models/ir_ui_view.py
@@ -102,7 +102,7 @@ class IrUiView(models.Model):
             'arch': arch,
             'inherit_id': self.id,
             'mode': 'extension',
-            'priority': 10000 + (diff['type'] == 'user' and 1 or 0),
+            'priority': 10000 + (diff['type'] == 'all' and 1 or 0),
             'user_ids': [(4, self.env.uid)] if diff['type'] == 'user' else [],
         })[0]
         result = self.create(values)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,66 @@
 language: python
-
-python:
-  - "2.7"
-  
 sudo: false
 cache: pip
 
+python:
+  # For branches <= 10.0, put `- "2.7.13"`
+  - "2.7.13"
+
 addons:
+  # By default postgresql-9.1 is installed but there is issue related:
+  #  https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935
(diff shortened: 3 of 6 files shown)


hbrunn:
please explain which symptom you're fixing, and motivate the huge diff. Also follow up on https://github.com/OCA/server-tools/pull/1039#issuecomment-513689527 for this addon to make it actually functional with the main branches

daramousk:
> please explain which symptom you're fixing, and motivate the huge diff

Basically this module on every function was expecting that `self.ensure_one()`, but that is not always the case since the `view_id` might not be passed https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/view_list.js#L49

Also, if for some reason userA sets a view for everybody then people are stuck using this view.
Because any change from the front end translates to a change by default as a user. Then nothing changes when they click the buttons.

And in any case userB should be able to set a personal view which overrides the `to_all` one.

https://github.com/OCA/web/blob/8.0/web_listview_custom_column/models/ir_ui_view.py#L75
https://github.com/OCA/web/blob/8.0/web_listview_custom_column/models/ir_ui_view.py#L105

hbrunn:
when there is no view_id, it's a corner case we don't care about. If we needed it, a better implementation would have been to prepend the call to `custom_column` on the js side with a call that creates a default view for the model, and then leave the rest untouched.

So the only relevant change is the wrong priority, this is a valuable correction of the code. The rest not so much. In general, try to keep your PRs' diffs minimal, if it wasn't my job to review this, I wouldn't.

We'll also need a migration script that changes the priorities of existing user/all custom views. according to the new scheme.

Further, if you address different things in the same PR, split it in different commits.

And when you change code formatting or similar, you need to be able to point to a PEP or OCA convention why it should be different, otherwise it's personal taste and we can't keep changing code back and forth because different people have different tastes.

daramousk:
> when there is no view_id, it's a corner case we don't care about.

I went on and researched when that happens. Now, this happens when 1) the `action` passed into the `ViewManagerAction.init()` either does not contain views at all or 2) the action does not contain any views (`_views_get_fnt()` returns {}`)

1) Now an action can very well not specify ids for the views it requires.
2) `_views_get_fnc() returns an empty value if `view_ids` is empty

I was under the impression that the `view_id` was empty only when there was no view in the system and the `fields_view_get` has to generate one on the fly, but that does not seem to be the (only) case.

So unless I am missing something in this exchange between the web client and the act_window model, this module will fail for every action that does not explicitly specify a `view_id`, for a concrete example you can go ahead and crash this module by trying to edit the `res.groups` listview vs `res.users` list view which will work.

https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/res/res_users_view.xml#L115
https://github.com/OCA/OCB/blob/10.0/odoo/addons/base/res/res_users_view.xml#L262


> If we needed it, a better implementation would have been to prepend the call to custom_column on the js side with a call that creates a default view for the model, and then leave the rest untouched.

Ideally, there should be only one call to the backend/custom_column, that would be to save the view (either user or all). but in our case we have this custom_column which conditionally creates a customized view. Now, what I did is manipulate this creation of that customized view to conditionally create a primary view. Everything is happening in the same place. How would putting this functionality in js would make it better?

> So the only relevant change is the wrong priority, this is a valuable correction of the code. The rest not so much. In general, try to keep your PRs' diffs minimal, if it wasn't my job to review this, I wouldn't.

With that being said, since I made the effort of doing some changes here I believe it would be of great service to just go ahead and add some docstrings. And raising coverage should outweigh the awfulness of some extra diff)

> We'll also need a migration script that changes the priorities of existing user/all custom views. according to the new scheme.

That is the plan, also since the naming scheme has changed I need to take care of this as well, but this will be done at a later stage if I get an ok for the rest of the stuff)

> Further, if you address different things in the same PR, split it in different commits.

When it comes to implementing new requirements that is the case, but this commit is basically a bugfixing one. I guess I could split this into two commits one being the priority and the other the rest.

> And when you change code formatting or similar, you need to be able to point to a PEP or OCA convention why it should be different, otherwise it's personal taste and we can't keep changing code back and forth because different people have different tastes.

I assume you speak of https://github.com/OCA/web/pull/1338/files#diff-fbe0fca8d8b2fe63c237912bd737bafdR58
While this is not in PEP8, from what I understand it exists at https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#33idioms
in the sentence which starts with "If a function call, dictionary..."

hbrunn:
this involves much more discussion than I consider okay.

The only way there's no view_id is if there's not view as in no view in the database. Or there's a model that generates its view on the fly, and both cases are not interesting for this module. If you can produce this differently, I'd like to see the code. Another sensible way to handle this case is to show an error message if there's no view_id, or not show the controls in the first place.

The answer to all other comments is no, please do as asked.

daramousk:
@hbrunn  Ok, so three things:

1) Changed the priority and provided a migration script
2) modified travis
3) In the case someone has set a global view and users cannot do anything, if they try to reset then they will get a descriptive message on what to do.

web_listview_custom_column/migrations/8.0.1.0.1/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+    model_data = env['ir.model.data']
+    model_view = env['ir.ui.view']
+    data = model_data.search([
+        ('module', '=', 'web_listview_custom_column'),
+        ('model', '=', 'ir.ui.view'),
+        ])
+    # find the views applying to everyone
+    views = model_view.browse(
+        data.filtered(
+            lambda rec: rec.name.endswith('_all')).mapped(
```
hbrunn, line 23: for the amount of data we're dealing here not relevant, but let postgres filter stuff if you can. You could have added `('name', '=like', '%_all')` to the search proposition above

web_listview_custom_column/migrations/8.0.1.0.1/post-migrate.py
```diff
@@ -0,0 +1,38 @@
...
+                'res_id'),
+    )
+    views.filtered(lambda rec: rec.priority == 10000).write({
+        'priority': 10001,
+    })
+    # find the views applying to individuals
+    views = model_view.browse(
+        data.filtered(
+            lambda rec: re.search(
+                '^custom_view_.*_user_.*$', rec.name)).mapped(
```
hbrunn, line 33: this would be `('name', '=like', 'custom_view_%_user_%')`

web_listview_custom_column/models/ir_ui_view.py
```diff
@@ -39,6 +39,21 @@ def custom_column(self, diff):
...
         elif diff['operation'] == 'reset':
+            # if empty, then search if a global view has been set on this
+            # and prompt user who tries to delete to talk to user who
+            # created those custom views.
+            if not customized_view:
+                views = self.search([
+                    ('model', '=', self.model),
+                    ('mode', '=', 'extension'),
+                    ('user_ids', '=', False),
+                    ('create_uid', '!=', self.env.user.id,),
```
hbrunn: no trailing commas in tuples with len>1: https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

web_listview_custom_column/models/ir_ui_view.py
```diff
@@ -39,6 +39,21 @@ def custom_column(self, diff):
...
+                    ])
+                raise exceptions.ValidationError(_(
+                    'Users %s have created views that apply to everyone. \n '
+                    'If you want to change this view you need to notify \n'
+                    'them to reset the views they have set by clicking this \n'
+                    'button') % (views.mapped('create_uid.name', )))
```
hbrunn: you don't need the brackets around `views....`, and also no comma there

web_listview_custom_column/models/ir_ui_view.py
```diff
@@ -102,7 +117,7 @@ def _custom_column_create_view(self, diff, arch):
             'arch': arch,
             'inherit_id': self.id,
             'mode': 'extension',
-            'priority': 10000 + (diff['type'] == 'user' and 1 or 0),
+            'priority': 10000 + (diff['type'] == 'all' and 1 or 0),
```
hbrunn, line 120: oh, so this has been wrong all along. Interesting nobody complained earlier

hbrunn (dismissed):
nice, thanks. Only a few nitpicks, and the first two is only advise for the future, no need to change now

web_listview_custom_column/models/ir_ui_view.py
hbrunn: wait, this will show `[user1, user2]` when there are multiple users (which probably never happens, but anyways). Better wrap in `', '.join()`

gfcapalbo:
@daramousk  :+1:

daramousk:
@gfcapalbo Please use the Approve mechanism so that the ocabot picks it up.

PR #1340, opened 2019-07-25
[12.0][FIX] Status Button invisible

keshrath:
Before:
image
After:
image

The problem is that these CSS attributes also affect other elements in the title.

https://github.com/OCA/web/commit/5aa9278b58a01e1119ae566c2340062f46a37c3f#diff-e053f829c6178b791df988b24e4282b4 >> https://github.com/OCA/web/blob/12.0/web_responsive/static/src/css/web_responsive.scss#L346

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -343,7 +343,7 @@ html .o_web_client .o_main .o_main_content {
         .oe_title {
             width: initial;
 
-            span {
+            span.o_field_widget {
                 max-width: 100%;
                 text-overflow: ellipsis;
                 white-space: nowrap;


pedrobaeza:
The screenshot if the before or the after? Put both, please

keshrath:
@pedrobaeza I've updated the text.

pedrobaeza:
Uhm, first weird thing is the position of the status... In v11 is on upper left part, not there. I let @Yajo to check this.

daramousk (approved):

yajo:
This is how a raw v12 looks like:

Captura de pantalla de 2019-08-12 10-31-46

I guess the title is longer in web_responsive and that's why the circle appears there.

OK for me.

yajo (approved):

pedrobaeza:
@Yajo if that's the original one, then the problem is another: you are using one extra line for the title and smart-button, and thus, this is a side effect. Why that extra line?

yajo:
I see no extra line... According to the screenshot from https://github.com/OCA/web/pull/1340#issue-301014611, it's all in the same line:

Captura de pantalla de 2019-08-12 12-44-25

pedrobaeza:
On the screenshot of raw v12 everything is in the same line:

62853374-70a9f100-bcec-11e9-9ec0-619416ef57fc

pedrobaeza:
And I see also a different h1 font/size.

yajo:
I don't see those as blockers for the current patch, which is actually a bugfix.

pedrobaeza:
OK, I'm not blocking, but not sure if it's related to the other problem. And you must solve that problem as well (on another PR if not related). If you tell me that it's not related, I'll merge this immediately.

yajo:
Yes, this one doesn't seem related.

pedrobaeza:
OK, please attend the other bug.

/ocabot merge minor

PR #1343, opened 2019-07-31
[11.0][FIX] web_action_conditionable: Adapt to new api

Tardo:
Current implementation of "web_action_conditionable" doesn't work because 'web.View' not exists.
This PR refactor all to use new api.

Affected Versions: 11.0 & 12.0


rafaelbn:
Hi @Yajo I think you could please review this PR, thanks!

web_action_conditionable/static/src/js/field_one2many.js
```diff
@@ -0,0 +1,38 @@
+/* global py */
+/* Copyright 2019 Alexandre Díaz
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
```
nikul-serpentcs: @Tardo LGPL -> AGPL
As Per README file

Tardo: Change done, thanks.

web_action_conditionable/static/src/js/field_one2many.js
```diff
@@ -0,0 +1,38 @@
...
+    "use strict";
+
+    var FieldOne2Many = require('web.relational_fields').FieldOne2Many;
+
+    FieldOne2Many.include({
+        init: function () {
+            var self = this;
+            try {
+                return this._super.apply(this, arguments);
+            } catch (error) {
```
yajo, line 14: I don't understand this structure. Are you only adding your stuff when upstream code fails? Why not adapt those attributes to booleans and avoid `_super()` to fail? :thinking:

pedrobaeza:
Please rebase and answer the question, but if it gets complicated, we don't need this at all, so we should close it.

web_action_conditionable/static/src/js/field_one2many.js
Tardo: This follows the original implementation. This allows to have the "normal" behavior and the "plus" added by this module.
I don't know how can be converted to boolean before initialize fields...

Tardo:
@pedrobaeza done

pedrobaeza:
OK, merging then /ocabot merge minor

PR #1345, opened 2019-08-05
[10.0] web_ckeditor4: Fix JS error in SCAYT plugin

p-tombez:
When in edit mode on a field with CKEditor configured, after enabling the SCAYT plugin, if you try to change the language, the following JS error is triggered : 
`TypeError: g.getLangList is not a function`

This intends to fix this issue without updating the whole CKEditor lib


simahawk:
@p-tombez is hard to understand the change w/ this diff.
The change seems to be only `s/getLangList/getSpellingLangList`. Is it correct?
Also, the file you edited I guess it comes from the bundle of ckeditor. Maybe is better to mention this change in the header.

asaunier:
> The change seems to be only s/getLangList/getSpellingLangList. Is it correct?

It seems so. I have seen no other changes than:
Screenshot from 2019-08-06 11-05-17

p-tombez:
You are correct : the only change is `s/getLangList/getSpellingLangList`
This is a quick and dirty fix to avoid updating the whole CKEditor lib (this version is more than 5 years old). I know the right way should have been to update it, but I don't understand JS well enough to manage it

simahawk (approved):

daramousk (commented):
Hi there thanks for this, I might be missing something on the plugin management, but I cannot find the definition of `getSpellingLangList`, where is this?

simahawk:
@p-tombez ?
@daramousk does this clash somehow w/ your work here? https://github.com/OCA/web/pull/1353
If not, can we merge this?

PR #1346, opened 2019-08-06
[ADD] web_widget_many2one_kanban

KKamaa:


web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
...
+            var field_name = this.name;
+            var params = [[this.res_id],{
+                [field_name]: parseInt(val)
+            }];
+            this._rpc({
+                model: this.model,
+                method: 'write',
+                args: params
+            }).then(function(result){});
+            },
```
thomaspaulb: This `.then()` is superfluous, it will just call a dummy function, so the whole `.then(...)` can be removed

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
...
+                fields: ['id', 'name'],
+                order: 'id'
+            }).then(function (data) {
+                // This is the first item empty for times when there is no data.
+                self.$el.find('.selectM2MField').append('<option' +
+                    ' value=""></option>');
+                for (var val of data) {
+                    self.$el.find('.selectM2MField').append(
+                        '<option value=' + val.id + '>' + val.name +
+                        '</option>');
```
thomaspaulb: Why not change to:

```
'<option value=' + val.id + ((val.id == curr_id) ? 'selected' : '') +  '>'
```

instead of looking up the element again with jQuery like you do below

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
...
+                method: 'search_read',
+                fields: ['id', 'name'],
+                order: 'id'
+            }).then(function (data) {
+                // This is the first item empty for times when there is no data.
+                self.$el.find('.selectM2MField').append('<option' +
+                    ' value=""></option>');
+                for (var val of data) {
+                    self.$el.find('.selectM2MField').append(
+                        '<option value=' + val.id + '>' + val.name +
```
thomaspaulb: I still think this should be officially with quotes:

```
'<option value="' + val.id + '">'

File: web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+            }];
+            this._rpc({
+                model: this.model,
+                method: 'write',
+                args: params
+            }).then(function(result){});
+            },
+        _updateField: function(curr_id) {
+            var self = this;
+            var model = this.field.relation
```
thomaspaulb: This line misses a `;`

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
+odoo.define('web.kanban_editable', function(require){
```
thomaspaulb: There should be a copyright line in this file such as [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/static/src/js/web_module_name.js#L1)

thomaspaulb (changes requested):

web_widget_kanban_many2one/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Web Widget Kanban Many2one',
+    'category': 'Extra Tools',
+    'author': 'Sunflower IT, Odoo Community Association (OCA)',
+    'website': 'www.sunflowerweb.nl',
```
Tardo:
```suggestion
    'website': 'https://github.com/OCA/web',
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
...
+
+    var many2one_kanban = AbstractField.extend({
+        template: _t('Kanban_editable_temp'),
+        events: {
+            'click .selectM2MField': '_onSelectField',
+        },
+
+        init: function(){
+            this._super.apply(this, arguments);
+            if(!this.recordData[this.name])
```
Tardo: This need be done on "_render"

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
...
+                this._updateField(this.recordData[this.name].id);
+            else
+                this._updateField(this.recordData[this.name].data.id);
+        },
+        _onSelectField: function(){
```
Tardo: Can't use '_setValue'?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
...
+                        self.$el.find(".selectM2MField " + "option[value='"+
+                            val.id+"']").prop('selected', true);
+                    }
+                }
+            });
+        },
+    });
+    field_registry.add('many2one_kanban', many2one_kanban);
+    return many2one_kanban;
+});
```
Tardo: Add empty line

web_widget_kanban_many2one/static/src/xml/web_widget_kanban_many2one.xml
```diff
@@ -0,0 +1,9 @@
+<template>
+    <t t-name="Kanban_editable_temp">
+        <span class="container">
+            <select class="form-control text-success selectM2MField" name="id">
+
+            </select>
+        </span>
+    </t>
+</template>
```
Tardo: Add empty line

web_widget_kanban_many2one/views/web_widget_kanban_many2one.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="assets_backend" name="web_widget_kanban_many2one"
+              inherit_id="web.assets_backend">
+        <xpath expr="." position="inside">
+            <script
+                    src="/web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js"></script>
+        </xpath>
+    </template>
+</odoo>
```
Tardo: Add empty line

Tardo (changes requested):
Thanks for the work, please revise if can use "Odoo API flow"...

thomaspaulb:
@Tardo reviews are clear except one: what do you refer to when you mention "Odoo API flow" ?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,62 @@
+odoo.define('web.kanban_editable', function(require){
+
+    var AbstractField = require('web.AbstractField');
+    var core = require('web.core');
+    var field_registry = require('web.field_registry');
```
thomaspaulb: Should we add a line like this:

https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/fields/basic_fields.js#L1785

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,64 @@
...
+            'click .selectM2MField': function(event){
+                event.stopPropagation();
+            },
+            'change .selectM2MField': '_onSelectField'
+        }),
+        _render: function () {
+            var el = this.$el;
+            var select = el.find('select');
+            var div = el.closest('.o_kanban_record');
+            var _text = div.find('h2 span').text();
```
thomaspaulb: These two are not needed anymore.

web_widget_kanban_many2one/static/src/xml/web_widget_kanban_many2one.xml
```diff
@@ -0,0 +1,9 @@
+<template>
+    <t t-name="Kanban_editable_temp">
+        <span class="container">
+            <select class="form-control text-success selectM2MField" name="id">
```
thomaspaulb: This "form-control" class induces the CSS issues that we talked about - either add some CSS to counter it, or lose the class here

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,64 @@
...
+            select.append('<option value=""></option>');
+            // options
+            for (var val of this._options) {
+                selected = (val.id == curr_id) ? 'selected' : '';
+                select.append('<option value="' + val.id + '" ' + selected +
+                    '>' + val.name + '</option>');
+            }
+        },
+        _onSelectField: function(){
+            var self = this;
```
thomaspaulb: We only need "self" (or sometimes people call it "that") if we wish to use "this" in a callback function also, for example after '_rpc' returns with success. Otherwise you can just use "this".

thomaspaulb (changes requested):

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -20,29 +20,29 @@ odoo.define('web_widget_kanban_many2one.kanban_editable', function(require){
...
             var self = this;
-            var val = self.$el.find('select[name="id"]').val();
-            val = self._parseValue(val).id;
-            self._setValue(val);
+            var val = this.$el.find('select[name="id"]').val();
+            val = this._parseValue(val).id;
+            this._setValue(val);
         },
         willStart: function() {
             var self = this;
```
thomaspaulb: Why do you still assign a value to 'self' if it is not used?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -20,29 +20,29 @@ odoo.define('web_widget_kanban_many2one.kanban_editable', function(require){
...
             // options
-            for (var val of self._options) {
+            for (var val of this._options) {
                 selected = (val.id == curr_id) ? 'selected' : '';
                 select.append('<option value="' + val.id + '" ' + selected +
                     '>' + val.name + '</option>');
             }
         },
         _onSelectField: function(){
             var self = this;
```
thomaspaulb: and here

thomaspaulb (changes requested):

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
thomaspaulb: @KKamaa is this done

KKamaa: what exactly was to be added in this line of code?

web_widget_kanban_many2one/static/src/css/web_widget_kanban_many2one.css
```diff
@@ -0,0 +1,5 @@
+select.selectM2MField:focus {
+    border: inherit !important;
+}
+
+
```
Tardo: :fire: this line

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
+/* Copyright 2019 Sunflower IT <www.sunflowerweb.nl>
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+ */
+
+odoo.define('web_widget_kanban_many2one.kanban_editable', function(require){
+
```
Tardo:
```suggestion
    "use strict";
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+    var AbstractField = require('web.AbstractField');
+    var core = require('web.core');
+    var field_registry = require('web.field_registry');
+
+    var _t = core._t;
+
+    var many2one_kanban = AbstractField.extend({
+        template: _t('Kanban_editable_temp'),
```
Tardo: _t it's related with translations, not with templates.

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+
+odoo.define('web_widget_kanban_many2one.kanban_editable', function(require){
+
+    var AbstractField = require('web.AbstractField');
+    var core = require('web.core');
+    var field_registry = require('web.field_registry');
+
+    var _t = core._t;
+
+    var many2one_kanban = AbstractField.extend({
```
Tardo: I think that better use "KanbanFieldMany2One" instead of "AbstractField" and use an option like "editable='true'"

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+        template: _t('Kanban_editable_temp'),
+        events: _.extend({}, AbstractField.prototype.events, {
+            'click .selectM2MField': function(event){
```
Tardo: Ugly name for the selector, don't use UpperCamelCase. And it's an Many2One field not Many2Many.
Please, don't use anonymous functions like this if not are strictly necessary, better implement it in the class to easy inheritance.

Please, follow the OCA javascript style rules, see https://raw.githubusercontent.com/OCA/pylint-odoo/master/pylint_odoo/examples/.jslintrc

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+                event.stopPropagation();
+            },
+            'change .selectM2MField': '_onSelectField'
+        }),
+        _render: function () {
+            var el = this.$el;
+            var select = el.find('.selectM2MField');
+            var field_name = this.name;
+            var field_data = this.recordData[field_name];
+            var curr_id = field_data ? field_data.data.id : false;
```
Tardo: Better use "short-circuit" evaluation
```suggestion
            var curr_id = field_data && field_data.data.id;
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+    var many2one_kanban = AbstractField.extend({
+        template: _t('Kanban_editable_temp'),
+        events: _.extend({}, AbstractField.prototype.events, {
+            'click .selectM2MField': function(event){
+                event.stopPropagation();
+            },
+            'change .selectM2MField': '_onSelectField'
+        }),
+        _render: function () {
+            var el = this.$el;
```
Tardo: really necessary? and the name it's not very intuitive... '$' means "jQuery Object"

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+            var select = el.find('.selectM2MField');
```
Tardo: widgets have '$' function that call to "this.$el.find("...
```suggestion
            var select = this.$('.selectM2MField');
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+            var field_name = this.name;
```
Tardo: really necessary?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+            var field_data = this.recordData[field_name];
+            var curr_id = field_data ? field_data.data.id : false;
+            // empty item
+            select.append('<option value=""></option>');
+            // options
+            for (var val of this._options) {
```
Tardo: Please, don't use ES6 "of"

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+                selected = (val.id == curr_id) ? 'selected' : '';
+                select.append('<option value="' + val.id + '" ' + selected +
```
Tardo: Better use underscore.string: _.str.sprintf()

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+                    '>' + val.name + '</option>');
+            }
+        },
```
Tardo: no "super" call needed?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+        _onSelectField: function(){
+            var val = this.$el.find('select[name="id"]').val();
```
Tardo:
```suggestion
            var val = this.$('select[name="id"]').val();
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+            val = this._parseValue(val).id;
+            this._setValue(val);
+        },
+        willStart: function() {
+            var self = this;
+            var model = this.field.relation;
+            _promise = self._rpc({
```
Tardo:
```suggestion
            return this._rpc({
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+                model: model,
+                method: 'search_read',
+                fields: ['id', 'name'],
+                order: 'id'
+            });
```
Tardo:
```suggestion
            })
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+            _promise.then(function (data) {
```
Tardo:
```suggestion
            .then(function (data) {
```

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
```diff
@@ -0,0 +1,63 @@
...
+                self._options = data;
+            });
+            return _promise;
```
Tardo:
```suggestion
```

web_widget_kanban_many2one/static/src/xml/web_widget_kanban_many2one.xml
```diff
@@ -0,0 +1,9 @@
+<template>
+    <t t-name="Kanban_editable_temp">
```
Tardo: Ummm... can't use a better name? and please, use UpperCamelCase.

web_widget_kanban_many2one/static/src/xml/web_widget_kanban_many2one.xml
Tardo: name="id"?

Tardo (changes requested):
Please, add a description to this PR and a screenshot to know how it should look.

Tardo:
@thomaspaulb i mean do the stuff in the correct place using the available "tools".

web_widget_kanban_many2one/static/src/xml/web_widget_kanban_many2one.xml
KKamaa: It a select like in forms so we need to define an id so that when a person selects a value it is captured as id=1,2,3,4 to be used by the widget.

web_widget_kanban_many2one/static/src/css/web_widget_kanban_many2one.css
KKamaa: @Tardo what do you mean by this...?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
KKamaa: "editable=true" in xml?

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
thomaspaulb: supportedFieldTypes: ['many2one'],

web_widget_kanban_many2one/static/src/css/web_widget_kanban_many2one.css
thomaspaulb: He means to remove these blank lines

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
thomaspaulb: @Tardo I disagree; this widget is meant to add editable functionality also in non-editable mode, just like for example the priority stars widget can do when you are in read-only kanban mode and click one of the stars. This is one of our customer's requirements, for faster many2one editing.

@Kkamaa What I think he meant is that we implement this widget as the editable version of the existing many2one widget in the Odoo source, using the "editable" boolean in the rendering function; this is not in XML but just in javascript. You can check some other widgets that use this functionality such as this one: https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/fields/relational_fields.js#L1487

pedrobaeza:
The module name should be `web_widget_many2one_kanban` or `web_kanban_widget_many2one`, but not a to intercalate both parts.

web_widget_kanban_many2one/static/src/js/web_widget_kanban_many2one.js
KKamaa: @thomaspaulb about this?

thomaspaulb: @Kkamaa I dont agree with Tardo, so unless he replies with a convincing argument, we dont work on this.

web_widget_many2one_kanban/static/src/js/web_widget_many2one_kanban.js
```diff
@@ -0,0 +1,58 @@
...
+            var model = this.field.relation;
+            return self._rpc({
+                model: model,
+                method: 'search_read',
+                fields: ['id', 'name'],
+                order: 'id',
+            }).then(function (data) {
+                self._options = data;
+            });
+            },
```
thomaspaulb: We have a performance problem here: if 140 items are on the kanban board, this function will be executed 140 times, or 280 times if two such widgets are on one kanban. The loading drag on opening the kanban view is noticable. Possible solutions could involve:

1. On opening the kanban view, look for these kind of widgets in the individual kanbans, and if there are, load their options once and store them on the kanban view object. Then when displaying the kanbans, take the options from there; so we load just 2 times and not 280 times.
2. Study the Odoo source code and see if those options are not already loaded somewhere else and we can already pick them from the Javascript
3. On loading this widget first time, cache the field results somewhere on the main kanban view, and on subsequent widget loads, look at the cache first before querying RPC.

web_widget_many2one_kanban/static/src/js/web_widget_many2one_kanban.js
```diff
@@ -0,0 +1,58 @@
...
+    var field_registry = require('web.field_registry');
+
+    var KanbanFieldMany2One = AbstractField.extend({
+        template: 'KanbanEditable',
+        supportedFieldTypes: ['many2one'],
+        events: _.extend({}, AbstractField.prototype.events, {
+            'click .oe_kanban_select': '_eventAction',
+            'change .oe_kanban_select': '_onSelectField',
+        }),
+        _render: function () {
```
daramousk: Keep in mind that `_render` gets called when you select an item as well so all the items will be appended again as well and you will end up with duplicate values.

daramousk (changes requested):
Hello there and thanks for this!

I have some questions on this:

1) If this new widget is supposed to work only on `FieldMany2One` fields, then why don't you extend `FieldMany2One` widget? You can avoid loading the data yourself and you probably wont need to write any js at all

2) I am not sure I understand what this module is supposed to do, but it would perhaps be enough for you to just create a custom xml template and load it? Since changing the appearance of the loaded records seems to be the major cause for this.

Thanks!

thomaspaulb:
@Kiplangatdan What happened here... you deleted the commits and then the bot merged 0 commits?

PR #1347, opened 2019-08-07
[12.0][IMP] Allow to specify a field on timeline view to define the element color.

apineux:
Feature splited from here: https://github.com/OCA/web/pull/1280


web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -421,11 +421,16 @@ odoo.define('web_timeline.TimelineRenderer', function (require) {
...
-                    self.color = color.color;
-                }
-            });
+            var color = null;
+            if (self.arch.attrs.color_field !== undefined) {
+                color = evt[self.arch.attrs.color_field];
+            } else {
+                _.each(self.colors, function (col) {
+                    if (eval("'" + evt[col.field] + "' " + col.opt + " '" + col.value + "'")) {
+                        color = col.color;
```
pedrobaeza, line 430: break in this case for not continuing evaluating, isn't it?

apineux: @pedrobaeza 
I agree that we could add a break there to avoid evaluate the other colors. 
But,
- There is no 'break' for the _.each
- This is a part of the existing code (I apply the original code when no color_field is set)
- This loop is done on the timeline options specified on the view, so there will be only a few color to check. The time consumed will be insignificant

I'm not sure that it is necessary to change that knowing the little change that it will bring.
Let me know what do you think.

pedrobaeza: Yeah, indeed this is improving existing code, but if no possible break here, let's keep it as is.

web_timeline/README.rst
```diff
@@ -66,6 +66,8 @@ the possible attributes for the tag:
...
 | colors             | No        | Allows to set certain specific colors if the expressed condition (JS syntax) is met.                                                                                                                                                                                      |
 +--------------------+-----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| color-field        | No        | Allows to set a field (on the current model) that contains the HTML color (e.g. #FFFFFF). This option has priority over 'colors'.                                                                                                                                         |
```
pedrobaeza, line 69: This should be put on the corresponding file inside `readme` folder. If not, it will be lost.

pedrobaeza (changes requested):
Squashing commits on merge, so ocabot can't be used unless you squash now in the PR.

pedrobaeza (approved):
Squashing commits on merge, so ocabot can't be used unless you squash now in the PR.

apineux:
Squash done :+1:

pedrobaeza:
Great, thanks!

web_timeline/readme/CONFIGURE.rst
```diff
@@ -26,6 +26,8 @@ the possible attributes for the tag:
...
 | colors             | No        | Allows to set certain specific colors if the expressed condition (JS syntax) is met.                                                                                                                                                                                      |
 +--------------------+-----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| color-field        | No        | Allows to set a field (on the current model) that contains the HTML color (e.g. #FFFFFF). This option has priority over 'colors'.                                                                                                                                         |
```
daramousk, line 29: Perhaps changing this to `color_field` would be more appropriate since this is how it's used

daramousk (approved):
Thanks for this, it seems to be doing what is supposed to be doing, a minor comment only

alexis-via (approved):

alexis-via:
BTW @apineux are you aware of this bug on colors : https://github.com/OCA/web/issues/1446
I made an code analysis on the bug; it's a problem for M2O fields.

marcelsavegnago (approved):

PR #1349, opened 2019-08-08
[IMP] web_responsive: Performance when hide menus

Tardo:
No more delays when open views with a lot of DOM nodes

Cc @tecnativa

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -115,7 +115,7 @@ odoo.define('web_responsive', function (require) {
 
         /**
          * Prevent the menu from being opened twice
-         * 
+         *
          * @override
          */
         _onAppsMenuItemClicked: function (ev) {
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Tried on customer DB with lots of session tracks and working with performance.

yajo (changes requested):
Add a little comment explaining that this is done for performance, please; to avoid future readers the inertia to revert this patch.

Tardo:
@Yajo comment added

yajo (approved):

HviorForgeFlow:
Can anyone clarify this improvement. I was noticing that there were a perform issue on web_responsive from v11 to v12.

In v11 when you click an App, menu is charged before display records of the default app view.
In v12 you must wait until default view is charged to see menus of that app.

Customers noticed that telling us that the performance in v12 seemed to be slower than in v11.

Is this PR solving that? Thanks,

pedrobaeza:
Can it be... Try it and tell us. If not, please specify through a video or similar what is the specific issue.

Tardo:
@hveficent Indeed the user experience has changed. The feeling in v11 is that when clicking on menu "something" happens ... in v12 there is a "load time" after clicking, this can be perceived by the user as a decrease in performance.
Actually the "effective" load time should be the same in both versions.

PR #754, opened 2017-10-13
[MIG] [11.0] web_widget_float_formula

hgminh95:
Hi @pedrobaeza, I have finished the migration with this module. Please review and tell me if you see anything wrong.

PR #1350, opened 2019-08-08
[MIG][12.0] web_widget_float_formula (from 11.0)

legalsylvain:
Hi all,

trying to migrate web_widget_float_formula from 11.0 ([PR](https://github.com/OCA/web/pull/754/)) to 12.0.

I have currently a problem with the events that are not raised.
I wrote the following lines, as it seems to be the new way to add events. (and because old code did'nt worked ;-).

```
        events: _.extend({
            "blur": "_compute_result",
            "focus": "_display_formula",
        }, FieldFloat.prototype.events),
```

**Result**

* The js file is well loaded;
* the events prototype is well overloaded. in my log : 

```
start
Object { blur: "_compute_result", focus: "_display_formula", keydown: "_onKeydown", input: "_onInput", change: "_onChange" }
```

but the events are not raised. (In fact, ``_onKeydown`` (from ``AbstractField``) and ``_onInput`` and ``_onChange`` (from ``InputFields`` are correctly raised).

Any idea ? 

thanks in advance.

CC : @lasley, @Yajo, @hbrunn.

#GRAPOCA


Tardo:
blur and focus events are triggered successfully...

Peek 09-08-2019 03-09
Tested on runbot.

legalsylvain:
Thanks @tardo for your investigation.

In fact, the problem is with Monetary field, not Float field. I guess that there was a change in the dependencies. monetary field does'nt inherit from float field. (So in V8 the widget was working for all fields).

legalsylvain:
Hi. finally working. just don't understand why I have to write two lines, after having computed the result. But if not, it does'nt work.

```
                    this._setValue(newText);
                    this.$input.val(newText);
```

Note : if you see improvment, feel free to improve the PR, making patch.

kind regards.

web_widget_float_formula/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Web Widget - Formulas in Float Fields',
+    'summary': 'Allow use of simple formulas in float fields',
+    'version': '12.0.1.0.0',
+    'category': 'Web',
+    'author': 'GRAP, LasLabs, Odoo Community Association (OCA)',
+    'website': 'http://www.grap.coop',
```
Tardo:
```suggestion
    'website': 'https://github.com/OCA/web',
```

web_widget_float_formula/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+A short demo video can be found at http://www.youtube.com/watch?v=jQGdD34WYrA.
```
Tardo, line 1: Can you add a gif?

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+            this._formula_text = '';
+        },
+
+        _process_formula: function(formula) {
+            try{
+                formula = formula.toString();
+            } catch (ex) {
+                return false;
+            }
+            var clean_formula = formula.toString().replace(/^\s+|\s+$/g, '');
```
Tardo, line 37:
```suggestion
            var clean_formula = formula.toString().trim();
```

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+            if (clean_formula[0] == '=') {
```
Tardo:
```suggestion
            if (clean_formula[0] === '=') {
```

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+                    var newText = field_utils.format.float(value);
+                    this._setValue(newText);
+                    this.$input.val(newText);
+                }
+            }
+        },
+
+        // Display the formula stored in the field to allow modification
+        _display_formula: function() {
+            if (this._formula_text !== '') {
```
Tardo:
```suggestion
            if (!_.isEmpty(this._formula_text)) {
```

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+            var thousands_sep = translation_params.thousands_sep;
+
+            var value;
+            formula = formula.replace(thousands_sep, '').replace(decimal_point, '.');
+            try {
+                value = eval(formula);
+            }
+            catch(e) {}
+
+            if (typeof value != 'undefined') {
```
Tardo, line 61:
```suggestion
            if (typeof value !== 'undefined') {
```

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+            this._clean_formula_text();
+
+            // Convert Text to valid formula
+            var text = this._getValue();
+            var formula = this._process_formula(text);
+
+            if (formula !== false) {
+                var value = this._eval_formula(formula);
+                if (value !== false) {
+                    this._formula_text = "=" + formula;
```
Tardo, line 77: What do you think of making the character '=' an attribute?

Something like ```CMD_CHAR: '='```

Tardo (changes requested):
Thanks for the migration work

gdgellatly:
@legalsylvain maybe check odoo master. They added this functionality to it. Maybe not merged yet.

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+        },
+
+        _eval_formula: function(formula) {
+            // Import localization values used to eval formula
+            var translation_params = core._t.database.parameters;
+            var decimal_point = translation_params.decimal_point;
+            var thousands_sep = translation_params.thousands_sep;
+
+            var value;
+            formula = formula.replace(thousands_sep, '').replace(decimal_point, '.');
```
tarteo, line 55: In javascript replace only replaces one occurrence. If you have 2 decimal separator in one formula it fails e.g.: =1.2/1.8

This will fix this problem
```
            if (decimal_point !== '.') {
                while (formula.indexOf(decimal_point) !== -1) {
                    formula = formula.replace(decimal_point, '.');
                }
            }
```

tarteo (changes requested):

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+*    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+**/
+
+odoo.define('web.web_widget_float_formula', function(require) {
+    "use strict";
+
+    var core = require('web.core');
+    var basic_fields = require('web.basic_fields');
+    var field_utils = require('web.field_utils');
+    var NumericField = basic_fields.FieldFloat;
```
daramousk, line 13: A `NumericField` is a different widget, did you intend to override the functionality on that one?
If not and this is only for `FieldFloat`, perhaps it would be more clear to change the name of `NumericField` to `FieldFloat`

web_widget_float_formula/static/src/js/web_widget_float_formula.js
```diff
@@ -0,0 +1,92 @@
...
+            // Convert Text to valid formula
+            var text = this._getValue();
+            var formula = this._process_formula(text);
+
+            if (formula !== false) {
+                var value = this._eval_formula(formula);
+                if (value !== false) {
+                    this._formula_text = "=" + formula;
+                    var newText = field_utils.format.float(value);
+                    this._setValue(newText);
```
daramousk, line 79: Perhaps you could `return newText` here and call super on `commitChanges()` so that you can avoid calling those two statements.

daramousk (commented):
Some comments, perhaps this could be of use here:
https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element

web_widget_float_formula/readme/USAGE.rst
legalsylvain: I personaly prefer video, as I see it more viewable. Feel free to replace by a gif, if you can generate one great.

quentinDupont (changes requested):
A part of changes from comments are in this PR https://github.com/legalsylvain/web/pull/1

alexey-pelykh:
Re-did the migration in #1592

PR #1351, opened 2019-08-08
[12.0][IMP] web_responsive: mouseover menu

Tardo:
Don't shows menus when mouse over if are waiting for an action to end

Cc @tecnativa


pedrobaeza (approved):
Tried on runbot and working great.

chienandalu (approved):
Code looks good. I don't know what to test in runbot though. Some steps to reproduce?

pedrobaeza:
@chienandalu: click on a menu that takes a bit to load. Move your cursor to other menus. Before: they got unfolded. Now: they are kept as is.

chienandalu:
Ok, tested on runbot as well :+1:

PR #1352, opened 2019-08-09
[ADD] Adds web_widget_bpmn

daramousk:
Some todo here:
1) The `bpmn-icon*`'s contents do not appear, these are defined in a new font and for some reason are not shown.
2) If you modify a graph, first you need to click outside on the form and then click save in order for it to be saved, this needs to be fixed after 1) is fixed
3) Some tests would be fine.
4) Perhaps give the opportunity to the user to pass options to the library via t-options.


web_widget_bpmn/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "web_widget_bpmn",
+    "version": "10.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "",
+    "summary": "",
```
MeetKD, line 10: Add category as a web and define summary.

Add website key.

web_widget_bpmn/static/src/js/web_widget_bpmn.js
```diff
@@ -0,0 +1,62 @@
...
+                Object.getOwnPropertyNames(
+                    this.bpmnjs.get('eventBus')._listeners).forEach(function(
+                        entry) {
+                        if (entry.match(/^element\..*$/)) {
+                            delete self.bpmnjs.get('eventBus')._listeners[
+                                entry]
+                        }
+                    })}
+            this.bpmnjs.importXML(this.view.datarecord[this.name]);
+            this.bpmnjs.get('canvas').zoom(0.5, 'auto');
```
MeetKD, line 32: It would be great if this zoom ratio can be configurable or we can pass as an option while defining this widget.

web_widget_bpmn/static/src/js/web_widget_bpmn.js
```diff
@@ -0,0 +1,62 @@
...
+//License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+odoo.define('web_widget_bpmn', function(require) {
+    var core = require('web.core');
+    var common = require('web.form_common');
+    var _t = core._t;
+
+    var FieldBPMN = common.AbstractField.extend(
+            common.ReinitializeFieldMixin, {
+
+
```
MeetKD, line 11: Can you please remove extra space.

MeetKD (changes requested):

Issue #1075, opened 2018-10-12
web_ckeditor4: update CKEditor lib source


daramousk:
- image 'upload' must still work. Note that this is not super efficient, it takes the image and embeds it as a data uri. Sometimes we want that, but not in the general case. Much more javascript that lets us select an existing ir.attachment or create a new one from an upload would be much better, but that's out of scope for a migration. I also never found anyone to fund this as I expect this to be some hours of work,
- this editor is known to mess up jinja2 expressions when cleaning html. In https://github.com/OCA/web/pull/158 I made an approach to fix this, maybe you can integrate that in your update and check if there were api improvements?

_Originally posted by @hbrunn in https://github.com/OCA/web/pull/820#issuecomment-424246939_

daramousk:
Created a ticket based on @hbrunn 's comments, If someone can assign this to me that would be great.

PR #1353, opened 2019-08-09
[IMP] Update ckeditor4 lib from 4.4.6 to 4.12.1 full version, readme generation

daramousk:
This is an implementation of @hbrunn 's request at https://github.com/OCA/web/issues/1075.
Pasting jinja expressions on the widget now does not seem to break them,
Everything else remained the same.

TODO: some tests could be nice, plus image uploading while working, it does show an `alert()`

cc: @StefanRijnhart, @hbrunn


hbrunn:
this is not reviewable. Please put the update of the CKeditor code in one commit, readme stuff in another and possible adaptions necessary in yet another one.

Further, update `.travis.yml` with the latest version from MQT (also in a separate commit) to have a working travis

NL66278 (changes requested):
:+1:  But should be rebased to solve conflict with updated travis configuration.

Issue #956, opened 2018-06-14
[10.0]web_many2one-clickable: context


whulshof:
The clickable link ends up in the wrong form and defining context in the field view does not work to direct to the right view.
There is a PR for 8.0 relating to this question, but I am not enough at ease with js to use this. Could anybody help me on the 10.0 version of the module?

pedrobaeza:
The action used is the default one, but there's no other way right now of expressing another different, sorry.

PR #354, opened 2016-05-13
[8.0] web_tree_many2one_clickable [ADD] context

kurt-schmitz:
When configure 'many2one_clickable' on invoice_id it open the wrong form view (supplier) instead
of the customer form view. On many2one fields you can use the context to open a specific view.
This is beeing ignored  by 'many2one_clickable' widet.

Ex.
<field name="invoice_id" widget="many2one_clickable" context="{'form_view_ref': 'account.invoice_form'}"/>

PR #1356, opened 2019-08-23
10.0 context clickable

whulshof:
Solution for Issue [10.0]web_many2one-clickable: context #956
When configure 'many2one_clickable' sometimes the wrong form view is opened. On many2one fields in forms you can use the context to open a specific view.
This is being ignored by 'many2one_clickable' widget. Current PR adapts the widget such that the context is evaluated.
inspired by [8.0] web_tree_many2one_clickable [ADD] context #354 from kurt-schmitz

### DIFF
--- a/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
+++ b/web_tree_many2one_clickable/static/src/js/web_tree_many2one_clickable.js
@@ -64,10 +64,8 @@ ListView.Column.include({
                     values.name = _.escape(row_data[this.id + '__display'].value ||
                                            options.value_if_empty);
                 }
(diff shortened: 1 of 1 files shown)


daramousk (commented):
Hi there and thank you for this, indeed people would expect have some control on the formview they view when clicking the item on the list.

While what you have done here will probably work for some cases, it will not work for cases in which you have things such as `{'default_field_id': field_on_view}`.

What if we check out how many2one fields are rendered on the FormView when it is on readonly mode? Checkout `FieldMany2One.execute_formview_action`, see how the context is build, make sure that the `get_formview_action` is called as well.

What do you think about this?

PR #1361, opened 2019-08-29
[IMP] Upgrade web-timeline vis.js version from 3.12 to 4.21

nzaterrence:


daramousk (commented):
Hello there and thanks for this contribution,

I was just checking this PR and I think there might be some duplicate files here.
Also, what prompted this migration? Do you have any specific use case for migrating specifically to 4.21? Because it seems that this version is EOL and the project has been archived.

As for the red travis, try rebasing, maybe the issue has been fixed.

Let me know what you think.

Thank you.

PR #1362, opened 2019-08-30
[ADD] Web widget Numeric step

quentinDupont:
Hello, 
That's my first widget and I find it complicated to find good documentation or ways of coding widgets so don't hesitate to tell me it's not good :smile: 

After first reviews, i made some changes and here is a new README :

### What's for ?
This widget changes input number field and make it easier to increment the number thanks to 2 buttons (+ and -). Use JS native logic for input number, so you can use the options min, max, step, placeholder.

### Usage

In your xml view, add ``widget="numeric_step"``
This will add the 2 buttons "+" and "-" just next to the input field in edit mode.
Iteration step by default is 1.

**Optional**

Add an option to choose the step iteration and limits (min and max values).

Example for an 0.25 step, min to -1 and max to 10 :

``<field name="sale_delay" widget="numeric_step" options="{'step': 0.25, 'min': -1, 'max': 10}" /> days``

**Examples**
Iteration with 0.25 step, min to -1 and max to 10.

Start to incremente with button, continue incrementing with scrolling mouse.

Iteration with 10 step, max limit 15, placeholder with onchange

cc @legalsylvain


pedrobaeza:
Interesting thing to check, @Tardo @Yajo @rafaelbn 

But I'm not sure about the module name

quentinDupont:
> Interesting thing to check, @Tardo @Yajo @rafaelbn
> 
> But I'm not sure about the module name

I was quite complicated to find a interesting name, so no problem to change if you have a better idea :smile:

web_widget_number_ux_choice/static/src/scss/number_ux_choice.scss
```diff
@@ -0,0 +1,11 @@
+
+.input_number_ux_choice {
+    width: 71% !important;
+}
+
+.btns_number_ux_choice {
+    width: 27% !important;
```
yajo:
```suggestion
    width: 27%;
```

web_widget_number_ux_choice/static/src/scss/number_ux_choice.scss
```diff
@@ -0,0 +1,11 @@
+
+.input_number_ux_choice {
+    width: 71% !important;
```
yajo: Avoid `!important` always if you can.

```suggestion
    width: 71%;
```

OTOH, to make sane widths and heights, I think it's better to leverage the bootstrap framework, even more in v12 where it's BS4 and uses flexboxes.

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
```diff
@@ -0,0 +1,74 @@
...
+* License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html) **/
+
+odoo.define('web.web_widget_number_ux_choice', function(require) {
+    "use strict";
+
+    var basic_fields = require('web.basic_fields');
+    var Registry = require('web.field_registry');
+    // var session = require('web.session');
+    // var rpc = require('web.rpc');
+    // var pyUtils = require("web.py_utils");
```
yajo: :fire: Remove garbage comments.

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
```diff
@@ -0,0 +1,74 @@
...
+
+    /*
```
yajo:
```suggestion
    /**
```

Marks the doc as a JSDoc string. Also the linter validates it using https://github.com/OCA/pylint-odoo/blob/master/pylint_odoo/examples/.jslintrc

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
```diff
@@ -0,0 +1,74 @@
...
+    * Get precision for the @input field
+    */
+    function getPrecision(input) {
+      if (!isFinite(input)) return 0;
+      var e = 1, p = 0;
```
yajo: Avoid meaningless variable names. They are hard to read. Longer vars are free! :money_with_wings:

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
```diff
@@ -0,0 +1,74 @@
...
+      while (Math.round(input * e) / e !== input) { e *= 10; p++; }
+      return p;
+    }
+
+    /*
```
yajo:
```suggestion
    /**
```

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
```diff
@@ -0,0 +1,74 @@
...
+        self.$input[0].value = newVal_s;
+    }
+
+    var NumberUxChoice = basic_fields.FieldFloat.extend({
+        template: 'NumberUxChoice',
+
+        _render: function () {
+            var self = this;
+            this._super();
+            this.$("button").parents().removeClass("o_field_integer o_field_number o_input o_required_modifier");
```
yajo: This kind of changes are clearer in the XML template itself.

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
```diff
@@ -0,0 +1,74 @@
...
+            this.$("button").click(function() {
```
yajo: No need to do this. You can use events. Check https://www.odoo.com/documentation/12.0/reference/javascript_cheatsheet.html#modifying-an-existing-field-widget

web_widget_number_ux_choice/static/src/xml/number_ux_choice.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
+@author: Quentin DUPONT <quentin.dupont@grap.coop>
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+<templates>
+    <t t-name="NumberUxChoice">
```
yajo: It's a common practice to prefix it with the module name.

```suggestion
    <t t-name="web_widget_number_ux_choice.field">
```

If the module name changes, adapt it here, of course.

yajo (changes requested):
This is a beautiful module.

I'm thinking that you should propose this upstream for master branch. This should be the default widget behavior for numeric inputs. Ping @ged-odoo on that matter (I bet he already considered this).

Also I'm wondering if you've tested [just using `<input type="number"/>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number), which supports step buttons and renders a numeric-only keyboard in mobiles. This would reduce a lot the JS logic here, as it would rely on native behavior.

If you still prefer to have it in a module, there are some code things to change below. Regarding the module name, I'd use `web_widget_numeric_step`.

Tardo:
As Yajo said, i think it would be better use native type="number" since it's compatible with IE11 (https://html5test.com/compare/browser/ie-11.html).

Maybe this module (big buttons) can go to the server-ux repository. ?¿??

quentinDupont:
I don't event think about native type="number" !
Thanks for your feedback, i'm gonna think about this and i come back :smile:

bealdav:
I love "due date" use case.
About widget name, shorter is probably better : `incremental` ?

Thanks

yajo:
>  Thanks for your feedback, i'm gonna think about this and i come back smile

That's great! Let me close this PR in the mean time. Reopen if needed. Thanks!

web_widget_number_ux_choice/static/src/scss/number_ux_choice.scss
quentinDupont: I can't manage to avoid ! important
And I also try flexboxes or to bootstrap it with container but you need to encapsule it in div and I didn't manage to put my field in div without breaking the widget magic..

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
quentinDupont: Done, thanks

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
quentinDupont: You're right !

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
quentinDupont: Odoo adds these classes after my template processing, I have no idea (and didn't find) how to avoid this !

web_widget_number_ux_choice/static/src/js/number_ux_choice.js
quentinDupont: With _onClick, it triggers for the input and i can't manage to differentiate my two buttons + and -

web_widget_number_ux_choice/static/src/xml/number_ux_choice.xml
quentinDupont: done !

quentinDupont:
Hello !
I would like you to reopen this PR please @Yajo :)

I made several changes : 
- use native input number (step, min, max, placeholder)
- remove native spinner and keep the buttons

And several points are being discussed above

You can see the code here : https://github.com/quentinDupont/web/tree/12.0_ADD_web_widget_ux_choice/web_widget_numeric_step

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -0,0 +1,95 @@
...
+        var newVal_s = newVal.toFixed(precision).toString();
+        self._setValue(newVal_s); 
+        self.$input[0].value = newVal_s;
+    }
+
+
+    NumericStep.include({
+        template: 'web_widget_numeric_step',
+
+        _render: function () {
```
daramousk, line 57: Since `var NumericStep = require('web.basic_fields').FieldFloat;`, by overriding `_render` you change the rendering of the `FieldFloat` as well, and by defining `template` you change the `template` of the FieldFloat.

Right now, when the module is installed all the float fields have this template rendered for them (even if the user is not using the widget) which looks like the picture attached and things break.
Screenshot from 2019-10-15 16-32-08

web_widget_numeric_step/static/src/scss/numeric_step.scss
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+/* Turn Off Number Input Spinners */
+input{
+    /* Firefox */
+    -moz-appearance: textfield;
+
+    /* Chrome */
+    &::-webkit-inner-spin-button {
```
daramousk, line 19: On chromium I am still able to see both sets of the buttons.

daramousk (commented):
Thanks for this,

I poked around with it a little bit and found some issues, also, I was under the impression that this could be done in a more straightforward way ex. overide the function below
https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/core/dom.js#L419
in order to deal with the error "The input element's type ('number') does not support selection" as shown here https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange
and force the type of the field to numeric.

In any case, if the issues are fixed I think this way is ok too.

web_widget_numeric_step/static/src/scss/numeric_step.scss
quentinDupont: Yep you're right.. can't fond the way in chromium. Every blog post about this problem are solving this for Firefox/Opera/Chrome but it seems to be different for Chromium..

web_widget_numeric_step/static/src/js/numeric_step.js
quentinDupont: Not good indeed !
I'm not sure to know how to handle this, do you have advices or good examples ?

Should I `extend` instead of `include` `FieldFloat` ?

legalsylvain: maybe something like that :   

https://github.com/OCA/web/blob/12.0/web_widget_color/static/src/js/widget.js#L9

quentinDupont: When I try to change from `include` to `basic_fields.FieldFloat.extend`, it never goes through my extending function like `_renderEdit` or `_prepareInput` (that are function of InputField that FieldFloat inherits in web.basic_fields)

I really try to copy what's working on `web_widget_color` but it doesn't seem to work for me. And I find this quite hard to debug.

One clue i find in debugging mode :
When it initiates my `basic_fields.FieldFloat.extend`, i find a difference in what the js return between my widget `NumericStep` and `FieldColor` where it returns `FieldColor`. 

Sans titre

I don't know what's that mean or if it's relevant..

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -0,0 +1,95 @@
...
+            this.$("button").parents().removeClass("o_field_integer o_field_number o_input o_required_modifier");
+            this.$("button").click(function() {
+                var node = $(this).parent()[0];
+
+                // Get step option or default is 1
+                if (typeof node.attributes['step'] !== 'undefined') {
+                    var step = parseFloat(node.attributes['step'].value);
+                } else {
+                    var step = 1;
+                }
```
Tardo, line 80:
```javascript
var step = Number((node.attributes['step'] || {}).value);
if (isNaN(step)) { step = 1.0; }
```

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+    * Increase input number with chosen step iteration
+    *
+    * @param {object} self Input field
+    * @param {Float} step Step iteration
+    * @param {String} minusOrPlus Choose "minus" to decrease. Default is "plus"
+    */
+    function addStep(self, step, minusOrPlus) {
```
Tardo, line 35: This method is not accessible out of the 'module' scope :(

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -0,0 +1,95 @@
...
+    var Registry = require('web.field_registry');
+    var NumericStep = require('web.basic_fields').FieldFloat;
+
+    /**
+    * Get decimal precision for the input field
+    *
+    * @param {object} input Input field
+    * @return {int} Precision
+    */
+    function getPrecision(input) {
```
Tardo, line 18: This method is not accessible out of the 'module' scope :(

web_widget_numeric_step/static/src/xml/numeric_step.xml
```diff
@@ -0,0 +1,19 @@
...
+    <t t-name="web_widget_numeric_step">
+        <t t-if="widget.mode != 'readonly'">
+            <input type="number" class="input_numeric_step"/>
+            <div class="btns_numeric_step">
+                <button class="fa fa-plus btn btn-secondary btn_numeric_step_plus" aria-label="Plus" title="Plus"/><button class="fa fa-minus btn btn-secondary btn_numeric_step_minus" aria-label="Minus" title="Minus"/>
+            </div>
+        </t>
+        <t t-if="widget.mode == 'readonly'">
+            <field/>
+        </t>
```
Tardo, line 17: I think that you can remove this.

Tardo (changes requested):
If you type a letter, the widget brokes.

PR #1369, opened 2019-09-10
[12.0][NEW] web_widget_plotly_chart

robyf70:
Add new widget for chart plotting using the plotly library, see https://github.com/plotly/plotly.py


web_widget_plotly_chart/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+{
+    'name': "Web Widget Plotly",
+    'summary': """Allow to draw plotly charts.""",
+    'description': """Allow to draw plotly charts.""",
+    'author': "Roberto Fichera, LevelPrime srl, "
```
pedrobaeza: Only company as author, please

web_widget_plotly_chart/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    'category': 'Web',
+    'version': '1.0.0',
```
pedrobaeza: Put full version: `12.0.1.0.0`

pedrobaeza (dismissed):
Thanks for the contribution!

One quick review + I miss the image you mention in the README

web_widget_plotly_chart/__manifest__.py
robyf70: @pedrobaeza Ok! Will do now

web_widget_plotly_chart/__manifest__.py
robyf70: @pedrobaeza All done

pedrobaeza:
OK, thanks. I miss a link to plotly documentation for reference of possible charts and its arguments, but I let real reviewers of the module to check deeply.

robyf70:
> OK, thanks. I miss a link to plotly documentation for reference of possible charts and its arguments, but I let real reviewers of the module to check deeply.

The library is very huge you can do much more things compared to Bokeh, here is where you can start reading https://github.com/plotly/plotly.py#overview

MarcoCalcagni (approved):

GSLabIt (approved):

andreampiovesana (approved):

web_widget_plotly_chart/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
daramousk: Please remove all the `coding` declarations from all the `*.py` files.

web_widget_plotly_chart/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    'name': "Web Widget Plotly",
+    'summary': """Allow to draw plotly charts.""",
+    'description': """Allow to draw plotly charts.""",
+    'author': "LevelPrime srl, "
+              "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    'category': 'Web',
+    'version': '12.0.1.0.0',
+    'depends': ['base'],
```
daramousk: `base` is always implicitly added, change this to `web`.  https://github.com/OCA/OCB/blob/12.0/odoo/modules/loading.py#L352

web_widget_plotly_chart/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    'data': [
+        'static/src/xml/assets.xml'
```
daramousk: The `assets.xml` needs to be moved outside the `static` folder, because now Odoo will send it to the client and there is no need for that. https://github.com/OCA/OCB/blob/12.0/odoo/http.py#L1345

daramousk (approved):
Hello and thanks for this, it works great.

 While on forms it works great it does not on a list view. (if you click lasso for example the click event propagates to the listview and opens the record.

Either than that from what I have seen on a form view this works great.

robyf70:
@daramousk Thanks for the review! I've pushed the changes you suggested

robyf70:
@pedrobaeza There is anything else to do to get this merged?

PR #1371, opened 2019-09-11
[12.0][ADD] web_widget_dropdown_dynamic

alexey-pelykh:
Dynamic dropdown widget that supports resolving options from backend of:

 * `fields.Char`
 * `fields.Integer`
 * `fields.Selection`

### DIFF
--- a/web_widget_dropdown_dynamic/readme/DESCRIPTION.rst
+++ b/web_widget_dropdown_dynamic/readme/DESCRIPTION.rst
@@ -1,0 +1,9 @@
+Dynamic dropdown widget that supports resolving options from backend of:
+
+ * ``fields.Char``
+ * ``fields.Integer``
+ * ``fields.Selection``
+
+**NOTE:** This widget is not intended to *extend* ``fields.Selection``, but to
+filter selection values. For fully-dynamic set of options, use ``fields.Char``
+instead.
--- a/web_widget_dropdown_dynamic/static/src/js/field_dynamic_dropdown.js
+++ b/web_widget_dropdown_dynamic/static/src/js/field_dynamic_dropdown.js
@@ -1,0 +1,140 @@
+/* 
+ * Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+ */
+odoo.define('web_widget_dropdown_dynamic.field_dynamic_dropdown', function (require) {
+    "use strict";
+
+    var core = require('web.core');
+    var AbstractField = require('web.AbstractField');
+    var field_registry = require('web.field_registry');
+
+    var _lt = core._lt;
+
+    var FieldDynamicDropdown = AbstractField.extend({
+        description: _lt('Dynamic Dropdown'),
+        template: 'FieldSelection',
+        specialData: '_fetchDynamicDropdownValues',
+        supportedFieldTypes: ['selection', 'char', 'integer'],
+        events: _.extend({}, AbstractField.prototype.events, {
+            'change': '_onChange',
+        }),
+        /**
+         * @override
+         */
+        init: function () {
+            this._super.apply(this, arguments);
+            this._setValues();
+        },
+    
+        //--------------------------------------------------------------------------
+        // Public
+        //--------------------------------------------------------------------------
+    
+        /**
+         * @override
+         * @returns {jQuery}
+         */
+        getFocusableElement: function () {
+            return this.$el.is('select') ? this.$el : $();
+        },
+        /**
+         * @override
+         */
+        isSet: function () {
+            return this.value !== false;
+        },
+        /**
+         * Listen to modifiers updates to hide/show the falsy value in the dropdown
+         * according to the required modifier.
+         *
+         * @override
+         */
+        updateModifiersValue: function () {
+            this._super.apply(this, arguments);
+            if (!this.attrs.modifiersValue.invisible && this.mode !== 'readonly') {
+                this._setValues();
+                this._renderEdit();
+            }
+        },
+    
+        //--------------------------------------------------------------------------
+        // Private
+        //--------------------------------------------------------------------------
+    
+        /**
+         * @override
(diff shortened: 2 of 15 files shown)


web_widget_dropdown_dynamic/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Dynamic dropdown widget that supports resolving options from backend.
```
pedrobaeza: Which field types are allowed in the widget?

alexey-pelykh: I plan to support character fields, initially. Maybe integer would be easy to support

pedrobaeza: Include selection as well

alexey-pelykh: What is usecase of that?

pedrobaeza: Restrict selection values according a criteria. Real example, in VAT declaration in Spain, you have 5 possible values in the result:

- To be debited in bank account
- To enter directly
- To return
- To be held in fiscal account
- To be compensated

First 2 values are selectable when the declaration is positive, and the other 2 when the declaration is negative. We need 3 fields (or at least 2) now for reflecting this + `attrs` in fields.

pedrobaeza:
@Tardo should be the technical reviewer, although he is on holidays, so it will be a bit until he does it.

foutoucour:
@alexey-pelykh could you add screenshots to illustrate the use cases? The module seems interesting but might be hard to understand the scope at the first read.
Thanks

alexey-pelykh:
@foutoucour not sure there's a good screenshot to express it. Main usecase is: there's a field or fields (e.g. `amount`) and selection dropdown options depend if field is positive or negative. Another example: there's an login/password to third-party service and profile selection that is fetched after login/password is entered.

tarteo (approved):
The code LGTM, it's very clean and clear.

It would be nice to add some demo data to make functional testing easier.

alexey-pelykh:
@tarteo another module with `_demo` suffix?

alexey-pelykh:
@pedrobaeza does this module fit in your usecase?

pedrobaeza:
I think so, but no time now for switching current approach that I exposed to use this new module. When migrating to v13 involved modules, we will be able to try it.

web_widget_dropdown_dynamic/static/src/js/field_dynamic_dropdown.js
```diff
@@ -0,0 +1,143 @@
...
+         * @returns {jQuery}
+         */
+        getFocusableElement: function () {
+            return this.$el.is('select') ? this.$el : $();
+        },
+        /**
+         * @override
+         */
+        isSet: function () {
+            return this.value !== false;
```
daramousk, line 45: Is this override needed?

daramousk (commented):
Works great for Chars and Integers. If I try to use this on an already existing Selection field, and I try to return values that do not already exist on the `selection=[]` definition of that field, then I will get a `ValueError`. You can try this with the product type field.

alexey-pelykh:
@daramousk but the case you've described effectively abuses the Selection field, so you should get ValueError.

web_widget_dropdown_dynamic/static/src/js/field_dynamic_dropdown.js
alexey-pelykh: Yes, `AbstractField` has different implementation of `isSet`: https://github.com/odoo/odoo/blob/856c2e9008f1af7bc1327d9c2900db1e109ab0fa/addons/web/static/src/js/fields/abstract_field.js#L276

tarteo:
> @tarteo another module with `_demo` suffix?

Not necessarily you could use a field of a model which is introduced in `base`.

alexey-pelykh:
@pedrobaeza any luck trying this module out?

pedrobaeza:
Sorry, no

artemsenko (approved):
Functional test via https://github.com/OCA/bank-statement-import/pull/212

alexey-pelykh:
@tarteo @pedrobaeza I've added a demo, please review and merge at earliest convenience

web_widget_dropdown_dynamic/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+Dynamic dropdown widget that supports resolving options from backend of:
+
+ * `fields.Char`
+ * `fields.Integer`
+ * `fields.Selection`
```
pedrobaeza: I doubt this one is supported when you save a value that is not on the original selection.

alexey-pelykh: You are correct, saving a value that does not exist in original `selection` will give an error, which is by design. This field can be used only to _filter_ selection, not to extend. For fully-dynamic fields, use `Char` or `Integer`

pedrobaeza: You can put a note about that.

web_widget_dropdown_dynamic_example/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from . import models
```
pedrobaeza: Line at the end

alexey-pelykh: :) Don't we have a linter for that?

pedrobaeza: That's the weird thing... It's not detected right now

alexey-pelykh: Yep, 100%, on other modules as well

alexey-pelykh:
@pedrobaeza comments attended

PR #1377, opened 2019-09-18
[10.0][FIX] make color_field actually work, add background_color_field

ntsirintanis:
There was a bug that made the option for coloring the entire tree not applicable. This PR fixes that bug, and also introduces new field `background_color_field`, which adds color to every row's background, while existing field `color_field` does the same but for text.
@hbrunn


lfreeke (approved):
Functional test :+1:

daramousk (approved):

PR #1380, opened 2019-09-25
[FIX][12.0] web_responsive: changed 'button box more' initial position

### LINKED ISSUE #1379: [12.0] web_responsive: smart buttons drop down issue when clicking twice to expand
When you have a lot of smart buttons and they are hidden in the "more" button, you can see that it displays differently comparing the first time you click and the second (if you collapse and expand again).

web_responsive-smart-button-issue

Tardo:
This PR fixes #1379 issue. Now the menu opens all times in the same place:
button_click_b

Cc @tecnativa

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -329,7 +329,14 @@ html .o_web_client .o_main .o_main_content {
     .o_form_view {
         .o_form_sheet {
             max-width: calc(100% - 32px);
-            overflow-x: scroll;
+            overflow-x: auto;
+
+            .oe_button_box {
+                .o_dropdown_more {
+                    padding: 0.1em;
+                    width: min-content;
+                }
+            }
         }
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But is this how it is shown on vanilla v12? I see that a bit odd...

Tardo:
Yes, it's what it looks like on vanilla v12 :/

pedrobaeza:
An can we improve it for having only the needed width or is it too difficult?

LoisRForgeFlow:
Yes, FWIW in CE is like this (all the time wider than the other buttons), but in EE it is always displaying with needed width.

Tardo:
Ok, now looks as expected:
button_click_b

pedrobaeza (approved):

LoisRForgeFlow (approved):
Tested, LGTM :+1:

yajo (approved):
This seems OK, although I think we should fix it upstream instead if it's upstream problem.

PR #1385, opened 2019-09-27
[12.0][IMP] web_widget_color: Apply required modifier

### LINKED ISSUE #1383: [12.0][web_widget_color] Required not working and deviation from Odoo look
The field is not width 100% like other fields:

image

If the field is not required you're not able to put nothing in the field.

tarteo:
Fixes #1383

The width can be altered by adding a style attribute to the field element.

Also removes warning message "Cannot link jscolor twice to the same element. Skipping."

### DIFF
--- a/web_widget_color/static/src/js/widget.js
+++ b/web_widget_color/static/src/js/widget.js
@@ -16,7 +16,23 @@ odoo.define('web.web_widget_color', function(require) {
 
         _renderEdit: function() {
             this.$input = this.$el.find('input');
-            this.jscolor = new jscolor(this.$input[0], {hash:true, zIndex: 2000});
+
+            if (!this.$input[0]._jscLinkedInstance) {
+                this.jscolor = new jscolor(this.$input[0], {
+                    hash: true,
+                    zIndex: 2000,
+                });
+            }
+
+            this.jscolor.fromString(this.value);
+        },
+
+        updateModifiersValue: function () {
+            this._super.apply(this, arguments);
+            if (this.mode !== 'readonly') {
+                this.jscolor.required = this.attrs.modifiersValue.required;
+                this.$el.removeClass('o_required_modifier');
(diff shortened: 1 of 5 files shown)


web_widget_color/static/src/js/widget.js
```diff
@@ -16,8 +16,20 @@ odoo.define('web.web_widget_color', function(require) {
 
         _renderEdit: function() {
             this.$input = this.$el.find('input');
-            this.jscolor = new jscolor(this.$input[0], {hash:true, zIndex: 2000});
+            this.jscolor = new jscolor(this.$input[0], {
```
daramousk: I believe we shouldn't create unnecessary difs just for formatting.

daramousk (commented):
I am not sure about this fix, if I make the field now required, it always gets a #FFFFFF value even though I explicitly delete it, I do not get an alert about a field that has been left empty though as I get with the other fields

web_widget_color/static/src/js/widget.js
tarteo: It fixes lint errors.

tarteo:
> I am not sure about this fix, if I make the field now required, it always gets a #FFFFFF value even though I explicitly delete it, I do not get an alert about a field that has been left empty though as I get with the other fields

Currently the module doesn't respect 'required == false'. So when I want to make an optional color field I cannot put an empty string into it. **It currently doesn't show an error when the field is required.**

So it always has a value when required (which is done by the colorjs library) and **you're not** able to make the field optional, or it always has a value when required and **you're able** to make the color field optional. I prefer the latter, it's objectively better.

NL66278:
There seems to be a conflict with this commit: https://github.com/OCA/web/commit/d1030754628b5ad777137c708a7c6c5399741388

Is this PR still relevant, or can it be closed?

PR #1386, opened 2019-09-27
[WIP] [10.0] [ADD] web_advanced_search_list

MPoch-PlanetaTIC:
New module to add the 'is in' option to advanced search of char fields. With this module, in a tree view you can filter by listing patterns separated by pipe char(|).

Usage:
 1. Select a Char field.
 2. Select "is in" operator
 3. Input the list of patterns separated by pipe char (|).

cc: @PlanetaTIC

### DIFF
--- a/web_advanced_search_list/__init__.py
+++ b/web_advanced_search_list/__init__.py
@@ -1,0 +1,3 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 PlanetaTIC - Marc Poch <mpoch@planetatic.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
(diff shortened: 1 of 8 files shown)


daramousk (commented):
Hi there and thank you for this.

I have some questions here, for going through the `search.js` and I think that the funtionality that this module introduces already exists. The default search operation as I see in `web.search_filters` is `ilike` and every piece of the domain is chained together with an `OR` operation.

Isn't this achieving the same thing as this module?

MPoch-PlanetaTIC:
> Hi there and thank you for this.
> 
> I have some questions here, for going through the `search.js` and I think that the funtionality that this module introduces already exists. The default search operation as I see in `web.search_filters` is `ilike` and every piece of the domain is chained together with an `OR` operation.
> 
> Isn't this achieving the same thing as this module?

Hello @daramousk 

You are right, you can chain search operations to obtain the same result.

But this module can save you time when, for example, you have a list of products references in an Excel file, and you want to list all of them in Odoo, then you copy all default codes, paste them in a text editor where you will separate all default code by pipe character ('|') and copy the string result, finally in Odoo's search domain of products list you can add a filter saying:
   Reference; "is in"; paste the string.
Useful to users who works with list of items.

PR #1390, opened 2019-09-30
[13.0][MIG] web_responsive: Migration to 13.0

Tardo:
Don't merge until do:
* Show chatter buttons-box like 12.0

Cc @Tecnativa TT19829


GSLabIt:
Seems there are small issue on mobile view regarding: 

- multi company dropdown
- smart buttons under "More" dropdown

see pic:

Schermata 2019-10-01 alle 19 30 04

web_responsive/models/res_users.py
```diff
@@ -0,0 +1,26 @@
...
+from odoo import models, fields
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    chatter_position = fields.Selection([
+        ('normal', 'Normal'),
+        ('sided', 'Sided'),
+    ], string="Chatter Position", default='normal')
```
yajo:
```suggestion
    ], string="Chatter Position", default='sided')
```

I think it's time to flip the switch...

web_responsive/static/src/js/web_responsive.js
```diff
@@ -172,14 +172,14 @@ odoo.define('web_responsive', function (require) {
             this._search_def.reject();
             this._search_def = $.Deferred();
```
yajo: I think you should use a `Promise` now.

web_responsive/static/src/js/web_responsive.js
```diff
@@ -442,26 +445,26 @@ odoo.define('web_responsive', function (require) {
...
             // Need close Mobile?
-            var menu_sections_mobile = document.querySelector(
+            const menu_sections_mobile = document.querySelector(
                 '.o_menu_sections.show');
             $(menu_sections_mobile).has(uniq_sel).collapse('hide');
         },
 
         _handleAction: function (action) {
-            return this._super.apply(this, arguments).always(
+            return this._super.apply(this, arguments).finally(
```
yajo, line 492: AFAIK, if super returns a `Promise` (I think it should now), then the direct replacement for `always()` is `then(fct).guardedCatch(fct)`, where `fct` is the method to be called.

Idea taken from https://github.com/odoo/odoo/pull/31887

web_responsive/readme/ROADMAP.rst
```diff
@@ -4,3 +4,5 @@
   new size. This is Odoo's own limitation.
 * App navigation with keyboard.
 * Make it more beautiful. Maybe OCA-branded?
+* The button for manage list view columns dissapears when scroll because
+  can't have sticky position.
```
yajo: You can do it :wink::

1. Override https://github.com/odoo/odoo/blob/21cccf9d7235b80fab30e19ce3fbd5df4a8ee686/addons/web/static/src/js/views/list/list_renderer.js#L916
1. Check if `.o_optional_columns_dropdown_toggle` exists.
1. If so, move it to the end of the last `thead>tr[0]>th` element.

The rest is playing a bit with CSS, and you're done.

You'd have to check what happens if the last col is hidden, though. Maybe you'd have to append it directly to the end of `thead>tr`.

yajo (commented):
Looks good.

web_responsive/static/src/js/web_responsive.js
Tardo: I think that in this case "finally" (then + catch) it's better than use "then" + "guardedCatch" (catch without errors) because i want to ensure that the menu is hidden in any circumstance. If not, the user can get stucked in the menu. What do you think?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

yajo: Oh definitely, I didn't realize that was the purpose. Thanks for the info :wink:

Tardo:
@GSLabIt multi company dropdown is broken in vanilla

GSLabIt:
@Tardo oh sorry, didn't figured it out ..

Tardo:
@GSLabIt no problem, thanks for report.
multi-company dropdown fixed here: https://github.com/odoo/odoo/pull/38146

Suggested changes applied.

GSLabIt:
@Tardo, "More" drop down looks much better.. maybe needs right border in last rigth btn? 

Screenshot_20191008_073620_com sec android app sbrowser beta

AdriaGForgeFlow:
Hi @Tardo do you think the feature I included in this commit [Eficent/odoo](https://github.com/Eficent/odoo/commit/5d4d517b340a34a2ca98eb50d828cb7734f2f636) can be added in this module?

It basically introduces the option to multi select in a list view using the shift key:
ListMultiSelection

I know the code is not the cleanest, I did my best. Maybe you can come up with a better approach or maybe it's better to propose this to Odoo directly. What are your thoughts?

pedrobaeza:
@ageficent that must be an independent module, and in fact, you already have it in https://github.com/OCA/web/tree/12.0/web_listview_range_select

AdriaGForgeFlow:
> @ageficent that must be an independent module, and in fact, you already have it in https://github.com/OCA/web/tree/12.0/web_listview_range_select

I hadn't seen it haha... Thanks! :+1:

GSLabIt (approved):
Functional LGTM

emagdalenaC2i (approved):
Code review

emagdalenaC2i:
@Tardo See https://github.com/OCA/maintainer-quality-tools/pull/613#issuecomment-541333517 to fix travis

install pre-commit (you need to do this once)
run pre-commit run --all-files, which will apply the necessary code formatting fixes
commit and push the changes

pedrobaeza (approved):
Seems to work reasonably well. I have detected that the smart-buttons are a bit more wider than on standard while on desktop, but I don't see that as something bad, as sometimes the text is cut.

The problem with the search panel in responsive is still there, but it's something that @Yajo will solve for 12.0 and then we forward-port it to 13.0.

Thanks for the work.

yajo:
I'm a bit late to the party, but @ageficent, regarding https://github.com/OCA/web/pull/1390#issuecomment-539400519, have you considered pushing that as a PR for Odoo? I'm pretty sure they'd accept that for master...

tdombos:
On mobile chrome  (Android) if you click on a menu item in the second menu (within-app menu), and then click on the drawer icon again, the menu doesn't appear.

pedrobaeza:
Please open a new issue with screenshots/video.

PR #1391, opened 2019-10-01
[IMP] web_responsive: Better support for long titles

Tardo:
long_title

Rules used:

* Small screens uses 100%
* Other screen sizes uses 100% - (132px * 3 + Some margin) => Size of button * max colums (+ added some margin) [132px it's a hardcoded size]

Cc @tecnativa

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -16,6 +16,15 @@
     transform: none !important;
 }
 
+// Support for long titles
+@include media-breakpoint-up(md) {
+    .o_form_view .oe_button_box + .oe_title,
+    .o_form_view .oe_button_box + .oe_avatar + .oe_title {
+        /* Button-box has a hardcoded width of 132px per button and have three columns */
+        width: calc(100% - 450px);
+    }
+}
+
 // Main navbar (with apps menu, user menu, debug menu...)
 @include media-breakpoint-down(sm) {
     .o_main_navbar {
@@ -348,8 +357,6 @@ html .o_web_client .o_main .o_main_content {
 
         // Support for long title (with ellipsis)
         .oe_title {
-            width: initial;
-
             span.o_field_widget {
                 max-width: 100%;
                 text-overflow: ellipsis;
@@ -370,6 +377,11 @@ html .o_web_client .o_main .o_main_content {
             .oe_title,
             {
                 max-width: 100%;
+            }
+
+            .oe_button_box + .oe_title,
+            .oe_button_box + .oe_avatar + .oe_title {
+                width: 100%;
             }
 
             // Avoid overflow on modals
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
Although not sure if `!important` is wasting other things

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -16,6 +16,18 @@
     transform: none !important;
 }
 
+// Support for long titles
+@include media-breakpoint-up(md) {
+    .o_content, .modal-content {
+        .o_form_view {
+            .oe_title {
+                /* Button-box has a hardcoded width of 132px per button and have three columns */
+                width: calc(100% - 450px) !important;
```
yajo: I tested this and it seems to have the same effect, so as a good rule of thumb, always avoid `!important` clauses:

```suggestion
                width: calc(100% - 450px);
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -372,6 +382,10 @@ html .o_web_client .o_main .o_main_content {
                 max-width: 100%;
             }
 
+            .oe_title {
+                width: 100% !important;
```
yajo: It seems like this `!important` should stay, however. Unless, maybe, if we switch to top full button row as suggested above.

yajo (changes requested):
On wide screens we get this strange behavior:

Captura de pantalla de 2019-10-02 11-13-10

I found interesting how this is handled in enterprise:

Captura de pantalla de 2019-10-02 11-14-47

It seems to me that, if we make the button box 100% wide as a top row, just like enterprise, then we can set oe_title also to 100% width always (no matter screen size), and we can get more comfortable and consistent form screens. As an example see the difference here:

Peek 02-10-2019 11-20

Versus this (I added all those "Active" buttons just to trigger the dropdown):

Peek 02-10-2019 11-32


Not sure what you think, but I think the enterprise one is more pleasant.

Well, just take this as an idea. The current behavior is not "wrong", and I wouldn't block this PR if you prefer it like it is. On that subject, just see the comments below.

LoisRForgeFlow:
For me @Yajo proposal is the best, also note that in v13 CE the button box looks like this with the top of the form dedicated to it.

pedrobaeza:
Yeah, but that means deeper changes. What do you think, @Tardo?

yajo:
@lreficent Thanks, I didn't notice this was also like this in vanilla v13 CE:

Captura de pantalla de 2019-10-09 13-21-39


~~Then AFAICS it would be actually *less* work to leave that part as it is upstream, right? I mean, it would mean actually code removal.~~ :point_left: Sorry, I didn't realize this PR is for v12. Then, maybe you can add that to the roadmap for v13, and accept this PR as it is for v12.

Tardo:
Ok, changes done... i know that it's not the best solution but is one that not require a lot of changes  in css and js.

GSLabIt (approved):

pedrobaeza:
Wrong rebase with extra unwanted commit

Tardo:
Oops! sorry, fixed commit history.

LoisRForgeFlow (approved):
Ok, this is already an improvement.

Thank you!

pedrobaeza:
@Yajo your call

yajo (approved):

PR #1393, opened 2019-10-02
12.0 mig web drop target

daramousk:


lfreeke (approved):
Functiona test.

I notice only after a reload you see the attached file. Maybe it should be added in the Known issues. If not easily can be solved.

etobella (changes requested):
Review why _onFileUploaded is not being called please

LoisRForgeFlow (approved):
Tested and it work perfectly in runbot (issues commented above are solved).  :+1: 

Thank you

PR #1394, opened 2019-10-04
[MIG] : Migrated web_dialog_size to v13

sudhir-erpharbor:


bealdav (approved):
Tested: OK

pedrobaeza (changes requested):
The change you have made in JS is affecting the functionality, as now 2 arrows appear when opening a popup dialog:

Selección_003

I suspect the previous instruction was waiting for solving full window before applying the conditions, and now you have removed that asynchronous part. The system has changed to use JS ES6 promises, so you have to adapt the code accordingly.

sudhir-erpharbor:
> The change you have made in JS is affecting the functionality, as now 2 arrows appear when opening a popup dialog:
> 
> Selección_003
> 
> I suspect the previous instruction was waiting for solving full window before applying the conditions, and now you have removed that asynchronous part. The system has changed to use JS ES6 promises, so you have to adapt the code accordingly.

Fixed the issue.

pedrobaeza (approved):

PR #1396, opened 2019-10-08
[ADD] web_view_calendar_column

etobella:
This module allows to configure multiple columns on calendar daily views.

Usage
=====

In order to define columns, we must access the action of the menu and edit the
context.
We should add the item column with the name of the field to use.

If we want to define a set of options, we can use `force_columns` on the
context.
With the force column, we will define a dictionary with the field value as key
and column name as value.

The libraries used for this module are GPL, so I think it is consistent with AGPL license. Any comments on this matter is welcome, as I don't have many knowledge on this.

@jbeficent @jarroyomorales


pedrobaeza:
Screenshot?

etobella:
imagen

etobella:
Drag and drop between columns work as expected

etobella:
Travis fails due to postgres. Fixing on #1397

pedrobaeza:
And what happened in month view?

etobella:
Mmmh, good point I will check it

etobella:
@pedrobaeza , What do you mean by Monthly view? I was trying to find a bug, but I imagined it. 

I don't know if it is a good idea to split by columns on weeks and months, because it can be hard to read. However, It can be done for weekly view, but I am not sure about monthly view.

pedrobaeza:
Yeah, exactly. What I mean is that these columns don't serve for that views, and maybe they are creating some kind of conflict.

etobella:
There should be no conflicts. The principal library are from fullcalendar, an addon from the library used by odoo. https://fullcalendar.io/docs/v3/vertical-resource-view
So, it should no changes should be applied there. Also, on monthly view, we are unable to split by resource, as the view would be crazy... We can add it to the weekly view, but I am not sure if it is a good choice.

Also, you can see that weekly view has no changes:
imagen

The same for monthly view:
imagen

pedrobaeza:
OK, if no split is done on that views, then perfect. Please explain this on the README.

JordiBForgeFlow:
@ageficent @lreficent

etobella:
I think this module is hard to test, so I will try to explain how I test it:

- Install **Sales management**
- Enter on developer mode
- Access Sales / Quotations
- Open Calendar view
- Edit the Calendar action (using the debug button) and set the field `'column': 'state'` on context
- Refresh
- Access daily view, now you will see it splitted

If you want to check it with other values, you could edit the calendar action again and add: 
`'force_columns': {'draft': 'Draft','sent': 'Send', 'sale': 'Sale', 'done': 'Locked'}` on context too

jarroyomorales (approved):
Tested and works as expected! :+1:

JordiBForgeFlow (approved):
:+1: Tested, works perfect!

AdriaGForgeFlow (approved):
Functional review :+1:

PR #1400, opened 2019-10-10
[10.0] [ADD] web_widget_mermaid

janverb:
This is a widget for rendering [mermaid](https://mermaidjs.github.io) diagrams, like flowcharts:

flowchart


web_widget_mermaid/static/src/js/web_widget_mermaid.js
```diff
@@ -0,0 +1,101 @@
...
+            barHeight: 20,
+            barGap: 4,
+            topPadding: 50,
+            leftPadding: 75,
+            gridLineStartPadding: 35,
+            fontSize: 11,
+            fontFamily: '"Lucida Grande", Helvetica, Verdana, Arial, '
+                        + 'sans-serif',  // Match Odoo's font choices
+            numberSectionStyles: 4,
+            axisFormat: '%Y-%m-%d',
```
hbrunn: this you can take on runtime from `_t.database.parameters.date_format`: https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/js/framework/translation.js#L12

web_widget_mermaid/static/src/js/web_widget_mermaid.js
```diff
@@ -0,0 +1,101 @@
...
+            // rid of it if it's still around. The id is based on the chartId.
+            $('#d' + this.chartId).remove();
+        }
+    })
+
+    core.form_widget_registry.add('mermaid', MermaidField);
+
+    return {
+        MermaidField: MermaidField,
+    };
```
hbrunn, line 102: export the default config here, then other plugins can change it to their needs

web_widget_mermaid/static/src/js/web_widget_mermaid.js
```diff
@@ -0,0 +1,101 @@
...
+            gridLineStartPadding: 35,
+            fontSize: 11,
+            fontFamily: '"Lucida Grande", Helvetica, Verdana, Arial, '
+                        + 'sans-serif',  // Match Odoo's font choices
+            numberSectionStyles: 4,
+            axisFormat: '%Y-%m-%d',
+        }
+    };
+
+    var MermaidField = core.form_widget_registry.get('text').extend({
```
hbrunn: better require `web.form_widgets` and use `FieldText` from there - you never know what ends up here

hbrunn (changes requested):
thanks. Please include one commit that updates this branch's `.travis.yml` to the latest version from [MQT](https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml).

Is it true that the included library is an unmodified version of mermaid's js code? If not, let's discuss the changes.

It looks to me that mermaid itself uses a versioning scheme of `x.y.z`, as this is the case for Odoo addons too, I think it makes sense to just use mermaid's version as the addon version. Otherwise, name the latest version used in the changelog file.

web_widget_mermaid/static/src/js/web_widget_mermaid.js
janverb: `form_widgets` [doesn't export](https://github.com/OCA/OCB/blob/acb74b37775870a6daab3dd77e2c15dcabd4f7ed/addons/web/static/src/js/views/form_widgets.js#L1760) `FieldText`, it only registers it. It does export `FieldChar`. Should I be using that instead?

hbrunn: interesting, keep it as it is then

hbrunn (approved):

hbrunn:
PS: please squash your commits so that we have just one adding the module, and another one updating `.travis.yml`. And if @gfcapalbo also reviews and approves, we're close to merging this

jarroyomorales (approved):
Very nice module! :+1:

PR #1404, opened 2019-10-14
[12.0][FIX] web_widget_x2many_2d_matrix: rendering in kanban mode

### LINKED ISSUE #1387: [12.0] web_widget_x2many_2d_matrix gives error on mobile/responsive
@hbrunn 

1) right click on a browser chrome/firefox etc. and click "inspect element"
2) click on phone icon to switch to responsive mode
3) open a form which uses web_widget_x2many_2d_matrix and you get error

observable in runbot....

Tardo:
Workaround to not break the widget on mobile.

Fixes #1387

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_view.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_view.js
@@ -1,0 +1,23 @@
+/* Copyright 2019 Alexandre Díaz
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixView', function (require) {
+    "use strict";
+
+    var BasicView = require('web.BasicView');
+
+    BasicView.include({
+        _processField: function (viewType, field, attrs) {
+            // Workaround for kanban mode rendering.
+            // Source of the issue: https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/views/basic/basic_view.js#L303 .
+            // See https://github.com/OCA/web/pull/1404#pullrequestreview-305813206 .
+            //  In the long term we should a way to handle kanban mode
+            // better (eg: a specific renderer).
+            if (attrs.widget === 'x2many_2d_matrix') {
+                attrs.mode = 'tree';
+            }
+            return this._super(viewType, field, attrs);
+        },
+    });
+
+});
--- a/web_widget_x2many_2d_matrix/README.rst
+++ b/web_widget_x2many_2d_matrix/README.rst
@@ -168,6 +168,8 @@ Known issues / Roadmap
 
 * Support extra invisible fields inside each cell.
 
+* Support kanban mode. Current behaviour forces list mode.
+
 Changelog
 =========
 
--- a/web_widget_x2many_2d_matrix/readme/ROADMAP.rst
+++ b/web_widget_x2many_2d_matrix/readme/ROADMAP.rst
@@ -12,3 +12,5 @@
   is merged.
 
 * Support extra invisible fields inside each cell.
+
+* Support kanban mode. Current behaviour forces list mode.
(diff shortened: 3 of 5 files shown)


pedrobaeza:
What was the problem?

hbrunn:
all of this is wrong - what we want to fix the original is an override of the *constructor* of the existing class that forces the list mode

Tardo:
@pedrobaeza  The previous PR changes XML in HTML way... this cause conflicts when try parse it after changes. 

@hbrunn I made changes to use "pure" JS... if you know a better way, please make a pr to improve it. Thanks!

pedrobaeza:
I suppose the problem is #1405. Then revert it the commit immediately in the repo, and then we continue in this PR.

simahawk:
hmm why do we hack form processing instead of changing the way the widget is initialized?
I guess this is what @hbrunn meant.

Tardo:
@simahawk  I tried, but I didn't get the expected behavior... i need some help here to do it in a better way.

simahawk:
what do you mean? What have you tried?

Tardo:
@simahawk I tried to change render mode in the widget class initialization... I'm missing something?

simahawk:
where and how? I mean: if you don't provide a piece of code to review how can we understand what have you tried and what went wrong? :man_shrugging:

Tardo:
If you are aware that I am doing it wrong, it is because you know how to do it right ... share your right way of doing it.
I can show you all my tries... but i don't see the benefit of it. For me this PR is the correct way to do it.

daramousk (commented):
Hi there and thank you for this!

Going through the issue, and the comments here is my take on this:

When the viewport width is reduced to less than 768px, we have this https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/views/basic/basic_view.js#L303
returning `kanban`, which means that the assumption `WidgetX2Many2dMatrix` will have `ListRenderer` as it's renderer will not be true.

Now, Odoo seems to be swapping lists for kanbans for small viewports due to kanban's `mobile_friendly: true` (verify this on quotations, open them with a 800px width and then with a 750 width, you will see that this is the behaviour, don't just scroll though because it does not seem to be dealing very well with this).

As to how this module can deal with this, perhaps we should not force rendering the listview and rather make sure that the `WidgetX2Many2dMatrix` can deal with it being rendered as `kanban`

With that being said, we need to get a mental picture of how this matrix should be rendered in kanban, and then the implementation should not be too hard.

Thanks.

simahawk:
> If you are aware that I am doing it wrong, it is because you know how to do it right

sorry, but here you are wrong. I don't know the right way to do it as you are not explaining where the problem comes from and how you tried to solve it before coming to current solution.
Please, don't give for granted that others knows what you think, what you investigated, what you've tried and what is your context. Also, don't get it personally: we are all here to try to improve what we have now -> thanks for your efforts so far anyhow :wink: 

Now, thanks to @daramousk we have a better picture and I agree with him: we should try to make it works smoothly and for doing so we have to decide how it should look like in kanban mode (we could add a specific renderer for it).

Tardo:
This PR it's a workaround to avoid view crashes: https://github.com/OCA/web/issues/1387#issuecomment-536479847

My original approach don't try to make the widget "responsive". Thanks for the comments @daramousk and @simahawk

web_widget_x2many_2d_matrix/static/src/js/2d_matrix_view.js
```diff
@@ -0,0 +1,19 @@
...
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define('web_widget_x2many_2d_matrix.X2Many2dMatrixView', function (require) {
+    "use strict";
+
+    var BasicView = require('web.BasicView');
+
+    BasicView.include({
+        _processField: function (viewType, field, attrs) {
+            // Force mode='list' in 'x2many_2d_matrix' widget
```
simahawk:
```js
// Workaround for kanban mode rendering.
// Source of the issue: https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/views/basic/basic_view.js#L303 .
// See https://github.com/OCA/web/pull/1404#pullrequestreview-305813206 .
// In the long term we should a way to handle kanban mode better (eg: a specific renderer).
```

simahawk (approved):

CasVissers-360ERP (approved):
LGTM, tested on EE iOS.

alexey-pelykh (approved):

pedrobaeza:
Let's merge this while we don't have a "kanban" 2D matrix view.

/ocabot merge minor.

PR #1407, opened 2019-10-15
[IMP] web_responsive: Adapt test to 13.0

Tardo:
Use 'with_user' instead of 'sudo'

Cc @tecnativa


pedrobaeza (approved):
Fast-tracking this

PR #1410, opened 2019-10-17
[12.0][MIG] web_listview_invert_selection

Martronic-SA:
Port web_listview_invert_selection to 12.0
set the version to 12
use of scss


Martronic-SA:
Wrong committer! Sorry

PR #1411, opened 2019-10-17
[12.0][MIG] web_listview_invert_selection

Martronic-SA:
migrate to 12,
pinned version,
swith to scss


Martronic-SA:
Is there a simple way to fix the @oca-clabot complain about user? (no github login found)
thanks

pedrobaeza:
That's because the email which you have committed is not registered in our DB linked to a GitHub user with a valid signed CLA.

Martronic-SA:
@pedrobaeza Do I have to delete and recreate th PR with the right committer?

pedrobaeza:
Or send another CLA with this email.

PR #1419, opened 2019-10-22
[12.0][MIG] web_widget_mermaid

janverb:


web_widget_mermaid/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "website": "https://github.com/OCA/web",
+    "depends": ['web'],
+    "data": [
+        "view/web_widget_mermaid_view.xml",
+    ],
+    "js": [
+        "static/lib/mermaid/mermaid.js",
+        "static/src/js/web_widget_mermaid.js",
+    ],
+    "qweb": [],
```
nikul-serpentcs: @janverb not needed

web_widget_mermaid/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
nikul-serpentcs: same here

web_widget_mermaid/__init__.py
```diff
@@ -0,0 +1 @@
+# -*- coding: utf-8 -*-
```
nikul-serpentcs: not needed
Add `License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)`

nikul-serpentcs (commented):
minor change

web_widget_mermaid/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Mermaid flowchart widget",
+    "category": "Web",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "version": "12.0.8.4.0",
```
daramousk, line 8: Since this is a migration, I believe we should start from `12.0.1.0.0`, or no? I see on the readme that you assosiate this with the version of mermaid you include here. Great idea but I am not sure if this is correct with the OCA guidelines on versioning. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#12version-numbers

web_widget_mermaid/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "license": "AGPL-3",
+    "summary": "Render mermaid markdown flowcharts",
+    "website": "https://github.com/OCA/web",
+    "depends": ['web'],
+    "data": [
+        "view/web_widget_mermaid_view.xml",
+    ],
+    "js": [
```
daramousk: I am not sure if this is needed, you already insert the widget and the library in your override of assets backend.

web_widget_mermaid/static/src/js/web_widget_mermaid.js
```diff
@@ -0,0 +1,100 @@
...
+        },
+        className: 'o_form_field_mermaid',
+        _renderReadonly: function() {
+            if (!this.value) {
+                return;
+            }
+            var config = _.extend(
+                {},
+                defaultConfig,
+                this.options
```
daramousk: `this.attrs.options`

daramousk (commented):
Hey! Great stuff, thanks for this.

I put some comments on this, also, what happens when the users enter a wrong input? You could use this https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/fields/abstract_field.js#L472

web_widget_mermaid/__manifest__.py
janverb: @hbrunn suggested it in #1400 - perhaps he has more thoughts?

daramousk: Ok then dismiss my comment here.

janverb:
When trying to display invalid input a `<pre>` with the error message is rendered instead:

image

I hadn't thought of showing an error message when saving. That seems friendlier.

I can make Odoo reject a change by throwing an error in `_parseValue`, but that doesn't let me show the actual error message. I suppose I could put a similar `<pre>` under the text box for that.

hbrunn (approved):
looks good, please squash your migration commits

hbrunn:
@janverb it would have been optimal to only squash your fixup commits, and leave the added functionality in a separate commit. But that's for next time, no need to disentangle this now

hbrunn:
PS: A nice future extension would be an option to turn on (or have it default on and allow to turn off) a live preview that's rendered where you now also render the error message. But no need to do now

tarteo (approved):
Works perfectly.

tarteo:
@janverb  Can you rebase commits, please.

janverb:
Done.

PR #1421, opened 2019-10-28
web_tree_image_tooltip: add image width and height limit in tooltip

DonatasV:
Before this PR, size of images in tooltips are not limited. If image is bigger than the screen, it cuts off like in example screenshot below.
The tooltip with large image might also cover up the thumbnail image, where the mouse is hovering, which makes the browser think, that the mouse moved away and tooltip should not be shown anymore. The tooltip gets removed, the mouse starts hovering the thumbnail again, which in turn displays the tooltip again. This results in flickering loop of tooltip appearing and disappearing.

Before:
image

After:
image


gaikaz (approved):

fuentes73 (approved):
Tested and works

daramousk (approved):

PR #1422, opened 2019-10-28
[12.0][ADD] web_widget_url_translatable

TDu:
This module extends the URL widget, so it works the same than other
widget with translatable fields.
In edition mode the button that allows to open the tranlation view
is displayed.

### DIFF
--- a/web_widget_url_translatable/__manifest__.py
+++ b/web_widget_url_translatable/__manifest__.py
@@ -1,0 +1,14 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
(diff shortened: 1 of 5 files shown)


i-vyshnevska (approved):

simahawk (approved):
LG. Would be nice to see a PR opened against Odoo master as the feature seems pretty easy and nice.

simahawk:
`USAGE.rst` is empty, should be removed

daramousk (commented):
Hello and thanks for this,

Can you please give me an example of a use case here for this widget?

Thanks

TDu:
Hello @daramousk, this is used on a specific Odoo instance that manages information that is used to generate web pages/api datas in multiple language. In this information there is urls that are different depending of the language.

PR #1423, opened 2019-11-01
[FIX] remove dependency to inexisting module 'web_graph' in V10

legalsylvain:
Hi,

trivial fix here.

I'm currently migrate from 8.0 to 10.0 with a DB with ``web_graph_improved`` installed. 

In V10, the update is failing because ``web_graph_improved`` is currently depending on ``web_graph`` module that doesn't exists anymore in V10.

Note that I'm changing an **uninstallable** module.


web_graph_improved/__manifest__.py
```diff
@@ -27,7 +27,7 @@
     'author': "Leonardo Donelli,Odoo Community Association (OCA)",
     'license': 'AGPL-3',
     'website': 'http://learts.glaucus.in',
```
nikul-serpentcs: @legalsylvain URL should be `https://github.com/OCA/web`

legalsylvain: @nikul-serpentcs, this module is uninstallable. The code will not work, and the module is not available in the app store.
So, this is not relevant to fix the code, and it is not the objective of this PR.

thanks.

pedrobaeza:
@legalsylvain this is a bit of a hack for avoiding your specific dependency problem during OpenUpgrade, but I think we should go for a general solution. One question: is this happening because the module is still in your addons path (although uninstallable)? Have you tried if you find the same problem removing it? If it's solved, then what we should do is to totally removed the module instead from this branch (that's one of the reasons why I pushed in the past for the new empty branch approach).

legalsylvain:
hi @pedrobaeza,

> Have you tried if you find the same problem removing it? If it's solved, then what we should do is to totally removed the module instead from this branch

No I didn't, but for sure, it will work.

> If it's solved, then what we should do is to totally removed the module instead from this branch.

I can change this PR, removing entirely the module, if you think it's the better approach.

> (that's one of the reasons why I pushed in the past for the new empty branch approach)

This is the best approach, indeed.

pedrobaeza:
Let's go for the removal!

legalsylvain:
done ! thanks for your advice.

pedrobaeza (approved):

PR #1428, opened 2019-11-06
12.0 mig web dashboard tile

legalsylvain:
- Migration of ``web_dashboard_tile`` to V12.
- Only the last commit has be reviewed : https://github.com/OCA/web/pull/1428/commits/698433e0b8bf45de2b2b87e9e8680e72139d7b6c

CC : co original author : @ivantodorovich 

CC : @quentinDupont, 

## Description

Adds a dashboard where you can configure tiles from any view and add them as short cut.

By default, the tile displays items count of a given model restricted to a given domain.

Optionally, the tile can display the result of a function on a field.

- Function is one of sum, avg, min, max or median.
- Field must be integer or float.

Tile can be:

- Displayed only for a user.
- Global for all users.
- Restricted to some groups.

**Note**: The tile will be hidden if the current user doesn't have access to the given model.

### Configuration

First, you have to create tile categories.

* Go to "Dashboards > Settings > Dashboard Categories"
* Click on Create
* Set a name, and save.

Odoo menu and action are automatically created. You should refresh your browser to see new menu items.

tile_category_form

Then you can create tiles.

* go to "Dashboards > Settings > Dashboard Tiles"
* create a new tile, set a name, a category and a model.
* You can optionally define colors, domain a specific action to use.
* Setting a user, or a group in "Security" tab will restrict the display of the tile.

tile_tile_form

You can optionally define a secondary value, for that purpose :

* Select a field, a function to apply.
* You can define a specific format. (.format() python syntax)

tile_tile_form_secondary_value

## Usage

* Go to "Dashboard > Overview" and select a category
* The tile configured is displayed with the up to date count and average values of the selected domain.

tile_tile_kanban

* By clicking on the item, you'll navigate to the tree view of the according model.

tile_tile_2_tree_view

**Note**

When you are in a tree view, with a domain, you can save it in the favorite menu, but the configuration is limited.

favorite_menu_create_tile

favorite_menu_create_tile_result

**Migration Changes**

- Rename "dashboard tile" to "overview dashboard", that make more sense, and is more easy to translate.
- refactor tile category
- improve description
- add legalsylvain as maintainer
- update code and translation


web_dashboard_tile/__manifest__.py
```diff
@@ -0,0 +1,33 @@
...
+    "GRAP, "
+    "Odoo Community Association (OCA)",
+    "maintainers": ["legalsylvain"],
+    "category": "web",
+    "license": "AGPL-3",
+    "contributors": [
+        "initOS GmbH & Co. KG",
+        "GRAP",
+        "Iván Todorovich <ivan.todorovich@gmail.com>",
+    ],
```
ivantodorovich: Is this key even used? :thinking: I guess we should move this to `CONTRIBUTORS.rst`

web_dashboard_tile/models/tile_category.py
```diff
@@ -0,0 +1,105 @@
...
+        return {
+            'name': self.name,
+            'res_model': 'tile.tile',
+            'type': 'ir.actions.act_window',
+            'view_mode': 'kanban',
+            'domain': "["
+            "('hidden', '=', False),"
+            "'|', ('user_id', '=', False), ('user_id', '=', uid),"
+            "('category_id', '=', %d)"
+            "]" % self.id,
```
ivantodorovich:
```suggestion
            'domain': f"""
                [
                    ('hidden', '=', False),
                    '|', ('user_id', '=', False), ('user_id', '=', uid),
                    ('category_id', '=', {self.id})
                ]
            """,
```

This is probably more readable :)

web_dashboard_tile/models/tile_category.py
```diff
@@ -0,0 +1,105 @@
...
+                    category._prepare_action())
+            if not category.menu_id:
+                category.menu_id = IrUiMenu.create(category._prepare_menu())
+
+    def _delete_ui(self):
+        for category in self:
+            if category.menu_id:
+                category.menu_id.unlink()
+            if category.action_id:
+                category.action_id.unlink()
```
ivantodorovich, line 80:
```suggestion
        self.menu_id.unlink()
        self.action_id.unlink()
```

ivantodorovich (approved):
quick review but LGTM :+1:

web_dashboard_tile/models/tile_category.py
legalsylvain: indeed ! thanks !

web_dashboard_tile/models/tile_category.py
legalsylvain: is ``unlink`` working for 0 items ? i added test in V8 because there was some trouble if no action or menu was present.
maybe it's an obsolete problem. (don't tested).

ivantodorovich: Yes it works :) `unlink` on an empty recordset simply doesn't do anything

https://github.com/odoo/odoo/blob/c164a46500e35b0120dedbcfe8e3faa2001461bf/odoo/models.py#L3158-L3159

web_dashboard_tile/__manifest__.py
legalsylvain: I put values in ``authors`` keys.

web_dashboard_tile/models/tile_category.py
legalsylvain: thanks for your answer.

web_dashboard_tile/models/tile_category.py
legalsylvain: FIY, I removed the "f" syntax, that was making travis failing. Otherwise LGTM.

quentinDupont (approved):
Functionnal review :heavy_check_mark:
We use this module in production server since one year and it is working greeaat

marcelsavegnago (approved):
LGTM

PR #1429, opened 2019-11-07
web_company_color: change odoo color by variables

DonatasV:
Before this PR, the colors are changes only on navbar and with custom scss. 
After this PR is changed odoo variables, so color changes in all system

### DIFF
--- a/web_company_color/hooks.py
+++ b/web_company_color/hooks.py
@@ -1,13 +1,21 @@
 # Copyright 2019 Alexandre Díaz <dev@redneboa.es>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import base64
 from odoo import api, SUPERUSER_ID
 from .models.res_company import URL_BASE
 
-
 def uninstall_hook(cr, registry):
     env = api.Environment(cr, SUPERUSER_ID, {})
+    values = """
+        $o-community-color: #7C7BAD; 
+        $o-brand-odoo: $o-community-color;
+        $o-brand-primary: $o-community-color;
+        $o-main-text-color: #4c4c4c;
+    """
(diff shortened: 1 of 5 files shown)


pedrobaeza:
So does it mean that reports are also affected by this color change?

DonatasV:
I just checked and yes, reports are also affected by this color

pedrobaeza:
Then I don't think it's a good idea.

gaikaz:
@pedrobaeza Before finding how well changing the primary color variable works for changing all the interface colors, @DonatasV also had made an extended version of this module, where it simply changed all places one by one overriding them, the way the module does it for the navbar. But the solution doesn't feel very elegant.

Now I'm wondering, do you agree with the general idea of extending this module to colorize more of the UI?
Assuming, it doesn't affect the reports (though honestly, the standard reports look pretty nice colored like that).

pedrobaeza:
Well, I'm not really comfortable with something that modifies the visual ergonomics of your system through a user color pick. One thing is to modify the navbar, which is for highlighting something, and another that touches a lot of things.

Why don't you see how Odoo has implemented this in v13, that offers the same kind of feature?

gaikaz:
Changing the SCSS variable like this PR does works really well. The color pick field should probably be considered technical and only available for admins, yet it doesn't require much designer knowledge to make the whole interface look and feel more personalized for the company. Say company's branding color is green, just by picking similarly shaded green to Odoo's purple, the result could be:
image
This screenshot is from PR's runbot instance. If by any chance this is interesting, please do play around with it. 

Just an idea, but what if the module would try and suggest a more fitting shade of the color chosen?

I will do check v13 implementation. Didn't know it now has something like that. :+1:

PR #1430, opened 2019-11-07
[ADD] web_widget_child_selector

etobella:
This module allows to use a new widget for many2one fields that are made hierarchically.
Now, you can navigate easily between the hierarchy.

When we are using readonly, the behaviour is similar than Many2one, but on edition
we are able to select the parents and childs in order to navigate properly.

imagen

Usage
------
Use child_selection widget on a field.
If necessary we can use child_selection_field on options in order to define
which field we will show on edition only.
For example, name instead of display_name.


web_widget_child_selector/models/base.py
```diff
@@ -0,0 +1,38 @@
...
+        if not hasattr(self, self._parent_name):
+            return False
+        parent = getattr(self, self._parent_name)
+        if not parent:
+            return False
+        return parent.name_get()[0]
+
+    def _get_record_parents(self, field):
+        if not self:
+            return [(False, _('Root'))]
```
jarroyomorales: I guess it is used to be able to set the field to False when using this widget but I think it can cause confusion since the word Root doesnt have anything to do with setting a field to false

web_widget_child_selector/static/src/xml/widget_child_selector.xml
```diff
@@ -0,0 +1,36 @@
...
+        </t>
+        <div t-if="widget.mode === 'edit'" class="o_field_widget">
+            <span class="o_input_value"/>
+            <div class="o_input_dropdown"/>
+        </div>
+    </t>
+    <t t-name="FieldChildSelectorChild">
+        <div class="row">
+            <div t-foreach="childs" t-as="key" class="col-xs-12">
+                <button t-att-data-id="childs[key][0]" t-att-data-index="key"
```
jarroyomorales: Maybe a group of radio buttons would be nicer

jarroyomorales (changes requested):
Just a couple of comments. It would be nice if you could fix travis warnings regarding js formatting and also add some tests to the python part at least to ensure js gets the correct information.

web_widget_child_selector/static/src/xml/widget_child_selector.xml
jarroyomorales: imagen

web_widget_child_selector/models/base.py
jarroyomorales: Maybe a good option is to add a X icon next to the name when in edit mode that sets the field to false and we can get rid off that 'Root' element.

etobella:
Comments attended @jarroyomorales

web_widget_child_selector/static/src/js/widget_child_selector.js
```diff
@@ -0,0 +1,100 @@
...
+                });
+                self.$input_dropdown.append(qweb.render('FieldChildSelectorChild', {
+                    'childs': self.childs,
+                }));
+                self.$input_value.append(qweb.render('FieldChildSelectorParent', {
+                    'parents': self.parents,
+                }));
+            });
+        },
+        _onChildSelectionParent: function(event) {
```
jarroyomorales: Where is this function used?

web_widget_child_selector/static/src/js/widget_child_selector.js
```diff
@@ -0,0 +1,100 @@
...
+            this.parents = {};
+            this.$input_dropdown.empty();
+            this.$input_value.empty();
+            var resources = [];
+            if (this.value.res_id)
+                resources = [this.value.res_id]
+            this._rpc({
+                model: this.field.relation,
+                method: 'get_record_direct_childs_parents',
+                args: [ resources, this.nodeOptions],
```
jarroyomorales: Here you should probably pass the domain as a parameter so python can filter which results returns

etobella: Comment attended

web_widget_child_selector/static/src/js/widget_child_selector.js
etobella: It was a previous function not used anymore

jarroyomorales (approved):
LGTM now :+1: :green_apple:! Could you please squash the commits?

tonygalmiche (approved):
I tested on my computer with this : 
`<field name="categ_id" string="Product Category" widget="child_selector"/>`

image

and 
`<field name="categ_id" string="Product Category" widget="child_selector" options="{&quot;child_selection_field&quot;: &quot;name&quot;}"/>`




And everything works fine

Thanks for this new widget

etobella:
This is already on 12.0 @OCA/web-maintainers , How should we proceed?

PR #1438, opened 2019-11-17
[13.0][MIG] web_tree_many2one_clickable

Daemo00:
standard migration


lfreeke (approved):
Functional test :+1: 

Thank you

pedrobaeza (dismissed):
I have tried on runbot and I don't get it working: no button when passing the mouse over the many2one like in 12.0:

Peek 18-11-2019 18-18

Daemo00:
Thanks @pedrobaeza and @lfreeke for the review! I can confirm that it is not working in the SO lines but it works for instance in the invoice lines.. Investigating deeper :mag:

lfreeke:
> Thanks @pedrobaeza and @lfreeke for the review! I can confirm that it is not working in the SO lines but it works for instance in the invoice lines.. Investigating deeper

It is rather odd because everywhere else it is working.

pedrobaeza:
Hehe, I went directly to the only case it's failing by chance!

hbrunn:
that's because of https://github.com/OCA/OCB/blob/13.0/addons/sale/views/sale_views.xml#L332 defined in https://github.com/OCA/OCB/blob/13.0/addons/account/static/src/js/section_and_note_fields_backend.js - I think it's out of scope for this addon to fix those

pedrobaeza:
thanks for the pointer, @hbrunn. I don't think so, unless that widget is defined in `sale` module. If it's in `web` module, we should inherit that widget too for adding the corresponding stuff for making it work.

hbrunn:
it's in `account` (see link above). But I wonder why this doesn't pick up the changes in the prototype, theoretically, this should just work

pedrobaeza:
Well, if it can be done compatible without requiring glue module, then better to have it. But if not, at least add it to the ROADMAP for being aware.

Daemo00:
Hi, what I discovered is that what causes the strange behavior is actually the `product_configurator` widget on `product_id` that extends `relationalFields.FieldMany2One`, but this module overrides the methods of `ListFieldMany2One`.
I verified this by changing https://github.com/odoo/odoo/blob/22a11a6f4d0b08a56712fafd0f87a2e7bb5b00ec/addons/sale/static/src/js/product_configurator_widget.js#L18 in order to let it extend `relationalFields.ListFieldMany2One`.

With this change, the `>>` button is still not shown on field `product_id` because condition https://github.com/OCA/web/pull/1438/files#diff-f06f62892caad3befc7bded8af2e3c50R33 is not satisfied since the field `product_id` has `noOpen = true`, but I think this is anyway the expected behavior for this module.

I do not completely understand the `product_configurator` widget, would it make sense to propose a change in Odoo in order to let it inherit from `relationalFields.ListFieldMany2One`?
This is the first solution that I can see because as far as I know this module is explicitly intended for lists so it can't `include` `relationalFields.FieldMany2One`.

pedrobaeza:
Well, if no_open is defined, then it doesn't matter about the inheritance, so for me we can continue. I can't say about JS inheritance though.

Daemo00:
Yep I can confirm that field `product_id` has `no_open`: https://github.com/odoo/odoo/blob/1b95bfe202576583d6458c6bd1a707dd988e6d95/addons/sale/views/sale_views.xml#L445

Noticing that in the beginning would have saved a few headaches about JS debugging :)

pedrobaeza:
@hbrunn do you approve this as is then or do you think it's worth to mention something in the ROADMAP?

Daemo00:
Anyway I'll try to submit a PR with the mentioned change to JS inheritance of the `product_configurator` because I've noticed that it is only used in there, we'll see what (or better, *if*) they respond.

hbrunn:
I'm not sure the proposal to change the base class will be successful, because then it would be awkward to use the widget in some other view type. Like kanban, this is what lists degrade to nowadays in responsive mode. Ah, this is something you should check: How does the addon perform in responsive mode? I'd expect we won't have a link, so you'll probably have to override the kanban version too.

We'll run into this kind of situations repeatedly in the future due to the way widget classes are selected by the new framework (which is super well done btw in my opinion). I think you can solve this by adding some function `_tree_many2one_clickable_setup_list` to `FieldMany2One`, and override `_renderReadonly` there to call this function if we're rendered by a list or kanban view. In the override of `ListFieldMany2One`, call the function unconditionally. Then if some widget doesn't do something wrong or intentionally very different, it should work.

In this specific case the widget probably relies on https://github.com/odoo/odoo/blob/22a11a6f4d0b08a56712fafd0f87a2e7bb5b00ec/addons/sale/static/src/js/product_configurator_widget.js#L35 being called when rendering in read mode, you'll have to see how much trouble this causes.

Another option would be to iterate through the fields registry and depending on some conditions (inherits from `FieldMany2One` but not from `List...` and doesn't have it's own `List...`) just patch the function into the objects there, but I feel this might be a path to pain and misery.

Daemo00:
> How does the addon perform in responsive mode? I'd expect we won't have a link

Exactly (invoice view): 
image
But maybe this addon named `web_tree_...` should only focus on tree views and leave kanban views to other dedicated addons..

This said, big thanks for the proposed solutions :) that comes from some serious problem solving skills

pedrobaeza:
As it is planned (appearing when you hover over the field), I don't think we should provide an alternative on responsive mode (where there's no hover as you are on a mobile). For me it's OK to limit to current scope.

web_tree_many2one_clickable/README.rst
```diff
@@ -0,0 +1,95 @@
...
+=====
+
+Put the mouse pointer over a many2one cell and click the button.
+
+.. image:: https://raw.githubusercontent.com/OCA/web/13.0/web_tree_many2one_clickable/static/img/clickable.gif
+
+Known issues / Roadmap
+======================
+
+This widget is currently not working on the product field in the lines tree of the sale order form, see https://github.com/OCA/web/pull/143 for further details.
```
pedrobaeza: Incorrect PR number

Daemo00: :man_facepalming: thanks, fixed

PR #1440, opened 2019-11-19
10.0-[FIX]-web_ir_actions_act_window_message-fixes #864

### LINKED ISSUE #864: [10.0] web_ir_actions_act_window_message calling non-existing recursive_reload
The close button tries to call self.inner_widget.active_view.controller.recursive_reload(); but this method only exists on form views.

We have a "More" action on a list view were we would return a message, but using the web_ir_actions_act_window_message module results in an error when the user clicks the Close button:

Uncaught TypeError: self.inner_widget.active_view.controller.recursive_reload is not a function
http://odoo.local:1111/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js:28
Traceback:
TypeError: self.inner_widget.active_view.controller.recursive_reload is not a function
    at Class.click (http://odoo.local:1111/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js:28:41)
    at HTMLButtonElement.<anonymous> (http://odoo.local:1111/web/static/src/js/framework/dialog.js:89:41)
    at HTMLButtonElement.dispatch (http://odoo.local:1111/web/static/lib/jquery/jquery.js:4641:58)
    at HTMLButtonElement.elemData.handle (http://odoo.local:1111/web/static/lib/jquery/jquery.js:4309:63)

The simple solution could be to test for the existence of the method before calling it, but sometimes a reload might also be needed on a list view...

https://github.com/OCA/web/blob/10.0/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js#L28

ntsirintanis:
This PR takes care of [#864](https://github.com/OCA/web/issues/864)

### DIFF
--- a/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
+++ b/web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_message.js
@@ -24,8 +24,9 @@ odoo.define('web.web_ir_actions_act_window_message', function(require)
                     text: action.close_button_title || _t('Close'),
                     click: function() {
                         // refresh the view before closing the dialog
-                        self.inner_widget.active_view
-                            .controller.recursive_reload();
+                        var controller = self.inner_widget.active_view.controller,
+                            reload = controller.recursive_reload || controller.reload;
+                        reload.apply(controller);
                         dialog.close()
                     },
                     classes: 'btn-default',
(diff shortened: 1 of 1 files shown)


ntsirintanis:
@hbrunn

hbrunn (approved):

PR #1442, opened 2019-11-20
[FIX] web_responsive: modify action handling

Tardo:
Previous this commit, AppMenu doesn't hide in some situations.

**Steps to reproduce:** (sometimes works, sometimes doesn't)
From Discuss App (refresh) open AppMenu and search some menu, select the second in the list and press enter.
_The action is launched correctly, but the AppMenu doesn't hide._

cc @tecnativa TT20688

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -442,22 +442,24 @@ odoo.define('web_responsive', function (require) {
         * The executed action
         */
         _hideMenusByAction: function (action) {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It worths to mention steps to reproduce the problem previous to this commit (and in the commit message itself apart from here).

Tardo:
It is difficult to reproduce, but done!

yajo (approved):

pedrobaeza (approved):

PR #1447, opened 2019-11-22
[12.0][ADD] web_widget_map

AdriaGForgeFlow:
This module add the possibility to insert a map into Odoo form views using the OpenLayers library.

image

cc ~ @Eficent


pedrobaeza:
Shouldn't this be in OCA/geospatial?

And there's something similar in https://github.com/OCA/geospatial/tree/12.0/web_view_google_map

AdriaGForgeFlow:
> Shouldn't this be in OCA/geospatial?
> 
> And there's something similar in https://github.com/OCA/geospatial/tree/12.0/web_view_google_map

Well, it's different. In geospatial there is a map view. This is a widget that can be included inside an Odoo form.

I don't know where this belongs, since it's a widget I thought it should be here in OCA/web. What do you think?

tarteo:
> Well, it's different. In geospatial there is a map view. This is a widget that can be included inside an Odoo form.

base_geoengine has a map form widget, which is also editable (users are able to draw polygons, points, etc).

legalsylvain:
> Shouldn't this be in OCA/geospatial?

My preference too. but I don't have a clear point of view. I think that @OCA/geospatial-maintainers should give a good point of view about that point.

> And there's something similar in https://github.com/OCA/geospatial/tree/12.0/web_view_google_map

This module is about OpenStreetMap, not google Map. (that is great, by the way ;-))

ps-tubtim (approved):

web_widget_map/README.rst
```diff
@@ -0,0 +1,99 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To insert a Bokeh chart in a view proceed as follows:
```
legalsylvain, line 39: Bokeh chart ? I think there is an error in a copy / paste. Don't you think ?

web_widget_map/README.rst
```diff
@@ -0,0 +1,99 @@
...
+.. figure:: https://raw.githubusercontent.com/OCA/web/12.0/web_widget_map/static/src/img/view_map_button.png
+    :align: center
+    :width: 600 px
+
+.. figure:: https://raw.githubusercontent.com/OCA/web/12.0/web_widget_map/static/src/img/view_map.png
+    :align: center
+    :width: 600 px
+
+By default the center of the map is set to latitude 0 and longitude 0.
+Refer to the module hr_attendance_geolocation_map to see an example on how
```
legalsylvain, line 58: greats to have an exemple, without refering on a another module.
or at least, set an url to find easily this module. (https://github.com/OCA/hr-attendance)

regards.

web_widget_map/static/src/js/web_widget_map.js
```diff
@@ -0,0 +1,85 @@
...
+        },
+        init: function () {
+            this._super.apply(this, arguments);
+            this.mapLayers = [new ol.layer.Tile({
+                source: new ol.source.OSM()
+            })];
+            this.centerMap = [0, 0];
+            this.markers = [];
+        },
+        _initializeMap: function () {
```
legalsylvain, line 20: I'm not sure about the design. 
you are calling _initializeMap and in the hr module, it makes an rpc call.
https://github.com/OCA/hr/pull/731/files#diff-4f79c3ed43c3d1ec84dd1097f991c34aR17

but if you have two modules that depends on this widget, I think that the call will be done for both model, and generate errors. 
I think that refactoring is needed. don't you think ?

legalsylvain (changes requested):
Question about design inline.

regards.

yvaucher:
It's great to see some more GIS initiative in Odoo.

I would be in favor to have this in OCA/geospatial

It is quite similar to the googlemap modules in the way that it doesn't need postgis in opposition to geoengine module. And + :100: for using openstreetmap.

I see there is a demand for GIS modules that are simpler and with less features than geoengine.
IMO both are suited for the OCA/geospatial repository.

PR #1448, opened 2019-11-22
[FIX] web_widget_x2many_2d_matrix: Wrong widget

fuentes73:
In previous versions `_setSubViewLimit` was defined at AbstractView widget but in v12 is defined at FormView widget

**Previous**
Captura de pantalla 2019-11-22 a las 17 56 39

**Now**
Captura de pantalla 2019-11-22 a las 18 00 14


pedrobaeza:
What is the difference?

fuentes73:
> What is the difference?

The difference is than without this patch the function isn't fired as you can see in the first screenshot

pedrobaeza (approved):

Abranes (approved):

PR #1450, opened 2019-11-26
[11.0][ADD] web_refresher

josueBulle:
Adds a button next to the pager in list views. It updates the displayed page of the list.

refresh


web_refresher/__manifest__.py
```diff
@@ -0,0 +1,42 @@
+##############################################################################
```
tarteo: Can you remove the banner, please.

tarteo (changes requested):
One remark otherwise :+1: Tested it on Runbot works perfectly

tarteo (approved):
Thanks!

ecino (approved):
:+1:

PR #1454, opened 2019-12-02
[11.0][IMP] web_tree_dynamic_colored_field: more options

jarroyomorales:
* Add attribute ``bg_color_field`` on the tree element's ``colors`` to use as background-color.
* In the tree view declaration, put ``options='{"bg_color_field": "color_field"}`` attribute in the ``field`` tag to have column's background colored in color_field's value.

@etobella @damdam-s @hbrunn @guewen


hbrunn:
we have https://github.com/OCA/web/pull/1377 already, can you better forward port this?

jarroyomorales:
@hbrunn This module is very different from 10 to 11, and I implemented background-color the same way this module was migrated, so I dont see how forward-porting #1377 would be any different from this. Of course I can change the attribute's name so it is the same in both versions, but do you need something else?

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -61,6 +81,8 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
             }
             this.applyColorizeHelper($td, nodeOptions, node, 'fg_color', 'color', ctx);
             this.applyColorizeHelper($td, nodeOptions, node, 'bg_color', 'background-color', ctx);
+            this.applyDirectColorizeHelper($td, nodeOptions, node, 'fg_color_field', 'color', ctx);
```
hbrunn, line 108: this would be `color_field`, right?

hbrunn (commented):
yes, just the name please. The code is totally different indeed. But then if I understand the diff correctly, the foreground color never really worked after the migration? It might be better to indeed then make everything uniform as you already did in the code, `bg_color`, `fg_color`, `bg_color_field`, `fg_color_field`, I think this will have least impact. Please update the documentation then accordingly.

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
etobella: No, this is a color_field selected for the column and line (background_color is for all the items of the line).
Sometimes it is hard to create the specifications on a single line and you would like to compute it with python. It is used inside options of the field.

jarroyomorales:
@hbrunn I tested the module and it is all working in 11.0. Do we keep it this way then? I already updated the readme.

hbrunn (approved):
let's keep it the way it is, yes

etobella (approved):

jarroyomorales:
This last change allows this module to work with trees that are groupby by default, as it wasnt working properly in previous versions.

etobella:
@hbrunn

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -10,21 +10,37 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
          * @override
          */
-        _renderBody: function () {
+        _renderView: function () {
             if (this.arch.attrs.colors) {
-                var colorAttr = this.arch.attrs.colors.split(';')
-                    .filter(color => color.trim().startsWith('color_field'));
+                var colors_attr = this.arch.attrs.colors.split(';')
+                var colorAttr = colors_attr.filter(
+                    attr => attr.trim().startsWith('color_field')
```
Tardo: Arrow functions are not supported in IE
https://www.odoo.com/documentation/11.0/setup/deploy.html#supported-browsers
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -10,21 +10,37 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
+                );
+                var backgroundColorAttr = colors_attr.filter(
+                    attr => attr.trim().startsWith('background_color_field')
```
Tardo: Arrow functions are not supported in IE
https://www.odoo.com/documentation/11.0/setup/deploy.html#supported-browsers
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

Tardo (changes requested):
Thanks for the efforts!

jarroyomorales:
@Tardo Changes done! Tested on IE11 and worked. Thank you!

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -10,21 +10,41 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
          *
          * @override
          */
-        _renderBody: function () {
+        _renderView: function () {
             if (this.arch.attrs.colors) {
-                var colorAttr = this.arch.attrs.colors.split(';')
-                    .filter(color => color.trim().startsWith('color_field'));
+                var colors_attr = this.arch.attrs.colors.split(';')
```
Tardo: It's not mandatory for the language.... but in OCA we still using semicolon.
```suggestion
                var colors_attr = this.arch.attrs.colors.split(';');
```

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -10,21 +10,41 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
+                var colorAttr = colors_attr.filter(
+                    function(attr) {
+                        return attr.trim().lastIndexOf('color_field', 0) === 0
```
Tardo:
```suggestion
                        return attr.trim().lastIndexOf('color_field', 0) === 0;
```

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -10,21 +10,41 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
+                    }
+                );
+                var backgroundColorAttr = colors_attr.filter(
+                    function(attr) {
+                        return attr.trim().lastIndexOf('background_color_field', 0) === 0
```
Tardo:
```suggestion
                        return attr.trim().lastIndexOf('background_color_field', 0) === 0;
```

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,39 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
         _renderBodyCell: function (record, node, colIndex, options) {
             var $td = this._super.apply(this, arguments);
             var ctx = this.getEvalContext(record);
             this.applyColorize($td, record, node, ctx);
             return $td;
         },
 
+        _setDecorationClasses: function (record, $tr) {
+            this._super.apply(this, arguments);
+            this.applyAllColorize($tr, record)
```
Tardo:
```suggestion
            this.applyAllColorize($tr, record);
```

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,39 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
             var treeColor = record.data[this.colorField];
             if (treeColor) {
-                $td.css('color', treeColor);
+                $tr.css('color', treeColor);
             }
+            var treeBackgroundColor = record.data[this.backgroundColorField];
+            console.log(this.backgroundColorField)
+            console.log(treeBackgroundColor)
+            if (treeBackgroundColor) {
+                $tr.css('background-color', treeBackgroundColor)
```
Tardo:
```suggestion
                $tr.css('background-color', treeBackgroundColor);
```

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,39 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
-        applyColorize: function ($td, record, node, ctx) {
+        applyAllColorize: function ($tr, record) {
             // safely resolve value of `color_field` given in <tree>
             var treeColor = record.data[this.colorField];
             if (treeColor) {
-                $td.css('color', treeColor);
+                $tr.css('color', treeColor);
             }
+            var treeBackgroundColor = record.data[this.backgroundColorField];
+            console.log(this.backgroundColorField)
```
Tardo: :fire:

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,39 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
+            console.log(treeBackgroundColor)
```
Tardo: :fire:

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -88,6 +124,12 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
             }
         },
 
+        applyDirectColorizeHelper: function ($td, nodeOptions, node, nodeAttribute, cssAttribute, ctx) {
+            var field = nodeOptions[nodeAttribute];
+            if (field && ctx[field])
+                $td.css(cssAttribute, ctx[field]);
```
Tardo:
```suggestion
            if (field && ctx[field]) {
                $td.css(cssAttribute, ctx[field]);
            }
```

Tardo (changes requested):
Thanks for the work! only few little comments....

jarroyomorales:
@Tardo Done!

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,37 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
          */
+
         _renderBodyCell: function (record, node, colIndex, options) {
             var $td = this._super.apply(this, arguments);
             var ctx = this.getEvalContext(record);
             this.applyColorize($td, record, node, ctx);
             return $td;
         },
 
+        _setDecorationClasses: function (record, $tr) {
```
Tardo, line 69: Please, add docstring

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,37 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
+            this._super.apply(this, arguments);
+            this.applyAllColorize($tr, record);
+        },
+
         /**
          * Colorize the current cell depending on expressions provided.
          *
          * @param {Query Node} $td a <td> tag inside a table representing a list view
```
Tardo: Update this docscript

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -33,25 +53,37 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
...
             if (treeColor) {
-                $td.css('color', treeColor);
+                $tr.css('color', treeColor);
             }
+            var treeBackgroundColor = record.data[this.backgroundColorField];
+            if (treeBackgroundColor) {
+                $tr.css('background-color', treeBackgroundColor);
+            }
+        },
+        applyColorize: function ($td, record, node, ctx) {
```
Tardo, line 98: Add docstring

web_tree_dynamic_colored_field/static/src/js/web_tree_dynamic_colored_field.js
```diff
@@ -88,6 +122,13 @@ odoo.define('web_tree_dynamic_colored_field', function (require) {
             }
         },
 
+        applyDirectColorizeHelper: function ($td, nodeOptions, node, nodeAttribute, cssAttribute, ctx) {
```
Tardo, line 144: Add docstring

Tardo (changes requested):
Looks good! Please, update the docstrings.

etobella:
@Tardo I think all comments have been attended

PR #624, opened 2017-05-16
Initial commit, port of help_popup to 10.0

daramousk:
Ported help_popup to v10. Some changes where needed on the template and on the js. Let me know what you think.

PR #1460, opened 2019-12-11
[ADD] web_external_help: Add web_external_help module

hiren-serpentcs:
The Idea behind this module is to allow user to give detailed description about a particular field or a Model. It can also redirect user to a web page.

Fields Help

Model Help


legalsylvain:
Hi @hiren-serpentcs. This module looks nice ! 

In the meantime, I think it is similar to the actual V8/10 OCA modules help_online / help_popup. Did you take a look on them ?
Do you think it could be possible to have a unique module that handle the features ? (display on model and field + possibililty to edit if member of a group, etc...)

- V8 [help_popup](https://github.com/OCA/web/tree/8.0/help_popup) (+ [V10 PR](https://github.com/OCA/web/pull/624))
- V10 [help_online](https://github.com/OCA/web/tree/10.0/help_online) 

Interesting discussion between help_online and help_popup : https://github.com/OCA/web/pull/417

CC authors : 
- @bealdav (help_popup)
- @lmignon, @Cedric-Pigeon (help_online)

Thanks ! 

Partial review : 
- about the name, I don't understand exactly what is "external". except that it is possible to add an url...
- You can use now readme syntax and avoid to write the manifest file. See [OCA guidelines.](https://github.com/OCA/maintainer-tools/tree/master/template/module/readme)

Swapnesh-SerpentCS:
@max3903  Can you please add your thoughts here?

max3903:
@b-kannan Can you help us here? It looks like we could have used/migrated/built on top of existing module.

PR #1104, opened 2018-11-19
[12.0][MIG] web_ir_actions_act_view_reload

alexey-pelykh:

PR #1462, opened 2019-12-16
[MIG] web_ir_actions_act_view_reload: Migration to 13.0

guewen:
Port #1104


web_ir_actions_act_view_reload/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+# Copyright 2018 Brainbean Apps
+# License LGPLv3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
+
+{
+    "name": "Web Actions View Reload",
+    "summary": "Enables reload of the current view via ActionManager",
+    "category": "Web",
+    "version": "13.0.1.0.0",
+    "license": "LGPL-3",
+    "author": "Modoolar, " "Brainbean Apps, " "Odoo Community Association (OCA)",
```
tarteo: This doesn't seem right. Does this work?

web_ir_actions_act_view_reload/static/src/js/web_ir_actions_act_view_reload.js
```diff
@@ -0,0 +1,38 @@
+// Copyright 2017 - 2018 Modoolar <info@modoolar.com>
+// Copyright 2018 Modoolar <info@modoolar.com>
+// License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
tarteo: The module uses LGPL license according to the module manifest. I think OCA always uses AGPL.

tarteo (dismissed):

web_ir_actions_act_view_reload/__manifest__.py
guewen: It works because of python's auto-concatenation of strings.
Black produced this. I'll fix.

web_ir_actions_act_view_reload/static/src/js/web_ir_actions_act_view_reload.js
guewen: I didn't change anything here.
@pnajman-modoolar can we consider it a mistake and change it to LGPL to be consistent with the manifest?

pnajman-modoolar: @guewen sure

web_ir_actions_act_view_reload/__manifest__.py
pnajman-modoolar: @guewen  while we are here and i'm just wondering, how did "Brainbean Apps" become author of this module? If i recall correctly they just migrated it to 12.0 without adding any new functionality. Are there any OCA guidelines about this?

pedrobaeza: The effort of migrating a module can be big depending the version. There's no strict rule for saying if someone work grants the co-authorship. @alexey-pelykh can you tell us arguments for you to be there?

alexey-pelykh: @pedrobaeza this module is quite heavy for migration, v12 was nearly a rewrite if I remember correctly

pedrobaeza: OK, if that was the case, then I think it's reasonable for you to be in the author list

alexey-pelykh: https://github.com/OCA/web/commit/633e0c8a2d3e8e2a7d9cda18b1b2ee5c98481450#diff-b3ac0207ff2950bb6f972a8d4de64810 not much remained

alexey-pelykh: I even did a mistake with copyright :D

pnajman-modoolar: @pedrobaeza @alexey-pelykh guys, thanks for clarification :)

web_ir_actions_act_view_reload/static/src/js/web_ir_actions_act_view_reload.js
pnajman-modoolar: @guewen , @tarteo guys, What's the proper licence for OCA modules `LGPL-3` or `AGPL-3` ?
I'm asking because [here](https://github.com/OCA/web/blob/13.0/web_dialog_size/__manifest__.py) we have AGPL-3 and [here](https://github.com/OCA/web/blob/13.0/web_responsive/__manifest__.py) LGPL-3.

pedrobaeza: It's the one you want to choose to fit your needs. Boths are valid in OCA.

guewen: Yep. I updated the file to match the manifest.

alexey-pelykh:
@guewen I don't remember exactly, but promises look differently in v13, as far as I remember, so `$.when()` should be something else, but I'm not sure

alexey-pelykh:
Or was that `Promise.resolve()` changed to `$.when()`...

guewen:
@alexey-pelykh that's right, thanks! `$.when()` is replaced by `Promise.resolve()`.

edit: added a note about this on https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

simahawk (approved):
LG.
@guewen can you squash?
@alexey-pelykh can you pass by again?

alexey-pelykh (approved):

pnajman-modoolar (approved):

pnajman-modoolar (approved):

simahawk:
@tarteo your comments have been addressed. We are good to merge.

PR #1465, opened 2019-12-20
[9.0][IMP] web_m2x_options: new option 'search_delay' on Many2one

sebalix:
This option allows to delay the 'name_search' RPC request done when the
user write in Many2one fields.

By default in Odoo the request is performed after 200ms, which can be too fast,
generating a lot of RPC queries stressing uselessly the server.
Increasing this value (for instance to 800ms) allows to get better performance
as the server is less stressed.

As other options, it can be set globally (`ir.config_parameter`) or at
the field level.

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -47,7 +47,7 @@ in the field's options dict
 
   Number of displayed record in drop-down panel
 
-``delay`` *int* (Default: openerp default value is ``200``)
+``search_delay`` *int* (Default: openerp default value is ``200``)
 
   Number of milliseconds after a user input to trigger the ``name_search``.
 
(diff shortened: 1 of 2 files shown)


sebalix:
ping @simahawk I need your lights on this one <s>regarding the call to `get_options`. The `this.view.ir_options` is not defined if I don't call this method, although it is called in `start`. Any idea?</s>
EDIT: I moved the RPC call on `ir.config_parameter` at the View level to ensure having the option values when the fields widgets are starting (these options were stored in the view BTW). This also allows to remove the duplicated code `get_options` in the Many2manyTags overload.

StefanRijnhart (approved):

sebalix:
@i-vyshnevska @StefanRijnhart I prefer to close this PR for the moment as we encountered some errors on some 9.0 view forms (with custom JS widgets like reconcilation one) with this module installed.

PR #1468, opened 2019-12-24
[11.0][ADD] web_search_date

etobella:
Allows to search on dates with before and after directly on the search bar
without the "Extra filters".

imagen

@ForgeFlow @jarroyomorales


jarroyomorales (commented):
LGTM! But will this automatically activate when adding a date/datetime field to a search view or you have to specify in some way you want to have this behaviour?

etobella:
It will be automatically added

AdriaGForgeFlow (changes requested):
I don't seem to find a way to make it work. All you need to do is add a date/datetime field to the search view right?
I added date_order in the sale order search view:
image
But the filter does not appear when I search. Am I doing something wrong?

etobella:
@AdriaGForgeFlow What did you write? At least, you need to write Year and month o something like this: like 2020-01 for 1st January 2020

AdriaGForgeFlow:
> @AdriaGForgeFlow What did you write? At least, you need to write Year and month o something like this: like 2020-01 for 1st January 2020

It worked when I typed 2020-01, maybe this should be somehow mentioned in the README? 
What do you think?

Thanks for the answer, btw :+1:

AdriaGForgeFlow (approved):

jarroyomorales (approved):
LGTM! Maybe it would be nice if you could specify in options which filters you want to use instead of adding the 3 automatically, but it is also ok for me if we keep it like this.

PR #1471, opened 2019-12-30
[11.0][ADD] web_widget_open_tab

etobella:
This addon creates a new widget that can be used to open on a new tab a record from a tree.

Usage
=====

Edit the tree view and add the widget as the first field, usually, we should use:

    <field name="id" widget="open"/>

Now, we can open when clicking with the mouse wheel over the eye.
On a usual click the record will be opened without changes (keeping the breadcrumbs).

@jarroyomorales


jarroyomorales (approved):
LGTM :+1:! I think this is the closest we can be to getting the 'open in new tab' behavior. It has to be used in controled situations and be well-tested since it is probably losing the context when opening in a new window/tab.

etobella:
video
This is a small example

sudhir-erpharbor:
LGTM. I would suggest adding image/gif of the functionality in the readme/index file.

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,46 @@
...
+        }),
+        isSet: function () {
+            return true;
+        },
+        _getReference: function () {
+            return 'model=' + this.model + '&id=' + this.res_id;
+        },
+        _renderReadonly: function () {
+            var $content = $(
+                '<a href="#' + this._getReference() + '">'
```
JordiBForgeFlow: Using `<a href="#' + this._getReference() + '" onclick="window.open(this.href);">` will effectively open in a new tab. This looks more intuitive to me, isn't it?

JordiBForgeFlow:
cc @AdriaGForgeFlow

web_widget_open_tab/static/src/js/widget.js
etobella: I implemented the same idea but on the `onclick` function. Just to simplify it :smile:

AdriaGForgeFlow:
I have done a PR with some suggestions here [tegin/web](https://github.com/tegin/web/pull/1)

etobella:
I think your proposal is much better. Changes applied :smile:

AdriaGForgeFlow (approved):
LGTM :+1:  

Nice feature, thanks!

HviorForgeFlow (commented):
I noticed that if I click on wheel button to open the tab in background, new tab goes to discuss

web_widget_open_tab/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Usage
+=====
+
+Edit the tree view and add the widget as the first field, usually, we should use:
+.. code-block:: xml
+
+    <field name="id" widget="open"/>
+
+Now, we can open when clicking with the mouse wheel over the eye.
```
tarteo, line 43: When I click with the left mouse button it opens the record in a new tab. When I click with the mouse wheel it opens a new tab not showing the record instead the discuss module.

tarteo (changes requested):

web_widget_open_tab/README.rst
etobella: @HviorForgeFlow @tarteo Now this have been fixed.

jarroyomorales:
@tarteo @HviorForgeFlow Could you please review again? Thank you!

HviorForgeFlow (approved):
Now YES! :+1:

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,53 @@
...
+        supportedFieldTypes: ['integer'],
+        events: _.extend({}, AbstractField.prototype.events, {
+            'click': '_onClick',
+        }),
+        isSet: function () {
+            return true;
+        },
+        _getReference: function () {
+            var url = new URL(window.location.href);
+            var hash = url.hash;
```
MeetKD: You can remove this unused variable.

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,53 @@
...
+            var $th = this._super.apply(this, arguments);
+            if (node.attrs.widget === 'open') {
+                $th.removeClass('o_column_sortable');
+                $th[0].width = 1;
+            }
+            return $th;
+        },
+    });
+
+    field_registry.add('open', FieldOpen);
```
MeetKD: Can you please replace 'open' as FieldOpenTab or similar to it.

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,53 @@
...
+            return true;
+        },
+        _getReference: function () {
+            var url = new URL(window.location.href);
+            var hash = url.hash;
+            return url.hash.replace(/view_type=\w+/i, "view_type=form") + '&id=' + this.res_id;
+        },
+        _renderReadonly: function () {
+            var $content = $(
+                '<a href="'+ this._getReference() + '">'
```
MeetKD: it would be better if you can also add tooltip.

etobella:
Thanks @MeetKD all comments attended!!! :wink:

etobella:
@tarteo All comments have been attended so far, could you review it again :wink:

web_widget_open_tab/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+Edit the tree view and add the widget as the first field, usually, we should use:
+.. code-block:: xml
+
+    <field name="id" widget="open"/>
```
jarroyomorales: The widget is `open_tab`!

tarteo: Please change this to `open_tab`

tarteo (changes requested):
LGTM only 1 small remark on the readme.

etobella:
@tarteo Change applied

etobella:
There was an issue with boards. If you rewrite the data of the url it does not rewrite the model.
Now, we have changed it. If the model is the right one, we use the old logic, as we want to maintain the action. Otherwise, we will recreate the URL

etobella:
@tarteo Can you review?

etobella:
@tarteo

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,82 @@
...
+    var AbstractField = require('web.AbstractField');
+    var field_registry = require('web.field_registry');
+    var ListRenderer = require('web.ListRenderer');
+    var core = require('web.core');
+    var config = require('web.config');
+    var qweb = core.qweb;
+    var _t = core._t;
+
+    var FieldOpenTab = AbstractField.extend({
+        description: "",
```
simahawk, line 13: any nice description instead?

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,82 @@
...
+        supportedFieldTypes: ['integer'],
+        events: _.extend({}, AbstractField.prototype.events, {
+            'click': '_onClick',
+        }),
+        isSet: function () {
+            return true;
+        },
+        _getReference: function () {
+            var url = new URL(window.location.href);
```
simahawk: what about using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams ? 
Not a blocker, just check if you can do it easier w/ it :wink:

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,82 @@
...
+                url.hash.search(re) === -1
+            ) {
+                return '#view_type=model&model=' + this.model + '&id=' + this.res_id;
+            }
+            return url.hash.replace(/view_type=\w+/i, "view_type=form") + '&id=' + this.res_id;
+        },
+        _renderReadonly: function () {
+            var $content = $(
+                '<a href="'+ this._getReference() + '">'
+            ).addClass('open_tab_widget fa fa-eye');
```
simahawk:
```suggestion
            var $content = $(
                "<a />", {"href": this._getReference(), "class": "open_tab_widget fa fa-eye"}
            );
```

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,82 @@
...
+                            }
+                        },
+                    });
+                },
+            });
+            this.$el.append($content)
+        },
+        _onClick: function (ev) {
+            console.log(this);
+            console.log(ev);
```
simahawk:
```suggestion
```

web_widget_open_tab/static/src/js/widget.js
```diff
@@ -0,0 +1,82 @@
...
+            ev.preventDefault();
+            ev.stopPropagation();
+            var element = $(ev.currentTarget).find('a');
```
simahawk, line 58: can't you check the tagName?

web_widget_open_tab/views/assets.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
simahawk: `templates/assets.xml`

web_widget_open_tab/static/src/less/open_tab.less
```diff
@@ -0,0 +1,10 @@
+.open_tab_widget {
```
simahawk: can't you toggle `.invisible/.visible` class on the element via JS instead of loading another resource?
Not a blocker but we reduce the size of the assets (and save 1 request in debug=assets mode).

simahawk (changes requested):

web_widget_open_tab/static/src/js/widget.js
etobella: I cannot use URLSearchParams as the original url of odoo is `/web#...` This is not an standard url that URLSearchParams can process :thinking:

web_widget_open_tab/static/src/js/widget.js
etobella: If description is shown, it will be written on the top of the tree, I think it is better to keep it blank

web_widget_open_tab/static/src/js/widget.js
etobella: Sure

web_widget_open_tab/static/src/js/widget.js
etobella: I prefer to check in order to ensure that we have not selected any other element it is just for security.

web_widget_open_tab/views/assets.xml
etobella: Sure

web_widget_open_tab/static/src/less/open_tab.less
etobella: True. Also, tooltip is doing it directly, so it has no sense.

etobella:
@simahawk Thanks for your review. All comments attended or answered :wink:

web_widget_open_tab/static/src/js/widget.js
simahawk: well, you should take only the part after the hash. In any case not a blocker...

web_widget_open_tab/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2019 Creu Blanca
```
simahawk: 2020?

web_widget_open_tab/static/src/js/widget.js
simahawk: a comment for your future self will help :wink:

web_widget_open_tab/static/src/js/widget.js
simahawk: quick example:

image

simahawk (approved):

web_widget_open_tab/static/src/js/widget.js
etobella: Thanks, I didn't know this, much easier now :wink:

web_widget_open_tab/static/src/js/widget.js
etobella: You are right :+1:

web_widget_open_tab/__manifest__.py
etobella: It was made on 2019, so 2019-2020

PR #1475, opened 2020-01-07
[13.0][IMP] web_responsive: Document Viewer

Tardo:
Forward Document Viewer feature

cc @tecnativa TT21158


pedrobaeza:
TT?

yajo:
We have no runbot?

pedrobaeza:
It seems so... @gurneyalex have checked it and I think if we retrigger somehow the PR, it will generate everything correctly. We need to forced push for that.

Tardo:
Updated with the improvements/fixes added in 12.0

pedrobaeza (approved):
Fast-tracking as this is direct from v12

PR #1483, opened 2020-01-10
[8.0][CI] Fix travis distro

andhit-r:
Use  distro trusy


pedrobaeza:
The best way to fix this is not to change distro, but to put `postgresql: "9.6" in `addons` section. See full up-to-date travis template in https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml

PR #1486, opened 2020-01-14
[13.0][MIG] web_drop_target

NuriaXifre:
Migration of the module web_drop_target to v13.
cc ~ @ForgeFlow


HviorForgeFlow (changes requested):
Functional review:
Feature works awesome, I've just found a little issue.
If you attach a file on a Draft model with no id, attachment is recorded in the system but it is not linked with the new record (obviously). Is there any chance to limit that behavior on New ID records? @etobella @AdriaGForgeFlow

AdriaGForgeFlow:
> Functional review:
> Feature works awesome, I've just found a little issue.
> If you attach a file on a Draft model with no id, attachment is recorded in the system but it is not linked with the new record (obviously). Is there any chance to limit that behavior on New ID records? @etobella @AdriaGForgeFlow

Should be handled here but the function does not appear anywhere else...
https://github.com/OCA/web/pull/1486/files?file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.png&file-filters%5B%5D=.py&file-filters%5B%5D=.rst&file-filters%5B%5D=.xml#diff-cf5abc94ebf71eacb63a60823a4a986bR152

I'm going to check this.

AdriaGForgeFlow:
What if we allow drag&drop only in readonly mode? @HviorForgeFlow @etobella 

With this would do:
``` javascript
        _on_dragenter: function(e) {
            e.preventDefault();
            if (this.mode == "readonly") {
                this._add_overlay();
            }
            return false;
        },
```

This way we make sure user does not drag&drop on a New record

HviorForgeFlow:
Couldn't it be done by checking id=<empty> on url?

hbrunn:
don't do any of the above. https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR154 should take care of this, so see why this test fails

AdriaGForgeFlow:
> don't do any of the above. https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR154 should take care of this, so see why this test fails

The function `_get_drop_file` does not appear anywhere else so is never called.

hbrunn:
so this should go away then anyways, and you can move the check somewhere else

AdriaGForgeFlow:
> so this should go away then anyways, and you can move the check somewhere else

How do you suggest we do it? Why we don't just limit dropping in non-readonly mode?

hbrunn:
override https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR52 instead of `_get_drop_file` (delete this one), check if https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR164 is set. If not, return null, otherwise return super

AdriaGForgeFlow:
> override https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR52 instead of `_get_drop_file` (delete this one), check if https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR164 is set. If not, return null, otherwise return super

Makes sense but then from the user perspective will see the drop overlay, then will drop the file but nothing will happen. This may be confusing. Maybe adding a message or an alert could be enough, what do you think?

hbrunn:
ideally we'd already check if a drop is allowed in https://github.com/OCA/web/pull/1486/files#diff-cf5abc94ebf71eacb63a60823a4a986bR41 and show a different overlay if it isn't

AdriaGForgeFlow:
Can you update reviews @hbrunn @HviorForgeFlow

LoisRForgeFlow (approved):
LGTM

HviorForgeFlow (approved):
:+1:

AdriaGForgeFlow (approved):

PR #1487, opened 2020-01-14
web: add module web_disable_delete_group

abdounasser202:
This commit introduces a new module modelled on web_disable_export_group 
that prevents data deletion by users


pedrobaeza:
This can be done already with ACLs. Why the need of a module?

abdounasser202:
Using ACLs we do it model by model. In this case, we want to apply this right to a user for all the models he has access to. The goal here is to hide the "delete" button just like the web_disable_export_group module does for the "export" button.

pedrobaeza:
Well, I don't find this useful and it's weird from usability perspective, but if others do, I won't oppose.

PR #1498, opened 2020-01-24
[12.0] [FIX] Tree m2o clickable: call to get_formview_action()

qgroulard:
When clicking on a m2o field on a form view, the method 'get_formview_action()' is called in order to get the action.
This method can be overridden in order to return specific action for a certain model.

IMO, we should expect the same behaviour when clicking on the m2o field on a tree view.


sbidoul (approved):
LGTM.

Calling `get_formview_action` avoids the hard-coded the action, and provides for better handling of polymorphic models.

AdriaGForgeFlow (approved):
LGTM :+1:

simahawk (approved):

PR #1500, opened 2020-01-24
[12.0] [ADD] m2x_options: Option open for One2Many fields

qgroulard:
With 'open' option, clicking on a row of a o2m field will open the view for the record in fullscreen rather than in a popup.


sbidoul (approved):
LGTM.

This is a useful addition to provide form navigation from o2m widgets in read-only mode, instead of the usual pop-up.

sbejaoui (approved):
Works fine, 

Thanks

sbidoul:
I still need to investigate why the bot wakes up at 8:00 UTC :)

PR #1508, opened 2020-01-30
[13.0][Migrate] web_export_view migrated 13.0

sinandurgut07:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

StefanRijnhart:
Is this perhaps the same functionality as is now included in core Odoo? 
image showing the 'Export all' button in the upper right corner of the Odoo list view widget
When clicking this button, an XLSX file is fetched containing the records and the column layout from the current list view.

StefanRijnhart:
Confirmed by Pedro's comment in the alternative PR

pedrobaeza:
There are some slight differences:

- Odoo core feature exports also the group information. Very good to have it!
- Odoo core always export full available recordset, while OCA's module is only exporting the selection. One approach or the other is not better or worse. It's simply the way it works and you have to get used to.

PR #1509, opened 2020-01-31
[12.0][IMP] web_responsive: Payment info popover position

Tardo:
This PR try to mitigate the problem with payment info popover placement when users have the sided chatter enabled...

Before:
popover_bad

After:
popover_good

cc @tecnativa TT21756


pedrobaeza:
This is not something in native Odoo, only in web_responsive?

pedrobaeza (approved):

Tardo:
Yes, only happens with web_responsive installed... but caused by https://github.com/odoo/odoo/blob/12.0/addons/account/static/src/js/account_payment_field.js#L83

yajo (approved):

PR #1510, opened 2020-02-02
error when trying to save current search filter on an older one

eng-eslam-a-y:
has to explicit check for type of value !== null as the typeof null is object
https://i.stack.imgur.com/FzI1R.png


tarteo (approved):

eng-eslam-a-y:
Should it be merged now

HviorForgeFlow (approved):

PR #1511, opened 2020-02-03
IMP 12.0 add char and text fields for numpad dot fuctionality

ljimenezsidoo:
Allow to use numpad dot code in two more fields: Char and Text


pedrobaeza:
This is not valid, as char and text fields can contain the dot for several purposes and you want to keep then even if coming from the numpad part.

What do you think about this, @chienandalu ?

chienandalu:
I think the same. It'd be annoying

ljimenezsidoo:
The regular dot and the numeric keypad dot have a different keycode. You can use the dot in these fields with the regular dot equally and for other situations where you need to use the l10n parameters and change the numpad dot to a comma, you can too

angelmoya:
Please @pedrobaeza and @chienandalu  check @acm1pt-colorado comment, as keypad dot will allways a comma in this case, and you still have regular dot key.

pedrobaeza:
As said, it's confusing to do that replacement in text fields IMO.

PR #1514, opened 2020-02-06
[FIX] if nocreate=1 then also see that as true

thomaspaulb:
Some people in their custom modules have used 'no_create=1' instead of True to disable create or edit, which [works](https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/js/views/form_common.js#L235) in standard Odoo. But when `web_m2x_options` is installed, it will stop working, which is IMO undesired/unintended behavior.

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -87,7 +87,7 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
             var is_string = typeof option === 'string'
             var is_bool = typeof option === 'boolean'
             if (is_string) {
-                return option === 'true' || option === 'True'
+                return option === 'true' || option === 'True' || option === '1'
             } else if (is_bool) {
                 return option
             }


web_m2x_options/static/src/js/form.js
```diff
@@ -87,7 +87,7 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
             var is_string = typeof option === 'string'
             var is_bool = typeof option === 'boolean'
             if (is_string) {
-                return option === 'true' || option === 'True'
+                return option === 'true' || option === 'True' || option === '1'
```
hbrunn, line 90: please just pyeval the string, which is the reason it works in core

hbrunn (changes requested):

web_m2x_options/static/src/js/form.js
hbrunn: ...pyeval or json?

thomaspaulb: [pyeval](https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/js/views/form_common.js#L662) is used to convert the options dictionary into a Javascript object, after which a simple [falsy/truthy test](https://github.com/OCA/OCB/blob/10.0/addons/web/static/src/js/views/form_common.js#L235) is used to determine the value, which is why it works in core.

Since I am dealing with the `options` object that is already filled using `pyeval`, maybe the correct way is then just to replace the `is_option_set` check by a falsy/truthy test?

thomaspaulb: (not all `is_option_set` checks, but eg. [this one](https://github.com/OCA/web/blob/10.0/web_m2x_options/static/src/js/form.js#L400))

hbrunn: but I wonder how strings end up there in the first place? If somebody writes `<... options="{'test': True}" />`, this should come out as boolean, not string, right? This seems to me we do the parsing somewhere else wrongly. Do you agree? So we should pyeval when we read this?

hbrunn: it's not in the options attribute anyways as you write above which makes part of my reaction bullshit, but i stick to that we should look at parsing and not add stuff that happens to end up there as a string

hbrunn (approved):
why do we need the switch between types in the first place? I believe in some earlier versions we wanted to be nice for input, and then started that. But that can't be asked for a contribution to change, so I applaud fixing the symptom, this will be fine here. Thanks!

web_m2x_options/static/src/js/form.js
thomaspaulb: As an alternative I could do this: https://github.com/sunflowerit/web/pull/3. Do you have a preference?

NL66278 (approved):
:+1: LGTM

gjotten:
@hbrunn

etobella:
@thomaspaulb Can you check the issues with travis?

thomaspaulb:
@simahawk Rebased

thomaspaulb:
Okay I don't really know what to do here :/

simahawk:
It's failing in the same way on 10.0.

image

@sbidoul do we still expect builds for 10.0 to work?

sbidoul:
> do we still expect builds for 10.0 to work?

@simahawk dunno. This is Travis & MQT, it's been ages since I looked into that. So I guess the answer is no.

thomaspaulb:
I've tried to use [guide](https://odoo-community.org/groups/contributors-15/contributors-253387) to change this to Github actions, but I'm running into too many issues to fix. Leaving this up to @OCA/web-maintainers to decide whether to manually force-merge this, skipping CI, or close.

NL66278:
@thomaspaulb You can switch to github actions if you remove the pre-commit job from it. GitHub no longer supports python2. pre-commit can still be used locally to check before pushing to github.

dreispt:
I'm doing a manual merge.

PR #1516, opened 2020-02-11
[12.0][add] web_widget_mpld3_chart

JordiBForgeFlow:
This module add the possibility to insert mpld3 charts into Odoo standard views.
This is an interactive D3js-based viewer which brings matplotlib graphics to the browser.

If you want to see some samples of mpld3's capabilities follow this `link
<http://mpld3.github.io/>`_.

To insert a mpld3 chart in a view proceed as follows:

#. Import the required libraries::

    import matplotlib.pyplot as plt, mpld3

#. Declare a text computed field like this::

    mpld3_chart = fields.Text(
        string='Mpld3 Chart',
        compute='_compute_mpld3_chart',
    )

#. In its computed method do::

    def _compute_mpld3_chart(self):
        for rec in self:
            # Design your mpld3 figure:
            plt.scatter([1, 10], [5, 9])
            figure = plt.figure()
            rec.mpld3_chart = mpld3.fig_to_html(figure)

#. In the view, add something like this wherever you want to display your
   mpld3 chart::

    <div>
        <field name="mpld3_chart" widget="mpld3_chart" nolabel="1"/>
    </div>


AdriaGForgeFlow (approved):
Code review :+1: 

You need to add the dependency to the requirements.txt file

web_widget_mpld3_chart/static/src/js/web_widget_mpld3_chart.js
```diff
@@ -0,0 +1,17 @@
+odoo.define('web_widget_mpld3_chart', function (require) {
+"use strict";
+
+    var fieldRegistry = require('web.field_registry');
+    var AbstractField = require('web.AbstractField');
+
+    var Mpld3ChartWidget = AbstractField.extend({
+        start: function() {
```
MeetKD, line 8: _renderEdit will be the right place.

MeetKD (changes requested):

web_widget_mpld3_chart/static/src/js/web_widget_mpld3_chart.js
JordiBForgeFlow: @MeetKD tried your suggestion and did not work. The current implementation drives the correct rendering.

JordiBForgeFlow:
@LoisRForgeFlow can you review?

requirements.txt
```diff
@@ -2,3 +2,6 @@
 bokeh==1.1.0
 # web_widget_plotly_chart
 plotly==4.1.0
+# web_widget_mpld3_chart
+matplotlib>=2.0.0
+mpld3>=0.3
```
LoisRForgeFlow: Aren't you adding in the module a specif version of the library? You should freeze it here.

LoisRForgeFlow (approved):
LGTM

Issue #1518, opened 2020-02-13
[12.0][Advanced search] - Can't combine several OR operators


sbejaoui:
Hi,

To reproduce this bug:

1. Open the advanced search wizard and combine multiple OR operator
![1](https://user-images.githubusercontent.com/12665409/74428605-a07fb000-4e59-11ea-91d0-57bba1bace04.png)
2. When you save you will see that the second or is transformed to an AND
![2](https://user-images.githubusercontent.com/12665409/74428684-c4db8c80-4e59-11ea-956f-55653d665689.png)
3. Save the search and take a look to the filter, you can see a difference between the original search and the saved one
![3](https://user-images.githubusercontent.com/12665409/74428856-17b54400-4e5a-11ea-8486-ec0d00e459bd.png)
4. If change the domain using the standard widget I see that It combine correctly OR operator
![4](https://user-images.githubusercontent.com/12665409/74428933-44695b80-4e5a-11ea-8105-4e4d77eaf04d.png)

I conclude that this bug is in "Advanced search" module.

Regards.

yk2kus:
How do you open advance search wizard?

pedrobaeza:
There's a fix in #1519

Tardo:
This issue is solved no?

PR #1519, opened 2020-02-13
[12.0][FIX] - advanced search, combination of several OR operators

sbejaoui:
Fixes: #1518 

There is a problem when converting a domain object to an array.

in the case of a combination of several OR operators, the retroconversion into an array
from the domain object is not equal to the original array used for initialization.

The problem can be related to the initialization of the domain object or to the
conversion to an array, but in any case, it is out of this PR scope.

The humanDomain needs an array domain, which mean that the conversion and the
retro-conversion are useless in this case.


sbejaoui:
cc/ @gva-acsone

gva-acsone (approved):
LGTM +1 functional test

yajo (approved):

ValentinVinagre (approved):
Functional testing 👍

PR #1521, opened 2020-02-17
[IMP] web_environment_ribbon: Allow to generate ribbon name with code

hbrunn:


pedrobaeza:
Just a note: IMO, this is not `[ADD]`, but `[IMP]`, as the module exists and you are improving it with a new feature (for me, `[ADD]` is when you add a new module to the repository).

pedrobaeza (commented):
Why not evaling always and if fails, then use plain values?

ivantodorovich (approved):

pedrobaeza (approved):
Let's merge it, as the feature itself is valuable:

/ocabot merge minor

PR #1525, opened 2020-02-19
[12.0][MIG] web_search_date

etobella:
Migration of #1468 

No major changes


dalonsod (changes requested):
First of all, readme doc is not updated, still shows 11.0

Tested on runbot, after installing `hr_timesheet` for testing purposes, I tried to do the following "Date at" filter and this error was fired:

imagen

Could you check?

Issue #1526, opened 2020-02-20
[13.0] web_domain_field invalid path to static/lib/js


lck:
Path to pyeval.js is invalid:

13.0/web_domain_field/static/lib/js/pyeval.js

the valid path should be:

13.0/web_domain_field/static/**src**/js/pyeval.js

as declared in web_domain_field/views/web_domain_field.xml

legalsylvain:
could you make a PR ? thanks !

sudhir-erpharbor:
Thanks for reporting the issue. The issue is fixed in the PR #1527

PR #1527, opened 2020-02-20
[13.0][FIX] Fixed the pyeval.js path in assets_backend

sudhir-erpharbor:
Fixed the issue of the path in assets_backend #1526


lmignon (approved):

legalsylvain (approved):
thanks.

PR #1529, opened 2020-02-23
[12.0][FIX] web_responsive: Change AppDrawer close method

### LINKED ISSUE #1528: web_responsive Apps Drawer not closing!
Version [12.0] 
Moduel web_responsive

App Drawer is not closing if click on the app icon and menu which is a direct action menu from the top bar.
It is working fine if it is menu which has child menu.It doesn't work on the cases of direct action menu

Tardo:
Previous of this commit, AppDrawer closes when the action have an id  that is defined in a menu item and no rpc calls are running.
Now AppDrawer closes in a more generic way, is enough that a controller is attached to consider that the AppDrawer needs to be closed.

This fixes #1528 and other related issues with the AppDrawer.

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -16,12 +16,28 @@ odoo.define('web_responsive', function (require) {
     var Chatter = require('mail.Chatter');
     var DocumentViewer = require('mail.DocumentViewer');
 
-    /*
-     * Helper function to know if are waiting
-     *
+
+    /* Hide AppDrawer in desktop and mobile modes.
+     * To avoid delays in pages with a lot of DOM nodes we make
+     * sub-groups' with 'querySelector' to improve the performance.
      */
-    function isWaiting () {
-        return $('.oe_wait').length !== 0;
+    function closeAppDrawer () {
+        _.defer(function () {
+            // Need close AppDrawer?
+            var menu_apps_dropdown = document.querySelector(
+                '.o_menu_apps .dropdown');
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@SaravananSakthivel please try this for solving your issue.

pedrobaeza:
I'm still having problems with this use case:

* From starting point (discuss), invoke app drawer with Shift+Alt+A.
* Search for a menu like the one in the GIF.
* Press Enter.
* App Drawer is not closed.

Peek 23-02-2020 21-14

pedrobaeza:
Not sure if you have pushed a fix for what I have said, but if so, the problem is still there.

Tardo:
Now should be fixed.

I can only reproduce it on runbot + firefox ... and in firefox you can't edit remote javascript "on the fly" :/ ... but the problem was that the AppDrawer can't be closed if an rpc tasks are still running. Now uses another method that doesn't need check that: not closes it when an app is clicked and closes it when a controller is attached. Simple + Less lines = KISS

pedrobaeza (approved):
Functional working, although my feeling is that we have lost the feature of not hiding the app drawer until the load is completed. Can it be?

Tardo:
The feature should still working.... :/  (Tested with a throttled connection)

pedrobaeza:
Great, thanks for confirming!

SaravananSakthivel:
> @SaravananSakthivel please try this for solving your issue.

@pedrobaeza yes it is working fine with the issue

pedrobaeza:
Great, let's merge it then:

/ocabot merge patch

@Tardo please move it to v13

PR #1531, opened 2020-02-24
[12.0][IMP] web_notify: add channels only if not done already

nilshamerlinck:
Hi there,

We faced a use case that can lead to a resources exhaustion both on client and server sides when `web_notify` is installed on the instance.

Here's how to reproduce the issue (in your local):
- open a first tab
- open `DevTools > Network`
- log in the instance as `admin`
- duplicate this tab twice (required)
- in the third tab, replace the url in the location bar by `/web/login`
- log in as `demo`

As a consequence:
- Odoo will get stuck in an infinite loop, continuously adding/removing `notify_*` channels (due to the native [crosstab mechanism](https://github.com/odoo/odoo/blob/12.0/addons/bus/static/src/js/crosstab_bus.js#L328))
- these changes will trigger thousands of xhr requests to `/longpolling/poll`, that will:
  - on the client side: accumulate and start using a lot of memory, making the tab very slow ("net:ERR_INSUFFICIENT_RESOURCES") and finally crash
  - on the server side: high load the `gevent` process (reaching limits: "Too many open files" (`ulimit -n`); "The Connection Pool Is Full" (`db_maxconn`))

Here's how it looks:

out

Some might argue:
- that the use case is not legit: why would a user log in as a different user in the same browser session?
  - testers do (they should not but they do)
  - some real life  use cases, like the creation and the testing of new users by the admin can lead to that case
- that this should be fixed in odoo instead: true, if you have an idea how to fix it let me know

Meanwhile, here's a commit that fixes the issue.

### DIFF
--- a/web_notify/static/src/js/web_client.js
+++ b/web_notify/static/src/js/web_client.js
@@ -26,15 +26,26 @@ odoo.define('web_notify.WebClient', function (require) {
                 this.channel_info,
                 this.channel_default,
             ];
-            this.call('bus_service', 'addChannel', this.channel_success);
-            this.call('bus_service', 'addChannel', this.channel_danger);
-            this.call('bus_service', 'addChannel', this.channel_warning);
-            this.call('bus_service', 'addChannel', this.channel_info);
-            this.call('bus_service', 'addChannel', this.channel_default);
+
+            // startPolling to get this new tab registered,
+            // in order to be able to call isMasterTab
+            this.call('bus_service', 'startPolling');
+
+            // - no need to add channels again if it was done already
+            // - this is also a workaround for the infinite loop issue
+            //   that occures when user logs in as a different user
+            //   while still being logged in 2+ other tabs
+            if(this.call('bus_service', 'isMasterTab')) {
+                this.call('bus_service', 'addChannel', this.channel_success);
+                this.call('bus_service', 'addChannel', this.channel_danger);
+                this.call('bus_service', 'addChannel', this.channel_warning);
+                this.call('bus_service', 'addChannel', this.channel_info);
+                this.call('bus_service', 'addChannel', this.channel_default);
+            }
+
             this.call(
                 'bus_service', 'on', 'notification',
                 this, this.bus_notification);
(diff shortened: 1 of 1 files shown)


jjscarafia:
Hi @nilshamerlinck ! I'm facing exactly the same issue on v13.
You save my day!

Some comments:

- I really don't know anything about js and/or bus notifications
- Regarding that this should be fixed in odoo, I'm not sure, because on odoo standard I couldn't replicate the issue
- I've try this patch and now the server is working ok when I login with new user on a new tab, but I don't receive notifications on this new tab/session till I close all the other tabs and then reload the new one. I still believe the fix is much better than the current behaviour.

I've create similar PR for v13 here https://github.com/OCA/web/pull/1648
Thanks!

hoangtrann:
@nilshamerlinck I tried the patch, but the issue which I face still remain. As I check the problem further, this work well on every browsers except one running on iOS mobile devices. In this case, the `StorageEvent` is fired within the same tab and right from the beginning. I don't think this behavior is expected and submit a [PR](https://github.com/odoo/odoo/pull/55654) for that  

Also, I mention it in this [issue](https://github.com/OCA/web/issues/1636#issuecomment-671195450)

gurneyalex (approved):
Tested on a production instance that was going bonkers because of this issue.

Thanks @nilshamerlinck

simahawk (approved):

hoangtrann:
@gurneyalex loved to see that we merged this PR ❤️  

Has you tried using the instance with a ios mobile device? I reported an [issue](https://github.com/OCA/web/issues/1636) which happen only on ios mobile devices and still happen after this fix. I also propose a [PR](https://github.com/odoo/odoo/pull/55654) to fix in odoo repo since it's related to StorageEvent using in CrossTab feature, but they seem to ignore this. Would love to apply this patch in OCA instead but I haven't found a way yet

simahawk:
@hoangtrann you can open it here https://github.com/OCA/OCB

gurneyalex:
@hoangtrann I unfortunately don't own an ios device...

PR #1532, opened 2020-02-25
[MIG][12.0] web_m2x_option: delay search

i-vyshnevska:
port feature from https://github.com/OCA/web/pull/1465

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -22,7 +22,10 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
         if (_.isUndefined(option))
             return false;
         if (typeof option === 'string')
-            return option === 'true' || option === 'True';
+            if (option === 'true' || option === 'True')
+                return true;
+            else
+                return JSON.parse(option);
         if (typeof option === 'boolean')
             return option;
         return false
@@ -91,6 +94,15 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
             var m2o_dialog_opt = is_option_set(this.nodeOptions.m2o_dialog) || _.isUndefined(this.nodeOptions.m2o_dialog) && is_option_set(ir_options['web_m2x_options.m2o_dialog']) || _.isUndefined(this.nodeOptions.m2o_dialog) && _.isUndefined(ir_options['web_m2x_options.m2o_dialog']);
             if (this.can_create && this.floating && m2o_dialog_opt) {
                 new M2ODialog(this, this.string, this.$input.val()).open();
+            }
+        },
+
+        _bindAutoComplete: function () {
+            this._super.apply(this, arguments);
+            var delay = is_option_set(ir_options['web_m2x_options.m2o_search_delay']);
+            if (delay &&
+                (this.field.relation in delay)){
+                this.$input.autocomplete("option", "delay", delay[this.field.relation]);
             }
         },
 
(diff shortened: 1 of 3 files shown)


web_m2x_options/static/src/js/form.js
```diff
@@ -21,6 +21,47 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
...
+        if (typeof option === 'string')
+            return option === 'true' || option === 'True';
+        if (typeof option === 'boolean')
+            return option;
+        return false
+    };
+
+    var settings;
+
+    web_m2x_options.done(function(res) {
```
sebalix: So I suppose that by moving this query at the JS module level will trigger the query when the JS client is loaded? Make sense after all.

web_m2x_options/static/src/js/form.js
```diff
@@ -21,6 +21,46 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
         method: 'get_web_m2x_options',
     });
 
+    function is_option_set(option){
```
TDu:
```suggestion
    var is_option_set = function is_option_set(option){
```

StefanRijnhart: @TDu For my information, what is the benefit of declaring the function variable this way (or which style guide are we honoring)?

If I understand correctly, in your suggestion it is declared twice, because `function is_option_set(option) {})` already declares the variable by itself.

TDu: Well looking at it again I should have suggested this
```suggestion
var is_option_set = function (option){
```
Because you should be always worried of the risk to declare something into the global object in js.
I am not sure if it is here or if my suggestion would break the call to this function below

StefanRijnhart: Thanks for the clarification! I read in https://dmitripavlutin.com/6-ways-to-declare-javascript-functions/ that using `function is_opton_set(option)` declares a variable in the current scope, so I'm still not sure what the difference is.

Scope is under require('web_m2x_options.web_m2x_options'), not global so I don't think that's an issue.

TDu: So indeed probably no differences, this https://softwareengineering.stackexchange.com/questions/160732/function-declaration-as-var-instead-of-function looks specific to our discussion

StefanRijnhart: Thanks for finding that!

web_m2x_options/static/src/js/form.js
StefanRijnhart: Checked this, and it does.

web_m2x_options/static/src/js/form.js
```diff
@@ -78,41 +118,19 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
             this._super.apply(this, arguments);
         },
     });
-
+   
     FieldMany2One.include({
 
-        start: function () {
+        _bindAutoComplete: function () {
```
StefanRijnhart: ~~Looks like you are overriding this obscure, private method because it is only called for widgets in edit mode, and then only once per widget. Maybe you can document this (or for whatever other reason you are changing this) in a docstring?~~ Oh never mind. Of course this feature is directly related to autocompletion.

StefanRijnhart (approved):
Thanks for the port @i-vyshnevska and thanks for the feature @sebalix! Tested and code review.

yvaucher (approved):
LGTM but rebase to do

sebalix (changes requested):
I prefer to block this PR until we have tested some compatibility issues we got on 9.0, as explained on the original PR for 9.0:
https://github.com/OCA/web/pull/1465#issuecomment-650946674
With this new feature we were unable to open some forms.

PR #1535, opened 2020-02-25
[FIX] web_widget_numeric_step: Float Precision

Tardo:
Uses Odoo tools to format the value.

Before this PR, if a float fields doesn't define the digits the widget prints the number as an integer. This PR uses Odoo utils to format the value that given a default float precision.

cc @tecnativa


yajo (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please migrate the module to 13.0

PR #1536, opened 2020-02-26
[13.0][MIG] web_widget_image_url

phuctranfxvn:


tonygalmiche (changes requested):
It works for form view.
But it might be nice to add a size limit or an option to set the size.
In the list view, I find that it is too big and there are messages:

image

Thank you for this migration

claudiagn (approved):
LGTM!
Would be great to add size limits so as to be equal in both views.

bosd:
Can you force push to recreate runboat?

PR #1538, opened 2020-02-27
[13.0][MIG] web_widget_digitized_signature

phuctranfxvn:


web_widget_digitized_signature/models/res_users.py
```diff
@@ -0,0 +1,25 @@
...
+
+from odoo import api, fields, models
+
+
+class ResUsers(models.Model):
+    _name = "res.users"
+    _inherit = ["res.users", "mail.thread"]
+
+    digital_signature = fields.Binary(
+        string="Digital Signature", oldname="signature_image", attachment=True
```
mileo: parameter 'oldname' is not longer supported; use an upgrade script instead.

mileo (changes requested):

web_widget_digitized_signature/models/res_users.py
phuctranfxvn: @mileo it has been fixed. Thanks

mileo: Thanks!!! Could you see the travis errors too?

tonygalmiche (changes requested):
The signature works.
But the width and height options are not taken into account : 
image

eLBati:
Following https://github.com/OCA/project/pull/664#issuecomment-638771702 , probably we should drop `web_widget_digitized_signature` in favor of [web.name_and_signature](https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/widgets/name_and_signature.js)

ThomasBinsfeld:
> Following [OCA/project#664 (comment)](https://github.com/OCA/project/pull/664#issuecomment-638771702) , probably we should drop `web_widget_digitized_signature` in favor of [web.name_and_signature](https://github.com/odoo/odoo/blob/13.0/addons/web/static/src/js/widgets/name_and_signature.js)

Hello @eLBati, do you have more information about how the Odoo widget would be reusable in custom modules?

eLBati:
@ThomasBinsfeld I think like done by portal https://github.com/odoo/odoo/blob/13.0/addons/portal/static/src/xml/portal_signature.xml

bodedra:
@phuctranerp Are you planning continue this PR and complete migration? If not, let me know I will take over from here. And Thanks for your contribution.

phuctranfxvn:
> @phuctranerp Are you planning continue this PR and complete migration? If not, let me know I will take over from here. And Thanks for your contribution.

Hi, @bodedra

I am not currently following this PR as seem that the function is available as a widget in Odoo already.

Thanks

bodedra:
@phuctranerp Thanks for your update. I will take look that widget and do clean up in project_task_digitized_signature module.

claudiagn (approved):
LGTM!

But would be great if sizes were taken into account.

eLBati:
So `project_task_digitized_signature` does not depend on this module and this PR can be closed, right?

etobella:
@phuctranerp Do you need some help migrating this module, it seems that only the pre-commit is necessary.

pedrobaeza:
Is it still needed? Odoo has one for this

etobella:
It seems so, I will check if this is not needed anymore:wink:

Saran440:
Hello, all.
I still need this module for some feature
this module is useful for keep signature of user
and other module can depend it to do something.

I think this module should be RFC to call function signature from core odoo. (or change name module)
WDYT?

etobella:
Which functionality is specific for the OCA module?

Saran440:
Amm, It's similar OCA base module.
Add new field (signature) on uses view. (res.users)

**Use My Case**
when user A validate document, we can get user A signature to print out report.

it will be generic module and other module can depend it.

PR #1540, opened 2020-03-02
pre-commit update: prettier xml indent

sbidoul:
- let prettier indent xml files, assuming white space is not significant (https://github.com/OCA/maintainer-quality-tools/pull/640)
- isort configuration update (no change to python sources)

### DIFF
--- a/web_dialog_size/static/src/xml/web_dialog_size.xml
+++ b/web_dialog_size/static/src/xml/web_dialog_size.xml
@@ -5,12 +5,12 @@
             <i class="fa fa-close" />
         </t>
         <t t-jquery="button.close" t-operation="before">
-            <button type="button" class="dialog_button_extend close"><i
-                    class="fa fa-expand"
-                /></button>
-            <button type="button" class="dialog_button_restore close"><i
-                    class="fa fa-compress"
-                /></button>
+            <button type="button" class="dialog_button_extend close">
+                <i class="fa fa-expand" />
+            </button>
+            <button type="button" class="dialog_button_restore close">
+                <i class="fa fa-compress" />
+            </button>
         </t>
     </t>
 </templates>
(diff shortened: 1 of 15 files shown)


yajo (approved):

web_dialog_size/static/src/xml/web_dialog_size.xml
```diff
@@ -5,12 +5,12 @@
...
-                /></button>
-            <button type="button" class="dialog_button_restore close"><i
-                    class="fa fa-compress"
-                /></button>
+            <button type="button" class="dialog_button_extend close">
+                <i class="fa fa-expand" />
+            </button>
+            <button type="button" class="dialog_button_restore close">
+                <i class="fa fa-compress" />
+            </button>
```
yajo, line 13: I wonder if in cases like this whitespace actually makes a difference... :thinking:

pedrobaeza: Why it should be different?

yajo: It could render a whitespace surrounding the icon that could break the layout.

pedrobaeza: I don't think so, ad the space is only being added inside `<i>` tag, or am I missing something?

sbidoul: I think such white space is not significant *by default*: https://www.w3.org/TR/css-text-3/#white-space-phase-2. But there are css properties that influence that (e.g. `whitespace: pre`) so in theory this could break some files. Such breakage is very unlikely in the OCA codebase IMO.

yajo: No, well, you know sometimes inside an `<a>` if there are whitespaces they form an ugly link with spaces underlined.

It doesn't seem to be the case here anyways:



(pic from runbot).

yajo (approved):

simahawk (approved):
LG

PR #1543, opened 2020-03-04
[FIX] web_widget_numeric_step: Fix _onChange input fields methods disabled by the widget

sergio-teruel:
cc @Tecnativa
ping @Tardo @pedrobaeza 
TT22667

### DIFF
--- a/web_widget_numeric_step/static/src/js/numeric_step.js
+++ b/web_widget_numeric_step/static/src/js/numeric_step.js
@@ -13,13 +13,14 @@ odoo.define('web_widget_numeric_step.field', function (require) {
     var NumericStep = FieldFloat.extend({
         template: 'web_widget_numeric_step',
         className: 'o_field_numeric_step o_field_number',
-        events: _.extend({}, FieldFloat.prototype.events, {
+        events: _.extend({}, _.omit(FieldFloat.prototype.events, ['change', 'input']), {
             'mousedown .btn_numeric_step': '_onStepMouseDown',
             'touchstart .btn_numeric_step': '_onStepMouseDown',
             'click .btn_numeric_step': '_onStepClick',
             'wheel .input_numeric_step': '_onWheel',
             'keydown .input_numeric_step': '_onKeyDown',
             'change .input_numeric_step': '_onChange',
+            'input .input_numeric_step': '_onInput',
         }),
         supportedFieldTypes: ['float', 'integer'],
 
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks. Lgtm. No test.

pedrobaeza (approved):

pedrobaeza:
I'm now fearing that people click several times very quickly on the buttons + and -, and this triggers for each click an onchange. Can we set a reasonable time capture window for not launching the onchange until passed? I know this is done on website for the number of items in the cart before processing the update.

yajo:
~~That should be done in the widget itself IMHO.~~ :point_left: I'm not stupid, I just didn't check the code :laughing:

yajo:
Yes, you can do that with a deferred/promise (I don't remember which one is preferred for v12) and a timeout.

Tardo:
Ummm i see the problem here... my initial intention wasn't overwrite '_onChange' method, i did it as a totally new one.
Now with these changes we have two 'change' (InputField + NumericStep) event calling the same method.

Please, remove the 'change' event declared by 'InputField' class... and the 'input' event need by changed too to listen the '.input_numeric_step' element (that is the real 'input' element).
In the NumericStep widget the $el var. is related to the group of input+buttons not only for the input.

A solution to not 'flood' the change event can be use: https://underscorejs.org/#throttle

pedrobaeza:
@Tardo, what about the suggestion of @Yajo's about deferred/promise with a timeout? Can you take over this as you are more experienced with JS?

Tardo:
Updated with some changes... finally i prefer use macrotask (setTimeout) instead of microtask (Promises) because in this case we don't need be "fast as possible". Underscore offers us 'debounce' that for me is exactly that we need in this case.

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -13,20 +13,25 @@ odoo.define('web_widget_numeric_step.field', function (require) {
...
             'change .input_numeric_step': '_onChange',
+            'input .input_numeric_step': '_onInput',
         }),
         supportedFieldTypes: ['float', 'integer'],
 
         // Values in milliseconds used for mouse down smooth speed feature
         DEF_CLICK_DELAY: 400,
         MIN_DELAY: 50,
         SUBSTRACT_DELAY_STEP: 25,
+        //
```
yajo: :wastebasket:

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -198,10 +213,11 @@ odoo.define('web_widget_numeric_step.field', function (require) {
         /**
          * Sanitize user input value
          *
-         * @param {ChangeEvent} ev
+         * @override
          */
         _onChange: function (ev) {
             ev.target.value = this._sanitizeNumberValue(ev.target.value);
+            this._super.apply(this, arguments);
```
yajo: Possibly not strictly needed, but this is the common pattern:

```suggestion
            return this._super.apply(this, arguments);
```

yajo (changes requested):
Looks neat! Minor details.

Tardo:
Changes done!

yajo (approved):

PR #1544, opened 2020-03-08
[12.0][FIX] web_translate_dialog: load translation for text field

i-vyshnevska:
translated field can be saved in input or textarea as there no other possibility to find which type of field has an initial field check for both of them

### DIFF
--- a/web_translate_dialog/static/src/js/web_translate_dialog.js
+++ b/web_translate_dialog/static/src/js/web_translate_dialog.js
@@ -215,7 +215,7 @@ var TranslateDialog = Dialog.extend({
                 });
                 if (code === self.view_language) {
                     _.each(text, function(value, key) {
-                        var view_elem = self.view.$('input[name="'+ key + '"]');
+                        var view_elem = self.view.$( ":input[name='" + key +"']")
                         view_elem.val(value).trigger('change');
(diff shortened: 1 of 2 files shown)


web_translate_dialog/static/src/js/web_translate_dialog.js
```diff
@@ -216,6 +216,9 @@ var TranslateDialog = Dialog.extend({
                 if (code === self.view_language) {
                     _.each(text, function(value, key) {
                         var view_elem = self.view.$('input[name="'+ key + '"]');
```
simahawk: you could use `:input` to get the input no matter the type https://api.jquery.com/input-selector/

simahawk (approved):

jcoux (approved):

oihane (approved):
LGTM

anajuaristi:
Hello, yesterday we detected a problem with this module. When you launch the translate dialog, the translations are not shown on fields. They are empty. But, saving the form, data is correctly stored on fields.
It seems that this PR is solving the problem. We tested in a local environment and it works right, so Could it be merged?
Thank you:
Ana

anajuaristi (approved):

PR #1545, opened 2020-03-11
[FIX] web_widget_numeric_step: Change number parse method

Tardo:
Previous this commit the widget only works with 'dots'. With this commit works using odoo parse method that support other separators.

** Can test it changing the separator in the language settings.

cc @tecnativa TT21495

### DIFF
--- a/web_widget_numeric_step/static/src/js/numeric_step.js
+++ b/web_widget_numeric_step/static/src/js/numeric_step.js
@@ -31,7 +31,6 @@ odoo.define('web_widget_numeric_step.field', function (require) {
 
         DELAY_THROTTLE_CHANGE: 200,
 
-        _prevValue: null, // Used to know if the value was really changed
 
         init: function () {
             this._super.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


Tardo:
ping @pedrobaeza

web_widget_numeric_step/static/src/js/numeric_step.js
```diff
@@ -133,10 +133,8 @@ odoo.define('web_widget_numeric_step.field', function (require) {
          * @param {String} mode can be "plus" or "minus"
          */
         _doStep: function (mode) {
-            var cval = Number(this.$input.val());
-            if (_.isNaN(cval)) {
-                return;
-            }
+            var field = this.record.fields[this.name];
+            var cval = field_utils.parse[field.type](this.$input.val())
```
legalsylvain: Hi . Thanks for this fix. using field_utils is a good idea. But don't you have to keep the like if isNan ? Because, field_utils can return a not a Number, if the input is invalid. Don't you think ? 

kind regards.

In other word, my proposal is : 
```
            var field = this.record.fields[this.name];
            var cval = field_utils.parse[field.type](this.$input.val())
            if (_.isNaN(cval)) {
                return;
            }
```

Tardo: Hi @legalsylvain the 'parse' method throws and exception if can't do the job... 
I know that this is too ugly for final user, perhaps best something like:
```
var cval = 0;
try {
    cval = field_utils.parse[field.type](this.$input.val());
} catch {
    return;
}
```

What do you think?

pedrobaeza: For me the previous value should be put in case of error.

Tardo: Ok, updated. Now if the user input a invalid value and do an "step" call, the method will put the previous value.

pedrobaeza (approved):
Working as expected and fix the problem. Fast-tracking this for avoiding the nasty problem that avoids the usage of the widget in every comma decimal separator region.

PR #1547, opened 2020-03-12
[12.0][ADD] web_pivot_computed_measure

Tardo:
Adds support for computed measures on the pivot view.

Go to pivot view and click on the "Measures" menu, you will see a new option called 'Computed Measure'.
These measures can be mixed and save as favorite.

measures

cc @tecnativa TT22499


pedrobaeza:
@ged-odoo I think this is something that is going to interest you.

web_pivot_computed_measure/README.rst
```diff
@@ -0,0 +1,89 @@
...
+.. contents::
+   :local:
+
+Usage
+=====
+
+Go to pivot view and click on the "Measures" menu, you will see
+a new option called 'Computed Measure'.
+
+These measures can be mixed and save as favorite.
```
pedrobaeza: This text should go on the description, not the usage.

web_pivot_computed_measure/README.rst
```diff
@@ -0,0 +1,89 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+Go to pivot view and click on the "Measures" menu, you will see
+a new option called 'Computed Measure'.
```
pedrobaeza, line 39: Add a bit more description of the usage (select first measure, second, the operation to be done, etc).

What about the developer option to put the formula?

web_pivot_computed_measure/README.rst
```diff
@@ -0,0 +1,89 @@
...
+
+* Tecnativa
+
+Contributors
+~~~~~~~~~~~~
+
+* `Tecnativa <https://www.tecnativa.com/>`_:
+
+  * Alexandre Díaz
+  * Pedro Baeza
```
pedrobaeza:
```suggestion
  * Pedro M. Baeza
```

web_pivot_computed_measure/README.rst
pedrobaeza: Add also:

- that you can use a computed measure as operand for other computed measures.
- explain that involved measures are added in the pivot automatically and they are needed for proper computed measure calculus, so if one of them is removed, the computed measure will be removed from the pivot table as well. This requirement is real, isn't it?

ivantodorovich (approved):
Nice PR! :+1:

legalsylvain (approved):
Tested functionnaly with purchase module (and report)
That is a very great job ! Thanks ! 

It is finally fixing Odoo bug, on average values ! (see bellow with the average price value). 
- Odoo is wrong,
- With that module, the value is OK

image

:+1: Functional Review. (+ light code review).

legalsylvain (changes requested):
Hi. 
I tried again your module, and in a different way, the loading of the computed measures from the context is not done properly.

Step to reproduce : 
- Add a computed measures in the context of an action like this : 
image
- go the menu that enable the action : KO : the field is not computed
image
- add any filter, domain, group by, etc... : OK : the field is computed
image

pedrobaeza:
Well, this is a bit out of scope, as the initial intention is not to load that computed measures from the context. @Tardo check a bit if it's simple to be done, but if not, we will just annotate that on the known issues.

legalsylvain:
> Well, this is a bit out of scope, as the initial intention is not to load that computed measures from the context.

Well, that is not the use case proposed by this module, but this is exactly the intention of this module, as the context is the tools used by this module to save and restore computed measures. ;-)
- It works from filter context
- It half works from action context. (the column is displayed, but the computation is not done).

pedrobaeza:
As said, if easy to be added, we'll do it. If not, we have set the base, other contributors can add that support.

web_pivot_computed_measure/static/src/xml/web_pivot_computed_measure.xml
```diff
@@ -0,0 +1,117 @@
...
+            Sum (m1 + m2)
+        </option>
+        <option name="sub" value="m1-m2">
+            Sub (m1 - m2)
+        </option>
+        <option name="mult" value="m1*m2">
+            Mult (m1 * m2)
+        </option>
+        <option name="div" data-format="float" value="m1/m2">
+            Div (m1 / m2)
```
yajo, line 19: I wonder what would happen if m2 is zero. Did you handle that?

web_pivot_computed_measure/static/src/xml/web_pivot_computed_measure.xml
```diff
@@ -0,0 +1,117 @@
...
+    </t>
+
+    <!-- t t-name="web_pivot_computed_measure.ComputedMeasureStyles">
+        <option name="normal" value="normal">
+            Normal
+        </option>
+        <option name="colored" value="colored">
+            Colored
+        </option>
+    </t -->
```
yajo:
```suggestion
```

web_pivot_computed_measure/static/src/xml/web_pivot_computed_measure.xml
```diff
@@ -0,0 +1,117 @@
...
+                    <select class="o_input o_time_range_selector" id="computed_measure_format">
+                        <t t-call="web_pivot_computed_measure.ComputedMeasureFormats" />
+                    </select>
+                </div>
+                <!-- div class="dropdown-item-text">
+                    <label for="computed_measure_style">Style</label>
+                    <select class="o_input o_time_range_selector" id="computed_measure_style">
+                        <t t-call="web_pivot_computed_measure.ComputedMeasureStyles" />
+                    </select>
+                </div -->
```
yajo:
```suggestion
```

web_pivot_computed_measure/static/src/js/pivot_controller.js
```diff
@@ -0,0 +1,145 @@
...
+                    hideMenu = true;
+                    var field1 = this.$buttons_measures_ex.find('#computed_measure_field_1').val();
+                    var field2 = this.$buttons_measures_ex.find('#computed_measure_field_2').val();
+                    var oper = this.$buttons_measures_ex.find('#computed_measure_operation').val();
+                    if (oper === "custom") {
+                        oper = this.$buttons_measures_ex.find('#computed_measure_operation_custom').val();
+                    }
+                    var name = this.$buttons_measures_ex.find('#computed_measure_name').val();
+                    var format = this.$buttons_measures_ex.find('#computed_measure_format').val();
+                    //var style = this.$buttons_measures_ex.find('#computed_measure_style').val();
```
yajo:
```suggestion
```

yajo (changes requested):
It's a pleasure to read your JS code, :clap: :clap: 

Please remove some :wastebasket:, and one DIV/0 question.

web_pivot_computed_measure/static/src/xml/web_pivot_computed_measure.xml
Tardo: infinity

web_pivot_computed_measure/README.rst
Tardo: You can't remove "measures" that are being used in a "computed measure". Need disable the computed measure first.
** The same for "computed measures" that are being used in others "computed measures".

pedrobaeza: Then please comment this in the README

Tardo:
Thanks for comments and reviews! :heart: 

@legalsylvain i can't reproduce your issue :/
pivot_action

legalsylvain:
@tardo : thanks a lot for your screenshot. indeed, there is no bug in your module. my context was wrong. (forgot ``pivot_measures`` key).

image
Now ok.

legalsylvain (approved):

web_pivot_computed_measure/static/src/xml/web_pivot_computed_measure.xml
yajo: Seems good enough.

yajo (approved):

web_pivot_computed_measure/README.rst
```diff
@@ -0,0 +1,110 @@
...
+    - Mult: m1 * m2
+    - Div: m1 / m2 (Format: Float)
+    - Perc m1 / m2 (Format: Percentage)
+    - Custom: Special option only visible in debug mode to write a custom formula.
+ - Name: The name of the new measure (emtpy = auto-generated)
+ - Format: How will the value be printed
+    - Integer
+    - Float
+    - Percentage (value * 100)
+ - Formula*: Custom operation formula
```
pedrobaeza, line 56: Is this always visible?

Issue #1517, opened 2020-02-12
[12.0] Advanced search - Wrong summation when using group by in list view


JGarcia-Panach:
In runbot:

This is a list view grouped by salesperson using a filter but without using advanced search filter
![image](https://user-images.githubusercontent.com/13234598/74336401-79b07380-4d9e-11ea-97d4-923c0b78520c.png)

All is ok

This is a list view grouped by salesperson using advanced search filter
![image](https://user-images.githubusercontent.com/13234598/74336747-30145880-4d9f-11ea-8c6c-5ca4a7845d8f.png)

**As you see, total does not match.**

There are two sale order lines that meet the search criteria
![image](https://user-images.githubusercontent.com/13234598/74336928-98633a00-4d9f-11ea-8f14-4ea3915ac94f.png)

And as you can see the grouped total is the sale order total multiplied by two

----

If we add another sale order line, we can see the grouped total is total x3
![image](https://user-images.githubusercontent.com/13234598/74337052-dd876c00-4d9f-11ea-9a4e-e77c11ae3bf3.png)

6.552 x 3 = 19656

![image](https://user-images.githubusercontent.com/13234598/74337131-04de3900-4da0-11ea-8711-aa7ed071dd68.png)

Tardo:
I can't reproduce the issue.... can you please record a video to see the steps to reproduce? Thanks!

JGarcia-Panach:
Tested in this URL:
http://3408529-12-0-2931ed.runbot2.odoo-community.org/web/login

This is the Sale order I use to test: 
![image](https://user-images.githubusercontent.com/13234598/74639513-53a81c00-516e-11ea-87b3-574249b553f8.png)

I change Unit of measure in order to use advanced search (searching lines with Unit of measure = Dozens)

Now the video:
![error advanced](https://user-images.githubusercontent.com/13234598/74640168-6f5ff200-516f-11ea-8529-b03186824145.gif)

As you can see, when you use group by combined with advanced search, the numbers don't match.
When you use group by, it multiplies by the number of hits

In this case (SO021) yo have 2 sale order lines:
![image](https://user-images.githubusercontent.com/13234598/74640484-e7c6b300-516f-11ea-9115-db066377178e.png)

Where total is 18,604.80

If you multily TOTAL by the number of hits (2 order lines found where Unit of measure contains "dozen"): 18,604.80 * 2 = 37,209.60

[...]

PR #1551, opened 2020-03-18
[13.0][MIG] web_advanced_search: Migration to 13.0

Tardo:
As commented in https://github.com/OCA/web/issues/1517 this module is affected by https://github.com/odoo/odoo/issues/47950

cc @tecnativa TT19820


chienandalu (commented):
I miss the domain editor in debug mode: 
image

chienandalu (commented):
Also field names in debug mode (surely the same issue):
image

yajo:
@chienandalu AFAIK those are upstream features...

chienandalu:
I tested the standard widget in filters and it was ok

web_advanced_search/static/src/js/web_advanced_search.js
```diff
@@ -119,13 +115,19 @@ odoo.define("web_advanced_search", function(require) {
...
+            const proposition = new AdvancedSearchProposition(
                 this,
-                this.dataset.model,
+                this._modelName,
                 event.data.domain
             );
-            this.propositions = [proposition];
-            this._commitSearch();
+            // Necessary to ensure that the porposition have the 'fieldSelector'
+            // is filled
```
yajo, line 124:
```suggestion
            // Necessary to ensure that the proposition has the 'fieldSelector'
            // and it is filled
```

web_advanced_search/static/src/js/web_advanced_search.js
```diff
@@ -119,13 +115,19 @@ odoo.define("web_advanced_search", function(require) {
...
+            _.defer(
+                function() {
+                    this.propositions = [proposition];
+                    this._commitSearch();
+                }.bind(this)
```
yajo, line 129: Better use an arrow function, which is bound by default:

```suggestion
                () => {
                    this.propositions = [proposition];
                    this._commitSearch();
                }
```

web_advanced_search/static/src/js/web_advanced_search.js
```diff
@@ -283,7 +284,9 @@ odoo.define("web_advanced_search", function(require) {
                 return FieldManagerMixin._applyChanges.apply(this, arguments);
             }
 
-            return $.Deferred().resolve();
+            return new Promise(resolve => {
+                resolve();
+            });
```
yajo, line 289: I think this is not the same.

We were returning a resolved deferred before, so it should be something like:

```suggestion
            return new Promise().resolve();
```

web_advanced_search/static/src/js/human_domain.js
```diff
@@ -1,69 +1,58 @@
...
-                value = _.str.sprintf('"%s"', this.value);
+            const chain = [];
+            let operator = this.operator_mapping[this.operator],
+                value = `"${this.value}"`;
             // Humanize chain
-            this.chain.split(".").forEach(function(element, index) {
+            const chain_splitted = this.chain.split(".");
+            const len = chain_splitted.length;
+            for (let x = 0; x < len; ++x) {
+                const element = chain_splitted[x];
```
yajo: Is this worth it? Why not just:

```js
             this.chain.split(".").forEach((element, index) => {
```

Almost like before, but you avoid having to manually bind the function.

yajo (changes requested):

alexey-pelykh:
Just in case https://github.com/OCA/web/pull/1554 will be accepted and merged prior this one, it would be great to include in v13

yajo:
On its way...

manuelcalerosolis (approved):
Functional review!

pedrobaeza:
@Yajo is it OK for you?

@Tardo can you add #1554?

yajo (approved):

oscars8a:
Hi everyone,

This PR correct actually the problem with advanced search filter?

I have rebuilt the runbot 1551 and I could prove that the problem persists.

Which one is the state of this problem for Odoo 13.0 and 12.0?
I have this problem in a Odoo 12.0 version.

I put a git that show the problem in the runbot 1551, and this is the link http://3465288-1551-0732a9.runbot2-3.odoo-community.org/

Thanks a lot.

Peek 2021-02-15 13-10

pedrobaeza:
No, that problem is not notified anywhere AFAIK

PR #1552, opened 2020-03-19
[WIP][13.0][MIG] web_company_color

JoanSForgeFlow:
**web_company_color** migration to v13.
- web_widget_color dependency has been removed, now it's used new odoo's one

@ForgeFlow @LoisRForgeFlow


AdriaGForgeFlow (changes requested):
It's not updating the colors for me. I don't know if I'm doing something wrong.

web_company_color

LoisRForgeFlow:
@Tardo @simahawk @Yajo Hi guys, we are kind of stuck with this migration, it thinks our limited frontend skills are overwhelmed with this.

We cannot make it work for more than the default company, we don't really understand how it worked in v12... how Odoo could decide which attachment to use depending on the active company? they all were created for the base one plus there were no company rules for the attachments. If we understand this maybe we would be able to find the root cause in v13.

Can you bring some light here? Thanks you!

web_company_color/view/assets.xml
```diff
@@ -0,0 +1,25 @@
...
+<odoo>
+    <template
+        id="assets_backend"
+        name="web_company_color assets"
+        inherit_id="web.assets_backend"
+    >
+        <xpath expr=".">
+            <t
+                t-set="company_id"
+                t-value="request.env.user and request.env.user.company_id"
```
AdriaGForgeFlow, line 15: This maybe should change to this:
``` python
t-value="request.env.company"
```

Still not working after trying

Tardo:
I have bad news.... after debugging i see that assets template only is "compiled" the first time, so you will load the 'company css' for the first user.

I don't know if exists an option to force "recompile" the template all times :/

How the module works:
- When the module loads, checks if exists an css for all existing companies (if not exists, creates a new one).
- When the user load a page, in the assets template, we get the scss url for the active user (really for the user active company).

simahawk:
Maybe as a quick temporary fix we could place the file outside assets and load it alone. It will be cached by the browser anyway. Sub-optimal but should work. Meanwhile we can seek for a better solution.

simahawk:
@Tardo AFAIR we could use the same machinery used by the website editor: there you can edit assets on the fly and they get reloaded. Have you checked it as well? (I didn't, have no time ATM).

LoisRForgeFlow:
@Tardo then we may need to refactor the module :cry:. Thanks for looking into it!

Tardo:
Good news! done in https://github.com/OCA/web/pull/1568

PR #1554, opened 2020-03-21
[12.0][IMP] web_advanced_search: support hierarchy filtering

alexey-pelykh:
Add `is child of` and `is parent of` operators to filtering
image

### DIFF
--- a/web_advanced_search/static/src/js/web_advanced_search.js
+++ b/web_advanced_search/static/src/js/web_advanced_search.js
@@ -1,5 +1,6 @@
 /* Copyright 2015 Therp BV <http://therp.nl>
  * Copyright 2017-2018 Jairo Llopis <jairo.llopis@tecnativa.com>
+ * Copyright 2020 Brainbean Apps (https://brainbeanapps.com)
  * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
 
 odoo.define("web_advanced_search", function (require) {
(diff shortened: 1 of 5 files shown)


alexey-pelykh:
@Tardo @Yajo @chienandalu maybe you'd be interested in this

web_advanced_search/static/src/js/web_advanced_search.js
```diff
@@ -153,14 +156,23 @@ odoo.define("web_advanced_search", function (require) {
             // To make widgets work, we need a model and an empty record
             FieldManagerMixin.init.call(this);
             this.trigger_up("get_dataset");
-            // Make equal and not equal appear 1st and 2nd
+            // Unfortunately, it's impossible to check if model supports these
```
legalsylvain, line 159: Why ?  Checking if model has parent_id field is not possible ?

legalsylvain (commented):
One question. LGTM otherwise.
Thanks !
(Code review / no test)

web_advanced_search/static/src/js/web_advanced_search.js
alexey-pelykh: Technically, we should check if `_parent_store = True` or check for field with name stored in `_parent_name` (by default, that's `parent_id`), yet I haven't found any clues on how to get those in frontend or example of those values being used in js.

legalsylvain: OK ! Could you just add it in the roadmap section ?

alexey-pelykh: Done

legalsylvain (approved):

yajo (approved):

PR #1559, opened 2020-03-23
[FIX] web_widget_numeri_step: Change click timeout

Tardo:
Previous this commit, the task for auto-step can fall in a infinite loop. Now only creates a new task if no other task is running.

cc @tecnativa TT21495


legalsylvain (approved):
make sense. Thanks for the fix.

Code review / no test.

yajo (approved):

Issue #1469, opened 2019-12-26
web_responsive wrong shortcuts in template


kos94ok-3D:
Steps to reproduce:
1. Install web_responsive
2. Open Shortcuts menu
![изображение](https://user-images.githubusercontent.com/33932458/71486627-fb186a80-281f-11ea-8cf3-898f04ada81b.png)
![изображение](https://user-images.githubusercontent.com/33932458/71486641-0a97b380-2820-11ea-9efd-551ba597d22c.png)

I think is a good idea update UserMenu.shortcuts template
Hi, @VincentSchippefilt, @Yajo, @Tardo  what do you think about it?

yajo:
Absolutely.

Tardo:
Feel free to do a PR ;)

PR #1560, opened 2020-03-23
[FIX] web_responsive: fix hotkey combinations in a shortcut template

kos94ok-3D:
Before:
изображение

After:
изображение

Closed: #1469

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -24,6 +24,7 @@
     'qweb': [
         'static/src/xml/apps.xml',
         'static/src/xml/form_view.xml',
+        "static/src/xml/menu.xml",
         'static/src/xml/navbar.xml',
         'static/src/xml/document_viewer.xml',
     ],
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
Can you avoid a total replacing so that you don't invalidate possible additions of extra modules?

kos94ok-3D:
> Can you avoid a total replacing so that you don't invalidate possible additions of extra modules?

Is it ok?

pedrobaeza (approved):
Great, thanks!

Tardo (approved):
Thanks!

Cedric-Pigeon:
@pedrobaeza @Yajo I think this one should be forward-ported to V13, isn't it?

pedrobaeza:
Yes, please. I think it's only cherry-picking the commit.

Cedric-Pigeon:
ok thanks, we'll do in the next coming days.

PR #1563, opened 2020-03-31
[12.0][ADD] proxy_action

kevinkhao:


legalsylvain (changes requested):
I think that the description should be improved : 
- as pywebdriver is not a OCA lib 
- as this module should work with iotbox also, if I understand correctly

Also maybe make a little developper section to enumerate functions available.

hparfr (approved):

PR #1570, opened 2020-04-07
[12.0][FIX][web_m2x_options] when all is undefined, fall back to core's default

hbrunn:
the module as it is fails web's tests, namely https://github.com/OCA/OCB/blob/12.0/addons/web/static/tests/fields/relational_fields_tests.js#L1002

What I propose here is to fall back on https://github.com/OCA/OCB/blob/12.0/addons/web/static/src/js/fields/relational_fields.js#L129 if none of the options are set.

### DIFF
--- a/web_m2x_options/static/src/xml/base.xml
+++ b/web_m2x_options/static/src/xml/base.xml
@@ -6,7 +6,7 @@
     <t t-extend="FieldMany2One">
         <t t-jquery=".o_external_button" t-operation="attributes">
             <attribute name="t-if">
-                !(widget.nodeOptions.no_open || widget.nodeOptions.no_open_edit)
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

NL66278 (approved):
:+1: LGTM

hbrunn:
I disagree with codecov in this case.

pedrobaeza:
That can be because there has been a period where it didn't receive data (due to a problem in MQT), so now comparison are not right.

hbrunn:
can we force merge somehow while still involving ocabot?

pedrobaeza:
Yes, codecov is not a requirement for ocabot merge

/ocabot merge patch

hbrunn:
hey ocabot, did you fall asleep in between? I'll try this again.

/ocabot merge patch

PR #1576, opened 2020-04-14
[IMP] web_tree_many2one_clickable: Don't break keyboard navigation

Tardo:
Before, you can't use "TAB" to navigate through all fields:
before_c

After, you can!
after_c

cc @tecnativa TT22262

### DIFF
--- a/web_tree_many2one_clickable/static/src/css/web_tree_many2one_clickable.scss
+++ b/web_tree_many2one_clickable/static/src/css/web_tree_many2one_clickable.scss
@@ -1,10 +1,17 @@
 td.o_many2one_cell {
-  a {
-    margin-left: 0.5em;
-    visibility: hidden;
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza:
@Tardo this patch is breaking `decoration-` attributes over tree on the many2one fields. Example:
```xml
<tree decoration-muted="id">
    <field name="id" invisible="1"/>
    <field name="many2one_field"/>
</tree>
```

Tardo:
Thanks @pedrobaeza should be currently fixed here: https://github.com/OCA/web/pull/1585 and fixed @carlosdauden suggestions about the focus too

PR #1585, opened 2020-04-21
[12.0][FIX] web_tree_many2one_clickable: Compatible with decoration colors

Tardo:
Using ``<span>`` tag again = Affected by colors
Improved keyboard navigation, now the element is not focusable in readonly mode

cc @tecnativa TT22262

### DIFF
--- a/web_tree_many2one_clickable/static/src/css/web_tree_many2one_clickable.scss
+++ b/web_tree_many2one_clickable/static/src/css/web_tree_many2one_clickable.scss
@@ -1,14 +1,7 @@
 td.o_many2one_cell {
-    a {
-        color: $o-main-text-color !important;
-        &:hover {
-            color: $o-main-text-color !important;
-        }
-
-        .many2one_clickable {
-            margin-left: 0.5em;
-            visibility: hidden;
-        }
+    .many2one_clickable {
+        margin-left: 0.5em;
+        visibility: hidden;
     }
 
     &:hover a.many2one_clickable {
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The name of the module in the title and commit is incorrect.

pedrobaeza:
Are both patches applicable to v13?

Tardo:
Oops! changes done

Tardo:
@pedrobaeza after some tests in 13.0 keyboard navigation appears to work fine without this fix.

pedrobaeza (approved):

yajo (approved):
The code looks good, and functional test works, although I'd appreciate at least a screenshot that lets us see what this is fixing... I don't really know what to test.

pedrobaeza:
It's fixing the color of the row through decorators.

/ocabot merge patch

pedrobaeza:
@Yajo do you know why pylint fails?

yajo:
I see this: https://travis-ci.org/github/OCA/web/jobs/678053493#L771

```
BlockingIOError: [Errno 11] write could not complete without blocking
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
BlockingIOError: [Errno 11] write could not complete without blocking
```

My bet is that pylint doesn't work fine when being executed in parallel, although I never saw this before. That can be disabled in pre-commit. Any other ideas @sbidoul @moylop260?

pedrobaeza:
This is not being executed in pre-commit. This is 12.0. I think there's any code here to be analyzed that is blowing up the linter.

yajo:
Oh true.

This could have been some race condition. What about just retrying for now?

pedrobaeza:
It's happening the same on #1589, so it's something structural. I'm analyzing right now.

pedrobaeza:
Fixed with the pylint cleaning I have made:

/ocabot merge patch

PR #1587, opened 2020-04-22
[12.0] [MIG] web_advanced_filter

hbrunn:
This addon allows users to apply set operations on filters: Remove or add
certain ids from/to a selection, but also to remove or add another filter's
outcome from/to a filter. This can be stacked, so the filter domain can be
arbitrarily complicated.

The math is hidden from the user as far as possible, in the hope it's still
user friendly.

---

I renamed https://github.com/OCA/web/tree/7.0/web_advanced_filters in the process.


ivantodorovich (approved):
LGTM :+1:

lfreeke:
filters

I noticed that sometimes with a group by the advanced filter is not working correctly.
In this case I filtered on 'Individual' and did a group by on 'Company' in the Contacts menu. I save the advanced filter. 
I refresh the Contacts menu and I got to Favorites and click on my Test filter, I see an empty kanban board.

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+
+class IrFilters(models.Model):
+    _inherit = 'ir.filters'
+    _evaluate_before_negate = ['one2many', 'many2many']
+
+    is_frozen = fields.Boolean('Frozen', compute='_compute_is_frozen')
+    union_filter_ids = fields.Many2many(
+        'ir.filters', 'ir_filters_union_rel', 'left_filter_id',
+        'right_filter_id', 'Add result of filters',
+        domain=['|', ('active', '=', False), ('active', '=', True)],
```
janverb, line 20: This works too:
```suggestion
        context={'active_test': False},
```
(idem for `complement_filter_ids`)
The field context isn't inherited by the returned recordset, so I think it's exactly equivalent.

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+    )
+    evaluate_always = fields.Boolean(
+        'Always evaluate this filter before using it',
+        compute='_compute_flags',
+        help='This is necessary if this filter contains x2many fields with'
+        '_auto_join activated',
+    )
+    save_as_public = fields.Boolean(
+        'Share with all users', default=False,
+        compute=lambda self: None, inverse=lambda self: None,
```
janverb, line 43: Could this field have `store=False` instead of being computed?

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+                'user_id',
+                False if values.get('save_as_public') else self.env.user.id,
+            )
+        return super(IrFilters, self).create(vals_list)
+
+    @api.multi
+    def _evaluate(self):
+        self.ensure_one()
+        return self.env[self.model_id].with_context(
+            safe_eval(self.context)
```
janverb, line 168: This destroys the context of the caller, including the `lang` and `tz` keys. The default for `context` is `'{}'`.
Perhaps:
```suggestion
            **safe_eval(self.context)
```
I'm also wondering if the old context should be restored after the `.search()`.

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+            if removed_filters:
+                self.env.cr.execute(
+                    '''delete from ir_filters
+                    where
+                    not active and id in %s
+                    and not exists (select right_filter_id
+                    from ir_filters_union_rel where left_filter_id=id)
+                    and not exists (select right_filter_id
+                    from ir_filters_complement_rel where
+                    left_filter_id=id)''',
```
janverb, line 198: Isn't this backwards? On my system `left_filter_id` seems to contain the main filter and `right_filter_id` the helper filters.

web_advanced_filter/static/src/js/web_advanced_filter.js
```diff
@@ -0,0 +1,256 @@
...
+                default_context: JSON.stringify(ctx),
+                default_filter_id:
+                this.search_view.dataset.context.default_filter_id || false,
+            },
+        },
+        {
+            on_close: this.search_view._reloadFilters.bind(this.search_view),
+        });
+    },
+    _advanced_filter_update: function(event) {
```
janverb, line 195: Opening the "Advanced filters" menu twice always reveals both sections. This method isn't executed when you do that, so I'm guessing some other part is touching the visibility and the logic needs to be triggered there as well.

web_advanced_filter/static/src/js/web_advanced_filter.js
```diff
@@ -0,0 +1,256 @@
...
+
+SearchView.include({
+
+    start: function () {
+        var self = this;
+        return this._super.apply(this, arguments).then(function () {
+            self.advanced_filters_menu = self._createAdvancedFiltersMenu();
+            return self.advanced_filters_menu.insertBefore(
+                self.favorite_menu.$el
+            );
```
janverb, line 223: Why does this return a value?

web_advanced_filter/views/ir_filters.xml
```diff
@@ -0,0 +1,49 @@
...
+                <field name="id" invisible="True" />
+                <group string="Add the result of following filters">
+                    <field name="union_filter_ids" nolabel="1" domain="[('user_id', 'in', [False, uid]), ('id', '!=', id), ('model_id', '=', model_id)]" />
+                </group>
+                <group string="Remove the result of following filters">
+                    <field name="complement_filter_ids" nolabel="1" domain="[('user_id', 'in', [False, uid]),('id', '!=', id), ('model_id', '=', model_id)]" />
+                </group>
+            </group>
+            <field name="domain" position="attributes">
+                <attribute name="attrs">{'readonly': ['|', ('union_filter_ids', '!=', [[6, False, []]]), ('complement_filter_ids', '!=', [[6, False, []]])]}</attribute>
```
janverb, line 24: Checking in this way that a 2many relation is empty doesn't seem to work, `domain` is always read-only and `domain_this` always visible no matter the filter.

web_advanced_filter/views/ir_filters.xml
```diff
@@ -0,0 +1,49 @@
...
+            </field>
+            <field name="domain" position="after">
+                <field name="domain_this" attrs="{'invisible': [('union_filter_ids', '=', [[6, False, []]]), ('complement_filter_ids', '=', [[6, False, []]])]}" />
```
janverb, line 27: Should this have `widget="domain"` like the `domain` field?

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+        for this in self:
+            try:
+                domain = safe_eval(this.domain)
+            except (SyntaxError, TypeError, ValueError, NameError):
+                domain = [expression.FALSE_LEAF]
+            this.is_frozen = (
+                len(domain) == 1 and
+                expression.is_leaf(domain[0]) and
+                domain[0][0] == 'id'
+            )
```
janverb, line 59: Should this also check that `domain[0][1] == 'in'`? The rest of the module seems to be written with that in mind. A filter that uses `'not in'` seems useful enough to anticipate, though I don't know how easy it is to make one in practice.

web_advanced_filter/wizards/ir_filters_combine_with_existing.py
```diff
@@ -0,0 +1,68 @@
...
+    filter_id = fields.Many2one('ir.filters', 'Filter', required=True)
+
+    @api.multi
+    def button_save(self):
+        self.ensure_one()
+        this = self
+        domain = json.loads(this.domain)
+        is_frozen = (len(domain) == 1 and
+                     expression.is_leaf(domain[0]) and
+                     domain[0][0] == 'id')
```
janverb, line 30: The comment for `ir.filters._compute_is_frozen` applies here too, especially given the merging later on.

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+        '''combine our domain with all domains to union/complement,
+        this works recursively'''
+        def eval_n(domain):
+            '''parse a domain and normalize it'''
+            try:
+                domain = safe_eval(domain)
+            except (SyntaxError, TypeError, ValueError, NameError):
+                domain = [expression.FALSE_LEAF]
+            return expression.normalize_domain(
+                domain or [expression.FALSE_LEAF])
```
janverb, line 76: This would be useful in `_compute_is_frozen` and `_compute_flags`, so maybe it should be a method so it can be used in those.
`_compute_flags` doesn't currently catch exceptions. If that's not deliberate then using a common method would help with that too.

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+
+        for this in self:
+            if this.model_id not in self.env:
+                this.domain = '[]'
+                _logger.error('Unknown model %s used in filter', this.model_id)
```
janverb, line 81: Maybe the filter ID should be logged too
(line 132 too)

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+import logging
+from odoo import _, api, fields, models
+from odoo.tools.safe_eval import safe_eval
+from odoo.osv import expression
+
+
+_logger = logging.getLogger(__name__)
+
+
+class IrFilters(models.Model):
```
janverb, line 12: I think it could have been better to create a new model to hold union filters and complement filters, instead of making them `ir.filters` records. As far as I see there's very little overlap between the fields that ordinary filters use and that helper filters use, it's just `domain` (and technically `model_id` but that's unnecessary).
That would remove the need to mess around with `active`, remove the need to put timestamps in names, and leave a cleaner database when uninstalling the module.
I don't know if it's worth changing at this point though.

web_advanced_filter/models/ir_filters.py
```diff
@@ -0,0 +1,215 @@
...
+                    ])
+            this.domain = str(expression.normalize_domain(domain))
+
+    @api.multi
+    def _inverse_domain(self):
+        for this in self:
+            this.domain_this = this.domain
+
+    @api.multi
+    @api.depends('domain')
```
janverb, line 117: I would add `'model_id'`, just in case

web_advanced_filter/static/src/xml/web_advanced_filter.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates>
+    <t t-name="AdvancedFilter">
+        <div role="separator" class="dropdown-divider"/>
+        <button type="button" class="dropdown-item o_advanced_filter o_closed_menu" aria-expanded="false">Whole selection (criteria)</button>
+        <div class="dropdown-item-text o_advanced_filter-domain">
+            <button type="button" class="dropdown-item o_advanced_filter-add-new">To new filter</button>
+            <button type="button" class="dropdown-item o_advanced_filter-add-existing">To existing filter</button>
+            <button type="button" class="dropdown-item o_advanced_filter-remove">Remove from existing filter</button>
```
janverb, line 9: This wording confused me at first.
Instead of "To existing filter" I think I'd write "Add to existing filter", to differentiate it from "To new filter" and make it clear that you're not replacing anything.
I first interpreted "Remove from existing filter" as something you'd use to remove a helper filter that's already part of the filter. Maybe "Subtract from existing filter" would be better. Or another word instead of "subtract".

janverb (changes requested):
Works well, but I had a bunch of notes about fiddly details

web_advanced_filter/models/ir_filters.py
hbrunn: should work nowadays: https://github.com/OCA/OCB/blob/12.0/odoo/fields.py#L1133

web_advanced_filter/models/ir_filters.py
hbrunn: you can use this module in two ways: Either you use the JS part that creates extra `ir.filters` records on the fly, but another use case is you already have carefully curated filters (possibly created/managed/administered by different people) being used on their own, that you combine manually as admin in the settings menu. The latter won't be possible then.

hbrunn:
@janverb thanks for your review. I stopped working on the project involving this module however, so I probably won't invest time in this PR any time soon, and can just hope somebody else picks up your points. Keeping this open until there's a new PR with the fixes

PR #1592, opened 2020-04-25
[12.0][MIG] web_widget_float_formula

alexey-pelykh:


legalsylvain:
Hi. Is it ready to review ?

alexey-pelykh:
@legalsylvain I'd say almost, Monetary field is failing tests still, integer and float should be good

legalsylvain:
first functional test, on runbot, with sale installed.
- works fine on float field
- the computation works also on monetary field, but not the "redisplay" of the formula, if the user set the focus out and on again.
So, it is not consistent. Is it desired ? or is it a side effect ? 

Anyway. .Thanks for this great job & refactoring.

alexey-pelykh:
@legalsylvain thanks for testing, `monetary` field is now working as expected

alexey-pelykh:
@pedrobaeza is it me, or previously js tests were executed by runbot automatically?

tonygalmiche (commented):
I tested on my personal computer
It works with integers, floats and monetary
But for integers, the formula = 3.9 / 2 = 1
It's normal ?
(I did this test with the sequence field of mrp.bom)
Thank you for this work

alexey-pelykh:
@tonygalmiche what behaviour would you expect?

tonygalmiche:
@alexey-pelykh ,
= 3.9 / 2 = 1.95 = 2 not 1 ?

alexey-pelykh:
@tonygalmiche it depends on if that's rounding or if that's clamping, both answers are valid

legalsylvain:
> But for integers, the formula = 3.9 / 2 = 1

the widget is making computation interpretating formula with a Javascript eval function. if 3.9 / 2 = 2 in Javascript, so it's OK for me. 
- I don't think it works differently in V8.
- I don't think that Odoo V13 works differently. (no tested, did you tested ?)
- I don't think we can do anything to "fix" that. 

kind regards.

alexey-pelykh:
@legalsylvain it seems to mimic py eval or even calls it? v13 should behave the same I guess.

tonygalmiche (approved):
no problem.
It was a small remark

alexey-pelykh:
Finally, tests are working as well

alexey-pelykh:
@legalsylvain @tarteo a review would be much appreciated!

ecino (approved):
I did functional testing and it works fine :+1:

alexey-pelykh:
@pedrobaeza can we just merge it?

PR #1598, opened 2020-05-06
[FIX] web_timeline: Change vis-timeline library import

Tardo:
Fixes https://github.com/OCA/project/pull/656 reported issues

cc @tecnativa

PR #1599, opened 2020-05-06
[13.0] web_timeline: Fix crash when loading due to mismatched CSS reference and update icon

chrisandrewmann:
Fix crash when loading due to mismatched reference to CSS and update icon to hourglass

-Fix crash when loading timeline as vis-timeline-graph2d.min.css and vis-timeline-graph2d.css were both referred to but min version was missing
-Deleted non minified version and added min version, update references to refer to minified version in all cases
-Changed icon for timeline view to hourglass as I feel works better for both enterprise and community Odoo-The old icon was identical to the Gantt icon used in enterprise and is confusing


tonygalmiche:
with this pr, I have this icon : 
image

but that solved the bug of this pr :)
-> https://github.com/OCA/project/pull/656

pedrobaeza:
Closing in favor of #1598

PR #1602, opened 2020-05-18
[12.0] ADD web_pwa

eLBati:
Make Odoo an installable Progressive Web Application.

Progressive Web Apps provide an installable, app-like experience on desktop and mobile that are built and delivered directly via the web.
They're web apps that are fast and reliable. And most importantly, they're web apps that work in any browser.
If you're building a web app today, you're already on the path towards building a Progressive Web App.

After having installed this module, browsing your odoo on mobile you will be able to install it as a PWA.

When a Progressive Web App is installed, it looks and behaves like all of the other installed apps.
It launches from the same place that other apps launch. It runs in an app without an address bar or other browser UI.
And like all other installed apps, it's a top level app in the task switcher.

In Chrome, a Progressive Web App can either be installed through the three-dot context menu.

This module also provides a "install" button in Odoo header navigation bar.

Remote Debugging Android Devices and Access Local Servers
------------------------------------------------------------------------------------

https://developers.google.com/web/tools/chrome-devtools/remote-debugging/local-server
https://developers.google.com/web/tools/chrome-devtools/remote-debugging


AdriaGForgeFlow (approved):
Tested on iphone with safari and works as expected. I opened Odoo in Safari then pressed the share icon and then Add to Home Screen:



After that a new icon is added to home screen and you can use Odoo without the need of typing URL again and without the navigation bar which is great.

Great addition IMO! Nice work. :+1:

pedrobaeza:
Is this interesting to be included directly in `web_responsive`? cc @Tardo @Yajo

web_responsive_pwa/controllers/main.py
```diff
@@ -0,0 +1,70 @@
+from odoo.http import request, Controller, route
+
+SERVICE_WORKER_SCRIPT = """
```
simahawk: Instead of harcoding this in python:

1. move this to `service-worker.js` file
2. rename controller to `service-worker-setup.js`
3. make it set only variables (eg: files_to_cache) or render a template like https://github.com/odoo/odoo/blob/13.0/addons/point_of_sale/views/pos_templates.xml#L30
4. use those variables in `service-worker.js`

simahawk (changes requested):
nice!

web_responsive_pwa/controllers/main.py
eLBati: Thanks, done with template:
https://github.com/OCA/web/compare/b16aec8a6424de99027b2f6c2edeb776080d44a7..51bbc19d7008a1cc59c5219d074f5fa293aff646

kenayagi:
Very nice, thanks! :smiley:
Is this also useful to get native Android notifications?

eLBati:
Hi @kenayagi yes it is:
https://codelabs.developers.google.com/codelabs/pwa-integrating-push/index.html#0

web_responsive_pwa/static/src/js/pwa_install.js
```diff
@@ -0,0 +1,56 @@
...
+"use strict";
+
+var core = require('web.core');
+var session = require('web.session');
+var SystrayMenu = require('web.SystrayMenu');
+var Widget = require('web.Widget');
+var QWeb = core.qweb;
+
+if ('serviceWorker' in navigator) {
+  window.addEventListener('load', () => {
```
yajo: Sorry, only ES5 syntax is allowed for v12. Please make linters happy: https://travis-ci.org/github/OCA/web/jobs/688422849#L805-L806

This applies to the whole file.

web_responsive_pwa/static/src/js/pwa_install.js
```diff
@@ -0,0 +1,56 @@
...
+
+var PwaInstall = Widget.extend({
+    name: 'pwa_install',
+    template:'web_responsive_pwa.systray.install',
+    events: {
+    },
+    start: function () {
+        window.addEventListener('beforeinstallprompt', this.saveBeforeInstallPromptEvent);
+        document.getElementById('pwa_install_button').addEventListener('click', this.installPWA);
+        return this._super();
```
yajo:
```suggestion
        return this._super.apply(this, arguments);
```

web_responsive_pwa/static/src/js/pwa_install.js
```diff
@@ -0,0 +1,56 @@
...
+let deferredInstallPrompt = null;
+
+var PwaInstall = Widget.extend({
+    name: 'pwa_install',
+    template:'web_responsive_pwa.systray.install',
+    events: {
+    },
+    start: function () {
+        window.addEventListener('beforeinstallprompt', this.saveBeforeInstallPromptEvent);
+        document.getElementById('pwa_install_button').addEventListener('click', this.installPWA);
```
yajo:
```suggestion
        this.el.addEventListener('click', this.installPWA);
```

web_responsive_pwa/static/src/js/pwa_install.js
```diff
@@ -0,0 +1,56 @@
...
+        return this._super();
+    },
+    saveBeforeInstallPromptEvent: function(evt) {
+        deferredInstallPrompt = evt;
+        document.getElementById('pwa_install_item').removeAttribute('hidden');
```
yajo:
```suggestion
        this.el.removeAttribute('hidden');
```

web_responsive_pwa/static/src/js/pwa_install.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+    installPWA: function (evt) {
+        deferredInstallPrompt.prompt();
+        // Hide the install button, it can't be called twice.
+        document.getElementById('pwa_install_item').setAttribute('hidden', true);
```
yajo:
```suggestion
        this.el.setAttribute('hidden', true);
```

web_responsive_pwa/static/src/json/manifest.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "Odoo PWA",
+  "short_name": "Odoo PWA",
+  "icons": [{
+    "src": "/web_responsive_pwa/static/img/icons/icon-128x128.png",
+      "sizes": "128x128",
+      "type": "image/png"
```
yajo: Is it possible to just use 1 SVG icon?

web_responsive_pwa/static/src/xml/pwa_install.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<templates>
+    <t t-name="web_responsive_pwa.systray.install">
+        <li class="o_mail_systray_item" id="pwa_install_item" hidden="1">
+            <a title="Install PWA" href="#" role="button" id="pwa_install_button">
+                <i class="fa fa-download" role="img" aria-label="Install PWA"/>
```
yajo: Don't mobile browsers already include a button like this natively?

web_responsive_pwa/views/webclient_templates.xml
```diff
@@ -0,0 +1,68 @@
...
+        </xpath>
+    </template>
+    <template id="service_worker" name="PWA service worker">
+'use strict';
+const CACHE_NAME = '<t t-esc="pwa_cache_name"/>';
+const FILES_TO_CACHE = [
+  <t t-foreach="pwa_files_to_cache" t-as="file_to_cache">
+      '<t t-esc="file_to_cache"/>',
+  </t>
+];
```
yajo: Can't you just use a normal JS file for JS code and download this data from a normal JSON controller? :thinking:

web_responsive_pwa/controllers/main.py
```diff
@@ -0,0 +1,25 @@
+from odoo.http import request, Controller, route
+
+
+class PWA(Controller):
+
+    @route('/service-worker.js', type='http', auth="public")
```
yajo: According to a comment down below, possibly this controller should be renamed to something like this:

```suggestion
    @route('/web_responsive_pwa/cache.json', type='json', auth="public")
```

However, if it were to stay as it is, it should be renamed too:

```suggestion
    @route('/web_responsive_pwa/service-worker.js', type='http', auth="public")
```

yajo (changes requested):

web_responsive_pwa/controllers/main.py
```diff
@@ -0,0 +1,25 @@
...
+    def service_worker(self):
+        qweb = request.env['ir.qweb'].sudo()
+        assets = qweb._get_asset_nodes("web.assets_common", {}, True, True)
```
yajo: BTW why are we just caching these and not `web.assets_backend` for instance?

web_responsive_pwa/static/src/js/pwa_install.js
legalsylvain: Hi @Yajo. Could you elaborate why this syntax is better ? just a naive question. I see that syntax (``this._super()``) in some OCA / Odoo modules

yajo: Yes, of course :) 

It's the standard. It calls super() with any arguments that came in. Avoids the reader to have to think if super() needs args and this is a bug or not.

FernandoRomera: https://github.com/odoo/odoo/wiki/Javascript-coding-guidelines
...
Always use this._super.apply(this, arguments);
....

legalsylvain: thanks !

web_responsive_pwa/static/src/json/manifest.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "Odoo PWA",
```
legalsylvain: The name of the app is Odoo PWA. Do you think it is possible to have a setting for that name, in a ``ir.config.parameters`` for exemple ?
I guess rendering this file dynamically, via a controller...

legalsylvain (commented):
Hi @eLBati : thanks a lot for this great module ! 

**Test done on my phone :** (LineageOS, based on Android 7.1.2 ; on Samsung GT-I9300)

- on **Chromium (85.0.4149.0)**, I found the button to add a PWA on the desktop, and it works like a charm. 
- on **Firefox** 
-> i didn't found how to generate the shorcut via Firefox
-> I don't find the button in the header of Odoo. Could you point where this button is ? you are saying it is in the main header.

@pedrobaeza said :
> Is this interesting to be included directly in web_responsive? cc @Tardo @Yajo

In fact, inspecting the code, I don't see any reference to web_responsive. Maybe it is possible to change the name of the module to ``web_pwa`` and make a simple dependency to ``web`` module. don't you think ?

web_responsive_pwa/static/src/json/manifest.json
eLBati: It seems it's not: https://developer.mozilla.org/en-US/docs/Web/Manifest/icons

web_responsive_pwa/static/src/xml/pwa_install.xml
eLBati: Yes they do, this `Install PWA` button is at the moment little invasive, see the download button in the navigation bar and the "Add tho home screen" banner at the bottom:

Screenshot_20200519-153008


I added the button to give the opportunity to extend it with custom behaviors (like a more evident modal dialog for example).
I would like to read the opinion of other contributors about the default behavior of this functionality. We could also think to make it configurable (more or less invasive)...

web_responsive_pwa/views/webclient_templates.xml
eLBati: I think @simahawk was suggesting something similar, but I am not sure this would work. More info about service workers: https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers

Anyway if someone provides a patch , I test it gladly.

web_responsive_pwa/controllers/main.py
eLBati: Right, added

web_responsive_pwa/static/src/js/pwa_install.js
eLBati: Done, thanks

web_responsive_pwa/static/src/js/pwa_install.js
eLBati: Done, thanks

web_responsive_pwa/static/src/js/pwa_install.js
eLBati: Done, thanks

eLBati:
> Could you point where this button is ? you are saying it is in the main header

See https://github.com/OCA/web/pull/1602#discussion_r427338575
Also, check https://web.dev/customize-install/ and [installation criteria](https://web.dev/install-criteria/)

> In fact, inspecting the code, I don't see any reference to web_responsive. Maybe it is possible to change the name of the module to web_pwa and make a simple dependency to web module. don't you think ?

Right, I added the web_responsive dependency because with standard Odoo Community interface, not responsive, the experience would be very poor.
But releasing the dependency would give more freedom, it could also be used with Odoo Enterprise.

pedrobaeza:
There are 2 options IMO:

- `web_pwa` for that freedom you mention if no technical dependency.
-  Integrate it in `web_responsive` as extra feature. Does it take advantage of anything in this module?

eLBati:
> The name of the app is Odoo PWA. Do you think it is possible to have a setting for that name, in a ir.config.parameters for exemple ?
I guess rendering this file dynamically, via a controller

Right, done. Also for icons and colors

eLBati:
> Integrate it in web_responsive as extra feature

On second thought, I think `web_pwa` should be separated and not depend on `web_responsive`, as no technical dependency.
I added a phrase in `readme` suggesting to use a responsive layout.

pedrobaeza:
OK for me about `web_pwa`. Please check Travis.

eLBati:
> Please check Travis

Done

web_responsive_pwa/views/webclient_templates.xml
simahawk: @Yajo I discussed that already w/ @eLBati  (elsewhere) and he was not convinced it was going to work (for some sw caching issue AFAIR, I have no time now to investigate).

IMO we could even avoid service worker if it's just for caching because is not required despite you found docs only related to it. See https://developer.mozilla.org/en-US/docs/Web/API/Cache
> You don't have to use it in conjunction with service workers, even though it is defined in the service worker spec.

In any case, we could add a huge TODO in the roadmap and approve what we get here as the alpha stage of this thing.

web_responsive_pwa/static/src/xml/pwa_install.xml
yajo: I think that if browsers supply a native UI for this, we should remove it.

web_responsive_pwa/views/webclient_templates.xml
yajo: The TODO option seems OK for now :blush:

web_pwa/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2020 Lorenzo Battistini @ TAKOBI
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Progressive web application",
+    "summary": "Make Odoo a PWA",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
```
yajo, line 8: Didn't you say alpha?

web_pwa/controllers/main.py
```diff
@@ -0,0 +1,69 @@
...
+        assets = qweb._get_asset_nodes(asset_xml_id, {}, True, True)
+        urls = []
+        for asset in assets:
+            if asset[0] == 'link':
+                urls.append(asset[1]['href'])
+            if asset[0] == 'script':
+                urls.append(asset[1]['src'])
+        return urls
+
+    @route('/service-worker.js', type='http', auth="public")
```
yajo, line 17:
```suggestion
    @route('/web_pwa/service-worker.js', type='http', auth="public")
```

web_pwa/controllers/main.py
```diff
@@ -0,0 +1,69 @@
...
+        icon144x144 = config_param.get_param(
+            "pwa.manifest.icon144x144", "/web_pwa/static/img/icons/icon-144x144.png")
+        icon152x152 = config_param.get_param(
+            "pwa.manifest.icon152x152", "/web_pwa/static/img/icons/icon-152x152.png")
+        icon192x192 = config_param.get_param(
+            "pwa.manifest.icon192x192", "/web_pwa/static/img/icons/icon-192x192.png")
+        icon256x256 = config_param.get_param(
+            "pwa.manifest.icon256x256", "/web_pwa/static/img/icons/icon-256x256.png")
+        icon512x512 = config_param.get_param(
+            "pwa.manifest.icon512x512", "/web_pwa/static/img/icons/icon-512x512.png")
```
yajo, line 51: Please use a single SVG instead (or additionally). It seems it's possible: https://w3c.github.io/manifest/#example-7

web_pwa/__manifest__.py
legalsylvain: @Yajo : why do you say that ? once reviewed, I think that the development status will be Beta. 
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_development_status.rst#1introduction
regards.

yajo: I didn't say that :sweat_smile: 

See https://github.com/OCA/web/pull/1602#discussion_r427786417

web_responsive_pwa/static/src/json/manifest.json
legalsylvain: thanks !

LoisRForgeFlow (commented):
I tested the module and it works great (Android 10 + Brave browser) :clap:. My suggestions would be to make the  icon and title of the App editable, for instance through system parameters, so users could recognize their company icon and name.

web_pwa/__manifest__.py
legalsylvain: Didn't understood. "alpha" is for unstable module. Is it the case here ?

simahawk: I said "alpha" because for me the very 1st version of a module that nobody used in real life (AFAIK) and that likely will be refactored (see the discussion regarding sw) should be alpha.
That been said, @eLBati your call ;)

eLBati: I think Beta is more suitable for this

web_pwa/controllers/main.py
eLBati: Thanks, do we have an SVG version of the OCA logo?

eLBati:
@LoisRForgeFlow done: https://github.com/OCA/web/pull/1602#issuecomment-630890613

scigghia (approved):
:+1:

web_responsive_pwa/static/src/xml/pwa_install.xml
eLBati: I am not sure how to check this, unless we totally remove the `beforeinstallprompt` event handler.
I don't know if there are cases or browsers where the native UI does not show while `beforeinstallprompt` is fired (showing our "download" icon only).
Since the icon in this implementation is little invasive, I think we could keep it.

web_responsive_pwa/views/webclient_templates.xml
eLBati: Done

pedrobaeza:
At your service!

[oca_logo.zip](https://github.com/OCA/web/files/4663411/oca_logo.zip)

eLBati:
@pedrobaeza thanks!

@Yajo trying with
https://github.com/eLBati/web/compare/12.0-web_responsive_pwa...eLBati:12.0-web_responsive_pwa-test?expand=1
but the "add to home screen" banner does not show.

Can anybody test it ?

AaronHForgeFlow (approved):
:+1: Functional tests using Android 10 and Chrome. Works like a charm and the title is editable.

Edit: and works in v13 without any change :)

web_pwa/controllers/main.py
eLBati: Done

web_responsive_pwa/static/src/xml/pwa_install.xml
legalsylvain: Hi @eLBati. Regarding this button : what about to move it and set instead an new option in the top right dropdown menu 

My PoV : 
- this button is not usefull a lot of time, so it's not interesting to have it as a shortcut, allways accessible
- the icon is not very intuitive : it means "download" and the action is more an "installation"
- the alt text "Install PWA" is not available on mobile, so there's no way of knowing what this button is for, before clicking on it.

something like that : with "Install PWA" or "Install App"

image

What do you think ? Just a suggestion, not a blocking point.

kind regards.

yajo:
I'm testing in android with Samsung browser and it doesn't seem to be working. 

No install icon. See:
Screenshot_20200528-085655_Samsung Internet

This browser supports PWA, as you can see in this other page (see install button in the address bar):
Screenshot_20200528-085649_Samsung Internet

web_pwa/controllers/main.py
eLBati: Just noticed:

> The path of the provided scope ('/') is not under the max scope allowed ('/web_pwa/'). Adjust the scope, move the Service Worker script, or use the Service-Worker-Allowed HTTP header to allow the scope.

So, service worker must be served  at `/`

Modifying

yajo: Oh interesting detail.

eLBati: Done

web_responsive_pwa/static/src/xml/pwa_install.xml
eLBati: @legalsylvain I think you are right. Done:

Screenshot from 2020-05-28 12-22-05

eLBati:
> I'm testing in android with Samsung browser and it doesn't seem to be working

@Yajo maybe this depended on https://github.com/OCA/web/pull/1602#discussion_r431701870
or not matching [installation criteria](https://web.dev/install-criteria/), probably `Be served over HTTPS`

yajo:
Oh that must be it.

Would it make sense to allow to install this even for a public user that's browsing the website? :thinking:

eLBati:
> Would it make sense to allow to install this even for a public user that's browsing the website?

:thinking:  I would do  a `website_pwa` module for that, mainly because the set of assets to register and cache would be different.
Considering the currently cached files

image

front-end pages would have a different set, making it a different application, from the service worker's point of view.

pedrobaeza:
You should count portal users and put also assets on that resources, as portal doesn't require website.

eLBati:
Right, so, the name would be `portal_pwa`.

I  assume `web_pwa` will be used by back-end users (employees) and `portal_pwa` by front-end users (customers, suppliers...)

So, with a view to add other mobile functionalities, like [notifications](https://github.com/OCA/web/pull/1602#issuecomment-630596401), the 2 modules will contain different sets of functionalities.

Also, the 2 applications could be branded differently, according to their audience.

pedrobaeza:
Great, can you add it to the ROADMAP as reminder?

eLBati:
@pedrobaeza done

pedrobaeza:
@Yajo @simahawk is it OK for you then?

simahawk (approved):

JordiBForgeFlow (approved):
Works like a charm! @eLBati in your experience is this easy to migrate to v13?

eLBati:
@JordiBForgeFlow it should be: https://github.com/OCA/web/pull/1602#pullrequestreview-417259079 :smiley:

yajo (approved):

eLBati:
Just noticed that there is another `web_pwa` at apps.odoo.com :confused:

Does anybody know how to publish the module in these cases?

pedrobaeza:
Not possible I'm afraid, unless we rename it, but sincerely I refrain some time ago of Odoo Apps that has censored us several modules (like debranding ones) or has problems like this one. I use OCA apps "store" without caring about the other one: https://odoo-community.org/shop

bealdav:
What's happen if you make a link in the branch to your module with a new name ?

pedrobaeza:
Uhm, I think that doesn't work.

JordiBForgeFlow:
But perhaps it's best practice to check the apps store before choosing a
name

eLBati:
> But perhaps it's best practice to check the apps store before choosing a
name

That's for sure :sweat_smile: 

Since  the module, at the moment, is not yet published on OCA shop nor PyPI, could we consider "hard" renaming it to something different, like `pwa`?

JordiBForgeFlow:
I think it's a good idea.

pedrobaeza:
I think we shouldn't care about others' names. The important thing is that in OCA, the name is not used. Please don't rename it, as I have already deployed the module in some of my installations.

simahawk:
In theory I agree w/ Pedro, in practice I'd say we should keep our visibility there.
If there's no way to avoid name clash I'd rename this to web_pwa_oca or pwa only (I prefer this) and for existing installations you can use a symlink (AFAIR it works but won't solve already installed module issue) or uninstall the existing module and install the new one (which is not a huge pain indeed).

eLBati:
So, created https://github.com/OCA/web/pull/1612

legalsylvain:
> But perhaps it's best practice to check the apps store before choosing a name

Maybe a new task for oca-git-bot ? ;-) CC : @pedrobaeza, @sbidoul 

_"You are proposing a module whose name is in conflict with the following module [web_pwa](https://apps.odoo.com/apps/modules/12.0/web_pwa/). Your module will not be visible in the Odoo Appstore."_

PR #1605, opened 2020-05-19
update bokehjs to version 2.0.2.

liunux4odoo:
it can be used with latest holoviews(1.13.2 in test) now.


liunux4odoo:
I have made a mistake: specify bokeh==2.0.2 in manifest.
this make unsatisfied external dependence error.
just remove it should be ok.

etobella:
You need to update the requirements in order to make runbot and travis work:

https://github.com/OCA/web/blob/12.0/requirements.txt#L2

PR #1610, opened 2020-06-02
[12.0][FIX] web_translate_dialog: Search using '=' to target right field

grindtildeath:
Without this commit, ir.translation are searched using a LIKE,
meaning that for a field called 'title', we'd get the translations
of all the fields beginning with title on this model.
Therefore, the returned value for field 'title' that is displayed
in the widget will be the value of the last result returned by
search_read, what could be anything beginning with title, and not
necessarily the field title itself.


vrenaville (approved):

guewen (approved):

leemannd (approved):
Thanks!

PR #1611, opened 2020-06-02
[FIX] #696 web_shortcut: open menu as well as action

### LINKED ISSUE #696: [10.0]BUG web_shortcut the submenu doesn't update sometimes
Click shortcut menu item to switch different page, but sometimes the left side submenu doesn't update along with action changing, and client may raise TypeError: this.active_view is null exception.

thomaspaulb:
Fixes #696


thomaspaulb:
@tarteo Would you like to review?

PR #1620, opened 2020-06-16
[13.0][ADD] web_orgchart_library

AdriaGForgeFlow:
This PR adds the module web_orgchart_library which adds the OrgChart library (https://github.com/dabeng/OrgChart) to be used in other modules.

cc ~ @ForgeFlow

### DIFF
--- a/web_orgchart_library/__manifest__.py
+++ b/web_orgchart_library/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2020 ForgeFlow S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Web OrgChart Library",
+    "category": "Hidden",
+    "summary": "This module includes the OrgChart Library.",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "version": "13.0.1.0.0",
+    "maintainers": ["AdriaGForgeFlow"],
+    "website": "https://github.com/OCA/web",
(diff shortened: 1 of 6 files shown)


LoisRForgeFlow (approved):
LGTM

JordiBForgeFlow (approved):

pedrobaeza:
Are you going to use this library for more than one module? Even so, not sure this should be bundled separately from the main module using it. You know also the possible version conflicts.

AdriaGForgeFlow:
> Are you going to use this library for more than one module? Even so, not sure this should be bundled separately from the main module using it. You know also the possible version conflicts.

The main problem is that, in my case, I'm going to use this in a module I plan to include in OCA/hr (hr_orgchart_overview) and that repository has the lib folder included in the .gitignore. How should I do this in your opinion?

pedrobaeza:
You should change the gitignore, or at least do it less restrictive (like only /lib). Who put initially that ignore? Maybe it was a mistake. Any way, Odoo already contains an HR org chart. Do you know it?

LoisRForgeFlow:
@pedrobaeza The Odoo org chart is a little bit simpler, we want a "full" org chart overview for the whole company in only one screen

pedrobaeza:
OK, but I think you should put it in the same module then.

AdriaGForgeFlow:
Closing this PR, library was included inside the module as requested by @pedrobaeza in here [OCA/hr/pull/849](https://github.com/OCA/hr/pull/849)

PR #1622, opened 2020-06-17
web_widget_x2many_2d_matrix: Fix show_column_totals option

gaikaz:
Commit a5e6acc3 introduced a bug, where `show_column_totals="0"` wouldn't work.
This is because JS ternary operator took node's attribute value as part of the condition.

As an example simplified JS code (before change):
`'0' || true ? '1' : ''` -> `'1'`
And this is what was probably meant originally (after change)
`'0' || (true ? '1' : '')` -> `'0'`

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
@@ -11,6 +11,7 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
     var X2Many2dMatrixRenderer = require(
         'web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer'
     );
+    var utils = require('web.utils');
 
     var WidgetX2Many2dMatrix = relational_fields.FieldOne2Many.extend({
         widget_class: 'o_form_field_x2many_2d_matrix',
@@ -43,11 +44,13 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                 node.field_label_x_axis || this.field_x_axis;
             this.field_label_y_axis =
                 node.field_label_y_axis || this.field_y_axis;
-            this.x_axis_clickable = this.parse_boolean(
-                node.x_axis_clickable || '1'
+            this.x_axis_clickable = utils.toBoolElse(
+                node.x_axis_clickable,
+                true
             );
-            this.y_axis_clickable = this.parse_boolean(
-                node.y_axis_clickable || '1'
+            this.y_axis_clickable = utils.toBoolElse(
+                node.y_axis_clickable,
+                true
             );
             this.field_value = node.field_value || this.field_value;
             // TODO: is this really needed? Holger?
@@ -66,13 +69,15 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                     this.field_value
                 ));
             }
-            this.show_row_totals = this.parse_boolean(
+            this.show_row_totals = utils.toBoolElse(
                 node.show_row_totals ||
-                (this.is_aggregatable(field_defs[this.field_value]) ? '1' : '')
+                this.is_aggregatable(field_defs[this.field_value]),
+                false
             );
-            this.show_column_totals = this.parse_boolean(
+            this.show_column_totals = utils.toBoolElse(
                 node.show_column_totals ||
-                (this.is_aggregatable(field_defs[this.field_value]) ? '1' : '')
+                this.is_aggregatable(field_defs[this.field_value]),
+                false
             );
         },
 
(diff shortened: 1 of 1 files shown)


web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -68,11 +68,11 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
             }
             this.show_row_totals = this.parse_boolean(
                 node.show_row_totals ||
```
Tardo, line 70:
```suggestion
                Number(node.show_row_totals) ||
```

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -68,11 +68,11 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
...
-                this.is_aggregatable(field_defs[this.field_value]) ? '1' : ''
+                (this.is_aggregatable(field_defs[this.field_value]) ? '1' : '')
             );
             this.show_column_totals = this.parse_boolean(
                 node.show_column_totals ||
```
Tardo, line 74:
```suggestion
                Number(node.show_column_totals) ||
```

Tardo (changes requested):
Thanks! also convert the values to a number for a correct evaluation.

gaikaz:
> Thanks! also convert the values to a number for a correct evaluation.

Good idea, but this would probably break `show_column_totals="true"`, as `true` ends up being `NaN` when converted to Number, though `parse_boolean` (same file line 185) does accept string `"true"` as a truthy.

pedrobaeza:
I remember there's a function in Odoo JS to parse that input with all the possible values to become true/falsy

gaikaz:
If there is, then this would become redundant:
https://github.com/OCA/web/blob/74559098b4a1dcd6d0bf62549dd998f15cd4fd24/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js#L179-L189
But I would rather not remove code I didn't write, as I'm not sure where else it's used.
This is kind of going outside the scope of the fix.
Should we really investigate further?

pedrobaeza:
I don't think it's outside the scope, as at the end, you are trying to fix this parsing in one of the attributes. @Yajo do you remember such Odoo JS function for parsing "booleans"?

gaikaz:
Can't argue with @pedrobaeza :smile: 
Ok, I did some research. You must be talking about `addons/web/static/src/js/fields/field_utils.js`. There is one function called `formatBoolean`:
https://github.com/odoo/odoo/blob/790cde7cc466e62db03331ce00f310153d668077/addons/web/static/src/js/fields/field_utils.js#L47-L70

But it seems to be meant to make it human-readable. There are some parsing functions, but they don't include one for Booleans, only has some `// todo` comments:
https://github.com/odoo/odoo/blob/790cde7cc466e62db03331ce00f310153d668077/addons/web/static/src/js/fields/field_utils.js#L685

I'll try to look more later.

yajo:
I found [this](https://github.com/odoo/odoo/blob/790cde7cc466e62db03331ce00f310153d668077/addons/web/static/src/js/core/utils.js#L408-L421). Could that help?

gaikaz:
Yeah, this looks more like it. I'll try to use it.

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -66,13 +69,15 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
                     this.field_value
                 ));
             }
-            this.show_row_totals = this.parse_boolean(
+            this.show_row_totals = utils.toBoolElse(
                 node.show_row_totals ||
-                (this.is_aggregatable(field_defs[this.field_value]) ? '1' : '')
+                this.is_aggregatable(field_defs[this.field_value]),
+                false
```
Tardo, line 75:
```suggestion
                this.is_aggregatable(field_defs[this.field_value])
```

web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -66,13 +69,15 @@ odoo.define('web_widget_x2many_2d_matrix.widget', function (require) {
...
             );
-            this.show_column_totals = this.parse_boolean(
+            this.show_column_totals = utils.toBoolElse(
                 node.show_column_totals ||
-                (this.is_aggregatable(field_defs[this.field_value]) ? '1' : '')
+                this.is_aggregatable(field_defs[this.field_value]),
+                false
```
Tardo, line 80:
```suggestion
                this.is_aggregatable(field_defs[this.field_value])
```

Tardo (changes requested):
Looks good! i think that you can avoid short-circuit evaluation...

gaikaz:
@Tardo thanks for the rereview :)

> Looks good! i think that you can avoid short-circuit evaluation...

The fallback `false` value is needed to not get `undefined`. Refer to the implementation of `toBoolElse`.  It uses underscore's `toBool` which does return `undefined` in cases where it can't tell if a string is truthy/falsy.
For example, if someone would decide to write `show_column_totals="no"`.

Tardo:
@gaikaz Ok, so...
node.show_column_totals -> Has a value => fallback to false
node.show_column_totals -> Is undefined => fallback to 'is_aggregatable'

If this is the correct behaviour, all fine for me ;)

Tardo (approved):

gaikaz:
Well, I can't say if that's the correct behavior per se, but it looks like that's the intention of the original code. :)

alexey-pelykh (approved):
Code LGTM

PR #1628, opened 2020-06-29
[12.0][IMP] web_pwa_oca: Refactor to use OdooClass

Tardo:
This PR aims to use OdooClass in the ServiceWorker so that other modules can extend it.

cc @tecnativa TT23822


pedrobaeza (approved):

web_pwa_oca/controllers/main.py
```diff
@@ -1,11 +1,23 @@
...
 from odoo.http import request, Controller, route
 
 
 class PWA(Controller):
 
-    def get_asset_urls(self, asset_xml_id):
-        qweb = request.env['ir.qweb'].sudo()
-        assets = qweb._get_asset_nodes(asset_xml_id, {}, True, True)
+    def _get_pwa_scripts(self):
+        """ Scripts to be imported in the service worker (Order is important) """
```
yajo: I don't understand where does this pattern of whitespacing docstring come from... :thinking: anyway: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

```suggestion
        """Scripts to be imported in the service worker (Order is important)"""
```

web_pwa_oca/controllers/main.py
```diff
@@ -1,11 +1,23 @@
...
+        return [
+            '/web/static/lib/underscore/underscore.js',
+            '/web_pwa_oca/static/src/js/worker/libs/class.js',
+            '/web_pwa_oca/static/src/js/worker/pwa.js',
+        ]
+
+    def _get_asset_urls(self, asset_xml_id):
+        """ Get all url's that have 'asset_xml_id' """
```
yajo:
```suggestion
        """Get all url's that have 'asset_xml_id'"""
```

web_pwa_oca/controllers/main.py
```diff
@@ -14,53 +26,58 @@ def get_asset_urls(self, asset_xml_id):
                 urls.append(asset[1]['src'])
         return urls
 
-    @route('/service-worker.js', type='http', auth="public")
-    def service_worker(self):
-        qweb = request.env['ir.qweb'].sudo()
+    def _get_pwa_params(self):
+        """ Get javascript PWA class initialzation params """
```
yajo:
```suggestion
        """Get javascript PWA class initialzation params"""
```

web_pwa_oca/controllers/main.py
```diff
@@ -14,53 +26,58 @@ def get_asset_urls(self, asset_xml_id):
...
-        })
-        return request.make_response(content, [('Content-Type', mimetype)])
+        return [
+            cache_version,
+            urls
+        ]
+
+    @route('/service-worker.js', type='http', auth="public")
+    def render_service_worker(self):
+        """ Route to register the service worker in the 'main' scope ('/') """
```
yajo:
```suggestion
        """Route to register the service worker in the 'main' scope ('/')"""
```

web_pwa_oca/controllers/main.py
```diff
@@ -14,53 +26,58 @@ def get_asset_urls(self, asset_xml_id):
...
 
     @route('/web_pwa_oca/manifest.json', type='http', auth="public")
-    def manifest(self):
-        qweb = request.env['ir.qweb'].sudo()
-        config_param = request.env['ir.config_parameter'].sudo()
-        pwa_name = config_param.get_param("pwa.manifest.name", "Odoo PWA")
-        pwa_short_name = config_param.get_param("pwa.manifest.short_name", "Odoo PWA")
-        icon128x128 = config_param.get_param(
+    def render_manifest(self):
+        """ Render the manifest used to install the page as app """
```
yajo:
```suggestion
        """Render the manifest used to install the page as app"""
```

web_pwa_oca/static/src/js/pwa_manager.js
```diff
@@ -0,0 +1,38 @@
...
+    "use strict";
+
+    var Class = require("web.Class");
+
+
+    var PWAManager = Class.extend({
+        /**
+         * @param {OdooClass} parent
+         */
+        init: function(parent) {
```
yajo: If you're using parent-child relationship, would it make more sense to extend the `Widget` class, which includes this? Or at least including the `ParentedMixin`?

web_pwa_oca/static/src/js/worker/libs/class.js
```diff
@@ -0,0 +1,150 @@
...
+ *         return true;
+ *       }
+ *     });
+ *
+ * When extending a class, each re-defined method can use this._super() to call the previous
+ * implementation of that method.
+ *
+ * @class Class
+ */
+function OdooClass(){}
```
yajo, line 49: This is copied from upstream, right? Why is this needed? :thinking:

web_pwa_oca/static/src/js/worker/pwa.js
```diff
@@ -0,0 +1,82 @@
+/* Copyright 2020 Tecnativa - Alexandre D. Díaz
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+"use strict";
+
+var PWA = OdooClass.extend({
```
yajo, line 14: I see this file is not an odoo module. I guess it's a need for this module, but since this is awkward for the odoo developer, could you add some docs here explaining why is this different?

web_pwa_oca/templates/assets.xml
```diff
@@ -0,0 +1,58 @@
...
+    <template id="assets_backend" name="web service worker assets" inherit_id="web.assets_backend">
+        <xpath expr="." position="inside">
+            <script type="text/javascript" src="/web_pwa_oca/static/src/js/pwa_manager.js"></script>
+            <script type="text/javascript" src="/web_pwa_oca/static/src/js/pwa_install.js"></script>
+            <script type="text/javascript" src="/web_pwa_oca/static/src/js/webclient.js"></script>
+        </xpath>
+    </template>
+    <template id="manifest" name="PWA manifest">
+        {
+          "name": "<t t-esc="pwa_name"/>",
```
yajo: As told above, wouldn't it be better to just have one controller that returns a JSON object? That'd be much easier to implement  and extend.

Just think what would happen if anybody sets this parameter to `My very "cool" website`. Now think all the possible corner cases... :shrug:

web_pwa_oca/controllers/main.py
```diff
@@ -14,53 +26,58 @@ def get_asset_urls(self, asset_xml_id):
...
+
+    @route('/service-worker.js', type='http', auth="public")
+    def render_service_worker(self):
+        """ Route to register the service worker in the 'main' scope ('/') """
+        return request.render('web_pwa_oca.service_worker', {
+            'pwa_scripts': self._get_pwa_scripts(),
+            'pwa_params': self._get_pwa_params(),
+        }, headers=[('Content-Type', 'text/javascript;charset=utf-8')])
 
     @route('/web_pwa_oca/manifest.json', type='http', auth="public")
```
yajo: I explain this in some comments below:

```suggestion
    @route('/web_pwa_oca/manifest.json', type='json', auth="public")
```

web_pwa_oca/controllers/main.py
```diff
@@ -14,53 +26,58 @@ def get_asset_urls(self, asset_xml_id):
...
             "/web_pwa_oca/static/img/icons/icon-512x512.png")
-        background_color = config_param.get_param(
+        background_color = config_param_sudo.get_param(
             "pwa.manifest.background_color", "#2E69B5")
-        theme_color = config_param.get_param(
+        theme_color = config_param_sudo.get_param(
             "pwa.manifest.theme_color", "#2E69B5")
-        mimetype = 'application/json;charset=utf-8'
-        content = qweb.render('web_pwa_oca.manifest', {
+        return request.render('web_pwa_oca.manifest', {
```
yajo: This should return the JSON dict directly. Using a template to render a JSON is a bad idea.

web_pwa_oca/templates/service_worker.xml
```diff
@@ -0,0 +1,40 @@
...
+            console.log('[ServiceWorker] Activating...');
+            evt.waitUntil(oca_pwa.activateWorker());
+            self.clients.claim();
+        </t>
+        self.addEventListener('activate', evt =&gt; {
+            <t t-raw="evt_activate" />
+        });
+    </template>
+    <template id="service_worker" name="PWA Service Worker">
+        self.importScripts(...<t t-esc="str(pwa_scripts)" />);
```
yajo, line 34: This structure seems very delicate... I see many places where bad code injections can break things... :roll_eyes: 

Besides, we lose all features as linters or autoformatters for JS. And extending qweb text templates is very hard.

Have you considered this?

1. Create a JSON controller that returns all PWA parameters.
1. Have a "normal" JS file that loads that JSON with AJAX and loads the PWA.

yajo (changes requested):
Some little requirements, many questions :sweat_smile:

web_pwa_oca/controllers/main.py
Tardo: Docstrings are trimmed... so no real differences in the generated document. I like the blank space style because for me looks better.

And for me follows the single comment style:
```# My comment``` vs ```#My Commen```

Comment (pedrobaeza) [reply web_pwa_oca/controllers/main.py]: Visually it's not the same the glyph `#` that covers full height than `"` that only covers upper part. That's why in most conventions, you don't need to let an extra space between `"` and text, which is advised between `#` and text.

Comment (Tardo) [reply web_pwa_oca/controllers/main.py]: I have no problem to change this, i will do that, but see this docstring:
```"""Get all url's that have 'asset_xml_id'"""``` vs ```""" Get all url's that have 'asset_xml_id' """```
or the extreme case:
```"""Get all url's that have \"asset_xml_id\""""``` vs ```""" Get all url's that have \"asset_xml_id\" """```

Comment (pedrobaeza) [reply web_pwa_oca/controllers/main.py]: But this is like `black` auto-formatter: the main thing is to be homogeneous, not to be pretty for all cases. The same would apply here.

File: web_pwa_oca/static/src/js/worker/libs/class.js (hunk 7 above)
Comment (Tardo) [reply web_pwa_oca/static/src/js/worker/libs/class.js:49]: Yes is just a copy but without "require" system. I can't use 'Odoo Bootstrap' here because depends on jQuery and jQuery uses DOM operations that are incompatible with ServiceWorkers.

ServiceWorkers are a piece separated from the page and doesn't have direct access to the DOM.

File: web_pwa_oca/templates/service_worker.xml (hunk 12 above)
Comment (Tardo) [reply web_pwa_oca/templates/service_worker.xml:34]: In this line i use 'Spread Operator': https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Operadores/Sintaxis_Spread

Since we don't have 'Odoo Bootstrap' on SericeWorkers i don't see a better way to have a inheritable design. If i put the 'service worker' on a .js file we lost the possibility to extend these service worker initialization process.

Points to know with service worker:
- They have the same scope where are loaded: If you load /blabla/src/js/myserviceworker.js the scope will be '/blablabla/src/js', we want be on main scope ('/').
- The events assignment must be on the main service worker script

Other point is that with these changes we need wait to the json-rpc reply and this introduce a promise on the service worker registration process.

File: web_pwa_oca/controllers/main.py (hunk 10 above)
Comment (Tardo) [reply web_pwa_oca/controllers/main.py]: Need be 'http' type because is loaded through a "link" html tag that launch GET petition. But i moved all the manifest definition to python.

File: web_pwa_oca/static/src/js/worker/libs/class.js (hunk 7 above)
Comment (yajo) [reply web_pwa_oca/static/src/js/worker/libs/class.js:49]: Humm I see...

:warning: I know I might be saying something stupid!

You probably thought about this also, but I'm wondering... Could you mock the `require()` function instead? I'd say a simple implementation of that would be much less code and would give more reusability for other modules of Odoo.

In any case, some docs please explaining why all this is needed.

File: web_pwa_oca/templates/service_worker.xml (hunk 12 above)
Comment (yajo) [reply web_pwa_oca/templates/service_worker.xml:34]: Thanks for the details, they help to review! :blush: 

> Other point is that with these changes we need wait to the json-rpc reply and this introduce a promise on the service worker registration process.

I understand.

If I'm not wrong, these processes are executed just once, when the user installs the app, right? :thinking:  If that's the case, then having to wait for some milliseconds until a config JSON file is downloaded from a controller doesn't seem like a big deal, don't you think? The benefits would be still more IMHO.

**Comment by pedrobaeza** [comment]
@joao-p-marques you have performed a merge operation instead of a rebase. Please fix the commit history.

**Comment by joao-p-marques** [comment]
Yes, I just didn't want to cause any conflicts with @Tardo 's work, as we are going to be both working on the same branch. As soon as I am finished with this I will rebase and squash.

**Comment by pedrobaeza** [comment]
I don't want to squash commits from you and @Tardo's, but never perform a git pull/merge operation. They are different things.

**Comment by joao-p-marques** [comment]
Commit history fixed @pedrobaeza 

I also need to document some errors:
When you have more than 1 database available (for example, when you use [`click-odoo-initdb`](https://github.com/acsone/click-odoo-contrib#click-odoo-initdb-stable) and end up with several cache DBs), you will get a 404 error in the JS console, as the service worker won't be able to access the PWA manifest. 
When you delete all the cache DBs, it works as normal. 
This could be fixed by using the specific cookie on the service worker when making the call to get the manifest, or by enabling this module as global (@Yajo can you explain a bit more about this?) 

Where can I document this? In the `ROADMAP`?

**Comment by pedrobaeza** [comment]
@Tardo is your part finished?

**Comment by joao-p-marques** [comment]
Some GIFs to show the behaviour:

Peek 2020-10-13 14-06
Peek 2020-10-13 14-07
Peek 2020-10-13 14-09

**Comment by joao-p-marques** [comment]
I just need to add some documentation and tests.

**Comment by joao-p-marques** [comment]
Just missing some tests and small fixes as discussed with @Tardo

**Comment by joao-p-marques** [comment]
Tests added.

Just need to check if the uploaded file is a valid SVG file under settings. How can I do that?

**Comment by joao-p-marques** [comment]
My part is finished @pedrobaeza 

I just had to limit the icon upload to SVG type. 
If we want to keep/upload only one icon, that is the best solution. We could also upload a a big PNG and rescale it as needed, but do you think that is necessary? Or is SVG enough?

**Comment by pedrobaeza** [comment]
Please don't limit to SVG. Anybody should decide what to upload and which size. Just put a banner warning about the minimum size for covering all needs (256x256?).

**Comment by joao-p-marques** [comment]
Ok. Will do.
In that case, maybe 512x512 is better (it was the previous biggest size)

**Comment by pedrobaeza** [comment]
@eLBati do you agree with the changes?

**Comment by joao-p-marques** [comment]
I just added another check to validate the uploaded icon's dimensions.

File: web_pwa_oca/templates/service_worker.xml (hunk 12 above)
Comment (Tardo) [reply web_pwa_oca/templates/service_worker.xml:34]: Can be good for some very specific parts... but more of them are incompatible with the service worker because touch DOM parts.

I will add a note in the roadmap.

File: web_pwa_oca/static/src/js/pwa_install.js [hunk 13]
```diff
@@ -1,45 +1,38 @@
...
-                console.log('User dismissed the A2HS prompt', choice);
-              }
-              deferredInstallPrompt = null;
-            });
-    },
-});
+
+    UserMenu.include({
+        start: function() {
+            window.addEventListener('beforeinstallprompt', this.saveBeforeInstallPromptEvent);
```
thomaspaulb: For 10.0, we ran into a condition where the 'beforeinstallprompt' handler is not fired, because this event listener is attached too late: the UserMenu's "start" function gets executed far after "serviceWorker.register", so the event can be missed, and then the 'Install PWA' button is not shown.

[Someone else](https://stackoverflow.com/q/61133389) ran into this as well, then solved it by putting the two closer together, which is also what worked for me [in the 10.0 backport](https://github.com/OCA/web/pull/1745/files#diff-ef9b369bb7e8416ad7be0e69b3247823580cfbbc86bfc72a3f418e8363bb80e6R17).

In your code, I see that you've moved "serviceWorker.register" somewhere else, but the two are still not in the same place, so I don't know in which order they are called in your case... but as a warning

a.svg
```diff
@@ -0,0 +1 @@
+a
```
yajo: :wastebasket: :question:

joao-p-marques: Yes, already removed. Ty!

pedrobaeza:
Not sure if @Tardo has local changes to this.

joao-p-marques:
Yes, we need to check that. Everything is finished from my part.

web_pwa_oca/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+To configure your PWA:
+
+* Go to **Settings > General Settings > Progressive Web App**.
+* Set the parameters (*Note:* Icon **must be a SVG file**)
+* **Save**
```
yajo: If you're in USAGE, you shouldn't explain how to configure. That belongs to CONFIGURE.

web_pwa_oca/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
...
+* Open the Odoo web app using a supported browser (like Chrome/Chromium)
+* Install the PWA
```
yajo:
```suggestion
#. Open the Odoo web app using a supported browser (like Chrome/Chromium)
#. Install the PWA
```

web_pwa_oca/readme/ROADMAP.rst
```diff
@@ -1,5 +1,27 @@
...
+        .. code-block:: javascript
+
+            var MyClass = OdooClass.extend({
+                myFunc: function() {
+                    return new Promise(async (resolve, reject) => {
+                        const mydata = await ...do await stuff...
+                        return resolve(mydata);
+                    });
+                }
+            });
```
yajo: Some RST errors here:

```suggestion
* Current *John Resig's inheritance* implementation doesn't support ``async`` 
  functions because ``this._super`` can't be called inside a promise. So we 
  need to use the following workaround:

  - Natural 'async/await' example (This breaks "_super" call)

    .. code-block:: javascript

        var MyClass = OdooClass.extend({
            myFunc: async function() {
                const mydata = await ...do await stuff...
                return mydata;
            }
        });

  - Same code with the workaround:

    .. code-block:: javascript

        var MyClass = OdooClass.extend({
            myFunc: function() {
                return new Promise(async (resolve, reject) => {
                    const mydata = await ...do await stuff...
                    return resolve(mydata);
                });
            }
        });

* Fix issue when trying to run in localhost with several databases. The browser 
  doesn't send the cookie and web manifest returns 404.
  
```

web_pwa_oca/static/src/js/worker/base/cache_manager.js
```diff
@@ -0,0 +1,45 @@
+/* Copyright 2020 Tecnativa - Alexandre D. Díaz
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+"use strict";
+
+const CacheManager = OdooClass.extend({
```
yajo: Some docs here would be appreciated.

web_pwa_oca/static/src/js/worker/base/tools.js
```diff
@@ -0,0 +1,17 @@
+/* Copyright 2020 Tecnativa - Alexandre D. Díaz
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+ "use strict";
```
yajo: Using this outside of a function can have side effects if the file is minimized and bundled with other ones. Such as how odoo assets do. Is that the case? If so, better declare a namespace (either with `odoo.define` or just `function(){...}`.

web_pwa_oca/static/src/js/worker/base/tools.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ *  Execute various promises in sequential order and return a promise with all results
+ *  Code from: https://decembersoft.com/posts/promises-in-serial-with-array-reduce/
+ *
+ * @param {Array[Promise]} tasks
+ * @returns {Promise[Array]}
+ */
+function execSequentialPromises(tasks) {
+    return tasks.reduce((promiseChain, currentTask) => {
+        return promiseChain.then(chainResults => currentTask.then(currentResult => [...chainResults, currentResult]));
```
yajo: I observe you're using JS syntax that isn't supposed to exist on v12. Is that on purpose?

web_pwa_oca/static/src/js/worker/pwa.js
```diff
@@ -0,0 +1,74 @@
+/* Copyright 2020 Tecnativa - Alexandre D. Díaz
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+"use strict";
```
yajo: Same about use strict outside of a namespace.

web_pwa_oca/templates/assets.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="web_layout_pwa" name="Web layout PWA" inherit_id="web.layout">
+        <xpath expr="//meta[@name='viewport']" position="after">
+            <!-- Add link rel manifest -->
+            <link rel="manifest" href="/web_pwa_oca/manifest.webmanifest"/>
+            <!-- Add iOS meta tags and icons -->
+            <meta name="apple-mobile-web-app-capable" content="yes"/>
+            <meta name="apple-mobile-web-app-status-bar-style" content="black"/>
+            <meta name="apple-mobile-web-app-title" content="Odoo PWA"/>
```
yajo: Shouldn't this get the name from the settings?

web_pwa_oca/tests/test_web_pwa_oca_controller.py
```diff
@@ -0,0 +1,141 @@
...
+            b"Q3MjE3LDAgMC45NTEyLC0wLjAxMiAxLjQxMDE4LC0wLjA0NTggMzAuNzg1MTYsLTA" \
+            b"uNDU4NDEgNTkuMTUzODIsLTE3LjY5NTM1IDc0LjA2NjQ5LC00NC44NTYyMSAzLjcy" \
+            b"NDYsLTYuNzQ3MjMgMS4yNjMyNiwtMTUuMTk1NjMgLTUuNDgzOTcsLTE4Ljg5MTU3I" \
+            b"C02LjcwNDI3LC0zLjY4MTYyIC0xNS4xNjY4NywtMS4yMjg0MyAtMTguODkxNDcsNS" \
+            b"40NjE1MSAtMTAuMTI4MDEsMTguNDk0MDIgLTI5LjM0NjQzLDMwLjE3OTAyIC01MC4" \
+            b"xNjExNSwzMC41MDg1MSBsIC0wLjkzOTc1LDAgYyAtMjEuMjQ0NDgsMCAtNDAuNzc0" \
+            b"MzcsLTExLjYxMzI4IC01MS4xNDU5NCwtMzAuNTA4NTEgLTIuNzUwNDYsLTUuMDE3N" \
+            b"DMgLTguMjA1NTQsLTcuNjQ2MTggLTEzLjYwOTA3LC03LjA5NTUyIHoiIHN0eWxlPS" \
+            b"JmaWxsOiNmZmZmZmY7ZmlsbC1vcGFjaXR5OjE7ZmlsbC1ydWxlOm5vbnplcm87c3R" \
+            b"yb2tlOm5vbmUiLz4KICA8L2c+Cjwvc3ZnPgo="
```
yajo: wouldn't it be more maintainable to drop a svg file somewhere and doing something similar to `b64encode(open(that_file, "b").read())`?

web_pwa_oca/controllers/main.py
```diff
@@ -1,75 +1,156 @@
...
+            "/web_pwa_oca/static/src/js/worker/base/tools.js",
+            "/web_pwa_oca/static/src/js/worker/base/cache_manager.js",
+            "/web_pwa_oca/static/src/js/worker/pwa.js",
+        ]
 
-    def get_asset_urls(self, asset_xml_id):
-        qweb = request.env['ir.qweb'].sudo()
-        assets = qweb._get_asset_nodes(asset_xml_id, {}, True, True)
+    def _get_asset_urls(self, asset_xml_id):
+        """Get all url's that have 'asset_xml_id'"""
```
yajo:
```suggestion
        """Get all urls that have 'asset_xml_id'"""
```

web_pwa_oca/controllers/main.py
```diff
@@ -1,75 +1,156 @@
...
+                (192, 192),
+                (256, 256),
+                (512, 512),
+            ]:
+                icons.append(
+                    {
+                        "src": "/web_pwa_oca/static/img/icons/icon-%sx%s.png"
+                        % (str(size[0]), str(size[1])),
+                        "sizes": "128x128 144x144 152x152 192x192 256x256 512x512",
+                        "type": "image/svg+xml",
```
yajo: I find it weird that the `src` ends on `png` but type is svg. Is this some required hack? If so, please add a comment.

web_pwa_oca/controllers/main.py
```diff
@@ -1,75 +1,156 @@
...
-            'background_color': background_color,
-            'theme_color': theme_color,
-        })
-        return request.make_response(content, [('Content-Type', mimetype)])
+    @route("/web_pwa_oca/manifest.webmanifest", type="http", auth="public")
+    def pwa_manifest(self):
+        """Returns the manifest used to install the page as app"""
+        return request.make_response(
+            json.dumps(self._get_pwa_manifest()),
+            headers=[("Content-Type", "text/javascript;charset=utf-8")],
```
yajo: Shouldn't it be something like this?

```suggestion
            headers=[("Content-Type", "text/json;charset=utf-8")],
```

yajo (changes requested):
Only code review. Some things are just doubts where I might be wrong. Good job! :blush:

joao-p-marques:
Thank you @Yajo 
Changes made on my part

yajo:
Thanks, those parts look better now and I can see there were some actual bugs :blush:. There are still comments left though.

joao-p-marques:
Thank you! Yes, those are from @Tardo's part. I think he is already working on it.

web_pwa_oca/static/src/js/worker/base/tools.js
Tardo: Yes.
- Technically, workers are not part of Odoo.
- If you can run services workers, you can use ES6+ https://caniuse.com/serviceworkers
- Current infrastructure can't testing service workers :/
- To work with service workers you need use a lot of asynchronous calls... so use async/await makes the life more easier and the code more maintainable and readable = Use ES6+

web_pwa_oca/static/src/js/worker/base/tools.js
Tardo: These scripts are not minified, are injected directly by the browser.

See: https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts

Tardo:
Changes done, ready for final review.

web_pwa_oca/static/src/js/pwa_install.js
Tardo: Thanks for the advice. I see that custom pwa installation flow has some problems.
If you refresh the page you still see the install option? I need restart the browser to see it again :/

thomaspaulb: For us it was also erratic before but now it always comes up, also after browser refresh. You'll need to do something like [this](https://github.com/OCA/web/pull/1745/commits/7bf04cd9bf4b5ad3a408f07cfc771e199412c6fe).

Tardo: I have tried your PR for 10.0 and I also have the same issues ... I don't know if it's a problem because I'm in localhost.

web_pwa_oca/readme/ROADMAP.rst
```diff
@@ -1,5 +1,37 @@
...
+            myFunc: function() {
+                return new Promise(async (resolve, reject) => {
+                    const mydata = await ...do await stuff...
+                    return resolve(mydata);
+                });
+            }
+        });
+
+* Fix issue when trying to run in localhost with several databases. The browser
+  doesn't send the cookie and web manifest returns 404.
```
yajo, line 33: Did you try adding `server_wide_modules = web,web_pwa` to your `odoo.conf`?

Once that's done, you can make your controllers with `auth="none"` and infer the database from a parameter, similar to [what is done in mail_tracking](https://github.com/OCA/social/blob/da180c747df57c05bd1be0e504d0f1f13c94b36b/mail_tracking/controllers/main.py#L42-L52).

Well, this is not a requirement because IMHO the roadmap note is good enough for now.

web_pwa_oca/static/src/js/worker/pwa.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Services workers are a pice of software separated from the user page.
+ * Here can't use 'Odoo Bootstrap', so we can't work with 'require' system.
+ * When the service worker is called to be installed from the "pwa_manager"
+ * this class is instantiated.
+ */
+var PWA = OdooClass.extend({
+
+    init: function () {
+        // To be overrided
```
yajo:
```suggestion
        // To be overriden
```

web_pwa_oca/templates/assets.xml
```diff
@@ -0,0 +1,24 @@
...
+            <!-- Add link rel manifest -->
+            <link rel="manifest" href="/web_pwa_oca/manifest.webmanifest"/>
+            <!-- Add iOS meta tags and icons -->
+            <meta name="apple-mobile-web-app-capable" content="yes"/>
+            <meta name="apple-mobile-web-app-status-bar-style" content="black"/>
+            <t t-set="pwa_name" t-value="request.env['ir.config_parameter'].sudo().get_param('pwa.manifest.name')"/>
+            <meta name="apple-mobile-web-app-title" t-att-content="pwa_name"/>
+            <link rel="apple-touch-icon" href="/web_pwa_oca/static/img/icons/icon-152x152.png"/>
+            <!-- Add meta theme-color -->
+            <meta name="theme-color" content="#2E69B5" />
```
yajo: Shouldn't these values be computed from ICP?

yajo (approved):
Good! :tada:

web_pwa_oca/static/src/js/worker/pwa.js
thomaspaulb: `Overridden`, to be absolutely correct in English :-)

web_pwa_oca/static/src/js/pwa_install.js
thomaspaulb: Actually yes, that can be an issue:

image

thomaspaulb: You probably don't have HTTPS on your localhost.

web_pwa_oca/readme/ROADMAP.rst
Tardo: I can't solve this problem following your instructions :/

Tardo:
All changes done!

joao-p-marques:
Same with me! @pedrobaeza WDYT?

thomaspaulb (approved):

marcelsavegnago (approved):

PR #1633, opened 2020-07-09
web_ir_actions_act_view_reload: fix README

KENLYLST:
Unresolved reference `self` and module `odoo.api` has no attribute `multi`

### DIFF
--- a/web_ir_actions_act_view_reload/README.rst
+++ b/web_ir_actions_act_view_reload/README.rst
@@ -39,8 +39,7 @@ To use this functionality you need to return following action:
 
 .. code-block:: python
 
-      @api.multi
-      def foo():
+      def foo(self):
          self.ensure_one()
          return {
             'type': 'ir.actions.act_view_reload',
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):

simahawk (changes requested):
@KENLYLST thanks for your contrib!

This change will be overridden because the README is computed w/ readme fragments.
It must be fixed here https://github.com/OCA/web/blob/e4004cccac2e10637527405eded7ba1eb4c8b3ef/web_ir_actions_act_view_reload/readme/USAGE.rst

Also, please, include module name in commit and PR title, thanks! :pray:

KENLYLST (commented):
I have fixed https://github.com/OCA/web/blob/e4004cccac2e10637527405eded7ba1eb4c8b3ef/web_ir_actions_act_view_reload/readme/USAGE.rst.

simahawk (approved):

PR #1635, opened 2020-07-10
[12.0][MIG] web_advanced_filter

janverb:
Continuation of #1587.

This addon allows users to apply set operations on filters: Remove or add certain ids from/to a selection, but also to remove or add another filter's outcome from/to a filter. This can be stacked, so the filter domain can be arbitrarily complicated.

The math is hidden from the user as far as possible, in the hope it's still user friendly.


NL66278:
@janverb Hi Jan, could you increase test coverage to get this branch completely green?

janverb:
@NL66278 If I'm reading the report correctly, the coverage within `web_advanced_filter` is 90.54%, which is enough, and the check is failing because of unexpected coverage changes in other modules. Is that possible? I'm tempted to rebase and see if that jiggles anything into working again.

web_advanced_filter/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Holger Brunn <hbrunn@therp.nl>
```
NL66278, line 1: @janverb you can put your name here!

web_advanced_filter/models/ir_filters.py
```diff
@@ -66,70 +72,87 @@ def _compute_is_frozen(self):
...
     @api.multi
     def _inverse_domain(self):
         for this in self:
             this.domain_this = this.domain
 
     @api.multi
-    @api.depends('domain')
+    @api.depends('domain', 'model_id')
     def _compute_flags(self):
         """check if this filter contains references to x2many fields. If so,
```
NL66278: This is an explanation that really does not explain anything. What goes wrong with negation? What does negation mean here? How is this solved by removing resulting ids?

NL66278:
@janverb warning on codecov/project indeed caused by other modules. Why do they show up here? A bug in coverage check?

web_advanced_filter/models/ir_filters.py
janverb: I'm having trouble figuring out the history of this.

It was added in #83, with a reference to [this `therp-backports` bug report](https://bugs.launchpad.net/therp-backports/+bug/1062795). But I can't reproduce the original issue, neither in OpenERP 7 nor in Odoo 12. I'm trying it by directly running running variants of `res.partner.search([("category_id", "!=", ...)])`. It's possible it's no longer necessary, but maybe I'm just missing the reason it was necessary originally.

#100 dealt with `auto_join`. I can't reproduce the underlying issue for that either, though maybe I'm interpreting it wrong—it sounds to me like `res.partner.search(['|', ('name', '=', 'something'), ('message_ids', '=', some_id)])` would only return results that match the second term, because `message_ids` has `auto_join`, but it also returns records that only match the first term, like you'd expect.

It almost seems like this stuff wasn't necessary for version 7, in which case it wouldn't be necessary for 12 either.

Do you have any insight?

NL66278: I have no insights to add here. Maybe just add tests for the problems that were supposedly solved, and see that they do not re-occur. It would be good if the code can be drastically simplified, or organized in such a way that it is more understandable.

hbrunn: @janverb to reproduce the first one, you need a partner with two labels set, and say `[("category_id.name", "!=", "label1")]`. You won't get the inverse of `[("category_id.name", "=", "label1")]`. So the issue is how Odoo treats dotted paths in domains.

The second is similar, when you have a path of auto_joined fields and union this with a domain that doesn't, you expect the resulting records of both filters. But if you put this in one domain, the resulting statement is a join that only yields results of the join. You should find an issue by the customer a from around the time of the original commit with an example.

janverb: @hbrunn Thank you! I've been able to reproduce both issues on version 12. I'll write tests and do some cleanup.

I'm still unsure about `has_x2many |= bool(field.compute)` and `has_auto_join |= bool(field.compute)`. Are those there to be more careful about fields with custom searching logic? If so, should they check for `field.search` instead?

hbrunn: yes, that's for custom searching and yes, you'll be better off checking for explicit search functions

janverb: Something else I'm now wondering about: shouldn't the `domain_this` of the main filter be subject to the same check for autojoined fields? It doesn't matter whether the autojoined search is in a union helper filter, or in the main filter, or even in a complement filter, as long as it's getting ORed at some point.

It might be a safer, simpler setup to merge `evaluate_before_negate` and `evaluate_always` into a single field and look at it to evaluate any of the own domain and the domains of the union and complement filters.

janverb: A redesign seems out of scope for the migration so I've just cleaned it up and added more documentation and tests.

janverb:
@NL66278 I'm not sure. It might be that, or it might be that some nondeterministic test or environment change caused code not to run that did run in the past. Let's see what the rebase does.

janverb:
This was causing trouble:
```python
    @api.depends(
        'domain_this', 'union_filter_ids.domain',
        'complement_filter_ids.domain',
    )
    def _compute_domain(self):
```
The `union_filter_ids.domain` and `complement_filter_ids.domain` dependencies meant that the cache was constantly getting invalidated. This was especially troublesome when getting the domains of all the filters in a recordset (as happens when loading a view), because the prefetching then caused domains to be calculated many times.

I'm changing it to this:
```python
    @api.depends(
        'domain_this', 'union_filter_ids', 'complement_filter_ids',
    )
    def _compute_domain(self):
```
That's slightly less correct, but I think it should typically be fine in practice, and it gives us a 15× speedup (30s → 2s) in one scenario.

(@NL66278)

NL66278:
@janverb Great catch!

NL66278 (approved):
:+1: LGTM And working in practice. @janverb Just need to still put your name to the contributors! ;-)

NL66278:
@janverb The problem is actually all the extra language files, not just the Dutch ones.

janverb:
@NL66278 I dropped the Transifex commit in the rebase, so only Dutch and German should be left now.

NL66278:
@janverb  OK, thats fine.

NL66278:
@hbrunn Can you approve of this PR now?

hbrunn (approved):

web_advanced_filter/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Authors
+~~~~~~~
+
+* Therp BV
+
+Contributors
+~~~~~~~~~~~~
+
+* Holger Brunn <hbrunn@therp.nl>
```
hbrunn, line 78:
```suggestion
* Holger Brunn <mail@hunki-enterprises.com>
```

hbrunn: @janverb if you see this early enough, I'd be thankful if you add this change

web_advanced_filter/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
...
```
hbrunn, line 1:
```suggestion
* Holger Brunn <mail@hunki-enterprises.com>
```

web_advanced_filter/README.rst
NL66278: @hbrunn maybe you can make a PR that changes your email address over the whole repo. Normally we have a commit / PR per module, but I have seen at least one other example of such a PR.

Anyway thanks for the approval and the merge,

PR #1642, opened 2020-07-17
[13.0] [MIG] web_sheet_full_width

feketemihai:


pedrobaeza:
Aren't you using `web_responsive`?

feketemihai:
@pedrobaeza not on this client, I will test it and if it's fine, than I will close it.

feketemihai:
@pedrobaeza It's not the same, probably the sheet is maximize but also it rearrange the layout, and i only need the sheet to be wider.

With Responsive"
Screenshot from 2020-07-17 10-02-14


With Sheet Full Width
Screenshot from 2020-07-17 10-02-45

So i would like to keep the PR.

pedrobaeza:
Well, you have the chatter sided as preference, but you can leave it on the bottom at user preference. I see that having response interface is a must today, but as you prefer.

feketemihai:
@pedrobaeza I didn't knew about the option, good, so we can close it. Thanks

LoisRForgeFlow:
@pedrobaeza is there any inconvenient of using web_responsive on top of web_enterprise?

pedrobaeza:
Probably a lot of incompatibilities, and I don't see any sense on that, as all the features (and more) are already on `web_responsive`. If you want such module only for the ugly color scheme, better to add it as extra module only with that on top of `web_responsive`.

LoisRForgeFlow:
What I want is the sheet full width, so to use in EE, is it ok to migrate this module?

pedrobaeza:
But why not removing `web_enterprise`and add `web_responsive`? I personally hate the first one and uninstall as soon as i can. Anyway, you can migrate the module if you want. It's not forbidden here in OCA to do it. I only advise about my personal opinion on this.

LoisRForgeFlow:
Yes, I understand your recommendation, but sometimes you cannot change everything you'd like, so I'm in favour of giving the greatest flexibility in OCA.

I will reopen this PR, thanks!

LoisRForgeFlow:
Well, I was about to open a new copying this one, but maybe it is easier if you reopen this PR (I don't have rights), could you do it @pedrobaeza ?

pedrobaeza:
Better that you start from this branch, but create a new PR depending on your remote, as you won't be able to do any requested change if needed.

LoisRForgeFlow:
Fair enough, will do it!

feketemihai:
@pedrobaeza  @LoisRForgeFlow probably would be good to use "excludes" in here or in web_responsive, not to have issue when someone install both modules

LoisRForgeFlow:
is there really an issue? they overlap but there is no problem if you have both installed.

Issue #1553, opened 2020-03-20
[12.0] web_advanced_search: When using a custom filter with 'is_equal_to' and then searching for more records, the filter is not executed.


marylla:
Hi there,
there might be a bug in this module when using the 'is_equal_to' comparison option in custom filters.

**Here is what I did (in version 12.0):**
* install account
* install web_advanced_search

* open the list of all customer invoices (Invoicing > Customers > Invoices)
* click on 'Filters’ > 'Add Custom Filter’ below search bar
* choose 'Account’ and 'is equal to’
* click in the third field to enter the comparison value and click on 'Search more...'
* the search more dialog opens, search for a account and choose one

**Defect:**
The 'search more' dialog closes and the dropdown for the custom filter is not shown anymore.

(If I click again on 'Filters’ below search bar, the chosen account and filter setting is displayed again and can be used.)

**Desired behavior:**
The filter dropdown should be still visible in order to execute the filter.

Rad0van:
+1
I can confirm this also doesn't work on 13.0 with for example Product Category.

Rad0van:
@Tardo maybe you could have a look please. Thank you.

PR #1647, opened 2020-07-22
[12.0][IMP] web_advanced_search: Prevent the menu from closing when using a relational field

Tardo:
Fix: https://github.com/OCA/web/issues/1553

Is a workaround because I don't see a better way to handle this issue.

cc @tecnativa


yajo (approved):
Seems legit.

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Should this be moved to v13 as well?

PR #1654, opened 2020-08-04
Design improvements for the backend user interface

androettop:
I added an enterprise-like background and a file that makes it easy to change the UI color.

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -9,8 +9,8 @@ $chatter_zone_width: 35%;
     display: flex;
     flex-direction: column;
     height: calc(100vh - #{$o-navbar-height});
-    max-height: calc(100vh - #{$o-navbar-height});
     position: fixed;
+    margin: 0;
     width: 100vw;
     z-index: 100;
     // Inline style will override our `top`, so we need !important here
@@ -131,16 +131,47 @@ $chatter_zone_width: 35%;
         }
     }
 }
+.o_main_navbar {
+    color: color-yiq($o-brand-odoo);
 
+    > ul > li > a,
+    > ul > li > label {
+        color: color-yiq($o-brand-odoo);
+    }
+    .dropdown-menu.show {
+        max-height: calc(100vh - #{$o-navbar-height});
+    }
+}
 // Iconized full screen apps menu
 .o_menu_apps {
     .search-input:focus {
         border-color: $o-brand-primary;
     }
 
+    a.full {
+        width: $o-navbar-height;
+        text-align: center;
+    }
+
     .dropdown-menu.show {
+        opacity: 1;
+        visibility: visible;
+    }
+
+    .dropdown-menu {
         @include full-screen-dropdown();
 
+        opacity: 0;
+        visibility: hidden;
+        transition: visibility 100ms ease, opacity 100ms ease;
+        background: url("../img/home-menu-bg-overlay.svg"),
+            linear-gradient(
+                to bottom,
+                $o-brand-odoo,
+                desaturate(lighten($o-brand-odoo, 20%), 15)
+            );
+        background-size: cover;
+        border-radius: 0;
         // Display apps in a grid
         align-content: flex-start;
         display: flex;
@@ -160,6 +191,12 @@ $chatter_zone_width: 35%;
             display: flex;
             flex-direction: column;
             justify-content: flex-start;
+            background: none;
+            img {
+                box-shadow: none;
+                transition: 300ms ease;
+                transition-property: box-shadow, transform;
+            }
 
             // Size depends on screen
             width: 33.33333333%;
@@ -171,6 +208,11 @@ $chatter_zone_width: 35%;
             }
         }
 
+        .o_app:hover img {
+            transform: translateY(-3px);
+            box-shadow: 0 9px 12px -4px rgba(0, 0, 0, 0.3);
+        }
+
         // Hide app icons when searching
         .has-results ~ .o_app {
             display: none;
@@ -178,9 +220,15 @@ $chatter_zone_width: 35%;
 
         .o-app-icon {
             height: auto;
-            max-width: 7rem;
+            max-width: 6rem;
         }
 
+        .o-app-name {
+            color: white;
+            margin-top: 4px;
+            font-size: 1.25rem;
+            text-shadow: 1px 1px 1px rgba(0, 0, 0, 0.4);
+        }
         // Search input for menus
         .form-row {
             width: 100%;
@@ -197,17 +245,22 @@ $chatter_zone_width: 35%;
             white-space: normal;
         }
 
+        .search-container {
+            margin-bottom: 1.5rem;
+        }
         // Allow to scroll only on results, keeping static search box above
         .search-container.has-results {
             height: 100%;
-
-            .search-input {
-                height: 3em;
-            }
-
+            margin-bottom: 0;
             .search-results {
-                height: calc(100% - 3em);
+                max-height: calc(100vh - 47px - 4em);
+                overflow-y: hidden;
+                overflow-x: scroll;
                 overflow: auto;
+                background: url("../img/home-menu-bg-overlay.svg"),
+                    linear-gradient(to bottom, gray("200"), gray("white"));
+                background-position: center;
+                background-size: cover;
             }
         }
     }
--- a/web_responsive/static/src/xml/apps.xml
+++ b/web_responsive/static/src/xml/apps.xml
@@ -19,7 +19,7 @@
         </t>
         <!-- Search bar -->
         <t t-jquery="[t-as=app]" t-operation="before">
-            <div class="search-container form-row align-items-center mb-4 col-12">
+            <div class="search-container align-items-center col-12">
                 <div class="search-input col-md-10 ml-auto mr-auto mb-2">
                     <div class="input-group">
                         <div class="input-group-prepend">
@@ -34,8 +34,8 @@
                             class="form-control"
                         />
                     </div>
+                    <div class="search-results ml-auto mr-auto rounded" />
                 </div>
-                <div class="search-results col-md-10 ml-auto mr-auto" />
             </div>
         </t>
     </t>
--- a/web_responsive/static/src/img/home-menu-bg-overlay.svg
+++ b/web_responsive/static/src/img/home-menu-bg-overlay.svg
@@ -1,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="2000" height="1128" viewBox="0 0 2000 1128">
+    <polygon fill-opacity=".03" points="0 1077.844 392.627 778.443 1504.99 1127.745 0 1127.745"/>
+    <polygon fill-opacity=".02" points="392.216 778.443 283.294 0 0 0 0 666.504"/>
+    <polygon fill-opacity=".03" points="1000 0 2000 1009.98 2000 439.94 1749.817 0"/>
+</svg>
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -172,7 +172,8 @@ odoo.define("web_responsive", function(require) {
          */
         _searchFocus: function() {
             if (!config.device.isMobile) {
-                this.$search_input.focus();
+                // This timeout is necessary since the menu has a 100ms fading animation
+                setTimeout(() => this.$search_input.focus(), 100);
             }
         },
 


web_responsive/static/src/css/variables.scss
```diff
@@ -0,0 +1,2 @@
+$o-brand-odoo: $o-community-color;
+$o-brand-primary: $o-community-color;
```
androettop: In this file you can modify the UI colors and variables related to font-size.

yajo (changes requested):
:heart_eyes: Gorgeous!



However I observe one problem. When doing the search, the result items have very low contrast with the background:



I think it's important for usability. Could you fix that please?

Oh and also pre-commit is yelling.  Please execute `pre-commit run -a` before pushing (also use `pre-commit install` to let pre-commit forbid your commits if some linters or formatters fail).

androettop:
Thanks you! I'm going to improve the search results style

.prettierrc.yml
```diff
@@ -5,4 +5,4 @@ printWidth: 88
 proseWrap: always
 semi: true
 trailingComma: "es5"
-xmlWhitespaceSensitivity: "ignore"
+htmlWhitespaceSensitivity: "ignore"
```
pedrobaeza: Why are you changing this?

androettop: Because pre-commit issued the following warnings
image

pedrobaeza: @Yajo do you know about this?

pedrobaeza:
It's convenient to not hammer Travis and runbot with continuous commits that launch a new build and waste CPU cycles for other PRs. Try to minimize then, please. Also, you will need eventually to squash all the commits together before merging.

androettop:
> It's convenient to not hammer Travis and runbot with continuous commits that launch a new build and waste CPU cycles for other PRs. Try to minimize then, please. Also, you will need eventually to squash all the commits together before merging.

Understood, I have no more changes to make.
Now i'm going to simplify all the changes in a single commit. Do I keep the change of .prettierrc.yml?

pedrobaeza:
Wait until tomorrow to see @Yajo's answer. Right now, I have made a quick test and I see a problem: when you click on the AppSwitcher application, the focus is not put on the search bar, so if you type, you don't get the search.

.prettierrc.yml
yajo: Yes, those warnings are "correct" because the `xmlWhitespaceSensitivity` setting is only used when the XML prettier plugin is loaded, and prettier runs twice: once "normal" and once with the XML plugin loaded, so it's normal that the 1st time warns about this.
Just revert this change and ignore them.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -8,9 +8,10 @@ $chatter_zone_width: 35%;
     box-shadow: none;
     display: flex;
     flex-direction: column;
-    height: calc(100vh - #{$o-navbar-height});
-    max-height: calc(100vh - #{$o-navbar-height});
+    height: calc(100vh - #{$o-navbar-height}) !important;
+    max-height: calc(100vh - #{$o-navbar-height}) !important;
```
yajo: Usually there will be theming problems in submodules when we use `!important`. Is it possible to avoid that here? Maybe by using a more specific CSS selector...

If not possible, always put a comment explaining why `!important` is required.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -160,6 +190,11 @@ $chatter_zone_width: 35%;
             display: flex;
             flex-direction: column;
             justify-content: flex-start;
+            background: none !important;
```
yajo: The same about `!important` above.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -131,16 +132,45 @@ $chatter_zone_width: 35%;
...
+        visibility: visible;
+        opacity: 1;
+    }
+
+    .dropdown-menu {
         @include full-screen-dropdown();
 
+        visibility: hidden;
+        opacity: 0;
+        transition: visibility 50ms linear, opacity 50ms linear;
```
yajo: You're using `ease` transition below, which is more elegant. Also, animating `visibility` is pointless because it's not a numeric value.

```suggestion
        opacity: 0;
        transition: opacity 50ms ease;
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -171,16 +206,27 @@ $chatter_zone_width: 35%;
...
+
         // Hide app icons when searching
         .has-results ~ .o_app {
             display: none;
         }
 
         .o-app-icon {
             height: auto;
-            max-width: 7rem;
+            max-width: 70px;
```
yajo: It's better to stick with `rem` units, which provide a better sense of responsiveness and accessibility. Icons should be proportional to the font size.

However, I agree these could be a little bit smaller:

```suggestion
            max-width: 6rem;
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -171,16 +206,27 @@ $chatter_zone_width: 35%;
...
         }
 
+        .o-app-name {
+            color: white;
+            margin-top: 4px;
+            font-size: 15px;
```
yajo: Use `rem` units here too.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -202,12 +248,19 @@ $chatter_zone_width: 35%;
             height: 100%;
 
             .search-input {
-                height: 3em;
+                height: 100%;
             }
```
yajo: I can see this code is unused. You can remove this block.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -202,12 +248,19 @@ $chatter_zone_width: 35%;
...
 
             .search-results {
-                height: calc(100% - 3em);
+                max-height: calc(100vh - 47px - 4em);
+                overflow-y: hidden;
+                overflow-x: scroll;
                 overflow: auto;
+                background: url("../img/home-menu-bg-overlay.svg"),
+                    linear-gradient(to bottom, #f9f9f9, #ffffff);
```
yajo: Better to use theme variables here:

```suggestion
                    linear-gradient(to bottom, gray('200'), gray('white'));
```

web_responsive/static/src/xml/apps.xml
```diff
@@ -34,8 +34,8 @@
                             class="form-control"
                         />
                     </div>
+                    <div class="search-results ml-auto mr-auto" />
```
yajo: Add this class to remove above styling:

```suggestion
                    <div class="search-results ml-auto mr-auto rounded" />
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -202,12 +248,19 @@ $chatter_zone_width: 35%;
...
+                background-position: center;
+                background-size: cover;
+                border-radius: $border-radius;
```
yajo: You can remove this styling by adding a new class as explained in other comment below:

```suggestion
```

web_responsive/static/src/css/variables.scss
yajo: You don't need to declare these variables, as [they are already found in Odoo core](https://github.com/odoo/odoo/blob/528e10572195bda2102648b935d6d8647334fd4b/addons/web/static/src/scss/primary_variables.scss#L16-L17).

web_responsive/views/assets.xml
```diff
@@ -25,4 +25,16 @@
...
+        name="Brand color odoo"
+        inherit_id="web._assets_primary_variables"
+    >
+        <xpath expr=".">
+            <link
+                rel="stylesheet"
+                href="/web_responsive/static/src/css/variables.scss"
+            />
+        </xpath>
+    </template>
```
yajo: As indicated above, all this is unnecessary:

```suggestion
```

yajo (changes requested):
Thanks for this very good improvement.

Some code comments.

web_responsive/static/src/css/web_responsive.scss
androettop: When visibility is not animated strange behavior occurs in chrome.
3163f00a-92e3-4bb5-a313-3702ff78b6ef

But if I add the animation also to the visibility this no longer happens

captured-min

web_responsive/static/src/css/web_responsive.scss
androettop: Isn't it better to leave it in the scss to make it easier to overwrite for example in a theme?

web_responsive/static/src/css/web_responsive.scss
androettop: > Wait until tomorrow to see @Yajo's answer. Right now, I have made a quick test and I see a problem: when you click on the AppSwitcher application, the focus is not put on the search bar, so if you type, you don't get the search.

@pedrobaeza, the problem was occurring due to the transition to display the menu, browsers prevent focus from an element that is not yet visible. I removed the transition to avoid sacrificing functionality.

web_responsive/static/src/css/web_responsive.scss
androettop: I removed the transition to avoid sacrificing auto focus functionality, possibly in a future pull request I will add the transition keeping the auto focus.

web_responsive/static/src/css/web_responsive.scss
yajo: Something I found out when developing https://github.com/OCA/openupgradelib/pull/148 (utility to migrate from Bootstrap 3 (odoo 11-) to 4 (odoo 12+)) is that it is much easier to understand how an element is styled when you use helper classes.

When using custom css, it's much harder to know if a selector still matches an element, and if the property it is defining is stil needed or is being used fine. You need to do those checks manually always (I have migrated customer themes too), or you might end up having a lot of dead code that increases the browser payload and decreases performance, and having to refactor every now and then (as I actually did with `web_responsive` in https://github.com/OCA/web/pull/1066. With classes, it can be automated.

And I guess that some day we will have Bootstrap 5, so... well, I recommend to use classes where possible now.

---

If Odoo were not using a CSS framework, then yes. All would be easier in plain CSS. But it uses Bootstrap, so it's better to use it wherever possible.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -161,8 +163,6 @@ $chatter_zone_width: 35%;
 
         visibility: hidden;
         opacity: 0;
-        transition: visibility 50ms linear, opacity 50ms linear;
```
yajo: I still think you could leave the opacity transition. It makes it look good.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -131,16 +131,46 @@ $chatter_zone_width: 35%;
...
     }
 
+    a.full {
+        width: $o-navbar-height;
+        text-align: center;
+    }
+
     .dropdown-menu.show {
+        visibility: visible;
+        opacity: 1;
```
yajo: No need to handle the `visibility` property, but we need to animate `opacity`:

```suggestion
        transition: opacity 100ms ease;
        opacity: 1;
```

(I used 100ms because it looks better IMHO, but you can use 50ms if you prefer. I'm no designer.)

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -131,16 +131,46 @@ $chatter_zone_width: 35%;
...
+    }
+
+    .dropdown-menu {
         @include full-screen-dropdown();
 
+        visibility: hidden;
```
yajo: Again, no need to deal with this:

```suggestion
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -160,6 +190,11 @@ $chatter_zone_width: 35%;
             display: flex;
             flex-direction: column;
             justify-content: flex-start;
+            background: none;
+            img {
+                box-shadow: none;
+                transition: 300ms all ease;
```
yajo: And you have to animate only the animated properties here. Notice that you're animating all, which includes the inherited `opacity` value, at a higher length that its parent (300ms vs. 100ms above). That's what produces the glitch from https://github.com/OCA/web/pull/1654#discussion_r466406100:

```suggestion
                transition: 300ms ease;
                transition-property: box-shadow, transform;
```

yajo (changes requested):
Follow comments below to fix the transition glitch you commented at https://github.com/OCA/web/pull/1654#discussion_r466406100.

yajo:
I'm seeing a problem specifically with the Discuss app: when you chose it, the drawer doesn't disappear automatically:

Peek 07-08-2020 08-57

web_responsive/static/src/css/web_responsive.scss
androettop: It is necessary to keep the visibility property because when opacity is zero the menu is invisible but it is still there. Visibility is more similar to applying a display: none;

web_responsive/static/src/css/web_responsive.scss
androettop: Thank you!, I missed that transition, I did not understand why it happened 😅

web_responsive/static/src/css/web_responsive.scss
androettop: As it is necessary to leave the visibility property, it is also necessary to "animate" the visibility property since if it changes without a delay, the menu disappears instantly.

androettop: You can read this stackoverflow answer that explains this very well
https://stackoverflow.com/questions/27900053/css-transition-with-visibility-not-working/37905071#37905071

androettop:
> I'm seeing a problem specifically with the Discuss app: when you chose it, the drawer doesn't disappear automatically:
> 
> Peek 07-08-2020 08-57

@Yajo , I am seeing that this problem is from before and is not caused by my changes.
I don't know the module in depth, in fact, I'm just starting with odoo, but I think this could just as the event of pressing the enter key is being handled.

captured (1)-min

Captura de pantalla de 2020-08-07 12-52-06

web_responsive/static/src/css/web_responsive.scss
yajo: That is a good theory, but in practice, you can see that it's not needed:

Peek 10-08-2020 08-59

The difference between `visibility: invisible` and `display: none` is that an invisible block still occupies its space, while an undisplayed block doesn't.

So, this property is unneeded because Bootstrap itself is correctly managing the `display` property, and making an undisplayed block invisible is useless because it's not going to be displayed anyway. Since `display` is not animated, it changes instantly, before `opacity` starts being animated. And that's all you really need here, so `visibility` is not needed.

web_responsive/static/src/css/web_responsive.scss
yajo: I think you forgot to include this change.

web_responsive/static/src/css/web_responsive.scss
androettop: Yes, if you do this the menu effect works, and visibility is not needed, but since the display:flex property is specified, bootstraps is not managing it.

If i only manage the opacity and no the visibility (or display) property, the fade effect works, but the menu is not disappearing. You can't use the apps ui because the invisible menu is on top of it. And animation is not possible if I use display: flex / none because the change is instantaneous and the animation does not have the effect it should.

yajo: Oh I see now! :+1: then, thanks!

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -160,6 +191,13 @@ $chatter_zone_width: 35%;
             display: flex;
             flex-direction: column;
             justify-content: flex-start;
+            background: none;
+            img {
+                box-shadow: none;
+                transition: 300ms ease;
+                transition-property: box-shadow, transform;
+                transition-property: transform, box-shadow;
```
yajo: This seems duplicated. Just remove one of them:

```suggestion
                transition-property: box-shadow, transform;
```

yajo (changes requested):
Please, just one little last change.

Also could you squash the commits?

androettop:
> Please, just one little last change.
> 
> Also could you squash the commits?

Sure, but I'm not sure how to do it, could you tell me how?

yajo:
Yes, here it is explained very well: https://gitbetter.substack.com/p/how-to-squash-git-commits

yajo (approved):

hoangtrann:
should we make these changes available in version `12.0`?

pedrobaeza:
If you want to do it, please do the PR.

PR #1657, opened 2020-08-11
[IMP][13.0] web_responsive

ivantodorovich:
Continuing after https://github.com/OCA/web/pull/1654, a few more style improvements

image

image

----

Also removed `display: flex` on search results because it was causing this issue:

**Pre:**
pre-display-flex

**Post:**
post-display-flex

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -38,7 +38,7 @@ $chatter_zone_width: 35%;
             display: flex;
 
             .navbar-toggler {
-                color: white;
+                color: gray("white");
             }
 
             .o_menu_sections,
@@ -144,9 +144,7 @@ $chatter_zone_width: 35%;
 }
 // Iconized full screen apps menu
 .o_menu_apps {
-    .search-input:focus {
-        border-color: $o-brand-primary;
-    }
+    user-select: none;
 
     a.full {
         width: $o-navbar-height;
@@ -181,21 +179,30 @@ $chatter_zone_width: 35%;
 
         @include media-breakpoint-up(lg) {
             padding: {
-                left: 20vw;
-                right: 20vw;
+                left: calc((100vw - 850px) / 2);
+                right: calc((100vw - 850px) / 2);
             }
         }
 
         .o_app {
             align-items: center;
             display: flex;
+            padding: 10px 0;
+            border-radius: 4px;
             flex-direction: column;
             justify-content: flex-start;
             background: none;
+            transition: 300ms ease;
+            transition-property: background-color;
+
             img {
                 box-shadow: none;
                 transition: 300ms ease;
                 transition-property: box-shadow, transform;
+            }
+
+            &:focus {
+                background-color: rgba(gray("white"), 0.05);
             }
 
             // Size depends on screen
@@ -210,7 +217,7 @@ $chatter_zone_width: 35%;
 
         .o_app:hover img {
             transform: translateY(-3px);
-            box-shadow: 0 9px 12px -4px rgba(0, 0, 0, 0.3);
+            box-shadow: 0 9px 12px -4px rgba(gray("black"), 0.3);
         }
 
         // Hide app icons when searching
@@ -224,10 +231,10 @@ $chatter_zone_width: 35%;
         }
 
         .o-app-name {
-            color: white;
+            color: gray("white");
             margin-top: 4px;
             font-size: 1.25rem;
-            text-shadow: 1px 1px 1px rgba(0, 0, 0, 0.4);
+            text-shadow: 1px 1px 1px rgba(gray("black"), 0.4);
         }
         // Search input for menus
         .form-row {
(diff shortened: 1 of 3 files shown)


web_responsive/static/src/css/web_responsive.scss
```diff
@@ -181,23 +177,31 @@ $chatter_zone_width: 35%;
...
         }
 
         .o_app {
             align-items: center;
             display: flex;
+            padding: 10px 0;
             flex-direction: column;
             justify-content: flex-start;
             background: none;
+            transition: all 0.3s ease 0s;
```
yajo: Remember that `transition: all` is dangerous. Could you be more specific?

yajo (changes requested):
One change suggested.

Also, I see a big usage of hardcoded colors and px sizes. Please  use color variables from the theme and rem untis wherever possible.

web_responsive/static/src/css/web_responsive.scss
ivantodorovich: Done, thanks!

ivantodorovich:
@Yajo what about now?

Not sure what you meant about hardcoded colors... they're just black/white with opacity. But I changed the rgb values for the keyword.. is that what you meant?

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -181,23 +179,32 @@ $chatter_zone_width: 35%;
...
+            transition-property: background-color;
+
             img {
                 box-shadow: none;
                 transition: 300ms ease;
                 transition-property: box-shadow, transform;
             }
 
+            &:focus {
+                background-color: rgba(white, 0.05);
```
yajo: The bootstrap theme you use includes a definition of what is "white" and "black", a few shades of gray and all the primary, secondary, complementary colors. It's better to use those, so your addon looks good if there's any theme change.

```suggestion
                background-color: rgba(gray('white'), 0.05);
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -210,7 +217,7 @@ $chatter_zone_width: 35%;
 
         .o_app:hover img {
             transform: translateY(-3px);
-            box-shadow: 0 9px 12px -4px rgba(0, 0, 0, 0.3);
+            box-shadow: 0 9px 12px -4px rgba(black, 0.3);
```
yajo:
```suggestion
            box-shadow: 0 9px 12px -4px rgba(gray('black'), 0.3);
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -227,7 +234,7 @@ $chatter_zone_width: 35%;
             color: white;
             margin-top: 4px;
             font-size: 1.25rem;
-            text-shadow: 1px 1px 1px rgba(0, 0, 0, 0.4);
+            text-shadow: 1px 1px 1px rgba(black, 0.4);
```
yajo:
```suggestion
            text-shadow: 1px 1px 1px rgba(gray('black'), 0.4);
```

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -240,20 +247,60 @@ $chatter_zone_width: 35%;
...
+                        color: white;
+                        display: block;
+                        padding: 1px 2px 2px 2px;
+                        box-shadow: none;
+
+                        &::placeholder {
+                            color: white;
+                            opacity: 0.5;
+                        }
+                    }
```
yajo: And well, the same here about color variables...

web_responsive/static/src/css/web_responsive.scss
ivantodorovich: Thanks for the clarification!
Yes, I was aware of the theme-colors, but wasn't aware of the white/black for this case. Even in odoo source code there are lots of cases where the hardcoded keyword black/white is used.  

https://github.com/odoo/odoo/search?l=SCSS&q=white

yajo: > Even in odoo source code there are lots of cases where the hardcoded keyword black/white is used

:grin: I know, but well... It's odoo, not OCA :stuck_out_tongue_winking_eye:

yajo (approved):

pedrobaeza:
Please squash commits

ivantodorovich:
> Please squash commits

done

pedrobaeza (approved):

patel1008b:
New Changes looks great, but something needs to do when its custom theme is used.


Screenshot 2020-08-13 at 6 49 01 PM

we are using a https://apps.odoo.com/apps/themes/13.0/backend_theme_v13/

pedrobaeza:
I think the problem is in the theme itself, not here.

androettop:
> New Changes looks great, but something needs to do when its custom theme is used.
> 
> Screenshot 2020-08-13 at 6 49 01 PM
> 
> we are using a https://apps.odoo.com/apps/themes/13.0/backend_theme_v13/

The theme developer must adapt the theme to the new web_responsive style

PR #1663, opened 2020-08-19
[MIG] web_access_rule_buttons: Migrate to v12

tirix:


web_access_rule_buttons/models/models.py
```diff
@@ -0,0 +1,37 @@
...
+                # allow the operations
+                result[operation] = True
+                continue
+            try:
+                self.check_access_rule(operation)
+            except exceptions.AccessError:
+                result[operation] = False
+            else:
+                result[operation] = True
+        return result
```
janverb, line 37: Maybe this could be simplified to only check writes, since that's the only thing the module needs? That'll also help with coverage.

web_access_rule_buttons/tests/test_access_rule_buttons.py
```diff
@@ -0,0 +1,19 @@
...
+        self.curr_obj = self.env['res.currency']
+        self.curr_record = self.env.ref('base.USD')
+
+    def test_check_access_rule_1(self):
+        res = self.curr_obj.check_access_rule_all(['write'])
+        self.assertTrue(res['write'])
+
+    def test_check_access_rule_2(self):
+        res = self.curr_record.check_access_rule_all(['write'])
+        self.assertTrue(res['write'])
```
janverb, line 19: Could you also add a test for when there's no access? (That'll again help with coverage)

janverb (changes requested):
This works great! Just two nitpicks.

PR #1667, opened 2020-08-24
[CHG][13.0] web_responsive: normal chatter compliant with enterprise version

benwillig:
This simply makes the normal chatter work in odoo enterprise

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -434,13 +434,24 @@ html .o_web_client .o_action_manager .o_action {
 
     // Form views
     .o_form_view {
+
+        display: block;
+
+        .o_form_sheet_bg {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you please specify what "odoo enterprise" means? (maybe screenshots of before and after).

You should check also Travis status.

Anyway, both maintainers are on holidays, so this won't be checked in at least a week.

benwillig:
At the moment, when having web_enterprise installed, the chatter was always on the side of the screen even when having "normal" in user preferences. This PR fixes this behavior in this situation.

pedrobaeza:
I think there's no sense in having both, as this is a replacement of `web_enterprise`.

yajo:
Yes, I agree. Sorry but this one is meant to be 100% incompatible with `web_enterprise` :shrug:

PR #1670, opened 2020-08-26
Need for a screenshot to illustrate the module

alexandreduf:
A screenshot would be useful here. It's hard for recent users to know what kind of bubbles were introduced back in the version 10.0
Using runbot doesn't really help to illustrate what the module does.

### DIFF
--- a/web_no_bubble/README.rst
+++ b/web_no_bubble/README.rst
@@ -26,7 +26,9 @@ Web No Bubble
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
 This module removes from the web interface the bubbles introduced in the version 10.0.
-The help boxes are not removed though.
+The help boxes are not removed though. 
+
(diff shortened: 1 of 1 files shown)


legalsylvain:
hi @alexandreduf 
- why don't you add directly a screenshot, instead of writing a todo in the readme ? 
- if you want to change the readme, you have to change files in the readme folder. other changes will be overwritten.

otherwise, LGTM.

kind regards.

alexandreduf:
> hi @alexandreduf
> 
> * why don't you add directly a screenshot, instead of writing a todo in the readme ?
> * if you want to change the readme, you have to change files in the readme folder. other changes will be overwritten.
> 
> otherwise, LGTM.
> 
> kind regards.

Hi @legalsylvain : the readme file can be improved by someone who knows what the module does and knows what to detail/screenshot, which is not my case. Even though I run a demo on runbot, I will hardly be able to know exactly what changes when the module is used or not. All I can do is suggest an improvement. I can't do the improvement myself since I don't know the module.

Regards,

PR #1675, opened 2020-09-03
[Updated] Icons to match Odoo's design

jonisunz:
[Added] .svg file


pedrobaeza:
Thanks for all these icons for enriching current modules (and following my suggestions about SVG and the rest). About this one, I'm not sure this icon will be the most suitable for expressing the goal of this module. I think that a mobile + a window representation/ apps switcher icon or similar can fit better the features of the module, don't you think?

jonisunz:
> I think that a mobile + a window representation/ apps switcher icon or similar can fit better the features of the module, don't you think?

Yes, that actually sounds right to me. Is this what you mean by app switcher icon?
image

pedrobaeza:
I meant this AppSwitcher icon (that is a FontAwesome one):

Selección_015

https://fontawesome.com/icons/th-large?style=solid

bosd (approved):

bofiltd (approved):

pedrobaeza (changes requested):
See my comments

PR #1650, opened 2020-08-01
Fix enterprise mobile bug

cormaza:
On enterprise mobile version, on m2o fields show error on variable search_val.length line 313 because this value come as undefined value

PR #1677, opened 2020-09-07
Fix enterprise mobile bug

cormaza:
On enterprise mobile version, on m2o fields show error on variable search_val.length line 313 because this value come as undefined value

Related with pr #1650


pedrobaeza:
As approved in the previous PR, but there was a formal problem with several commits, I merge

/ocabot merge patch

PR #1690, opened 2020-09-30
[12.0][FIX] web_responsive update scss

JulianaPoudou:
On the edit page of a channel, if the user is configured with the chat on the side, when clicking to add a subscriber, the list is truncated. Changing the overflow of the div o_chatter displays the list correctly.

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -537,7 +537,7 @@ html .o_web_client .o_main .o_main_content {
                     flex: 0 0 $chatter_zone_width;
                     max-width: initial;
                     min-width: initial;
-                    overflow: auto;
+                    overflow: initial;
 
                     .o_chatter_header_container {
                         padding-top: $grid-gutter-width * 0.5;


remi-filament (approved):
Thanks Juliana, for this fix !

pedrobaeza:
Can you please put screenshots of before and after?

JulianaPoudou:
Before the update 

image

After the update

image

pedrobaeza:
I haven't experimented such problem right now. Is there anything more on top of `web_responsive`? Is it reproducible on OCA runbot?

remi-filament:
Hi @pedrobaeza we did not face it neither before. It was reported by one of our customer who wants to play with channel followers (which we never did before...)
It seems the issue is only present on these channels followers since the followers section is in the same form as the channel one, and not under or sided with notes as for other objects....

The issue is reproducible on OCA runbot indeed : http://3441040-12-0-bc4a85.runbot1.odoo-community.org/web?#id=1&model=mail.channel&view_type=form&menu_id=72

pedrobaeza:
Thanks for the extra info. My colleague will check.

Tardo (approved):
LGTM :+1: Thanks!

pedrobaeza:
This has provoked a bigger problem: now full screen scrolls when chatter is on the side, so I'm reverting this. Please find another way to solve the initial problem.

pedrobaeza:
Reverted at 2c82fa895386fd00a4b6bfc3534df08c6aced8e7

PR #1696, opened 2020-10-05
[13.0] [FIX] Axis label functionality

### LINKED ISSUE #1680: [13.0][web_widget_x2many_2d_matrix] field_label_x_axis & field_label_y_axis not used.
Came across this issue and I am not entirely sure if this is behaviour that was planned for label option.

### To replicate

1. Create matrix with `field_label_x_axis` or `field_label_y_axis`, pointing to Char field.
2. Fill computed or not computed Char field with desired label text.
3. Make sure it's present in tree, example:

**_XML_** 

```xml
<field 
    name="line_ids" 
    widget="x2many_2d_matrix" 
    field_x_axis="product_id"
    field_y_axis="partner_id"
    field_label_x_axis="product_matrix_label"
    field_value="amount_received"
>

    <tree>
        <field name="partner_id"/>
        <field name="product_id"/>
        <field name="amount_received"/>
        <field name="product_matrix_label"/>
    </tree>
</field>
```

**_PYTHON_**
```python
#... on `line_ids` model
    product_matrix_label = fields.Char(compute="_compute_product_matrix_label")

    @api.depends("product_id")
    def _compute_product_matrix_label(self):
        for line in self:
            line.product_matrix_label = (
                f"{line.product_id.name} {_('Received')} (kg)"
                if line.product_id
                else "Total (kg)"
            )
#...
```

### Behaviour:

Label field is not getting used by matrix. In above example, product name will be used instead and if it's empty `Product` will be displayed. 

Please note that I tried non-computed field for label that I filled with text manually and same thing happens.

### Desired Behaviour:

Display label field text instead.

### Possible causes:
Not JS expert but apart from `init_params`, label fields are not present anywhere else...
image

adamwojt:
This fixes #1680 . It's not the prettiest as I am not JS expert. Feel free to suggest better solution.

### DIFF
--- a/web_widget_x2many_2d_matrix/README.rst
+++ b/web_widget_x2many_2d_matrix/README.rst
@@ -125,6 +125,7 @@ the field in the default function:
                     'planned_hours': 0,
                     'message_needaction': False,
                     'date_deadline': fields.Date.today(),
+                    'x_axis_label': f"This is {project.name}" # This could be computed too
                 })
                 # if the project doesn't have a task for the user,
                 # create a new one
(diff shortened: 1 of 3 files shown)


paulius-sladkevicius:
@simahawk could you take a look at it, many thanks.

PR #1702, opened 2020-10-13
[12.0][FIX] web_translate_dialog reload parent view

santostelmo:
This fixes the following isssue:
1. Open a record.
2. Use the translator wizard: action -> Translate and save your changes
3. When you land back into the form view, the changes of translations are not visible directly. They are visible only if you reload the page. 

Goal:
The webpage should be reloaded with the newest translations

### DIFF
--- a/web_translate_dialog/static/src/js/web_translate_dialog.js
+++ b/web_translate_dialog/static/src/js/web_translate_dialog.js
@@ -222,6 +222,7 @@ var TranslateDialog = Dialog.extend({
                 return done;
             });
         });
+        this.view.reload();
         this.close();
     },
     on_button_close: function() {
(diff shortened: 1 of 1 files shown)


grindtildeath (commented):
I tested this but it doesn't seem to work properly. :confused:

santostelmo:
> I tested this but it doesn't seem to work properly.

Yes, I realized this. I need to find another way to do it....

simahawk:
@santostelmo IMO you should bubble up a change event as in v13 https://github.com/odoo/odoo/blob/338f8d843832c339c52c00c7f4c41b5ab2d696ed/addons/web/static/src/js/widgets/translation_dialog.js#L176

TDu (approved):

santostelmo:
@simahawk modification done using `Mutex`.  Funtional test working.

chroi (approved):

PR #1703, opened 2020-10-13
[14.0][ADD] web_calendar_color_field

ivantodorovich:
Odoo introduced a mechanism to set the desired color for a calendar filter, through the field's `color` attribute.
Unfortunately, this only works for filters, but not if we want to use the calendar's `color` attribute.

This module makes the appropiate changes so that the filter's color is used, if available.

Usage example:

```xml
    <calendar color="category_id">
        <field name="category_id" filter="1" color="kanban_color"/>
    </calendar>
```

Where:
- `category_id` is a `Many2one` field in your record.
- `kanban_color` is an integer field in your category record.

### DIFF
--- a/web_calendar_color_field/readme/USAGE.rst
+++ b/web_calendar_color_field/readme/USAGE.rst
@@ -1,0 +1,13 @@
+Set a color attribute value on any many2one calendar field set as filter and
+set the calendar's color attribute to use the same many2one field.
+
+Example:
+
+.. code-block:: xml
+
+    <calendar color="category_id">
+        <field name="category_id" filter="1" color="kanban_color"/>
+    </calendar>
+
+- `category_id` is a `Many2one` field in your record.
+- `kanban_color` is an integer field in your category record.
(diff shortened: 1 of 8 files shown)


web_calendar_color_field/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+Set a color attribute value on any many2one calendar field set as filter and
+set the calendar's color attribute to use the same many2one field.
```
tarteo, line 2: I don't fully understand how I should adjust the calendar view. Can you please add an example?

tarteo (changes requested):

web_calendar_color_field/readme/USAGE.rst
ivantodorovich: Sorry about that :sweat_smile: 
Just added an example in the readme

tarteo (approved):
Thanks!!

PR #1705, opened 2020-10-14
[12.0] web_translate_dialog: Do not translate related readonly fields

santostelmo:
Purpose: Ensure related readonly fields are not taken by the translation wizard.


grindtildeath (approved):

ferranmateu (approved):
Functional review ok

PR #1723, opened 2020-11-04
[12.0][ADD] web_set_single_page_hidden

vvrossem:
In 12.0, on a _PO_, _SO_, _WH/IN_, _WH/OUT_, or, more generally, anything that has _lines_ in its `form` view uses the `Pager` widget to visualise the line total count  (i.e. `1-10/10`).

`Pager` has the following `@param {boolean} [options.single_page_hidden] (not) to display the pager if only one page`.
On `init` function, the `single_page_hidden` option is set to `false`.
On `_render` function, the `Pager` is hidden `if (size === 0 || (this.options.single_page_hidden && this._singlePage()))`

`FieldX2Many`, an extension of `AbstractField`, instantiates a `Pager` with `option` `single_page_hidden: true,` in its `_renderControlPanel` function. It's impossible to `require` this field since it has not been made available (i.e no `return { FieldX2Many: FieldX2Many}`) in Odoo.

Two relational fields - `FieldOne2Many` & `FieldMany2Many` - extends `FieldX2Many`.
Their `start` function is overriden to set `single_page_hidden = false` on their `Pager` widget.

Before: (without `1-4/4`)
before

After: (with `1-4/4`)
after


houssine78 (approved):

web_set_single_page_hidden/static/src/js/relational_fields.js
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    FieldMany2Many.include({
+        start: function () {
+            var self = this;
+            return this._super.apply(this, arguments).then(function () {
+                if (!_.isUndefined(self.pager)) {
+                    self.pager.set_single_page_hidden(false);
+                }
+            });
```
hparfr, line 26: I don't get why you bind this to _super and don't bind it in the promise.

web_set_single_page_hidden/static/src/js/pager.js
```diff
@@ -0,0 +1,14 @@
+odoo.define('web_set_single_page_hidden.Pager', function (require) {
+    "use strict";
+
+    var Pager = require('web.Pager');
+
+    Pager.include({
+        set_single_page_hidden: function (value) {
+            if (!_.isUndefined(this.options)) {
```
hparfr, line 8: why use _.isUndefined ? 

```suggestion
            if (this.options) {
```

hparfr (approved):
I don't get the use case.

I have few minor non-blocking remarks in the JS.

vvrossem:
@hparfr thank you for reviewing this PR :)

The use case is to force the display of the line total count. In standard Odoo, the Pager is hidden when - I think - there are < 40 items.

web_set_single_page_hidden/static/src/js/pager.js
vvrossem: Indeed, your suggestion has better readability. I'll change it :)

web_set_single_page_hidden/static/src/js/relational_fields.js
vvrossem: I'm not sure to understand what you mean. How to properly call parent function, which is:
```js
    start: function () {
        return this._renderControlPanel().then(this._super.bind(this));
    },
```

vvrossem:
@houssine78  thanks for the review!

houssine78:
Hi @OCA/web-maintainers some of you could have a look a this PR ? It brings back a small feature from 9.0

hparfr:
> The use case is to force the display of the line total count. In standard Odoo, the Pager is hidden when - I think - there are < 40 items.

It's not a use case, it's a description. My question can be rephrased :  " **why** you need to always display the pager ?"

vvrossem:
We need to always display the pager so the user doesn't have to count the items present in the document one by one

legalsylvain (approved):
Code review. no test. thanks !

vvrossem:
@legalsylvain Thank you for the review :)

MeetKD:
@vvrossem I would suggest to add a system parameter like display single page pager or something and check the condition in js accordingly.

houssine78:
Hi @pedrobaeza I think this PR is ready for merge.

houssine78:
many thanks @pedrobaeza  !

vvrossem:
Thank you everybody for your time reviewing this PR. Thank you @pedrobaeza for merging it !

Issue #723, opened 2017-09-10
[10.0]Request for using web_notify in the desktop 


flotho:
Hi,
When sending a message in the chatter on odoo, a message could appear in your notification taskbar. The web_notify module work perfectly but don't seem to send the notif the same way so that the user has to be in front of odoo to see the notif.

Is there a way to implement his? what could it be?

regards

hbrunn:
replace https://github.com/OCA/web/blob/10.0/web_notify/static/src/js/web_client.js#L35 resp https://github.com/OCA/web/blob/10.0/web_notify/static/src/js/web_client.js#L37 with a call to the [javascript notification api](https://developer.mozilla.org/en-US/docs/Web/API/notification). Should be very straightforward. Better not replace but check if you have permission for a desktop notification, if so, do that, otherwise, use what's here already.

But I think it's better to do it the other way around: Provide a module that makes calls to `on_message_*` create a desktop notification (or both depending on configuration) if there's permission.

pedrobaeza:
Closing as it was answered, but no later feedback.

PR #1727, opened 2020-11-08
[12.0][IMP]Add native notifications via desktop

flotho:
Also fix https://github.com/OCA/web/issues/723

### DIFF
--- a/web_notify/models/res_users.py
+++ b/web_notify/models/res_users.py
@@ -31,6 +31,10 @@ class ResUsers(models.Model):
     notify_warning_channel_name = fields.Char(compute="_compute_channel_names")
     notify_info_channel_name = fields.Char(compute="_compute_channel_names")
     notify_default_channel_name = fields.Char(compute="_compute_channel_names")
+    
+    allow_native_notifications = fields.Boolean(_('Allow Native notifications'),
+                                                help=_('Notifications will be send to your OS via the browser'),
+                                                default=False)
 
     def notify_success(
         self, message="Default message", title=None, sticky=False
@@ -80,6 +84,7 @@ class ResUsers(models.Model):
             "message": message,
             "title": title,
             "sticky": sticky,
+            "allow_native_notifications": self.allow_native_notifications,
         }
         notifications = [
             (record[channel_name_field], bus_message) for record in self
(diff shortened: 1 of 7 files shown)


flotho:
Hi @hbrunn , I would appreciate a review here.

web_notify/static/src/js/web_client.js
```diff
@@ -61,16 +60,26 @@ odoo.define('web_notify.WebClient', function (require) {
...
-                    title: message.title,
-                    message: message.message,
-                    sticky: message.sticky,
-                    className: message.className,
-                }
-            );
+        	var notif_method = 'notify';
+        	var service_class =  'notification' ;
+
+        	if(message.allow_native_notifications){
```
hbrunn, line 66: I miss a call to https://developer.mozilla.org/en-US/docs/Web/API/Notification/permission here

web_notify/models/res_users.py
```diff
@@ -80,6 +85,7 @@ def _notify_channel(
             "message": message,
             "title": title,
             "sticky": sticky,
+            "allow_native_notifications": self.allow_native_notifications,
```
hbrunn, line 88: this makes all the comparisons in tests fail. You need to adapt https://github.com/OCA/web/blob/f56f71155f23ce07bad57b3a7f9d812412ff5e3a/web_notify/tests/test_res_users.py#L24 and below

web_notify/static/src/js/services/notification_service.js
```diff
@@ -0,0 +1,27 @@
+odoo.define('web_notify.NotificationService', function (require) {
+"use strict";
+
+ 	require('bus.BusService');
+
+	var NotificationService = require('web.NotificationService');
+
+	NotificationService.include({
+
+		notify: function (params) {
```
hbrunn, line 10: please explain this override

web_notify/models/res_users.py
```diff
@@ -32,6 +32,11 @@ def _compute_channel_names(self):
     notify_info_channel_name = fields.Char(compute="_compute_channel_names")
     notify_default_channel_name = fields.Char(compute="_compute_channel_names")
 
+    allow_native_notifications = fields.Boolean(
```
hbrunn, line 35: I'm not sure if we need this at all - either the user has allowed desktop notifications, then I think we should use them too, or they didn't, and then we don't need this flag

hbrunn (changes requested):
nice feature. Is it true that in v14 we get this for free?

flotho:
> nice feature. Is it true that in v14 we get this for free?

don't really know, I'll test it soon.

pedrobaeza:
Since v13, installing mail_bot, it ask you for that and you have it indeed "for free".

flotho:
maybe I should better check how it has been done in v13 and backport this.

PR #1728, opened 2020-11-09
[12.0][IMP] web_m2x_options: Added limit for o2m field entries + efficiency

CarlosRoca13:
cc @Tecnativa TT26667

Added the system parameter "web_m2x_options.field_limit_entries" to set the limit of entries that can be displayed on One2many fields.

Improved the eficiency of code: With this improvement we achieve to just do an rpc call when the js file is called instead of do the call for each field.

Here you have an example:

Please @chienandalu @ernestotejeda can you review this

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -115,6 +115,10 @@ If you disable one option, you can enable it for particular field by setting "cr
 
   Whether the field should always show "Search more..." entry or not.
 
+``web_m2x_options.field_limit_entries`` *int*
+
+  Number of displayed lines on all One2many fields
+
 To add these parameters go to Configuration -> Technical -> Parameters -> System Parameters and add new parameters like:
 
 - web_m2x_options.create: False
@@ -122,6 +126,7 @@ To add these parameters go to Configuration -> Technical -> Parameters -> System
 - web_m2x_options.m2o_dialog: False
 - web_m2x_options.limit: 10
 - web_m2x_options.search_more: True
+- web_m2x_options.field_limit_entries: 5
 
 
 Example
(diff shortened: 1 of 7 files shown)


web_m2x_options/README.rst
```diff
@@ -115,13 +115,18 @@ If you disable one option, you can enable it for particular field by setting "cr
 
   Whether the field should always show "Search more..." entry or not.
 
+``web_m2x_options.field_limit_entries`` *int* (Default: odoo default value is ``40``)
```
pedrobaeza: There shouldn't be a default value, as if not defined, the limit is set by Odoo itself in their core, so IMO you shouldn't change that default value if not defined.

pedrobaeza (changes requested):
The improvement for reading only one time should be split in a different commit from the new feature for limiting the number of entries in the one2many.

CarlosRoca13:
@pedrobaeza Changes done

web_m2x_options/readme/USAGE.rst
```diff
@@ -69,13 +69,18 @@ If you disable one option, you can enable it for particular field by setting "cr
 
   Whether the field should always show "Search more..." entry or not.
 
+``web_m2x_options.field_limit_entries`` *int* (Default: odoo default value is ``40``)
```
pedrobaeza: This change is still in the wrong commit

CarlosRoca13:
Sorry, changed yet @pedrobaeza

pedrobaeza (approved):

joao-p-marques (approved):
LGTM :+1:

PR #1744, opened 2020-11-25
[FIX] Error when install button not found

thomaspaulb:
Fixes "Uncaught TypeError: Cannot read property 'removeAttribute' of undefined"


pedrobaeza:
@Tardo is this fixed in #1628 ?

thomaspaulb:
It's [not](https://github.com/OCA/web/pull/1628/files#diff-ef9b369bb7e8416ad7be0e69b3247823580cfbbc86bfc72a3f418e8363bb80e6R19), we havent been able to pinpoint exactly when this happens but we've seen it, so I think there must be some views in/around Odoo that don't have the UserMenu.

Tardo:
Deprecated. See: https://github.com/OCA/web/pull/1759

PR #1746, opened 2020-11-26
[13.0][MIG] web_pivot_computed_measure: Migration to 13.0

ernestotejeda:
cc @Tecnativa TT25958

Main changes:
- Adapt the javascript code to the new data structure used to manage the data to be displayed in the pivot view.

PR #1751, opened 2020-12-02
[13.0][IMP] web_pivot_computed_measure: Saving vertical space

ernestotejeda:
cc @Tecnativa TT25958

Put in the same line the label and the dropdowns for saving vertical space as @pedrobaeza commented in the PR of the migration of this module to v13 https://github.com/OCA/web/pull/1746#pullrequestreview-539258233
Current
image

Desired
image

### DIFF
--- a/web_pivot_computed_measure/static/src/scss/web_pivot_computed_measure.scss
+++ b/web_pivot_computed_measure/static/src/scss/web_pivot_computed_measure.scss
@@ -1,0 +1,9 @@
+#add_computed_measure_wrapper {
+    padding: 0 20px;
+    min-width: 300px;
+    white-space: nowrap;
+    .d-table-cell {
+        vertical-align: middle;
+        padding: 3px 0;
+    }
+}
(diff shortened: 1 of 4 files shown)


web_pivot_computed_measure/static/src/scss/web_pivot_computed_measure.scss
```diff
@@ -0,0 +1,8 @@
+#add_computed_measure_wrapper {
+    padding: 0 20px;
+    min-width: 300px;
```
Tardo, line 3: You can use the 'o-search-options-dropdown-custom-item' mixin.

Use 'white-space: nowrap' to force text in a single line :)

Tardo (changes requested):
I think that you can use bootstrap "sugar" class helpers here...
https://getbootstrap.com/docs/4.2/components/dropdowns/#forms
https://getbootstrap.com/docs/4.2/components/forms/#horizontal-form

joao-p-marques (approved):
Code and functional review :+1:

web_pivot_computed_measure/static/src/scss/web_pivot_computed_measure.scss
ernestotejeda: Thanks @Tardo , I added 'white-space: nowrap but not 'o-search-options-dropdown-custom-item' mixin because it restricts the element to have a maximum width of 250px and that is not what I want.

Tardo (approved):

PR #1760, opened 2020-12-13
[12.0][IMP] web_widget_numeric_step: better visualization in list view cells

ernestotejeda:
cc @Tecnativa TT25581

better visualization in list view cells. If a list view has many columns and the widget field (numeric step) has a very small width, editing it is very difficult.
Before:
image
After:

### DIFF
--- a/web_widget_numeric_step/static/src/css/numeric_step.scss
+++ b/web_widget_numeric_step/static/src/css/numeric_step.scss
@@ -1,3 +1,6 @@
 .widget_numeric_step {
     display: inline-flex;
 }
+.numeric_step_editing_cell {
+    min-width: 120px;
+}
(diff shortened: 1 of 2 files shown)


ernestotejeda:
@Tardo could you revew this, please?

legalsylvain:
Hi @ernestotejeda  thanks for the patch.
the problem I see with this implementation is if there are two fields with the same widget. the "plus" button of the left field with be joint to the "minus" button of the right field, that is very confusing for the end user.
I think that the minus and the plus buttons should be at the same side.

kind regards.

CC @quentinDupont

ernestotejeda:
I understand and it is a good topic for a debate. Anyway, even giving a solution to what you commented, I think that if we don't do something like what I am doing here, we would continue with the same problem of not seeing the number that we are editing because the input of the widget remains very small on some occasions. What do you think?

pedrobaeza:
Yes, the scope of this PR is for not hiding the value, not for the position of the buttons.

pedrobaeza (approved):

legalsylvain:
> Yes, the scope of this PR is for not hiding the value, not for the position of the buttons.

OK, let's talk about the position in other place. Thanks for the patch.

marcelsavegnago (approved):

HaraldPanten (approved):

PR #1766, opened 2020-12-28
[CHANGE] Wrap and center app name

letzdoo-js:
Before : Long app names are not wrapped and therefore are covering one each other

Capture d’écran 2020-12-28 à 11 06 36

After : Long app names are wrapped and do not cover each other anymore

Capture d’écran 2020-12-28 à 11 06 48


pedrobaeza:
How is this affecting to the second app row?

PR #1768, opened 2020-12-29
[12.0][FIX] web_widget_numeric_step: Better visualization in list view cells

Tardo:
Fix #1760 changes

Previous PR applies the css class to all elements, so one in "readonly" mode affects to other in "normal" mode.
For a strange cause i need use "defer" .. appears that "start" resolves the promise before insert DOM operation ends.
Now click event only affects if "left click" was pressed.
Now the "auto step" events are cleaned if the widget lost the focus.

cc @tecnativa TT27407

numeric_step

### DIFF
--- a/web_widget_numeric_step/static/src/css/numeric_step.scss
+++ b/web_widget_numeric_step/static/src/css/numeric_step.scss
@@ -1,6 +1,11 @@
 .widget_numeric_step {
     display: inline-flex;
 }
+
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why previous change didn't work?

Tardo:
First comment updated

pedrobaeza (approved):

victoralmau (approved):

PR #1769, opened 2020-12-29
[12.0][IMP] web_widget_one2many_product_picker: Prevent lost search input focus

Tardo:
cc @tecnativa TT27546

### DIFF
--- a/web_widget_one2many_product_picker/static/src/js/views/One2ManyProductPicker/renderer.js
+++ b/web_widget_one2many_product_picker/static/src/js/views/One2ManyProductPicker/renderer.js
@@ -154,7 +154,7 @@ odoo.define("web_widget_one2many_product_picker.One2ManyProductPickerRenderer", 
                             false,
                             true,
                             search_record_index
-                        )
+                        )[0]
                     );
                 }
             }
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I can't try in runbot, but is this really the expected behavior? I'm guessing reading code that this will return focus to text search after clicking on any button, but that includes the search button, which is not the desired behavior. Only clicking on rubber icon should do this, and note that when you do that, the focus can be in any other part.

Tardo:
Now the focus of the "search input" is restored when:
- Click the "eraser" button (the rubber)
- ~~When press on search button~~ When the search is launched by pressing the enter key
- When changes the search mode (Search button dropdown options)

pedrobaeza:
I think that when pressing search button, we shouldn't return the focus to the search input box, but only on the rest of cases you mention.

victoralmau (approved):

chienandalu (commented):
Agreed. The second case makes no sense

Tardo:
My fault. Second point is:
- When the search is launched by pressing the enter key

pedrobaeza (approved):

PR #1776, opened 2021-01-07
[12.0][FIX] web_pwa_oca: manifest and workers should only be present …

diggy128:
…in backend

### DIFF
--- a/web_pwa_oca/templates/assets.xml
+++ b/web_pwa_oca/templates/assets.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
-    <template id="web_layout_pwa" name="Web layout PWA" inherit_id="web.layout">
-        <xpath expr="//meta[@name='viewport']" position="after">
+    <template id="web_layout_pwa" name="Web layout PWA" inherit_id="web.webclient_bootstrap">
+        <xpath expr="//t[@t-call-assets='web.assets_common']" position="before">
             <!-- Add link rel manifest -->
             <link rel="manifest" t-attf-href="/web_pwa_oca/manifest.webmanifest"/>
             <!-- Add iOS meta tags and icons -->


pedrobaeza:
I don't think this is the proper solution, as `website` is a variable for controlling the website, but you can have portal views that are not backend. Please explain more the problem you are trying to solve.

diggy128:
Yeah, it was just a quick fix. I copied how web_favicon handles it.

The issue here is that pwa should be available for backend only. Thus, the manifest and worker scripts should not be on website.

The use case is that when using reverse proxy and have rules in place restricting access to backend for internet users, having pwa manifest and worker scripts on the website result in repetitive 404 errors on proxy (nginx) which results in fail2ban banning the ip altogether.

web_pwa_oca/templates/assets.xml
```diff
@@ -2,23 +2,27 @@
...
-            <!-- Add iOS meta tags and icons -->
-            <meta name="apple-mobile-web-app-capable" content="yes"/>
-            <meta name="apple-mobile-web-app-status-bar-style" content="black"/>
-            <t t-set="pwa_name" t-value="request.env['ir.config_parameter'].sudo().get_param('pwa.manifest.name')"/>
-            <meta name="apple-mobile-web-app-title" t-att-content="pwa_name"/>
-            <link rel="apple-touch-icon" href="/web_pwa_oca/static/img/icons/icon-152x152.png"/>
-            <!-- Add meta theme-color -->
-            <t t-set="pwa_theme_color" t-value="request.env['ir.config_parameter'].sudo().get_param('pwa.manifest.theme_color')"/>
-            <meta name="theme-color" t-att-content="pwa_theme_color" />
+            <t t-if="not website">
```
Tardo: I think it would be better to change the "web.webclient_bootstrap" (t-set='head_web') template instead of "web.layout"...
what do you think?

web_pwa_oca/templates/assets.xml
```diff
@@ -2,23 +2,27 @@
...
+                <t t-set="pwa_theme_color" t-value="request.env['ir.config_parameter'].sudo().get_param('pwa.manifest.theme_color')"/>
+                <meta name="theme-color" t-att-content="pwa_theme_color" />
+            </t>
         </xpath>
     </template>
     <template id="assets_backend" name="web service worker assets" inherit_id="web.assets_backend">
         <xpath expr="." position="inside">
-            <script type="text/javascript" src="/web_pwa_oca/static/src/js/pwa_manager.js"></script>
-            <script type="text/javascript" src="/web_pwa_oca/static/src/js/webclient.js"></script>
+            <t t-if="not website">
```
Tardo: This is not really necessary... "web.assets_backend" is called in "web.webclient_bootstrap" that is rendered when loads "/web".

Tardo (changes requested):

web_pwa_oca/templates/assets.xml
diggy128: Since web.webclient_bootstrap is used as a generic loader, I think it would be best to have all JS in assets_backend as it should be.
In that case someone can reroute JS below the fold if needed for backend.

web_pwa_oca/templates/assets.xml
diggy128: See comment below. I believe this should be moved to the top of web.assets_backend.
@pedrobaeza what do you think?

pedrobaeza: My colleague is the expert on this, so I don't have a technical opinion on this. @Yajo can also give an opinion.

web_pwa_oca/templates/assets.xml
yajo: IIUC what @Tardo means is that `assets_backend` implies that you're already in the backend, so no need to check for website existence.

See my comment above too.

web_pwa_oca/templates/assets.xml
```diff
@@ -2,23 +2,27 @@
 <odoo>
     <template id="web_layout_pwa" name="Web layout PWA" inherit_id="web.layout">
```
yajo: I agree with https://github.com/OCA/web/pull/1776#discussion_r557772635.

The change you should do to fix your problem is this line, and revert all other changes in the PR:

```suggestion
    <template id="web_layout_pwa" name="Web layout PWA" inherit_id="web.webclient_boostrap">
```

yajo (changes requested):

yajo (approved):

Tardo (approved):

PR #1780, opened 2021-01-13
14.0 mig web tree many2one clickable

Kev-Roche:


StefanRijnhart:
Thanks!

Non blocking: the second README.rst commit seems to be the revert of the first, so you can squash these two commits out of the way I think?

Non blocking #2: Oh, you left out committing the pre-commit changes separately, but there are no other changes so it does not make review any more difficult.

StefanRijnhart (approved):

Kev-Roche:
@StefanRijnhart Thank you for your advice, I made the squash.

rconjour (approved):

PR #1787, opened 2021-01-22
[13.0][web_widget_domain_editor_dialog][FIX] Error when used in wizards.

JordiBForgeFlow:
When clicking edit the list of record an error happened due to lack
of proper initialization of the default domain.

https://user-images.githubusercontent.com/7683926/105528833-c35f4900-5ce5-11eb-8ecb-f6dc6f185a6c.mp4

@chienandalu @CarlosRoca13

### DIFF
--- a/web_widget_domain_editor_dialog/static/src/js/basic_fields.js
+++ b/web_widget_domain_editor_dialog/static/src/js/basic_fields.js
@@ -16,6 +16,9 @@ odoo.define("web_widget_domain_editor_dialog.basic_fields", function(require) {
             if (this.mode === "readonly") {
                 return this._super.apply(this, arguments);
             }
+            if (!this.value) {
+                this.value = [];
+            }
             const dialog = new DomainEditorDialog(this, {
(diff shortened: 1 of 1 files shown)


web_widget_domain_editor_dialog/static/src/js/basic_fields.js
```diff
@@ -16,6 +16,9 @@ odoo.define("web_widget_domain_editor_dialog.basic_fields", function(require) {
             if (this.mode === "readonly") {
                 return this._super.apply(this, arguments);
             }
+            if (this.value == false) {
```
chienandalu: If it's `undefined` it won't be `false` wich is a very specific property, so maybe it would be more appropiate something like this:
```suggestion
            if (!this.value) {
```

JordiBForgeFlow: Fixed. Thanks @chienandalu !

chienandalu (approved):

PR #1792, opened 2021-01-27
[FIX] web_widget_one2many_product_picker: Update widget state

Tardo:
cc @tecnativa TT27848


Tardo:
Moved commits to https://github.com/OCA/web/pull/1802

PR #1794, opened 2021-01-28
[14.0][mig] web_drop_target

dsolanki-initos:


fshah-initos (approved):

dsolanki-initos:
@jarroyomorales can you please review this PR?

dsolanki-initos:
@LoisRForgeFlow Will you please review this PR?
Thanks in advance!

fuentes73:
Hi! Is this module necessary yet? In v14.0 by default you can drag&drop attachments 
This is a record from runbot.odoo.com
drop

pedrobaeza:
The idea of this module is to have the ability to drop objects in any place of the web client and react according that. It's for example used for dragging and dropping files in DMS. Said that, I don't know if this code is using core mechanism for that or implementing their owns.

fuentes73:
Hi @pedrobaeza I know the idea of this module.
I don't know if it's an easy tasks but will be better deprecate this module and adapt DMS to use core mechanism?

OTOH, with this module installed you can see the message duplicate, is not a problem because the attachment is created only once
captura-web-drop-target

pedrobaeza:
If DMS can use drag & drop mechanism, indeed it would be better, but I don't know if a great amount of code is needed for that and a module like this one is still desirable for containing such amount of code. Maybe you can draft something for using the core mechanism in DMS? :pray:

victoralmau:
According to mi tests that need it in https://github.com/OCA/dms/pull/66 work fine.

victoralmau (approved):

pedrobaeza:
I would remove that double entry in the UI, but I have also a doubt: it's supposed that the drag and drop feature for chatter is added in mail_drop_target, and this one doesn't depend on `mail`, so I don't know why this enables the other feature.

@etobella maybe you can help here.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -98,12 +98,13 @@ odoo.define("web_drop_target", function (require) {
             }).then(function () {
                 // Find the chatter among the children, there should be only
                 // one
-                var res = _.filter(self.getChildren(), "chatter");
-                if (res.length) {
-                    res[0].chatter._reloadAttachmentBox();
-                    res[0].chatter.trigger_up("reload");
-                    res[0].chatter.$(".o_chatter_button_attachment").click();
-                }
+                var chatterContainerComponent = self.getChildren()[1]
```
pedrobaeza, line 101: I think this code is the offending one that is adding this double.

victoralmau:
IMO it's not ready yet because if i try to drop image in "attachments chatter" appear 2 attachments (duplicate).

pedrobaeza:
@victoralmau can you please check my comment about removing that code to see if it disappears? This module must not handle anything from `mail` module.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -0,0 +1,200 @@
...
+                // Find the chatter among the children, there should be only
+                // one
+                var chatterContainerComponent = self.getChildren()[1]
+                    ._chatterContainerComponent;
+                var chatter = chatterContainerComponent.__owl__.refs.component.chatter;
+                chatter.thread.refresh();
+                chatter.update({
+                    isAttachmentBoxVisible: !chatter.isAttachmentBoxVisible,
+                });
+            });
```
victoralmau, line 108:
```suggestion
            })
```

victoralmau: I have testing it locally and remove it to prevent duplicate attachment creation.

victoralmau (changes requested):

pedrobaeza:
@victoralmau I think you can supersede this, so we can merge it ASAP and don't block dependent modules.

pedrobaeza:
Superseded by #1892

PR #1795, opened 2021-01-28
[10.0] [FIX] web_m2x_options:  Open m2m_tag form into the same view mode as the parent form

lmignon:
On many2many_tags widget open dialog with the same view mode as the parent view

Before this change the Form was always opened in edit mode when clicking on a many2many_tags. With this change, the Form is always open into the same mode as the parent form.


thomaspaulb (approved):

thomaspaulb:
@lmignon Do you have any idea why Travis is failing?

PR #1797, opened 2021-02-01
[12.0][IMP] web_responsive: Bigger checkboxes in list view

Tardo:
big_checkboxes

Needs a fw-port after merging!

cc @tecnativa TT27826


pedrobaeza:
They look a bit ugly... Can you check several things?:

- The border color is the same?
- Can you round a bit the corners?

legalsylvain (commented):
Hi @tardo

Regarding the UX, your PR is cool. (=more easy to click on checkbox, that was before too small)
From an aesthetic point of view, I agree with @pedrobaeza, the border are too wide. 

image






image

thanks.

Tardo:
Thanks. Yes... i'm trying to use "zoom" instead of "scale" to avoid 'pixelation'... but using zoom is a little more tedious because it changes the row height.

Tardo:
Updated!
- No pixelation
- New border style

pedrobaeza (approved):
That's way better!

legalsylvain:
Tested again.
From an aesthetic point of view, it's OK.
But the UX remains the same as before. (the button is not bigger, so there is a lot of space where it is possible to click without selecting).

Not a blocking point, though, just a remark.

image

kind regards.

pedrobaeza:
Something weird is happening then, as that doesn't match with the initial screenshot. I made a code review though.

Tardo:
Thanks @legalsylvain ... Appears that firefox doesn't like "zoom" property on checkboxes... :/

legalsylvain:
> Thanks @legalsylvain ... Appears that firefox doesn't like "zoom" property on checkboxes... :/

Indeed : 
- Firefox : :-1: 
- Chromium : :+1:

Tardo:
Updated!

big_checkboxes

legalsylvain (approved):
Thanks ! works like a charm on Firefox. (84.0.2)

pedrobaeza (approved):

pedrobaeza:
I think it's worth to check the before/after comparing them side by side for seeing that no extra height is consumed:

checkboxes

mamcode (approved):
LGTM :+1:

etobella (approved):
Awesome!!!

pedrobaeza:
/ocabot merge patch

Please fw-port it to v13

legalsylvain:
Hi @Tardo 

I just checked in a tree view with the widget ``widget="boolean_toggle"``
without your commit. The display is OK.
image
with your commit, the display seems KO
image

Could you check if you have a similar problem ? 

thanks ! 

(Occures on Chromium and Firefox).

PR #1798, opened 2021-02-02
[FIX][12.0] search_panel view: Do not apply model domain to comodel

katyukha:
Description
-----------

This commit fixes incorrect bechavior (error thrown) in case when
additional `domain` provided to action (`ir.actions.act_window`)
that displays view with search panel enabled.

Before this commit
------------------

For example we have following models:
- My Category
- My Model

And in category view, we have stat-button that display number of records
of records in this category. On click, it have to open view for My
Model, with domain like `[('category_id', '=', 42)]`.
In this case, following error will be raised:

```tracaback
Error:
Odoo Server Error

  ...
  File "/opt/odoo/custom_addons/web_view_searchpanel/models/base.py", line 60, in search_panel_select_range
    hierarchical_naming=False).search_read(model_domain, fields),
  File "/opt/odoo/odoo/odoo/models.py", line 4615, in search_read
    records = self.search(domain or [], offset=offset, limit=limit, order=order)
  File "/opt/odoo/odoo/odoo/models.py", line 1581, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/opt/odoo/odoo/odoo/models.py", line 4147, in _search
    query = self._where_calc(args)
  File "/opt/odoo/odoo/odoo/models.py", line 3939, in _where_calc
    e = expression.expression(domain, self)
  File "/opt/odoo/odoo/odoo/osv/expression.py", line 673, in __init__
    self.parse()
  File "/opt/odoo/odoo/odoo/osv/expression.py", line 854, in parse
    raise ValueError("Invalid field %r in leaf %r" % (left, str(leaf)))
ValueError: Invalid field 'category_id' in leaf "<osv.ExtendedLeaf: ('category_id', '=', 26) on bureaucrat_knowledge_category (ctx: )>"
```

Diagnostics
-----------

It seems that model domain was passed to comodel, thus system cannot
find field related to model in comodel. See code
(web_view_searchpanel/models/base.py", line 60, in
search_panel_select_range)

As tested, the `search_domain` causes this bug.

But if we look at implementation of same method
(`search_panel_select_range`) in Odoo 13.0 (see
[code](https://github.com/odoo/odoo/blob/13.0/addons/web/models/models.py#L214))
then we can see, that there is only empty domain applied for search.

Solution
--------

It seems, that variable `model_domain` could be simply removed, and
we could do the search in comodel without any extra domain in this case.

So, this commit, only makes imlementation of this method look same as in
Odoo 13.0

After this commit
-----------------

Everything is working fine.

### DIFF
--- a/web_view_searchpanel/models/base.py
+++ b/web_view_searchpanel/models/base.py
@@ -48,16 +48,10 @@ class Base(models.AbstractModel):
         if parent_name:
             fields.append(parent_name)
 
-        model_domain = expression.AND([
-            kwargs.get('search_domain', []),
-            kwargs.get('category_domain', []),
-            kwargs.get('filter_domain', []),
-        ])
-
         return {
             'parent_field': parent_name,
             'values': Comodel.with_context(
-                hierarchical_naming=False).search_read(model_domain, fields),
+                hierarchical_naming=False).search_read([], fields),
         }
 
     @api.model


katyukha:
Hi @keshrath, @etobella,

Can you review this pull-request?

Thanks!

katyukha:
It seems that change, that added this bug was added via this commit (https://github.com/muk-it/muk_web/commit/702eab721c930d43784068afb8078af86e7b7713#diff-e6c9cddaa493ae1763ae87e87ddadb4f2d9108d8686b5b21c1f33b39e0ac025b)

etobella (approved):
I think we need to maintain the same logic than odoo

katyukha:
@etobella thank you for review.

And i have a question, what are the next steps to get this PR merged?

etobella:
Well, It would be interesting that someone checks this PR besides me.
As this is used by DMS, maybe we can ask kindly a review from @victoralmau

victoralmau:
I have reviewed this change in dms 12.0 and ran the tests and it does not seem to affect this change.
If you @etobella  see the correct change from me too.
Ping @pedrobaeza  to confirm it in case you know other addons use it and it can have side effects.

pedrobaeza:
The only OCA module that I know using it is DMS, but seems legit.

victoralmau (approved):

PR #1802, opened 2021-02-03
[12.0][IMP+FIX] web_widget_one2many_product_picker: Multiple changes

Tardo:
- Now changes are applied without confirm/add the record (but only if modified)
- Less time for the 3D 'flip' transition
- Other minor Improvements (less memory usage, transitions, ...)
- Use correct state when update current records
- New button to add products fast
- Force to use qty 1.0 in "pure virtual" records to ensure get useful onchange data

product_picker

product_picker_anat

cc @tecnativa TT28012


chienandalu (approved):

marcelsavegnago (approved):

Tardo:
PR Ready, first comment updated.

pedrobaeza (approved):

PR #1807, opened 2021-02-04
[IMP] web_dialog_size: add task to roadmap

yajo:
It is important to let users decide their default dialog size. Some have ultrawide monitors and this module does more harm than benefit.

Here I add such task to roadmap, to avoid forgetting about it.

@Tecnativa

### DIFF
--- a/web_dialog_size/README.rst
+++ b/web_dialog_size/README.rst
@@ -43,6 +43,11 @@ If you want to set dialog boxes maximized by default, you need to:
 #. Add a new record with the text *web_dialog_size.default_maximize* in
     the *Key* field and the text *True* in the *Value* field
 
+Known issues / Roadmap
+======================
+
+* Allow setting default dialog size per user.
+
 Bug Tracker
 ===========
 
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

victoralmau (approved):

CarlosRoca13 (approved):

chienandalu (approved):

bealdav:
Just a question. Some users may have some ultra wide screens and very small screens.

Not sure a setting by user will help in this case

yajo:
Forward ported to v13 3542c5afefb86017e5afc5aff53800678e7eee08 and v14 edbd7dbd3fa724c844e08cdf08b60e7b5eda1cae.

@bealdav Yes, my intention is that just the system remembers last user's choice. So, if the user clicks on the "maximize" icon, from there onwards, the dialogs will appear maximized until he clicks on "minimize", and viceversa.

Both options will adapt to screen max after all, so that shouldn't be a big problem.

In any case, it'll be seen when implemented.

PR #1763, opened 2020-12-16
[13.0][FIX] web_responsive: port shortcuts

Tardo:
Superseded https://github.com/OCA/web/pull/1668

cc @tecnativa

PR #1711, opened 2020-10-17
[14.0][MIG] web_responsive

pierre-pizzetta:
- chatter is now an owl component
- template name for 'discuss.mobile' is now 'KanbanView.MobileTabs'
- added <viewport> tag in 'web.template' <head>
- css classes '.oe_chatter' and '.o_chatter' are now replaced with '.o_FormRenderer_chatterContainer'

know issues:
- [x] ~~composer on mobile breakpoint doesn't clear the input after sending~~ edit: actually the behavior is the same on current enterprise runbot and seems to be a bug in the new owl composer component
- [x] viewport information were removed from web.layout and prevent scaling on mobile
- [x] chatter has wrong width ratio on desktop

PR #1819, opened 2021-02-11
[14.0] [MIG] web_responsive

SplashS:
Hello I adapted web_responsive module to 14.0 version.
There are still some problems, but they look like the same as in 13.0 version.
I'm waiting for feedback.


cvinh:
Hello
Thanks a lot for your PR
There was already one here https://github.com/OCA/web/pull/1711
Did you consider it ?
Best regards

SplashS:
Yes, I’ve seen it.
I tried to contact it’s author (see my comment inside it) to append my work to his request,
But there was no answer and I spent a time to create my own PR from the scratch.

SplashS:
I've restored mobile kanban view. Missed it for the first time,
So to recap. My PR support all known to me `web_responsive` cases:
1. Apps menu works.
2. Local menu collapses, topbar minimizes.
3. Searchbar behavior restored close to 13.0.
4. Switchview collapses, quickactions collapses, action buttons minimizes.
5. Kanban mobile view restored as in 13.0.
6. Chatter & discuss look good for me.

Know issues:
1. List view selection suggests to expand selection from one page to all records. This state moves switchview out of small screen. But in this case I think it's not important.
2. With rtl lang on small screen list view optional column selector showed in wrong place. I was trying to solve this problem, but good solution lies in main odoo/odoo codebase I think.

I'm still waiting for feedback :)

cvinh:
Hello
I tested your PR, 
I have an error when opening the Discuss App
(chrome 88.0.4324.93 on android 10)
`Error: QWeb2 - template['KanbanView.MobileTabs']: No enumerator given to foreach`
I guess you can easily reproduce the issue
Best regards

SplashS:
Thanks, i've fixed.
It's interesting that fix is to remove all extension code for discuss from `web_responsive`.
Looks like community version still support mobile for discuss. So, it will be interesting task in future, when mobile discuss will be migrated in enterprise, in 15.0 for example.
Works with same issues as in Enterprise:
1. From full screen chat I can't go in form view (for example, res.user) (It's actually loaded, but fullscreen chat overlaps it).
2. In mobile channel there is no option to invite people, except `@` in message (no Invite button like in desktop).

web_responsive/models/res_users.py
```diff
@@ -0,0 +1,27 @@
...
+        string="Chatter Position",
+        default="sided",
+    )
+
+    def __init__(self, pool, cr):
+        """Override of __init__ to add access rights.
+        Access rights are disabled by default, but allowed on some specific
+        fields defined in self.SELF_{READ/WRITE}ABLE_FIELDS.
+        """
+        super(ResUsers, self).__init__(pool, cr)
```
Tardo:
```suggestion
        super().__init__(pool, cr)
```

web_responsive/readme/ROADMAP.rst
```diff
@@ -0,0 +1,7 @@
+* To view the full experience in a device, the page must be loaded with the
+  device screen size. This means that, if you change the size of your browser,
+  you should reload the web client to get the full experience for that
+  new size. This is Odoo's own limitation.
+* App navigation with keyboard.
+* Handle long titles on forms in a better way
+* Standard sticky headers seems to not work properly on iOS Safari/Chrome (see #1626).
```
Tardo, line 7: is this still happening?

web_responsive/static/src/js/web_responsive.js
```diff
@@ -0,0 +1,574 @@
...
+    const AbstractWebClient = require("web.AbstractWebClient");
+    const AppsMenu = require("web.AppsMenu");
+    const BasicController = require("web.BasicController");
+    const config = require("web.config");
+    const core = require("web.core");
+    const FormRenderer = require("web.FormRenderer");
+    const Menu = require("web.Menu");
+    const RelationalFields = require("web.relational_fields");
+    const ListRenderer = require("web.ListRenderer");
+    const DocumentViewer = require("mail.DocumentViewer");
```
Tardo: Now appears that needs use AttachmentViewer instead of DocumentViewer

Tardo (changes requested):
Looks good... but i miss some features/improvements:
- Focus on searchbox in the AppDrawer
- DocumentViewer improvements
- For some reason, now uses the kanban view by default in all screen sizes
- Maybe fix this issue:
v14_responsize_a

cvinh:
Hello
I checked your last commit
Small issue : the checkboxes are too big in list view on a computer... on a smartphone it's good size
Best regards

pedrobaeza:
> Hello
> I checked your last commit
> Small issue : the checkboxes are too big in list view on a computer... on a smartphone it's good size
> Best regards

That's on purpose, and it has been done that way on previous versions as well recently. The reasoning behind this is that the check was not easy to be marked with the mouse, and also for consistency, so that's more valued than the aesthetics.

web_responsive/models/res_users.py
SplashS: accepted

web_responsive/static/src/js/web_responsive.js
SplashS: > Looks good... but i miss some features/improvements:
> 
> * Focus on searchbox in the AppDrawer

It seems that this the initial idea according to comment `Autofocus on search field on big screens`.
Should we autofocus on small devices too? Why?
Or it doesn't work on your desktop environment? For me it works on desktop.

> * DocumentViewer improvements

OK, will work on it quite later.

> * For some reason, now uses the kanban view by default in all screen sizes

It seems like common odoo community problem. I reproduced it on small screen (only) on odoo 14.0 runbot.
Should we fix this here or report to upstream?

> * Maybe fix this issue:
>   v14_responsize_a

Fixed

I made some improvement to scrolling.
Will check Safari problem mentioned before later when runbot ready.

web_responsive/readme/ROADMAP.rst
SplashS: I've reproduced this bug and fixed. This was because `position: sticky;` was applied to both `thead` and `th`.
I left only `th`.

web_responsive/static/src/js/web_responsive.js
SplashS: There is a problem with reimplementing min/max buttons on AttachmentViewer.
AttachmentViewer is Component. And I know the only way to patch Component via PatchMixin.
The problem is that PatchMixin must be applied to _patched_ Component. And AttachmentViewer doesn't have it.
I don't know how to extend Component that doesn't have PatchMixin.
Of course I will think about it more.
Any suggestions?

SplashS:
I migrated DocumentViewer improvements to AttachmentViewer.
Please check functionality. May be I missed some.

To recap:
I've consider all comments except:
> Focus on searchbox in the AppDrawer

It seems that this the initial idea according to comment `Autofocus on search field on big screens`.
Should we autofocus on small devices too? Why?
Or it doesn't work on your desktop environment? For me it works on desktop.

> For some reason, now uses the kanban view by default in all screen sizes

As I understand it about problem that page reload switches view to kanban.
It seems like common odoo community problem. I reproduced it on small screen (only) on odoo 14.0 runbot.
Should we fix this here or report to upstream?

web_responsive/static/src/js/web_responsive.js
SplashS: AttachmentViewer patched.
See my comment below to recap.

Tardo (changes requested):
Looks better! but i'm still see some issue:

- Issues with searchbox and buttons in some resolutions
issues_web_responsive

- Issue with shortcuts: I can't open AppDrawer (ALT+SHIFT+A)

------

Now autofocus works well on desktop (maybe a problem with my local cache?)
Now the kanban view works well too... i see kaban in mobile and list on desktop :) (maybe a problem with my local cache?)

SplashS:
> Looks better! but i'm still see some issue:
> 
> * Issues with searchbox and buttons in some resolutions

I've double checked all resolutions and made many breakpoints for nice look of control panel.

> * Issue with shortcuts: I can't open AppDrawer (ALT+SHIFT+A)

This is not a problem of migration. On 13.0 runbot this also doesn't work.
I have some thoughts about fix, but this PR is not right place for them.
Cause is that Chrome doesn't support accesskey for some combinations.
For example, Alt + Shift + [A, B, I, T] - not work.
Simple fix is to downgrade to accesskey='h' for appmenu.
And extended accessibility with Alt + Shift looks quite unfinished.
I found different keys for Save button in list view and form view.
Shortcut help shows old combinations.
One more time: accessibility is the object of different discussion.

> 
> Now autofocus works well on desktop (maybe a problem with my local cache?)
> Now the kanban view works well too... i see kaban in mobile and list on desktop :) (maybe a problem with my local cache?)

Nice to hear it :)

pedrobaeza:
> This is not a problem of migration. On 13.0 runbot this also doesn't work.
I have some thoughts about fix, but this PR is not right place for them.
Cause is that Chrome doesn't support accesskey for some combinations.
For example, Alt + Shift + [A, B, I, T] - not work.

Indeed that problem does exist in previous versions, but on Chrome. On Firefox, it should work properly. I suggest to keep the `A`, but also add `H` as suggested (the enterprise option).

About the shortcuts, #1763 should draw them correctly.

SplashS:
> Indeed that problem does exist in previous versions, but on Chrome. On Firefox, it should work properly. I suggest to keep the `A`, but also add `H` as suggested (the enterprise option).

There is one more problem with A accesskey - Activity view switch also uses it.

> About the shortcuts, #1763 should draw them correctly.

Alt + Shift + displays correctly in help, but actual accesskeys are not overwritten.

I made separate issue for this problems. #1830

SplashS:
@Tardo I’m sorry. I’ve seen that you requested some changes. But I missed them and perhaps overwrote your changes request by force push.

Tardo (approved):
@SplashS LGTM Thanks for the work and explanations. Only a non-blocking comment about "prev" and "next" buttons in the attachment preview... they should be invisible if the cursor is outside of the preview panel.

I've been told about a possible enhancement to AppDrawer's search engine to be able to search menus ignoring accents.
I have seen that it currently uses "fuzzy.js" for search, but the library seems to be abandoned (https://github.com/myork/fuzzy).
We could add the use of "fuse.js" and do what they say here: https://github.com/krisk/Fuse/issues/415 or implement the "unnacent" feature with the current fuzzy.js library.

What do you think?

SplashS:
> Only a non-blocking comment about "prev" and "next" buttons in the attachment preview... they should be invisible if the cursor is outside of the preview panel.

Ok, I've found this functionality. Now I migrated all DocumentViewer styles to AttachmentViewer.

Also I've fixed issue with Action Menu in Form View on 'xs' screens.
And also I've set `sequence` for this module to 1. Without it some standard modules like `crm_iap_lead` overwrote `ListView.buttons` and broke consistency (of accesskeys f.x.) I think this module deserves to load as earlier as possible.

SplashS:
> I've been told about a possible enhancement to AppDrawer's search engine to be able to search menus ignoring accents.
> I have seen that it currently uses "fuzzy.js" for search, but the library seems to be abandoned (https://github.com/myork/fuzzy).
> We could add the use of "fuse.js" and do what they say here: [krisk/Fuse#415](https://github.com/krisk/Fuse/issues/415) or implement the "unnacent" feature with the current fuzzy.js library.
> 
> What do you think?

I think it's possible but it should be done not only for 14.0.
I will play with it quite later.

SplashS:
> I've been told about a possible enhancement to AppDrawer's search engine to be able to search menus ignoring accents.
> I have seen that it currently uses "fuzzy.js" for search, but the library seems to be abandoned (https://github.com/myork/fuzzy).
> We could add the use of "fuse.js" and do what they say here: [krisk/Fuse#415](https://github.com/krisk/Fuse/issues/415) or implement the "unnacent" feature with the current fuzzy.js library.

I've digged into fuzzy search.
So unaccent searching is hard to implement with current fuzzy library (which moved to https://github.com/mattyork/fuzzy but last commit dated as 2016 year).
Suggested by you `fuse.js` library can be used to implement unaccent search, but it must be very nice tuned. For example it's nessesary to rewrite not only accents but some letters too (`ё` in russian is different letter, not simply `е` with accent, same situation with polish `ł` and `l`). The goal can be reached by collecting all mappings between accented letters and their unaccented variants. Something like `deburredLetters` in https://github.com/lodash/lodash/blob/4.1.0-npm-packages/lodash.deburr/index.js but extended with different letters mapping, not only accented letters.

But I think that this feature requires separate PR.
What is nessesary to merge this PR?

pedrobaeza:
Yes, that can be elaborated in other PR, as a migration having the same features is enough. @Tardo, is it OK for you with all the features in v13?

pedrobaeza:
Some administrative commit squashing is missing, but I can do it if you want before merging for reducing a bit the commit number.

SplashS:
> Some administrative commit squashing is missing, but I can do it if you want before merging for reducing a bit the commit number.

Yes, do it please. I've tried to do it, but did not cope. It was hard for me even to add symlink to repository cause my OS is Windows.

pedrobaeza:
> > Some administrative commit squashing is missing, but I can do it if you want before merging for reducing a bit the commit number.
> 
> Yes, do it please. I've tried to do it, but did not cope. It was hard for me even to add symlink to repository cause my OS is Windows.

Wow, if you want to do some serious Odoo stuff, I recommend you to go to Linux, as if not, you have to deal with a lot more problems than on Linux platform.

Tardo (changes requested):
Thanks for the information! unnacent feature doesn't seem to be an easy task.

I detect some differences:
- The employees view doesn't looks good on small screens:
web_resp_fA

- Chatter Scrollbar doesn't look the same as in 13.0:
web_resp_scroll

- Attachment Preview buttons should appear when the cursor is hovered over, not when the element has the focus.

SplashS:
> * The employees view doesn't looks good on small screens:

Yep! This one I've missed. I implemented mobile searchpanel with behavior close to Enterprise.

> * Chatter Scrollbar doesn't look the same as in 13.0:
> * Attachment Preview buttons should appear when the cursor is hovered over, not when the element has the focus.

Sorry! I can't reproduce these issues. Give me please more details. I tested on Chrome / Windows.

pedrobaeza:
Another one to be included here :pray: 

#1853

Tardo:
@SplashS 
v12  chatter scroll
scroll_chatter_12

v13 chatter scroll
scroll_chatter_13

You can see that the scrollbar is not being displayed in the same element.

Please, add #1856 instead of #1853 :)

pedrobaeza:
Another PR to take into account: #1856

SplashS:
> v12 chatter scroll
> v13 chatter scroll

Ok, i've got it. Fixed.

> Attachment Preview buttons should appear when the cursor is hovered over, not when the element has the focus.

Still can't reproduce this issue. May be it is browser dependent?

SplashS:
> Another PR to take into account: #1856

Ok. I will do it soon. But to be sure I described my plan. Please correct me cause I have not a lot practice in Git (just theory :) )
I will cherry pick all commits after last commit from 13.0 that currently found in this PR. (There are several commits in 13.0 after I started migration).
And then reorganize commits order to make migration commit the last to be able amending possible future fixes for migration.
Right?

pedrobaeza:
Yes, it seems OK. Let me know if you have any trouble.

SplashS:
> Another PR to take into account: #1856

I rebased to append latest PRs from 13.0
Also tried to reduce commits history.

To recap. I haven't undone tasks for this PR except I still can't reproduce this:

> Attachment Preview buttons should appear when the cursor is hovered over, not when the element has the focus.

pedrobaeza:
@Tardo if you are not able to explain the way to reproduce the problem, and there's no more pending things, we can proceed to merge.

pedrobaeza:
About the administrative commit squashing, let me show you what I referred:
Selección_001
The idea is to squash "useless" commits with the real commit that generated them.

SplashS:
> The idea is to squash "useless" commits with the real commit that generated them.

I only merged allmost translation commits as described in https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests
There is only two translation commits left. Is it OK?
Squash or fixup for version change commits is better? Should I concat commit messages or just use the meaningful one?

SplashS:
> Squash or fixup for version change commits is better? Should I concat commit messages or just use the meaningful one?

I did fixup. Seems good. Only 103 commits.

I checked 2 newly included PRs ([FIX] web_responsive: Do not set overflow property on inline fields,
[IMP] web_responsive: Scrollable dropdowns). Seems that they work.

> Attachment Preview buttons should appear when the cursor is hovered over, not when the element has the focus.

Also I recorded my hover behavior for sided attachment viewer:
hover

Tardo (approved):
LGTM 👍🏻 Thanks for the work! now all seems to work fine :) 👏🏻 👏🏻

pedrobaeza:
Thanks for the excellent work!

/ocabot merge nobump

GSLabIt:
@SplashS is there any chance to back port to v 12?

PR #1822, opened 2021-02-15
[13.0][ADD] Boolean Button Ported from 12 to 13

SkiBY:


web_boolean_button/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Boolean Button Widget",
+    "version": "13.0.0.1.0",
```
thomaspaulb: It's customary to start with version 1.0.0, so `13.0.1.0.0`

web_boolean_button/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "category": "Web",
+    "summary": "",
```
thomaspaulb: Fill the summary here with a oneliner of what this module does

web_boolean_button/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "license": "LGPL-3",
+    "author": "Odoo Community Association (OCA)," "Sergei Ruzki,",
```
thomaspaulb: In the guidelines (sorry about so many customs but you will get used to them) we say that the company that made the module comes first, then other contributors, then OCA. So normally it would be:

`Bigdata BY,Odoo Community Association (OCA)`

However, because this is a forward port of some Odoo SA code, [there are some extra rules](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#10backporting-odoo-modules).

In this case I would say go for:

`Bigdata BY,Odoo SA,Odoo Community Association (OCA)`

Your name then comes in the `CONTRIBUTORS.rst` file, which should be in the `readme` folder.

The files you can put there are [shown in the module template](https://github.com/OCA/maintainer-tools/tree/master/template/module/readme), you dont have to add all just `USAGE.rst`, `CONTRIBUTORS.rst` and `CONFIGURE.rst` are the most commonly used ones.

Then, you run the [oca-gen-addon-readme](https://github.com/OCA/maintainer-tools/blob/master/README.md#readme-generator) script which will generate `README.rst` and `static/src/index.html` for you. It sound complex but actually will save a lot of time, you just write a few lines in those files and the generator will do all the markup for you.

web_boolean_button/README.rst
```diff
@@ -0,0 +1,98 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/web-13-0/web-13-0-web_boolean_button
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/162/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+This module change navbar colors based in the company logo colors.
```
thomaspaulb: This is old I guess, but anyway it will be overwritten by `oca-gen-addon-readme` once you run it.

web_boolean_button/README.rst
```diff
@@ -0,0 +1,98 @@
+==================
+Web Boolean Button
+==================
```
thomaspaulb: Then for the last general comment: the naming of your PR. This is not a migration because this module was before not in OCA. So you can use: `[ADD][13.0] Web Boolean Button module` to make it easy for reviewers to see.

Then when in future you also decide to contribute the 12.0 version it would be `[MIG][12.0] Web Boolean Button from 13.0` (or BCP for backport)

web_boolean_button/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Boolean Button Widget",
+    "version": "13.0.0.1.0",
+    "category": "Web",
+    "summary": "",
+    "license": "LGPL-3",
```
thomaspaulb, line 6: If I'm correct, this should be [OPL](https://www.odoo.com/documentation/user/14.0/legal/licenses/licenses.html#odoo-apps) instead, since it's a forward port of code from the 12.0 `web` module. @pedrobaeza Can I tap your knowledge on this one?

thomaspaulb (changes requested):
@SkiBY I'll go into the technical details once the license and guideline stuff has been tackled.

web_boolean_button/__manifest__.py
pedrobaeza: No, we can't publish anything OPL here in OCA. It should respect the previous license, so I think `LGPL-3` seems correct if this comes from Odoo core v12. Isn't this something of enterprise, right?

You should put anyway the disclaimer about being Odoo initial code, and also you need to make a first commit attributing the original code to "Odoo S. A. <info@odoo.com>", and then put successive commits from your part with the needed changes to adapt to the version. See for example https://github.com/OCA/manufacture/pull/501

thomaspaulb: No, it's CE 12.0. Thanks for the clarification!

SkiBY:
Changes after review added.

web_boolean_button/__manifest__.py
SkiBY: hope resolved

thomaspaulb: Yep well done this part.

web_boolean_button/static/tests/form_tests.js
```diff
@@ -0,0 +1,312 @@
+odoo.define("web_boolean_button.form_tests", function(require) {
+    "use strict";
+
+    /* global QUnit*/
+
+    var FormView = require("web.FormView");
+    var testUtils = require("web.test_utils");
+    var createView = testUtils.createView;
+
+    QUnit.module("fields", {}, function() {
```
thomaspaulb: @SkiBY nice work so far! I wish I could approve directly but I still have something to complain about this test:

- When I go to the runbot (Go to first page of PR with our conversation, go to the button, go to one of four checks which is called "Runbot", click the link, click on "connect", end up [here](http://3469276-1822-e928bb.runbot1-3.odoo-community.org/), login as admin, go to `/web/tests`, I cannot find the test for `web_boolean_button`. I think that's because you called this test `fields`. This comes from Odoo 12.0 since there, it's a test that tests all fields, but here it should only test the widget. I think if you just start directly with `QUnit.module("web_boolean_button", ...` such as in [this example](https://github.com/OCA/web/blob/13.0/web_widget_dropdown_dynamic/static/tests/web_widget_dropdown_dynamic_tests.js#L11), and you just leave the last part of your tests, then in your local it will appear in `/web/tests` and it will run. If it runs correctly there, then you can push and I will also see it run correctly here, and I can approve :-)

thomaspaulb (changes requested):

web_boolean_button/static/tests/form_tests.js
SkiBY: Thank you, @thomaspaulb
Naming is changed

thomaspaulb:
@hbrunn It seems something is up here with the Runbot, I can't access the logs. It also asks for approval of a maintainer. Do you know what is going on here?

SkiBY:
@thomaspaulb looks like all green

thomaspaulb (approved):

ntsirintanis (approved):

gurneyalex:
CI on the merge is broken 

* pre commit failure on the website field of the manifest file

@SkiBY can you fix this?

thomaspaulb:
@gurneyalex I fixed that small thing, then squashed the commits. Should be ready for merge now.

PR #1825, opened 2021-02-17
[12.0][FIX] web_widget_one2many_product_picker: Auto save

Tardo:
cc @tecnativa TT28269


chienandalu (commented):
As for me, I already tested it and seems to be ok, although I expecting for customer feedback

PR #1827, opened 2021-02-18
[12.0][FIX] web_widget_one2many_product_picker: Decimal precision

Tardo:
cc @tecnativa TT28298

### DIFF
--- a/web_widget_one2many_product_picker/static/src/js/views/One2ManyProductPicker/record.js
+++ b/web_widget_one2many_product_picker/static/src/js/views/One2ManyProductPicker/record.js
@@ -7,13 +7,13 @@ odoo.define("web_widget_one2many_product_picker.One2ManyProductPickerRecord", fu
     "use strict";
 
     var core = require("web.core");
+    var utils = require("web.utils");
     var Widget = require("web.Widget");
     var Domain = require("web.Domain");
     var widgetRegistry = require("web.widget_registry");
     var tools = require("web_widget_one2many_product_picker.tools");
     var ProductPickerQuickModifPriceForm = require(
         "web_widget_one2many_product_picker.ProductPickerQuickModifPriceForm");
-    var FieldManagerMixin = require('web.FieldManagerMixin');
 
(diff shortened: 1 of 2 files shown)


web_widget_one2many_product_picker/models/ir_http.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2020 Tecnativa - Alexandre D. Díaz
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models, tools
+from odoo.http import request
+
+
+class Http(models.AbstractModel):
+    _inherit = 'ir.http'
+
+    def get_currencies(self):
```
pedrobaeza: This seems a very low level place to put this. Can you do it other way? You already have the currency in the form, and you can pass it as option to the widget if needed (but I don't think so, as the default is to `currency_id` if I'm not wrong.

Tardo: Ok, now currencies are get from the widget.
Yes, by default uses "currency_id" and can be changed via options using "currency_field".

pedrobaeza (commented):
How the monetary field gets this? I think we should mimic it and I don't think we need to dump the roundings for all the currencies.

pedrobaeza:
I insist: can we avoid to retrieve all the currencies rounding. I guess the proper data should be around already thanks to monetary widget.

Tardo:
Know the rounding value is necessary when the widget calculates the subtotal using the discount... we lost the server side rounding.

pedrobaeza:
But how the monetary widget does it? I don't think that's a server operation.

Tardo:
@pedrobaeza For example here, Monetary rounds the value to use in the SQL: https://github.com/odoo/odoo/blob/12.0/odoo/fields.py#L1411

pedrobaeza (approved):
OK, I see

pedrobaeza:
Is this needed in v13, @Tardo ?

Issue #1830, opened 2021-02-23
Keyboard accessibility problems in web_responsive


SplashS:
Migrating `web_responsive` from 13 to 14 I faced some problems in keyboard accessibility.
Due to this is not migration issue I collected all my finds here.

1. Alt + Shift + A for Apps Menu doesn't work on Chrome. This is because `KeyboardNavigationMixin` does "nothing" on Chrome and lets browser process shortcuts by itself. Chrome doesn't support all acesskeys. My investigations show that combinations Alt + Shift + [A, B, I, T] and Alt + [E, F, D] don't work on Chrome (Windows). So the simplest way is to pick up new letter for Apps Menu.
2. `KeyboardNavigationMixin` has the ability to automatically assign acesskeys to buttons. We should extend this functionality to exclude not working shortcuts (Alt + Shift + [A, B, I, T]) from being used as auto accesskeys.
3. There are some interlapping accesskeys. My find is A accesskey used for both Apps Menu and Activity view in SwitchView.
This may be fixed when we pick up different letters for both buttons due to Alt + Shift + A doesn't work on Chrome (but Alt + A does so it's not upsream problem).
4. `web_responsive` overrides Edit accesskey from A to E and Discard accesskey from J to D, but not in all places. F.x. these buttons in editable list view not overwritten. We should verify all accesskey overrides.
5. Shortcuts help shows that control keys on Windows are Alt + Shift, but accesskeys not overwritten in help. F.x. for Edit shown A, but works E and so on. We should make help consistent.

So let's think if the juice is worth the squeeze :)
May be we should fall back to upstream Alt + behavior?

pedrobaeza:
For the last question, a big NO for me. It's a pain to have Alt+number that I use for switching between tabs in Firefox to get overwritten by Odoo an avoiding me to perform it. Each time a open a runbot is like a kick in...

SplashS:
I made some realistic accesskeys transformation.
It can be implemented.
I don't know how to deal with App Menu except to leave H accesskey for it.
I can add both shortcuts A and H but it strange behavior and can affect translation strings.
All other looks good for me:
To `Alt + Shift + 'E'` from `Alt + 'A'` // Odoo Edit (I think A was used cause `Alt + E` doesn't work on Chrome)
To `Alt + Shift + 'A'` from `Alt + 'A'` // Odoo Activity view (it won't work on Chrome, but it's rarely used)
[...]

Issue #1737, opened 2020-11-18
[13.0] web_responsive : Shift + Alt + I import shortcut is not working in Google Chrome. 


MohsenWaleed:
Shift + Alt + I import shortcut is not working in Google Chrome.

pedrobaeza:
Well, in that case, the only solution is to customize such shortcut on Chrome, as Chrome is very greedy with shortcuts and don't allow a web page to use them.

PR #1832, opened 2021-02-24
[13.0] [FIX]: making all accesskeys working on Google Chrome in web_responsive

SplashS:
This is fix for all problems described in #1830 and #1737.

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -797,3 +797,8 @@ html .o_web_client .o_action_manager .o_action {
         }
     }
 }
+
+// Shortcuts table ui improvement
+.o_shortcut_table {
+    width: 100%;
+}
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -16,6 +16,7 @@ odoo.define("web_responsive", function(require) {
     const Chatter = require("mail.Chatter");
     const ListRenderer = require("web.ListRenderer");
     const DocumentViewer = require("mail.DocumentViewer");
+    const ControlPanelRenderer = require("web.ControlPanelRenderer");
 
     /* Hide AppDrawer in desktop and mobile modes.
      * To avoid delays in pages with a lot of DOM nodes we make
@@ -506,6 +507,16 @@ odoo.define("web_responsive", function(require) {
                     newEvent.key = "Shift";
                 }
             }
+            // Fix to make `Alt + Shift + A` shortcut working on Google Chrome.
+            if (
+                alt &&
+                shift &&
+                newEvent.keyCode == 65 &&
+                this.BrowserDetection.isBrowserChrome()
+            ) {
+                setTimeout(() => $("a.full").click(), 0);
+                newEvent.keyCode = 0;
+            }
             return newEvent;
         },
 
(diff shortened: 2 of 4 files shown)


Tardo (approved):
LGTM :+1: Thanks!

web_responsive/static/src/xml/apps.xml
```diff
@@ -14,9 +14,9 @@
         <t t-jquery=".o_app &gt; t" t-operation="replace">
             <t t-call="web_responsive.AppIcon" />
         </t>
-        <!-- Same hotkey as in EE -->
+        <!-- Same hotkey as in EE (Alt + Shift + A doesn't work on Chrome) -->
         <t t-jquery=".full" t-operation="attributes">
-            <attribute name="accesskey">a</attribute>
+            <attribute name="accesskey">h</attribute>
```
pedrobaeza: This is a breaking change, so unless you provide both keys, I don't want to change it.

pedrobaeza (changes requested):

SplashS:
I've made some research about what accesskey is used in different versions of Odoo (without web_responsive) for apps menu:
12.0, community: no accesskey, enterprise: accesskey="h"
13.0,community: no accesskey, enterprise: accesskey="h"
14.0, both: accesskey="h"
master, both: accesskey="h"

So it's strange that web_responsive currently changes it to A. There is no breaking change here.
So I think you told about breaking change for current web_responsive users.
I don't know how to implement both accesskey's on Chrome because `KeyboardNavigationMixin` does nothing on this browser and lets Chrome to handle accesskeys by itself. And duplicate accesskey attributes are not supported.
For Chrome users, Alt + Shift + A currently doesn't work. And we will affect only non Chrome users of web_responsive.
In case of this change they will see visual hint for accesskey, displaying `H` on the icon. And they just learn to new shortcut in few seconds.
If it matter it will possible to make trick for backward compatibility and handle `Alt + Shift + A` in addition for `Alt + Shift + H` only for non Chrome users. But visual hint will show `H` for apps menu icon.
I can try to do it in few days. Do you think it's worth it?

pedrobaeza:
Yes, I mean breaking `web_responsive` habit since a lot of versions. We had that accesskey before Odoo granted one in their core. And we keep `e` as shortcut for edit as more natural and what we had in v8 in core. It would be good if you can find a way to compatibilize both.

SplashS:
OK, I'll try.

web_responsive/static/src/xml/apps.xml
SplashS: I added some hack to support both keys. It's quite dirty but works. Tested on Chrome and Edge on Windows.

pedrobaeza:
Thanks for the investigation. Trying on runbot with Firefox, I get an undesired effect: the opening of the Firefox `Archive` menu, which wasn't happening before:

https://user-images.githubusercontent.com/7165771/122168972-2ca24d00-ce7d-11eb-8be1-c683f37be0fe.mp4

Can you do it the contrary? Let accesskey as `A`, and add through the hack the `H`

SplashS:
I think you insist on this shortcut cause the convenience of pressing it. I agree.
Returning to the root cause I wanted to change it because it doesn't work on Chrome.
But this hack works on Chrome! So I think I can leave only `A` key and write additional support for `Alt + Shift + A` for Chrome and it will be the best solution.

I have a new question :)
Talking about finger convenience. Maybe make `Z` and `X` keys instead of `P` and `N` for prev/next page. According to usability it will be nice. We group all common keys for right-handed users. What do you think?

SplashS:
This is working variant for `Alt + Shift + A`. Tested on Chrome and Edge on Windows.

pedrobaeza:
Yes, @SplashS, one of the goals of keeping this is to have one-hand shortcuts, but also to be consistent across versions. Your proposal about using z/x for previous/next makes sense to me, but preserving as well the others for the second goal.

PR #1848, opened 2021-03-11
[12.0][IMP] web_pwa_oca: Add support for require system

Tardo:
Refactor module to support require system.

Service workers need to load "Odoo Bootstrap" to be able to use "require". The problem is that "Odoo Bootstrap" does operations on the DOM (by using jQuery) and to avoid errors a minimal compatibility layer "jquery-sw-compat.js" has been created to allow using "require" in the service worker.

Moved Service Worker (js) code from the xml to python.

cc @tecnativa


Tardo:
PR Updated

pedrobaeza (approved):
Please squash everything together

pedrobaeza:
And please put explanation in the commit message (not the PR comment).

Tardo:
changes done

chienandalu (approved):

PR #1861, opened 2021-03-18
[12.0] web_m2x_options: fix always show search more

ajaniszewska-dev:
Add fix to show 'search more' on many2one field always when such option is set on field.
Before that, even if field had attribute option={'search_more':True} it wasn't working due to conditions in the function.


simahawk (commented):
LG, can you include module name in commits pls? :pray:

grindtildeath (approved):
LGTM :+1:

simahawk (approved):

PR #1756, opened 2020-12-04
[WIP][12.0][ADD] web_pwa_cache

Tardo:
Enables support for configuring PWA offline/cached mode

- Adds a new form type called 'formPWA' that wins to "form" when user runs in standalone mode.
- Manage cache
- Support offline operations

cc @tecnativa TT23822

PR #1858, opened 2021-03-16
[13.0][MIG] web_widget_one2many_product_picker: Migration to 13.0

Tardo:
cc @tecnativa

PR #1863, opened 2021-03-18
[WIP][13.0][ADD] web_pwa_cache

Tardo:
Continue work from https://github.com/OCA/web/pull/1756

cc @tecnativa


eLBati:
> Depends on: #1858

Why?

pedrobaeza:
I think the dependent PR reference is incorrect. @Tardo can you fix it?

Tardo:
Updated, sorry. Depends on web_pwa_oca which is currently merged.

web_pwa_cache/i18n/es.po
```diff
@@ -0,0 +1,930 @@
...
+#. module: web_pwa_cache
+#. openerp-web
+#: code:addons/web_pwa_cache/static/src/xml/base.xml:0
+#, python-format
+msgid "Offline Mode"
+msgstr "Modo sin conexión"
+
+#. module: web_pwa_cache
+#: model:ir.model.fields,field_description:web_pwa_cache.field_pwa_cache__onchange_field_id
+msgid "Onchage field"
```
yvaucher, line 543:
```suggestion
msgid "Onchange field"
```

Seen in your Odoo days presentation, seems the source term is already fixed.

web_pwa_cache/static/src/js/worker/bus.js
```diff
@@ -0,0 +1,112 @@
...
+                try {
+                    await this._managers.sync.run();
+                } catch (err) {
+                    console.log(
+                        "[ServiceWorker] Can't finish synchronization process succesfully!"
+                    );
+                }
+                try {
+                    // Try prefetch data
+                    await this._components.prefetch.prefetchDataPost();
```
thomaspaulb, line 99: In Chrome Desktop, getting `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'prefetchDataPost')` in this line, after the previous try/catch has failed. Seems strange that it's not caught

web_pwa_cache/static/src/js/worker/bus.js
```diff
@@ -0,0 +1,112 @@
...
+        _doPrefetchDataPost: function() {
+            if (this._prefetch_promise) {
+                return this._prefetch_promise;
+            }
+
+            this._prefetch_promise = new Promise(async (resolve, reject) => {
+                this._prefetch_running = true;
+                // Try sync records first
+                try {
+                    await this._managers.sync.run();
```
thomaspaulb, line 91: Here I'm getting `TypeError: Cannot read properties of undefined (reading 'run')`. It seems that `sync` is not there in `_managers`, only `config`, and `activateWorker`, the place where it's supposed to be filled, does not seem to be called. As a result the UserMenu is not shown.

Tardo: Hi! Please, can you give me more info or/and steps to reproduce?

For example...
The user has the "Enable PWA Cache" group?
Browser?
This is present after pwa installation? before?

Thanks.

thomaspaulb: Actually it was because my `admin` user didnt have that `Enable PWA Cache` group enabled - I didnt know I was supposed to do that.

So I guess that solves my issue just maybe leads to some minor comments such as

- Add [Installation note](https://github.com/Tecnativa/web/tree/13.0-add-web_pwa_cache/web_pwa_cache/readme) mentioning that group, also that Chrome is required and Firefox wont work (and why)
- Better handling of this error and the other one I commented separately - these occur when you have the PWA installed on Chrome 93 on Ubuntu, but your user doesnt have the group.

thomaspaulb: Great work by the way :+1:

Tardo: Thanks! i will check the issues...

P.S: 'admin' user is not the best option to use with pwa cache... you will get all possible records.

web_pwa_cache/static/src/js/worker/core/config.js
```diff
@@ -0,0 +1,147 @@
...
+                        model_info_config,
+                        {
+                            param: param,
+                            value: value,
+                        },
+                        ["param"]
+                    );
+                    console.log(
+                        `[ServiceWorker] Configuration ${param} changed: ${this._cache[param]} -> ${value}`
+                    );
```
thomaspaulb, line 135: If my user does not have the PWA group, and the PWA is not installed, these messages still popup in the console, and when I try to download for example a report or a record, my requests are proxied by the service worker and passed on to Odoo's normal `fetch`.

Is this intended behaviour? It should probably be alright, but perhaps in weird cases it could break things for non-PWA users.

Maybe it is possible to disable the service worker in case the PWA is not installed? The internet says that this is possible if checking for Chrome running in [standalone mode](https://stackoverflow.com/q/21125337) or not.

web_pwa_cache/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,36 @@
...
+
+How Works
+~~~~~~~~~
+
+The module handle two modes:
+
+- Normal: Runs Odoo as normal
+- Standalone: When the app starts you will request to set the mode (offline or online).
+    - Online: All requests except Create, Update and Delete operations goes from cache first, if not found, tries from network.
+    - Offline: All requests goes from/to cache
```
thomaspaulb, line 29: This 'Online mode' still breaks the onchanges and other small things; would it make sense to separate the modes into 'Online Cached' and 'Fully Online'? Then the user can choose if he wants to live with the glitches and save bandwidth, or go for full functionality like in a browser.

web_pwa_cache/static/src/js/worker/core/osv/expression.js
```diff
@@ -0,0 +1,1723 @@
...
+                model.fields[left].type === "boolean" &&
+                ((operator === "!=" && !right) || (operator === "==" && right))
+            ) {
+                query = `(${table_alias}."${left}" IS NOT NULL and ${table_alias}."${left}" != 0)`;
+                params = [];
+            } else if (!right && operator === "!=") {
+                query = `${table_alias}."${left}" IS NOT NULL`;
+                params = [];
+            } else if (operator === "=?") {
+                if (right) {
```
thomaspaulb: Think I spotted a bug here - should be `(!right)`? Popped up when doing a `name_search` on `state` on partner form view, in PWA online mode, while having a country selected.

web_pwa_cache/static/src/js/worker/bus.js
Tardo: Updated Readmes ... i can't reproduce the issue with "sync" :/

web_pwa_cache/static/src/js/worker/core/config.js
Tardo: The service worker is installed "automatically" by the browser... we call PWA to the "Home Icon"/Standalone behaviour.
Check for standalone mode is currently implemented... the problem here comes when you enabled the group and disable it in the future... If you "do nothing" when user is not in the group, the service worker doesn't know that something has been disabled.

web_pwa_cache/static/src/js/worker/core/osv/expression.js
Tardo: Fixed! thanks!!

web_pwa_cache/readme/DESCRIPTION.rst
Tardo: Online mode should not break anything... when you call to an onchange/resource that is not cached the request goes to Odoo server.
I will add it to the roadmap, but in principle it is not a functionality we are looking for on our side.

sergio-teruel:
@AlexandreDz Can you rebase, please??

web_pwa_cache/readme/DESCRIPTION.rst
thomaspaulb: I think what broke was the request that used the `?=` operator - will watch for any other things breaking on Online mode.

Tardo:
@sergio-teruel rebased... 'AlexandreDz' is not me :P

thomaspaulb:
@Tardo We want to implement this for a customer who is currently on v12. Do you think it is doable for us to backport this in a few days, or there will be a lot of hiccups?

pedrobaeza:
I don't think it's easily doable, as it touches very core web client things, and there were heavy changes from v12 to v13.

web_pwa_cache/static/src/js/worker/components/importer.js
```diff
@@ -0,0 +1,209 @@
...
+            return this._db.indexeddb.userdata.put(values);
+        },
+
+        /**
+         * @param {String} model
+         * @param {Object} data
+         * @returns {Promise}
+         */
+        read: function(model, data) {
+            return this._db.write(model, [data.id], data);
```
thomaspaulb, line 102: Managed to end up in a state whereby the data returned on `read` was an array (just on browsing `res.partner` records in form view on a stock 13.0 demo database, nothing fancy), then came up with:

```
           if (_.isArray(data)) {
               var self = this;
               _.each(data, function(item) {
                   self._db.write(model, item.id, item);
               });
           } else {
               return this._db.write(model, [data.id], data);
           }
```

Not 100% sure yet if this is good or whether my problem originated elsewhere.

etobella:
One question, grouping is not working, isn't it? I mean, for maintenance request, we are unable to check it as it is automatically grouped :sob: Anyway, I think it is a great idea and solution :+1:

matnoublanche:
Hi team,

I see that this project is on WIP, I have few questions about it :

Actually it's on 13.0 branch, do you plan to switch it on 15.0 ?
Is it compatible with Enterprise edition ?

Thank you for your answers.

Mathieu

pedrobaeza:
No plan for now for v15 and not checked with enterprise interface.

andreampiovesana (approved):
LGTM

andreampiovesana:
merge?

pedrobaeza:
This PoC has been stopped for now. You can use it as is, that is pretty completed, but we are not going to continue it eventually.

thomaspaulb:
> This PoC has been stopped for now. You can use it as is, that is pretty completed, but we are not going to continue it eventually.

Reasons?

pedrobaeza:
No more budget.

Issue #1435, opened 2019-11-12
[web_responsive] Display the menu when logging in instead of the communications module.


mileo:
Display the menu when logging in instead of the communications module.

pedrobaeza:
I don't get what do you mean

mileo:
When a user first accesses odoo, they may not yet have been introduced to the system interface.

So it's much more intuitive to display the application menu than the discuss module.

![image](https://user-images.githubusercontent.com/1975978/68692906-0f94f200-0555-11ea-8df9-d7d94a0fca68.png)

with web_responsive we don't have a home screen like enterprise:

![image](https://user-images.githubusercontent.com/1975978/68692856-f724d780-0554-11ea-83b2-da9039c435bb.png)

pedrobaeza:
@Yajo @Tardo what do you think?

Tardo:
Since you can see unread messages from systray icon i see this suggestion a good improvement.

But is not a trivial change.

yajo:
`web_enterprise` has a specific screen for the apps selection, but `web_responsive` just extends the dropdown to make it fullscreen and behave similarly.

I consider `web_enterprise`'s approach better. One obvious advantage is that when you are using the website and hit the apps menu, you get to the same familiar screen, instead of having 2 different UIs as we have now. Another obvious advantage is what this issue is explaining.

However the change is going to be huge for `web_responsive`. Given Odoo v14 will have a completely refactored JS/Qweb system, I think we can defer this refactoring for that version.

amachu:
What to do in web_responsive module v 12, if we have to Land users to the icons list instead of first menu in the order?

pedrobaeza:
@amachu as said, it's difficult to be developed due to the architecture, so there's nothing you can do for that unless you develop it.

Tardo:
This can be closed?

pedrobaeza:
Well, the wishlist is still there. I'm not totally sure I prefer to start in the apps menu though...

yajo:
IMHO the request is quite valid. Indeed assuming the user wants discuss after login is worse than asking him what app to open, at lest as a default behavior.



> One obvious advantage is that when you are using the website and hit the apps menu, you get to the same familiar screen, instead of having 2 different UIs as we have now.

Just to make this point clear, this is the 2nd UI I meant:

![2020-04-20_08-31](https://user-images.githubusercontent.com/973709/79727345-2e599a80-82e4-11ea-9088-8e157ce18f36.png)

[...]

PR #1867, opened 2021-03-21
[14.0][IMP] Show AppsMenu by default

SplashS:
I made some hacks to display apps menu when user logged in system instead of Communications module.
Also I made Enterprise-style AppsMenu behavior on Website.
This is implementation of #1435

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -142,6 +142,19 @@ $chatter_zone_width: 35%;
     .dropdown-menu.show {
         max-height: calc(100vh - #{$o-navbar-height});
     }
+
+    &.o_app_drawer {
+        &.o_first_app {
+            .o_menu_apps .dropdown > a,
+            .o-menu-toggle {
+                display: none;
+            }
+        }
+        .o_menu_brand,
+        .o_menu_sections {
+            display: none;
+        }
+    }
 }
 // Iconized full screen apps menu
 .o_menu_apps {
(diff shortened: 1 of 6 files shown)


mileo:
LGTM tested on runbot, could you make this optional like: 

https://github.com/SplashS/web/blob/14.0-default-appdrawer/web_responsive/models/res_users.py#L10

mileo (approved):

SplashS:
I thought about this customization.
But there is one. You can choose default action for each user. In this case user will get this action on start instead of appsmenu.
This customization has sense only in case if we plan to leave current behavior by default.
I wanted to make new behavior by default.
What do you think?

GSLabIt:
@SplashS any chance to back port to 12.0?

SplashS:
> @SplashS any chance to back port to 12.0?

I can say about 13.0 that it is possible without serious changes in this PR.
I haven't seen 12.0 so deep.

pedrobaeza:
I don't think we should perform such changes in stable versions, as they change seriously the behavior, and the risk of bugs is high.

SplashS:
> I don't think we should perform such changes in stable versions, as they change seriously the behavior, and the risk of bugs is high.

Did you mean 12.0 or 14.0 too?

pedrobaeza:
12 and 13. 14 module has just arrived, so it's still a good moment to introduce these changes.

SplashS:
> 14 module has just arrived, so it's still a good moment to introduce these changes.

What do you think about default behavior? There is option to override default screen for each user already. Should we make another option to on/off behavior from this PR?

pedrobaeza:
I think it can be OK to leave as default behavior while not home action is defined.

SplashS:
@Tardo maybe you can review this?

andreampiovesana:
merge?

SplashS:
> merge?

I think @pedrobaeza is right. There is too late to provide this feature in stable version.

It was included in 15.0.

I consider this PR is outdated.

andreampiovesana:
how it works in version 15? and in version 16?

pedrobaeza:
It was partially reverted. It has the side effect of not allowing to put a home action.

PR #1878, opened 2021-03-24
[14.0][MIG] web_drop_target: Migration to 14.0

victoralmau:
Migration to 14.0

@Tecnativa TT28929


fuentes73:
Hello and thanks for your work!
But there is already a migration #1794

Issue #1870, opened 2021-03-21
web_responsive 14.0 issue on safari 


robinhli:
odoo 14.0, web_responsive 14.0 installed.
browse on android chrome, no issue
browse on apple safari, can't display. pic as follows:

Error:

Traceback:
Error: The string did not match the expected pattern.

![image](https://user-images.githubusercontent.com/34515177/111903365-ece5e800-8a7c-11eb-8c05-6f786f4aa654.png)

SplashS:
@robinhli Thank you for your feedback.
I've tested last 14.0 runbot on iOS Safari 14.4.1. Sorry, but I can't reproduce this problem.
Please give more details. The best case is to reproduce it on runbot. (Currently available here https://runbot.odoo-community.org/runbot/repo/github-com-oca-web-162).
It's interesting what you did to get this error, in which Odoo App it happened, what version of iOS do you have.
If you can't provide all of this information, please share the available information.

PR #1883, opened 2021-03-26
[14.0][FIX] web_responsive error on grouped mobile kanban with zero groups

SplashS:
When trying to reproduce #1870 I've found another problem.
When mobile kanban view rendered in grouped mode (f.x. CRM Pipeline) and there is no groups, JavaScript error appears.
This PR fixes it.

### DIFF
--- a/web_responsive/static/src/js/kanban_renderer_mobile.js
+++ b/web_responsive/static/src/js/kanban_renderer_mobile.js
@@ -26,6 +26,24 @@
             this._super.apply(this, arguments);
             this.isMobile = true;
         },
+        /**
+         * KanbanRenderer will decide can we close quick create or not
+         * @private
+         * @override
+         */
+        _cancel: function () {
+            this.trigger_up("close_quick_create");
+        },
+        /**
+         * Clear input when showed
+         * @override
+         */
+        toggleFold: function () {
+            this._super.apply(this, arguments);
+            if (!this.folded) {
+                this.$input.val("");
+            }
+        },
     });
 
     KanbanView.include({
@@ -345,6 +363,9 @@
          *   and displayed
          */
         _moveToGroup: function (moveToIndex, animate) {
+            if (this.widgets.length === 0) {
+                return Promise.resolve();
+            }
             var self = this;
             if (moveToIndex >= 0 && moveToIndex < this.widgets.length) {
                 this.activeColumnIndex = moveToIndex;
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Maybe you can reduce the diff and complexity returning early if no `this.widgets.length`.

web_responsive/static/src/js/kanban_renderer_mobile.js
```diff
@@ -345,6 +345,7 @@
          *   and displayed
          */
         _moveToGroup: function (moveToIndex, animate) {
+            if (this.widgets.length == 0) return;
```
pedrobaeza: Shouldn't you return with the promise resolved?

SplashS: Of course. My bad.

pedrobaeza (approved):

SplashS:
The easiest way to reproduce the problem:
Create a new project in the Project app. After that, you should set up the stages. There will be no columns by default.

Tardo:
It seems that it is not working properly. The add button should appear even without columns defined. Also the background does not seem to be set correctly.

porject screen

SplashS:
Indeed there are some other problems.

> It seems that it is not working properly. The add button should appear even without columns defined.

I prevented quick create hiding when there are no columns. I think it is better cause user is focused on next task.

> Also the background does not seem to be set correctly.

I made some tweaks for no content message and hided sample data for mobile.

web_responsive/static/src/js/kanban_renderer_mobile.js
```diff
@@ -345,6 +363,7 @@
          *   and displayed
          */
         _moveToGroup: function (moveToIndex, animate) {
+            if (this.widgets.length == 0) return Promise.resolve();
```
Tardo: Please use "strict equality" operator. 
```suggestion
            if (this.widgets.length === 0)  {
                return Promise.resolve();
            }
```

Tardo (changes requested):
:+1: Now works fine, thanks!

web_responsive/static/src/js/kanban_renderer_mobile.js
SplashS: Done

Tardo (approved):
Thanks!

PR #1886, opened 2021-03-31
[13.0][IMP] web_widget_numeric_step: Add auto_select option

Tardo:
cc @tecnativa TT28955

numeric_step

### DIFF
--- a/web_widget_numeric_step/README.rst
+++ b/web_widget_numeric_step/README.rst
@@ -55,6 +55,14 @@ Example for an 0.25 step, min to -1 and max to 10 :
 
   <field name="sale_delay" widget="numeric_step" options="{'step': 0.25, 'min': -1, 'max': 10}" /> days
 
+**Available Options**
+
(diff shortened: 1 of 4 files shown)


legalsylvain:
@quentinDupont : this PR could interest you. (if yes, should be backported in V12.)

pedrobaeza:
Not sure if this should be an option or always act as this one. Do you see any drawback on not doing? How other widgets work with this auto-selection?

chienandalu (approved):
Nice

quentinDupont:
I'm not seeing any drawback indeed. In my opinion, it should be even for every float field !
Bravo and thanks !

Tardo:
Odoo's default behavior is to select content in tree mode but not in form mode. Currently this widget only "auto select" if the corresponding option is used.

Maybe leaving the option, making the widget "auto select" in tree mode and making a change in "web_responsive" to "auto select" the inputs, could be a good solution to keep the "Odoo behavior".

sergio-teruel (approved):

PR #1889, opened 2021-04-05
[14.0][ADD] Added a web Hide menu.

AmmarOfficewalaSerpentCS:


pedrobaeza:
This seems quite similar to https://github.com/OCA/server-ux/tree/13.0/base_menu_visibility_restriction

AmmarOfficewalaSerpentCS:
@pedrobaeza ,
Yes, But they work with groups, In this module, we hide the menu based on user configuration.

pedrobaeza:
IMO, the granularity shouldn't be put by user. Anyway, this module lacks a proper explanation in the README, and this has nothing to do with the web client, as it's done on server logic.

AmmarOfficewalaSerpentCS:
With a new readme, this module need to switch on a server-Ux repo?

pedrobaeza:
Or even add the user granularity feature to the existing one. @max3903 do you really think is a good idea to add such granularity?

pedrobaeza:
But yes, the module should be moved out of this repo.

PR #1900, opened 2021-04-20
[8.0][FIX] web_export_view: Boolean fields aren't export

igallart:
In excel file, the boolean fields appear without value.

Runbot
http://3478259-8-0-6ea583.runbot3.odoo-community.org/web?debug=#page=0&limit=80&view_type=list&model=product.product&action=143


igallart:
The travis error is by database, the commit is correct.

PR #1903, opened 2021-04-23
[IMP] web_company_color: Support button and link customizing

hbrunn:


pedrobaeza:
This is not `[ADD]`, but `[IMP]`.

StefanRijnhart (approved):

StefanRijnhart:
@pedrobaeza I only see additions ;-)

pedrobaeza:
@StefanRijnhart @hbrunn `[ADD]` is for adding modules, not for adding things inside a module. If you add new features inside the same module, that's an improvement, so `[IMP]`.

hbrunn:
I didn't want to start a discussion about this, but if you have such strong intuitions about this, please propose them to https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message - otherwise we discuss forever about this on different PRs

pedrobaeza:
It's not my intuition. See in first place the initial message of such guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#1introduction

pointing to the Odoo ones, and there you can see such reference:

https://www.odoo.com/documentation/14.0/reference/guidelines.html#commit-message-structure

We can point directly in our guidelines if you prefer, and add `[MIG]` and other OCA specific codes.

hbrunn:
I think we should point to this, because 'loosely based' tells us where the inspiration comes from, not that all or any of it applies.

Thanks for this link, I must admit it's a while ago I read the Odoo ones, and never came across this. I like uniformity even though I'd do this differently, so here we go: https://github.com/OCA/odoo-community.org/pull/61

etobella (approved):
LGTM :+1: Functional review

StefanRijnhart:
Wut. OCA scripts depend on an unmaintained library from 2014? https://pypi.org/project/anybox.testing.openerp/#description

```
ERROR: Could not find a version that satisfies the requirement anybox.testing.openerp (from versions: none)
ERROR: No matching distribution found for anybox.testing.openerp
The command "pip install anybox.testing.openerp" failed and exited with 1 during .
```

pedrobaeza:
The problem now is that Pypi is having problems (the same as certain parts in GitHub).

https://www.githubstatus.com/
https://status.python.org/

PR #1909, opened 2021-04-29
[MIG] web_advanced_search: Migration to 14.0

rven:


rven:
Does anybody know what this is used for?
>     /**
>      * A search field for relational fields.
>      *
>      * It implements and extends the `FieldManagerMixin`, and acts as if it
>      * were a reduced dummy controller. Some actions "mock" the underlying
>      * model, since sometimes we use a char widget to fill related fields
>      * (which is not supported by that widget), and fields need an underlying
>      * model implementation, which can only hold fake data, given a search view
>      * has no data on it by definition.
>      */
>     const Relational = Char.extend(FieldManagerMixin, {

I removed it from the module and it is working fine without it.
Maybe a leftover from an older version?

Isn't this covered by the **DomainSelectorDialog**?

rven:
I notice that the code editor is missing
image
It should be like this:
image
The last screenshot is from a local environment with only the web_advanced_search module installed.
Possibly another module in the web repo is the causing this issue?

yajo (approved):
>  Possibly another module in the web repo is the causing this issue?

Nope, you have to enable developer mode, as always.

ioans73 (approved):
Functional review

asm8073:
Hi guys, AFAIR there is only review which remains to release the web_advanced_search for Odoo 14. If it is something quick to do, could you please finalize it? We use the module in our company and I cannot upgrade just because it is not ready yet. 
If I could help with the review, please let me know.
Thanks

rven:
@asm8073 it is already merged into v14.0!

asm8073:
> 
> 
> @asm8073 it is already merged into v14.0!

@rven , thank you! I did not check it recently, but now it is available. Thanks a lot for your efforts!

Issue #1910, opened 2021-04-29
[13.0] web_responsive: Error show image, pdf, video on mobile


daonamutc1:
**Steps:**
Login odoo on google chrome by mobile app.
On discuss open image/pdf/video
**Current behavior:**
User can't view image/pdf/video
**Expected behavior:**
User can view image/pdf/video

pedrobaeza:
Closing as there's a PR.

PR #1911, opened 2021-04-29
[FIX] The document preview can't be used properly because .o_modal_fullscreen.o_document_viewer has z-index: 10 small than z-index: 1051 of .o_thread_window

daonamutc1:
Fix issuse #1910

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -697,7 +697,7 @@ html .o_web_client .o_action_manager .o_action {
 .o_web_client.o_chatter_position_sided {
     .o_modal_fullscreen.o_document_viewer {
         // On-top of navbar
-        z-index: 10;
+        z-index: $o-mail-thread-window-zindex + 1;
 
         &.o_responsive_document_viewer {
             /* Show sided viewer on large screens */


pedrobaeza:
Why is that needed? I'm able to see images and video without it? Do you have any other module on top of this one?

daonamutc1:
Please, show my video and image.
You will see .o_modal_fullscreen.o_document_viewer has z-index: 10 small than z-index: 1051 of .o_thread_window
Casue that o_modal_fullscreen will hide.

https://user-images.githubusercontent.com/55737816/116641057-2839d780-a996-11eb-81d0-31caa22e4a9a.mp4

IMG_20210430_091915_hdr

pedrobaeza:
In your video, it seems you have another theme on top of `web_responsive`. Unless you prove that in standard `web_responsive`, this is needed, this is not needed here, but on your customization. Check the same on runbot, and provide detailed steps to reproduce the problem (not only a video of the last step).

daonamutc1:
> In your video, it seems you have another theme on top of `web_responsive`. Unless you prove that in standard `web_responsive`, this is needed, this is not needed here, but on your customization. Check the same on runbot, and provide detailed steps to reproduce the problem (not only a video of the last step).

Please, check again.

https://user-images.githubusercontent.com/55737816/116952721-87109100-acb5-11eb-9c6a-b09eb1efd0ac.mp4

Screen Shot 2021-05-04 at 08 43 50

pedrobaeza (approved):
OK, so it seems when you attach an image on a chat, it gets a higher z-index. Then it's OK for me to change it.

web_responsive/static/src/css/web_responsive.scss
```diff
@@ -697,7 +697,7 @@ html .o_web_client .o_action_manager .o_action {
 .o_web_client.o_chatter_position_sided {
     .o_modal_fullscreen.o_document_viewer {
         // On-top of navbar
-        z-index: 10;
+        z-index: 1052;
```
Tardo:
```suggestion
        z-index: $o-mail-thread-window-zindex + 1;
```

Tardo (changes requested):
Thanks for the fix! Please, use sass variables instead of hardcoded values when possible.

pedrobaeza:
Please squash both commits into one and change commit message to the proper one: it's not due to mobile, but previewing files from a chat window.

Tardo:
Please, normalize commit message to something like: 
```
[FIX] web_responsive: Change z-index of document preview

Previous this commit, the document preview can't be used properly because ...
```

Thanks!

pedrobaeza:
@daonamutc1 thanks for the change, but I'm afraid you have put only the second line. Can you please put the whole commit message? You need also to throw a comment for the rest to be aware that you have made the change.

Issue #1919, opened 2021-05-05
[RFC] web_responsive: Change behavior of html widget


etobella:
When using the html widget, it is hard to manage it as a second scroll bar is created if the data inside the widget is big enough, as you can see:
![old_behaviour](https://user-images.githubusercontent.com/28590170/117119541-84d03480-ad92-11eb-8b34-644569a9c9f3.gif)

We could make the widget overflow like this:

![new_behavior](https://user-images.githubusercontent.com/28590170/117119526-7da92680-ad92-11eb-941f-e1c98b983a48.gif)

the code necessary to do this is small, and we can make it work with options if needed.

One option was to create a specific module, but as we are talking of only 13 lines of JS I think we could insert it inside `web_responsive`

@Yajo @pedrobaeza @OCA/web-maintainers

etobella:
The necessary code to make this happen with options was:
```javascript
odoo.define("my_demo_module.fields", function(require) {
    "use strict";
    var backend = require("web_editor.backend");
    backend.FieldTextHtmlSimple.include({
        _getSummernoteConfig: function() {
            var result = this._super.apply(this, arguments);
            if (this.nodeOptions.automatic_summernote_height) {
                result.height = undefined;
            }
            return result;
        },
    });
});
```

pedrobaeza:
@Tardo @SplashS what do you think about this?

Tardo:
I think this will make other views more cumbersome to manage. Other modules can place fields after another field that uses this widget.

Would it be possible to disable the scroll action of elements inside the "main" container that are not active (no focus)?

etobella:
> I think this will make other views more cumbersome to manage. Other modules can place fields after another field that uses this widget.

Well, it will change the size according to the size of each field if it is bigger than the minimum px, The size is independent for each one. Right now it is harder to read all in this case, as you need to click on each field and start scrolling.

> Would it be possible to disable the scroll action of elements inside the "main" container that are not active (no focus)?

I am not sure, as the widget is generated on the start, not when clicking.

However, changing the behavior can be somehow crazy. If you have two HTML fields big enough, you will need to click on the first to read it all (the scroll is working only when clicking, but the new behavior will be making it big). When you start reading the second field, everything will be resized, I don't know if this is the expected behavior :disappointed:

SplashS:
I think we can make suggested by @etobella behavior only for mobile. In this case fields will be placed in one column and there will not be reordering of fields on size change.
[...]

PR #1921, opened 2021-05-05
[IMP] web_responsive: Automatic height for HTML fields

etobella:
Following comments from #1919 

@SplashS @Tardo @pedrobaeza

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -13,7 +13,7 @@
     "license": "LGPL-3",
     "installable": True,
     "depends": [
-        'web',
+        'web_editor',
         'mail',
     ],
     "data": [
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -16,7 +16,7 @@ odoo.define('web_responsive', function (require) {
     var Chatter = require('mail.Chatter');
     var DocumentViewer = require('mail.DocumentViewer');
     var ListRenderer = require('web.ListRenderer');
-
+    var web_editor_backend = require("web_editor.backend");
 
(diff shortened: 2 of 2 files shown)


web_responsive/__manifest__.py
```diff
@@ -13,7 +13,7 @@
     "license": "LGPL-3",
     "installable": True,
     "depends": [
-        'web',
+        'web_editor',
```
etobella, line 16: web_editor is auto_install, so it should be installed, I don't know if this is a problem for `web_responsive`

pedrobaeza: Well, changing dependencies is not the ideal indeed...

yajo (changes requested):
I personally don't like this feature here.

This is how vanilla Odoo 12 looks like:


https://user-images.githubusercontent.com/973709/117264589-477bad80-ae4b-11eb-9589-240df552b820.mp4

Mouse-wheel scrolling feels natural to me. Also, html widgets have a handle at the bottom that allow resizing if the user feels the need for more space.

Adding this option might make other modules depend on `web_responsive` to be able to make use of `automatic_summernote_height`, which is not something we desire, because then that'd make the modules not work in enterprise versions for example.

I think this fits better in a more specific module such as `web_field_html_customizable`, which can be used with `web`, `web_responsive` or `web_enterprise`, and which covers a specific need and can be extended to support further summernote customizations.

Sorry because I'm late to the party, but my vote is 👎🏼 here 🤷🏼‍♂️

etobella:
Sure!!! I knew that might be an issue, so don't worry, I will close this PR and make a new one with the new module.

Thanks for all the comments

yajo:
Thanks for your contribution and your understanding 😊

PR #1922, opened 2021-05-06
[ADD] web_field_html_customizable

etobella:
Following the comments from #1921 I made new module


LoisRForgeFlow (approved):
LGTM

albariera (approved):

etobella:
@OCA/web-maintainers This is approved. Might be able to merge this before github closes it?

pedrobaeza:
The name should be `web_widget_html_auto_height`. I know you want this to be a potential container for other options, but this has not happened yet, and at the end, it's better to isolate things unless they are developed at once or makes sense to be together.

etobella:
We are not currently using this module, and it was just an idea (I implement the necessary changes on my module). If it really has no interest, it has no sense to keep the PR as open (also, a lot of changes happens with the new html field, si I don't know if this has sense...)

PR #1930, opened 2021-05-20
[REM][requirement.txt] remove the requirement of matploit becase crea…

matteoopenf:
…te conflict when the repo is added on odoo.sh

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,4 @@
 bokeh==1.1.0
 # web_widget_plotly_chart
 plotly==4.1.0
-# web_widget_mpld3_chart
-matplotlib==3.0.3; python_version < '3.7'
-matplotlib==3.4.1; python_version >= '3.7'
 mpld3==0.3


pedrobaeza:
But this library is needed for one of the modules in this repository. You can't remove it. Any problem you have, should be solved other way.

matteoopenf:
> But this library is needed for one of the modules in this repository. You can't remove it. Any problem you have, should be solved other way.

which way,  the requirement as  is broken odoo.sh, said me how can I change and I will do?

matteoopenf:
> But this library is needed for one of the modules in this repository. You can't remove it. Any problem you have, should be solved other way.

I also propose The move the specifc version requirement on manifest of module, because if you keep there the requirement odoo.sh try to install both requirement and the build failed

pedrobaeza:
There is no such "move to the manifest of the module". You need to declare the requirements.txt the same. If odoo.sh doesn't interpret correctly both lines, that discriminates by Python version, then it's a problem of Odoo.sh, and you should report it there. Closing this as invalid.

PR #1932, opened 2021-05-25
[14.0][MIG] web_refresher: Migration to 14.0

joao-p-marques:
Standard migration. It can be simplified now due to the new definition of the buttons in the view: https://github.com/odoo/odoo/blob/14.0/addons/web/static/src/xml/base.xml#L513

@Tecnativa
TT29862

@Tardo @pedrobaeza can you review please?


chienandalu (approved):
:+1: tested

pedrobaeza (approved):

PR #1936, opened 2021-05-30
[12.0][ADD] New module web_edit_xmlid

ypapouin:
This module override the default method used to read current record metadata in order to provide two new actions: `Create` or `Edit` XML ID.

icon

**Create XML ID**
image

**Edit XML ID**
image

image


pedrobaeza:
I more suited name would be web_edit_xmlid, as metadata are create date, create uid, write date...

etobella (approved):
Tested and works as expected. Just as comment, once you create the record, it would be great to launch the editor, as it is closed automatically and you cannot review it again.

ypapouin:
Small edit to force `noupdate=True` by default to avoid having our record manually linked with a XML_ID to be deleted when the module that owns this record will be updated.

@etobella , there is now way to review the record before it is created since it is managed by the framework, but the edit button allows you to modify it immediately.

bofiltd (approved):

pedrobaeza:
@bofiltd putting 10 reviews in 2 minutes in the corresponding PRs is not trustful and can't be taken as a serious review, so please abstain for doing that.

PR #1940, opened 2021-06-02
Make sure the attachment is created without a website_id

bartvanbossche-dynapps:


Tardo:
Thanks! this PR solves a issue when uses multi-website. "website" module changes the behaviour when gets attachment using url's.
The commit adds a context used in "website" module, but "web_company_color" doesn't have a dependency with it.

Maybe it is better not to use Odoo "url" mechanism and implement a controller to serve the content of the generated .scss file?

PR #1942, opened 2021-06-03
[14.0][FIX] web_widget_x2many_2d_matrix - show_column_totals and show_row_totals

enriquemartin:
Needs to `parse_boolean()` the `show_column_totals` and `show_row_totals` attributes. 
Also it was aggregating every float, monetary or integer field even if the attribute was false, so OR operator has been changed for AND.

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
@@ -59,17 +59,13 @@ odoo.define("web_widget_x2many_2d_matrix.widget", function (require) {
                     )
                 );
             }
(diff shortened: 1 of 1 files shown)


petrus-v:
is it the same that as been fixed on version 12 in a different way ? https://github.com/OCA/web/pull/1622/files

enriquemartin:
not exactly @petrus-v we need to change the operator, otherwise the totals are always displayed

xavier-bouquiaux:
tested and works great thanks @enriquemartin

sbejaoui (approved):

sbejaoui (approved):
Code review. 
Thanks

OpenCode (approved):
It works

dariodelzozzo (approved):

PR #1950, opened 2021-06-09
 [IMP] web_tree_dynamic_colored_field : Add bg_color_field option

legalsylvain:
simple improvement to add the possibility to define a field in ``bg_color_field`` that will be used to dynamically set a color to a line.

CC : original developpers : @guewen, @damdam-s, @hbrunn

Sample of implementation below and review welcome here : https://github.com/OCA/margin-analysis/pull/96

with_module_web_tree_dynamic_colored_field


hbrunn (approved):
we have similar PRs for [v10](https://github.com/OCA/web/pull/1377) and [v11](https://github.com/OCA/web/pull/1454) that never made it to merge, so there's definitely interest in this functionality.

PR #1952, opened 2021-06-10
[14.0][ADD] web_widget_char_size

SimoRubi:
Covering the web client side of deprecation of `size=` attribute of `Char` field.

For more details, see https://odoo-community.org/groups/contributors-15/contributors-196199

### DIFF
--- a/web_widget_char_size/readme/DESCRIPTION.rst
+++ b/web_widget_char_size/readme/DESCRIPTION.rst
@@ -1,0 +1,11 @@
+This module allows to add an option `size` to Char fields in web client:
+
+.. code-block:: XML
+
+    <field name="sized_5_field" options="{'size': 5}"/>
+
+that allows to enter only the specified number of characters in the field.
+
+Note that adding the `size=` attribute in Char field definition raises the following warning:
+
+ unknown parameter 'size', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it
--- a/setup/web_widget_char_size/odoo/addons/web_widget_char_size
+++ b/setup/web_widget_char_size/odoo/addons/web_widget_char_size
@@ -1,0 +1,1 @@
+../../../../web_widget_char_size
--- a/setup/web_widget_char_size/setup.py
+++ b/setup/web_widget_char_size/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/web_widget_char_size/__init__.py
+++ b/web_widget_char_size/__init__.py
@@ -1,0 +1,1 @@
+#  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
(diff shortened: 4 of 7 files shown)


alan196 (approved):
LGTM 👍

SimoRubi:
@OCA/web-maintainers can this be merged?
Thanks!

SimoRubi:
@OCA/web-maintainers can this be merged?
Thanks!

web_widget_char_size/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,11 @@
...
+
+.. code-block:: XML
+
+    <field name="sized_5_field" options="{'size': 5}"/>
+
+that allows to enter only the specified number of characters in the field.
+
+Note that adding the `size=` attribute in Char field definition raises the following warning:
+
+ unknown parameter 'size', if this is an actual parameter you may want to override the method _valid_field_parameter on the relevant model in order to allow it
```
legalsylvain, line 11: Could be possible to add it to base model ? so all models could have the option available without override.

SimoRubi: Thanks for having a look!
I'm sorry I don't get what you mean: this is a modification to JS class `FieldChar` so there are no models involved but it simply applies to every `Char` field displayed in the backend

legalsylvain: Hi thanks for your answer. I just discovered the ``_valid_field_parameter`` function, so maybe it's not relevant, but I was thinking to overload base model, to avoid to override the method on each model. Something like : 

```
class Base(models.AbstractModel):
    _inherit = 'base'

    def _valid_field_parameter(self, field, name):
        return (
            name == "size" and field.type == "char"
            or super()._valid_field_parameter(field, name)
        )
```

ref another exemple : 

https://github.com/odoo/odoo/blob/14.0/addons/mail/models/ir_model.py#L8

Don't you think it will allow to remove the current warning in the description ?

SimoRubi: Thanks now I get what you meant and that would probably remove the warning too (that's actually what the warning itself suggests).

I still prefer the JS modification because I believe `size` has been declared as an invalid parameter for a reason, I think that it might have something to do with efficiency of VARCHAR column in postgresql.
So I don't think that implementing something that says 'just consider that as valid' is a good idea, that's why I just acted on the UI: in this way we have the same old UX with no impact on backend code's logic.

legalsylvain: Ok !

tafaRU (approved):
👍 LGTM Code and functional review (locally).
Thanks a lot!

andreampiovesana (approved):
merge?

tafaRU:
@legalsylvain are you ok with merging this? Could you please proceed with the merge?

legalsylvain (commented):
Thanks. 
/Ocabot merge patch

SimoRubi:
> Thanks.
> /Ocabot merge patch

Thanks @legalsylvain but @OCA-git-bot is not merging, probably because ocabot should be lower case, can you try again?

legalsylvain:
indeed.

/ocabot merge patch

PR #1688, opened 2020-09-29
[ADD][12.0] web_favorite_company module.

legalsylvain:
This extend the ``web_responsive`` module to have a better display of the companies menu.

It is usefull only if you have an Odoo Instance, with a lot of companies.

* Go to your preferences (Right Menu > Preferences)

* In the new field 'Favorite Companies', select the companies you work on a lot.

res_user_form

* click on the current company name

* The list of the companies are now displayed with the logo, and there are ordered by name, displaying first the favorites companies.

company_menu

Note : the css could be improved, but I'm not very confortable with that language. If you have ideas to improve, please make directly a PR against mine. Thanks ! 

CC : @SandieFavre, @quentinDupont #GRAPOCA

PR #1955, opened 2021-06-18
 [ADD] new module web_responsive_company

legalsylvain:
### Description
This module extends the ``web_responsive`` module to have a better display of the companies menu.

It is usefull only if you have an Odoo Instance, with a lot of companies.

### Install

This module depends on the other OCA module ``res_company_code`` present
in the multi-company OCA repository to have the possibility to make search
by company code.

Ref : https://github.com/OCA/multi-company/tree/12.0/res_company_code

However, it can be used without having codes defined on companies.

### Usage

* click on the current company name

* The list of the companies are now displayed with the logo and the code (if defined), and there are ordered by complete name. (code and name)

company_menu

* It is also possible to search by complete name, in the same way as for the menus.

company_menu_search

**Note** : A new access key is available on 'Y' to switch company.

shortcut

CC : @quentinDupont (ref : apps/deck/#/board/144/card/1675)

CC : web_responsive contributors : @sergio-teruel, @Tardo, @lasley

Supersed : https://github.com/OCA/web/pull/1688


web_responsive_company/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) 2020 - Today: GRAP (http://www.grap.coop)
```
sergio-teruel: Change to 2021 :smile:

web_responsive_company/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "version": "12.0.1.0.0",
+    "category": "Web",
+    "website": "https://github.com/OCA/web",
+    "author": "GRAP,Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "installable": True,
+    "depends": ["web_responsive", "res_company_code"],
+    "data": [
+        "views/assets.xml",
+
```
sergio-teruel: Remove extra blank line

sergio-teruel (commented):
Great module!!!
Little comment...
Waiting runbot to functional test

web_responsive_company/__manifest__.py
legalsylvain: it's an update of previous module I developped ``web_favorite_company`` in 2021.
https://github.com/OCA/web/pull/1688
(but after few monthes of use in my company, it requires extra settings, so I prefer a more simple "search functionnality")

but no problem, I'll change it.

sergio-teruel: Ok.. ok perfect!!

web_responsive_company/__manifest__.py
legalsylvain: done.

web_responsive_company/__manifest__.py
legalsylvain: done.

legalsylvain: done.

legalsylvain:
@sergio-teruel. Thanks for the review. I did some changes. tests are available on runbot. (weird thing : I had to install manually the module on runbot).

pedrobaeza:
Thanks for the PR, Sylvain. On a quick test from the mobile, seems to work ok. On the technical part, the dependency to res_company_code seems a bit forced for your customer requirement. The main problem I see with this module in the long term is that on 13.0+, you won't be able to select more than one company with this approach.

legalsylvain:
> Thanks for the PR, Sylvain. On a quick test from the mobile, seems to work ok

thanks for your review ! 

> On the technical part, the dependency to ``res_company_code`` seems a bit forced for your customer requirement

Yes & no. I understand your Point of view in a technical part. But in a functional part, this module is usefull only if you have a lot of company. Otherwise, the default dropdown list is enough. And if you have a lot of companies (in GRAP, there are ~ 60 companies), having short code is (I guess) a common thing.

If this feature is great and welcomed in the community, I think that in most recent version, we could : 
- introduce this feature in the core ``web_responsive`` module. (without dependency with ``res_company_code``)
- add an extra glue module ``web_responsive_res_company_code`` that only adds code on kanban view.

I can add such sentence in the ROADMAP.rst file if you want.

> The main problem I see with this module in the long term is that on 13.0+, you won't be able to select more than one company with this approach.

Yes. I currently use 12.0. I guess than when porting this module in V13+, UI should be reworked (introducing checkboxes I guess) but the main design (kanban view for companies + search field) could be kept.

pedrobaeza:
I think you can implement other thing for avoiding the dependency and having the same feature:

- Change name_search on `res_company_code` for searching also on the code. Similar to what is done here: https://github.com/OCA/l10n-spain/blob/6d545f3f3b7f7e59d3e67413c9edcd5855c62845/l10n_es_partner/models/res_partner.py#L39
- On this module, search companies with `name_search`, not with `search`.

legalsylvain:
> I think you can implement other thing for avoiding the dependency and having the same feature:
>   Change name_search on res_company_code for searching also on the code. Similar to what is done here: https://github.com/OCA/l10n-spain/blob/6d545f3f3b7f7e59d3e67413c9edcd5855c62845/l10n_es_partner/models/res_partner.py#L39
>     On this module, search companies with name_search, not with search.

makes senses ! I'll try the next week. Thanks !

quentinDupont (approved):
Tested on runbot, works like expected :ok_hand: :smiley:

legalsylvain:
Hi @pedrobaeza ! 

- I refactored the module ``web_responsive_company``. I removed the dependency to the ``res_company_code`` module, as you proposed, and replaced it by a lazy dependency in ``ir_http.py``. Module works well with and without ``res_company_code``.
I didn't implemented the change in ``name_search`` of the company, because it is changing the behaviour of deployed module and could not be desired in a stable version. It could be introduced, when porting ``res_company_code`` to V13+. I think it is acceptable, don't you think ?

- regarding the shortcut, for the time being I proposed "Y". (the second last letter of the alphabet, if we follow my  (weird?) logic. ;-)
(https://github.com/OCA/web/pull/1956#issuecomment-864827597). Do you have other idea ?
- "C" for "**C**ompany" or "**C**hange Company" conflicts with "Create"
- "S" for "**S**witch Company" conflicts with "Save"

If we want to have all the shortcut in the same hand I can change to "W". Any point of view ?

pedrobaeza (approved):
Tested on runbot.

Indeed there's no good accesskey, so Y seems good. We can set the mnemonics as compan**Y**.

For v13, seems good to change `res_company_code` for the `name_search`.

legalsylvain:
~~Just tested in production, and i have a bad css effect when company qty > 30. (that's weird, I used the same css as web_responsive, for the menu, but maybe there are never 30 main application installed). I have to investigate. please wait before merging.
kind regards.~~

UPD : all is good. it was just conflicting with my previous module ``web_favorite_company``. (#1688).

PR #1956, opened 2021-06-18
 [IMP+FIX] web_responsive : fix shortcut display and add new shortcuts (u) for the User menu

legalsylvain:
* Update shortcut form to display correct shortcut for "Edit a Record" (A >> E)
* Update shortcut form to display correct shortcut for "Discard a record modification" (J >> D)
* Add new entry in the form to display correct shortcut for "Select menu" (A)

image

* add a new shortcut ~"z"~ "u" for the user Menu. (that is used a lot for login out)

CC : @quentinDupont (ref : apps/deck/#/board/144/card/1678)

CC : ``web_responsive`` contributors : @sergio-teruel, @Tardo, @lasley

### DIFF
--- a/web_responsive/readme/USAGE.rst
+++ b/web_responsive/readme/USAGE.rst
@@ -1,6 +1,9 @@
 The following keyboard shortcuts are implemented:
 
-* Toggle app drawer - ``Alt + Shift + H``
+* Toggle app drawer - ``Alt + Shift + A``
 * Navigate app search results - Arrow keys
 * Choose app result - ``Enter``
 * ``Esc`` to close app drawer
+
+
+Also you can access to the user menu, with ``Alt + Shift + U``
--- a/web_responsive/static/src/xml/menu.xml
+++ b/web_responsive/static/src/xml/menu.xml
@@ -1,9 +1,57 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
 
+<t t-extend="UserMenu">
+    <t t-jquery=".dropdown-toggle" t-operation="attributes">
+        <attribute name="accesskey">u</attribute>
+    </t>
+</t>
+
 <div t-extend="UserMenu.shortcuts">
+    <!-- Replace "a" shortcut by "e" (for 'Edit a Record') -->
+    <t t-jquery="table.o_shortcut_table > tbody > tr > td:nth-child(2) > span:nth-child(2).o_key:contains('a')" t-operation="replace">
+        <span class="o_key">e</span>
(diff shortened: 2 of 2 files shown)


pedrobaeza:
> add a new shortcut "z" for the user Menu. (that is used a lot for login out)

I don't see this one. It's very error prone for something that user doesn't want. Logout is not frequently done.

legalsylvain:
> It's very error prone for something

I don't understand why it could be error prone.
1) Alt  + Shift is never done except for special action.
2) it is the unique menu that can not be accessed by accessed keys. (main menu with A ; sub menu with number ; company menu with Y (with web_responsive_company))
3) once Alt + Shift + Z is done, user have to down many times with arrow to log out . Not possible to log out by mistake. 

>  Logout is not frequently done.

That depends. in shared computed, it can be. and for IT users, with ``auth_admin_passkey`` it is very frequent.

Did you tried on runbot ?

pedrobaeza:
You press Z instead of A, and you get something that you don't want to get.

legalsylvain:
> You press Z instead of A, and you get something that you don't want to get.

So, you can click on Esc to exit the menu. I don't see the problem. That the same for all the shortcuts.

pedrobaeza:
I have tried on runbot and know I understand that the logout is not direct, but I still prefer to use other shortcut like U (for user), and more if we are debating to use them for previous/next records in https://github.com/OCA/web/pull/1832#issuecomment-862227148.

legalsylvain:
Hi ! 
thanks for your test on runbot. the choice of "Z" was based on the following thoughts : 
- if "A" is the shortcut for the first menu, "Z" is a good candidate to be the shortcut for the last menu
- "Z" is free. (not taken by odoo / oca modules)
- that's all !

I'm OK to switch to "U", it makes more sense. I updated the code. Thanks for the proposal ! 

(Regarding other moves : I think we are limited by Odoo limitations. AFAIK, we can not "translate" shortcuts, and it should be the solution. A good choice for a localization (language + keyboard) will be a bad choice for another localization. So "U" is perfect from a european / anglosaxon point of view. (User / Usuario / utilisateur / utilizzatore) But for other localization, it has no sense.
Maybe we could talk with odoo usability responsible regarding that point.)

pedrobaeza:
Well, conventions are always taken in source language, like model names, so this shouldn't be different. The `S` for saving doesn't fit with "Guardar" in Spanish, but it's OK as long as it's homogeneous and the reference is English.

pedrobaeza (approved):

remi-filament (approved):

PR #1961, opened 2021-06-23
[14.0] [MOD] Pass context defined on tree view field

rven:
When for instance a form_view_ref context is defined on a field in a specific tree view, the context should also be passed when opening the form.


legalsylvain (approved):
Makes sense.

StefanRijnhart (approved):
Thanks! I came looking around in this project just for this issue.

PR #1970, opened 2021-07-02
Fix issue with checkbox size in tree view

mohmdlabeeb:


pedrobaeza:
Which error? The big size is on purpose.

mohmdlabeeb:
The UI would be better if it is of normal size

ypapouin:
I agree with @pedrobaeza , this big size is more practical on a daily usage.
But maybe an in-between solution would be nice.

Look at Gmail:
image

The circle is the clickable part and its only appears on hover.

pedrobaeza:
On mobile/tablet, there's no hover. That's why we put it that size.

PR #1971, opened 2021-07-02
[11.0][ADD] web_clickjack_protection

KKamaa:
Clickjacking is a technique by which a malicious party embeds your website in an <iframe>, then hovers buttons over it to make the user think he is clicking on your site when in fact he is communicating with the parent.

Clickjacking can be prevented on the webserver side by adding headers, but there are [ways around this](https://github.com/niutech/x-frame-bypass>). This module prevents clickjacking more thoroughly by making it impossible for your site to be embedded. It does so by adding a small "framebreaker" Javascript which creates a CSS style element on the fly to hide the body of the current page by default. Then, if it doesn't detect a parent frame, it removes it again.


thomaspaulb (approved):

thomaspaulb:
@gfcapalbo Would you mind reviewing this? It implements a [framebreaker](https://en.wikipedia.org/wiki/Framekiller) script.

gfcapalbo:
Could not test a full clickjack scenario, but read it all and it does what described: if we are not top element, we are placed as top element therefore making embedded odoo impossible.   :+1

gfcapalbo (approved):

thomaspaulb:
@KKamaa Can you checkout why this failed and fix it

KKamaa:
> @KKamaa Can you checkout why this failed and fix it

@thomaspaulb this error is coming from travis not sure why, it seem its failing due to travis.yaml configuration but for this MR its the same as 11.0

thomaspaulb:
@KKamaa Have fixed on https://github.com/OCA/web/pull/2750, could you rebase to latest and remove last two commits, it should go green

dreispt:
Superseded by #2764

PR #1972, opened 2021-07-06
[ADD] web_disable_archive_group: New module to 13.0 

dani-xtendoo:
This module makes the option 'Archive' enabled only for the users that belong
to the Archive Data group.


manuelcalerosolis (approved):
Functional review

PR #1976, opened 2021-07-08
[14.0] [MIG] support_branding

KKamaa:
A migration update of support branding module from v10 to v14. Added few improvements.
see below sample of random test.
image

image


support_branding/demo/ir_config_parameter_data.xml
```diff
@@ -0,0 +1,25 @@
...
+            <field name="key">support_company_url</field>
+            <field name="value">https://www.sunflowerweb.nl</field>
+        </record>
+        <record id="demo_config_parameter_company_color" model="ir.config_parameter">
+            <field name="key">support_branding_color</field>
+            <field name="value">#fff</field>
+        </record>
+        <record id="demo_config_parameter_support_email" model="ir.config_parameter">
+            <field name="key">support_email</field>
+            <field name="value">info@sunflowerweb.nl</field>
```
thomaspaulb: Let's put "Odoo Community Association", "https://odoo-community.org" and "oca@example.com" here though

thomaspaulb (approved):
LGTM, code review

support_branding/static/src/js/support_branding.js
```diff
@@ -2,105 +2,169 @@
...
+                var $description = $(".support-desc");
+                var $button = $(".support-btn");
+                var $body = $(".sp-body");
+                var $header = $form.parents(".modal-dialog").find(".modal-header");
+                var $footer = $form.parents(".modal-dialog").find(".modal-footer");
+
+                $statement.prepend(inputs);
+                if (self.support_cp_email) {
+                    if (self.support_cp_name) {
+                        var title = "Support By " + self.support_cp_name;
```
thomaspaulb: `Support by` looks a bit better to me

support_branding/static/src/xml/base.xml
```diff
@@ -1,36 +1,64 @@
...
-        <t t-jquery="div" t-operation="after">
-            <div>
+        <t t-jquery="div.o_error_detail" t-operation="after">
+            <div class="container">
                 <form
                     class="support-branding-submit-form oe_form"
                     action="mailto:"
                     enctype="text/plain"
                     method="get"
                 >
```
thomaspaulb, line 10: In my Chrome I'm getting this message:

image

Could this be suppressed somehow? It also eclipses the 'Email to **company**' button

support_branding/static/src/js/support_branding.js
```diff
@@ -2,105 +2,169 @@
...
+                                        }
+                                    });
+                                }
+                            },
+                            function () {
+                                $body.val(desc + "\n" + $body.val());
+                                $btn.unbind("click");
+                                $btn.click();
+                            }
+                        );
```
thomaspaulb, line 174: Usability here is not optimal:

image

1. We are seeing an "OK" button and an "Email to **company**" button. Clicking "OK" closes the dialog without sending an email, which is not what a user would expect.
2. Clicking the "Email to **company**" button does send the email, but silently, there is no feedback and the dialog stays open. This will confuse users and let them press many times, so multiple mails being sent.

thomaspaulb (changes requested):

support_branding/static/src/xml/base.xml
KKamaa: for this wasn't able to recreate it on my end but added autocomplete=off as way to suppress also check version of chrome you are using or if any plugins might affect.

support_branding/static/src/js/support_branding.js
KKamaa: this has been restructured and notification sent to a user when email is created. Disabled _OK_ button since we already have close on right top of dialog, and also after successfully sending, dialog will close.

support_branding/views/res_config_settings.xml
```diff
@@ -0,0 +1,96 @@
...
+        <field name="model">res.config.settings</field>
+        <field name="priority" eval="10" />
+        <field name="inherit_id" ref="base.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[hasclass('settings')]" position="inside">
+                <div
+                    class="app_settings_block"
+                    data-string="Support Settings"
+                    string="Support Settings"
+                    data-key="support_company"
```
thomaspaulb: Found a small thing:

image

in Enterprise, this menu doesnt have an Icon. 

I think we can use the `static/description/icon.png` but I also saw that it had been changed back to the default whereas it was this cool lifeboat icon - could we get back that icon and also use it here?

support_branding/views/res_config_settings.xml
```diff
@@ -0,0 +1,96 @@
...
+                                        <label for="support_release" />
+                                        <div class="text-muted">
+                                            Add the support release
+                                        </div>
+                                        <div class="content-group">
+                                            <div class="mt16">
+                                                <field name="support_release" />
+                                            </div>
+                                        </div>
+                                        <br />
```
thomaspaulb, line 93: Was thinking to make it easy to test this module we can add a button here that says `Test exception` which then calls something that generates an exception eg just a function with `raise UserError('Test exception')`

thomaspaulb (changes requested):

support_branding/views/res_config_settings.xml
KKamaa: So the issue was `data-key` it needs to be the name of the module so as to find the correct icon. Though based on OCA the icon has to be their default not sure if its okay to change it or leave as below:

image

support_branding/views/res_config_settings.xml
KKamaa: added a js test for this and also a button function to generate division by zero error so that users can test. See below results of the test.

sample

support_branding/static/src/xml/base.xml
thomaspaulb: The autocomplete off didn't work for me, did some research and they suggest to remove the "mailto:" and use a roundabout way to do it by Javascript:

https://stackoverflow.com/questions/61472406/website-is-now-insecure-after-adding-mailto

support_branding/views/res_config_settings.xml
thomaspaulb: Ok lets use the OCA icon for now

support_branding/views/res_config_settings.xml
thomaspaulb: Nice! :-)

support_branding/static/src/xml/base.xml
KKamaa: So the issue was due to this line:
`$form.attr("action", "mailto:" + self.support_cp_email);`

So removed it so that we rely on send the mail via `mail.mail` on odoo. This just launches the the form later on submission by getting the default mail app installed.

thomaspaulb: Ahh great it now works. Didnt realize that we didnt need `mailto:` at all.

thomaspaulb (approved):

gjotten (approved):
LGTM, functional tests went well.

gjotten:
@OCA/web-maintainers this looks ready to be merged.

hbrunn (changes requested):
just looking at the screenshot: What's the functionality of the OK button right of the text box? And I think this text box should be `width=100%`

KKamaa:
> just looking at the screenshot: What's the functionality of the OK button right of the text box? And I think this text box should be `width=100%`

@hbrunn Hey this was updated as shown below

image

hbrunn (approved):
thanks

/ocabot migration support_branding
/ocabot merge nobump

PR #1755, opened 2020-12-03
[14.0][mig] web_ir_actions_act_multi

koenloodts:

PR #1990, opened 2021-07-18
[MIG] web_ir_actions_act_multi: Migration to 14.0

alessandro-fiorino:
Just a version bump nothing to fix


pedrobaeza:
What about #1755?

alessandro-fiorino:
Didn't see it it's a dupe. Thanks.

PR #1993, opened 2021-07-20
[IMP] web_refresh

SplashS:
There are beta version of web_refresh module.

It allows user to refresh current view without page reload.
Also it have two options:
1. Refresh every N secs.
2. Refresh on server changes.
The first one is simple, but the second one is not so.

There are many ui states of refresh icon that indicate current state.
To achieve live refreshing I used bus.bus and some heuristics to determine if current view is changed and have to be reloaded.
At this point there are heuristics for form, list, kanban and pivot. Graph, calendar, activity and discuss - in progress.

I would happy to any feedback. Especially on these points:
1. Naming. There is option to rebase this PR as 14.0.2 version of web_refresher module. Since base behavior is the same.
2. Suggestions on plans in ROADMAP file.
3. Testing of performance and usability.
4. General feedback :)


pedrobaeza:
We currently have `web_refresher` that manually launches the refresh. The first option can be included in that one, but it's very dangerous, as each tab will request a refresh every N seconds, which can exhaust the maximum number of allowed connections. Odoo did that in the past for the chatter, and was disabled after a bad experience. For the second option, I would wait for v15 OWL implementation, that has react type components that will be refreshed on demand.

SplashS:
@pedrobaeza Performance is the central point of this module.
But live update feature is quite useful to avoid it at all. So I tried to achieve balance:
1. "Refresh every N secs" feature is allowed after module install. But user must activate it manually on each view.
2. "Refresh on server changes" feature is disabled by default after module install. Performance warning is showed for this option. User still must activate it manually on each view. 
3. Both options can be enabled / disabled in General Setings serverwide or per user.
4. Default and minimum refresh periods can be configured.

So if both options are disabled by administrator - this module behavior is equal to web_refresher.
Also it is possible to disable both performance critical options by default. In this case module update in case of merging with web_refresher won't affect installations performance.

administration

SplashS:
> I would wait for v15 OWL implementation, that has react type components that will be refreshed on demand.

I don't understand clear what do you mean? Is it something close to new dashboards in master?

pedrobaeza:
> I don't understand clear what do you mean? Is it something close to new dashboards in master?

I mean that I was told that components will already be refreshed when needed, like chatter when a new message arrives.

tfossoul:
Hi @SplashS 

My case:
I use the queue_job module to create records ansynchronously via a wizard. When the user click on the "create multiple record" button, the user is redirected to a tree view. I would like the tree view to be refreshed automatically every x seconds to show the progression of records creation.

Is it possible to configure the refresh action by action, or via a context with your module ?

Thanks

PR #1997, opened 2021-07-30
[13.0][IMP] web_pwa_oca: Apple Icons

Tardo:
Improve support to display icons on IOS

cc @tecnativa TT31173


pedrobaeza:
Travis is red

Tardo:
Travis is happy

web_pwa_oca/templates/assets.xml
```diff
@@ -16,10 +16,113 @@
...
+            </t>
+            <t t-else="">
+                <!-- This icon must be 180x180 -->
+                <link
+                    rel="apple-touch-icon"
+                    href="/web_pwa_oca/static/img/icons/icon-180x180.png"
+                />
+                <link
+                    rel="apple-touch-icon"
+                    sizes="57x57"
```
pedrobaeza, line 81: What about if we don't declare all these variants? It's too much noise IMO.

web_pwa_oca/models/res_config_settings.py
```diff
@@ -133,21 +156,22 @@ def set_values(self):
...
+            if self._unpack_icon(self.pwa_icon).size < (
+                DEFAULT_ICON_SIZE,
+                DEFAULT_ICON_SIZE,
+            ):
                 raise exceptions.UserError(
-                    _("You can only upload PNG files bigger than 512x512")
+                    _(
+                        "You can only upload PNG files bigger "
+                        + "than {icon_size}x{icon_size}".format(
+                            **{"icon_size": DEFAULT_ICON_SIZE}
```
yajo, line 170:
```suggestion
                            icon_size=DEFAULT_ICON_SIZE
```

web_pwa_oca/models/res_config_settings.py
```diff
@@ -9,6 +10,25 @@
...
+    (256, 256),
+    (512, 512),
+    # APPLE
+    (57, 57),
+    (60, 60),
+    (72, 72),
+    (76, 76),
+    (114, 114),
+    (120, 120),
+    (180, 180),
```
yajo, line 29: All sizes are square. Would it make sense to just write one of the dimensions?

yajo (changes requested):
It seems ios can't handle transparency properly and applies a black background, so the little ant becomes invisible. Besides, that produces some animation glitches when closing the app. The icons should use a solid background. Other than that, it seems to work:

imagen

pedrobaeza:
The idea is also to try custom image, as the default one was previously working.

yajo:
Custom image works: 
imagen

joao-p-marques (approved):
Code and functional review :+1: 

As commented by @Yajo, it would be nice to solve the background issue, but it is not a blocking issue and it is working as expected I think.

AntoniRomera (approved):

AntoniRomera:
@Yajo We can merge this PR and open an issue for the background colour issue, I think it's not a bug that blocks usability.

yajo:
Yes, after all I guess nobody will want to use the default icon, so the fact it doesn't look good is not a big problem.

yajo (approved):

pedrobaeza:
I have aborted the merge. My comment still applies. This is adding too much noise and stored icons. I think only bigger one should be stored and declared, and Apple code should be smart enough to take the rest.

Tardo:
The problem with the background is because iPhone uses 'Maskable Icons' by default... currently chrome supports it. This  is the way that browsers adds support for "Adaptive Icons" in PWA.
So... one solution is to not use icons with alpha channel and follow the rule to make "maskable icons".

How "Maskable Icons" works:
https://storage.googleapis.com/web-dev-uploads/video/tcFciHGuF3MxnTr1y5ue01OGLBn2/mx1PEstODUy6b5TXjo4S.webm

pedrobaeza:
I insist in not putting that bunch of images, but only the minimal one. Can we put the bigger one and let iOS resize?

Tardo:
I will try with only 192x192 and 255x255 that its the minimum required by "lighthouse" to validate a PWA manifest.

And i will generate the new default icons with: https://maskable.app/editor

PR #2000, opened 2021-08-02
[14.0] Add web_m2x_options_manager: manage options via UI

SilvioC2C:


web_m2x_options_manager/models/ir_model.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def button_empty(self):
+        for ir_model in self:
+            ir_model.empty_m2x_create_edit_option()
+
+    def button_fill(self):
+        for ir_model in self:
+            ir_model.fill_m2x_create_edit_option()
+
+    def empty_m2x_create_edit_option(self):
```
simahawk:
```suggestion
    def _empty_m2x_create_edit_option(self):
```

web_m2x_options_manager/models/ir_model.py
```diff
@@ -0,0 +1,34 @@
...
+        """Removes every option for model ``self``"""
+        self.ensure_one()
+        self.m2x_create_edit_option_ids.unlink()
+
+    def fill_m2x_create_edit_option(self):
```
simahawk:
```suggestion
    def _fill_m2x_create_edit_option(self):
```

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,168 @@
...
+        "* Remove: option 'Create & Edit' is set to False only if not already"
+        " present\n"
+        "* Force Remove: option 'Create & Edit' is always set to False",
+        required=True,
+        string="Create & Edit Option",
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        self.clear_caches()
```
simahawk: wouldn't be better to clear only caches of related models?

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,168 @@
...
+        for opt in self:
+            opt.ir_model_name = opt.ir_model_id.model
+
+    def _inverse_ir_model_name(self):
+        ir_model_obj = self.env["ir.model"]
+        for opt in self:
+            # Using limit=1 since there's a SQL constrain to grant uniqueness
+            opt.ir_model_id = ir_model_obj.search(
+                [("model", "=", opt.ir_model_name)], limit=1
+            )
```
simahawk:
```suggestion
            opt.ir_model_id = ir_model_obj._get(opt.ir_model_name)
```

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,168 @@
...
+
+    @api.constrains("ir_model_name", "ir_model_field_name")
+    def _check_model_and_field_names(self):
```
simahawk: can't you avoid this by throwing an error on the inverse when no record is found?

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,168 @@
...
+
+    @api.constrains("ir_model_field_id")
+    def _check_field_type(self):
+        ttypes = ("many2many", "many2one")
+        if any(o.ir_model_field_id.ttype not in ttypes for o in self):
+            msg = _("Only Many2many and Many2one fields can be chosen!")
+            raise ValidationError(msg)
+
+    @api.constrains("ir_model_name", "ir_model_field_name")
+    def _check_field_model_uniqueness(self):
```
simahawk: why not an sql constrain to make them unique by model and field?

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,168 @@
...
+        fnames = ("ir_model_name", "ir_model_field_name")
+        for data in self.read(fnames):
+            dom = [(k, "=", v) for k, v in data.items() if k != "id"]
+            if sc(dom) > 1:
+                raise ValidationError(
+                    _("There is already an option for field '%s' in '%s'!")
+                    % (data["ir_model_field_name"], data["ir_model_name"])
+                )
+
+    def apply(self, node):
```
simahawk: apply what?
private?

simahawk (changes requested):
nice addition!
No test?

web_m2x_options_manager/models/m2x_create_edit_option.py
SilvioC2C: @simahawk I need to clear the cache of this model to avoid misbehavior when calling method `_get()` which is decorated with `ormcache`

web_m2x_options_manager/models/m2x_create_edit_option.py
SilvioC2C: @simahawk I believe a constrain makes it clearer and more manageable

web_m2x_options_manager/models/m2x_create_edit_option.py
SilvioC2C: @simahawk this way we can show the user an error message with field and model names, while we wouldn't be able to do that with a SQL constrain

web_m2x_options_manager/models/m2x_create_edit_option.py
simahawk: then clear the cache of that method only ;)

ivantodorovich: > @simahawk I need to clear the cache of this model to avoid misbehavior when calling method _get() which is decorated with ormcache

:point_up: I'd be good to put this in a comment

web_m2x_options_manager/models/m2x_create_edit_option.py
ivantodorovich: I agree with @simahawk here. I understand the UX improvement, but it's not worth the query in the loop IMO

web_m2x_options_manager/models/ir_model.py
```diff
@@ -0,0 +1,34 @@
...
+    def _empty_m2x_create_edit_option(self):
+        """Removes every option for model ``self``"""
+        self.ensure_one()
+        self.m2x_create_edit_option_ids.unlink()
+
+    def _fill_m2x_create_edit_option(self):
+        """Adds every missing field option for model ``self``"""
+        self.ensure_one()
+        existing = self.m2x_create_edit_option_ids.mapped("ir_model_field_id")
+        valid = self.field_id.filtered(lambda f: f.ttype.startswith("many2"))
```
ivantodorovich:
```suggestion
        valid_ttypes = self.env["m2x.create.edit.option"]._get_valid_ttypes()
        valid = self.field_id.filtered(lambda f: f.ttype in valid_ttypes)
        existing = self.m2x_create_edit_option_ids.ir_model_field_id
```

Assuming you define `_get_valid_ttypes` in `m2x.create.edit.option` :)
I think it's better to have a specific method for this, and also to avoid matching only on startswith

Hint: Also note that in 14.0 you can drop the `mapped` call :)

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,172 @@
...
+        for opt in self:
+            if opt.ir_model_field_id.model_id != opt.ir_model_id:
+                mname = opt.ir_model_name
+                fname = opt.ir_model_field_name
+                msg = _("'%s' is not a valid field for model '%s'!")
+                raise ValidationError(msg % (fname, mname))
+
+    @api.constrains("ir_model_field_id")
+    def _check_field_type(self):
+        ttypes = ("many2many", "many2one")
```
ivantodorovich, line 126: You can reuse `_get_valid_ttypes` here, as per my previous comment

web_m2x_options_manager/models/m2x_create_edit_option.py
```diff
@@ -0,0 +1,172 @@
...
+from odoo.exceptions import ValidationError
+from odoo.tools.cache import ormcache
+from odoo.tools.safe_eval import safe_eval
+
+
+class M2xCreateEditOption(models.Model):
+    _name = "m2x.create.edit.option"
+    _description = "Manage Options 'Create/Edit' For Fields"
+
+    ir_model_field_id = fields.Many2one(
```
ivantodorovich: Not blocking.. but here I would simply field names for the sake of readability

- `ir_model_field_id` -> `field_id`
- `ir_model_field_name` -> `field_name`
- `ir_model_id` -> `model_id`
- `ir_model_name` -> `model_name`

ivantodorovich (approved):
not blocking suggestions :+1:

web_m2x_options_manager/models/ir_model.py
SilvioC2C: @ivantodorovich I think that adding method `_get_valid_ttypes` would be an overkill, these features are meant to be used for `Many2many` and `Many2one` fields only... I'd say it's pretty safe to leave these values here

ivantodorovich (approved):

hkapatel-initos (approved):
Functional test: LGTM

hkapatel-initos:
Hello, @simahawk Can you please re-check this pull so.. we can merge these new module ASPS and add it our the list.

web_m2x_options_manager/tests/test_m2x_create_edit_option.py
```diff
@@ -0,0 +1,109 @@
...
+
+from lxml import etree
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import SavepointCase
+from odoo.tools.safe_eval import safe_eval
+
+
+class TestM2xCreateEditOption(SavepointCase):
+    def setUp(self):
```
simahawk, line 12: next time pls change this to `setUpClass` otherwise makes no sense to use `SavepointCase` :wink:

simahawk (approved):

PR #2001, opened 2021-08-03
[FIX] Conflicting modules

pepetreshere:


legalsylvain (changes requested):
Hi.

1) comments should be in english.

2) if there is a problem with ``odoo.sh``. It should be investigated, but removing a required lib is not possible.

PR #2002, opened 2021-08-09
[UPDATE] OCA web_ir_actions_act_multi v 14.0

Vitalii36:
Migrated module web_ir_actions_act_multi 13.0 ->  14.0, few changes about version


pedrobaeza:
Thanks for the contribution, but you are not preserving commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0, and moreover, there's another migration for this module: #1755, so I close it in favor of that one, that you can review instead.

Vitalii36:
Got it, thank you.

PR #2003, opened 2021-08-10
[14.0] [FIX] web_advanced_search: Many2one selection on Filters missing

### LINKED ISSUE #1957: [14.0] web_advanced_search Many2one selection on Filters missing
Looks like they didn't migrate this part to Odoo14 because it wasn't super clear from the comment where it was used. https://github.com/OCA/web/pull/1909#issuecomment-828997442 

The functionality this enabled in v13 is to have a many2one dropdown when adding a filter for a many2one field.

jeroenev:
Includes code for searching functionality on relational fields when adding filters.

Fixes: #1957

### DIFF
--- a/web_advanced_search/static/src/xml/web_advanced_search.xml
+++ b/web_advanced_search/static/src/xml/web_advanced_search.xml
@@ -1,11 +1,27 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<!-- Copyright 2017-2018 Jairo Llopis <jairo.llopis@tecnativa.com>
-     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <templates>
     <t t-inherit="web.FilterMenu" t-inherit-mode="extension" owl="1">
         <xpath expr="//CustomFilterItem" position="after">
             <li t-if="items.length" class="dropdown-divider" role="separator" />
             <AdvancedFilterItem fields="props.fields" />
+        </xpath>
+    </t>
+    <t t-inherit="web.CustomFilterItem" t-inherit-mode="extension" owl="1">
+        <xpath expr="//select[@t-elif]" position="after">
+            <t t-elif="fieldType === 'many2one'">
+                <t
+                    t-if="selectedOperator.symbol === '=' || selectedOperator.symbol === '!='"
+                >
+                    <Relational />
+                </t>
+                <input
+                    t-else=""
+                    type="text"
+                    class="o_input"
+                    t-att-value="condition.displayedValue"
+                    t-on-input="_onValueInput(condition)"
+                />
+            </t>
         </xpath>
     </t>
     <t t-name="web_advanced_search.AdvancedFilterItem" owl="1">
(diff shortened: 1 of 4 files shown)


web_advanced_search/static/src/xml/web_advanced_search.xml
```diff
@@ -1,13 +1,29 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<!-- Copyright 2017-2018 Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza: Please don't remove copyright, but add yours if needed

pedrobaeza (commented):
Thanks for restoring back this!

Several comments:

- On relational fields, it would be good to autoselect "Is equal to" operator by default.
- Using up and down arrows for moving between results on what you have typed makes the dropdown to be closed. I suppose any prevent propagation is missing.
- What about adding the feature on the popup (domain widget)?

jeroenev:
> On relational fields, it would be good to autoselect "Is equal to" operator by default.

I had this at first, but It would keep the data from the previous field if I switched to a different model
So if I had Contact selected, and then switch to a Company field, it would continue showing contacts in the widget

I tried using a custom `ComponentAdapter` to replace or update the `FieldWidget`, and tried adding the `FieldWidget` as a prop, but I couldn't get it working properly without errors using `ComponentAdapter` / `renderWidget` / `updateWidget`.
So I removed the code for auto-selecting `==`, so I could make it work reliably.

Probably overlooked something, but couldn't find what exactly. 

The other 2 things I can check later, at least the 'basic' functionality works now

pedrobaeza:
OK, thanks for the effort. When my colleague comes from holidays, he will check the technical part.

NL66278 (approved):
:+1: This is definitely a nice improvement and when tested worked well.

I am currently also trying to get some enhancements to work. Especially restoring the old functionality to have an "in selection" operator, where for x2x fields you can open a tree of a related model, enter search criteria, or select records, and then apply either the search criteria on the related model, or use the records selected... Have to learn a lot about how to interact with Owl. This PR provides some welcome pointers.

pedrobaeza:
@NL66278 I don't see such feature as needed if we improve the domain widget for accepting is equal + many2one input. Is there anything missing from this widget?

NL66278:
@pedrobaeza The original web_advanced_search_x2x made it very easy to enter advanced search criteria from the actual tree view + search of the related model. And to just select a few (or a lot) of records in the related model and say the related model should be in this list of records. I think with some work, you can get a part of this functionality from the domain widget -and it is great to have this widget at least, but I think not for all users this will be easy to do.

pedrobaeza:
True, selecting specific records from previous filtering is not available here, but we have it through https://github.com/OCA/web/tree/13.0/web_widget_domain_editor_dialog, so we can make a mention in the README and to not develop anything more. Check it and tell me what you think about it.

NL66278:
@pedrobaeza I will check it out. Thanks for the pointer!

rousseldenis (approved):

NL66278:
@pedrobaeza I checked out the web_widget_domain_editor_dialog. Definitely a nice module, with great functionality. But I think it is a little bit more geared to administrators / designers then to the educated end-users. You have to enter debug mode, manually select the appropriate model etc. Not a problem for you or me... but might be a hurdle for end-users / application managers. And I did not immediately see an equivalent for picking a number of records in a related model, but maybe I did not venture deep enough. Anyhow, this is my proposal to restore the old 'is in selection' logic: https://github.com/OCA/web/pull/2033

spadae22:
@pedrobaeza HI, I am having this module updated to match and improve searches for administrators / designers like myself as I used this module in v8 forever.  Thanks to all for comments and contributions. I am just writing the check to contribute back to the OCA.

PR #2004, opened 2021-08-11
[14.0][FIX] web_m2x_options: fix search_more behaviour

### LINKED ISSUE #1994: web_m2x_options: search_more not working if values under limit
Module: web_m2x_options

Odoo version: 14.0

Issue: The "Search more" button on a many2one field doesn't show up when setting "search_more" to True if the number of values in the dropdown is under the limit (7 by default).

Expected: When I set the search_more to true, I expect the button to appear regardless of the limit.

Solution: I've changed the condition to always show if search_more is set. 

https://github.com/OCA/web/blob/cef972152c969d0347232048e1e13ccf74516441/web_m2x_options/static/src/js/form.js#L238

Currently:
`if (values.length > self.limit)`

Proposed change:
`if (values.length > self.limit || can_search_more) `

I can submit a PR if you think this is an appropiate solution.

SilvioC2C:
In README.rst it is said that ``search_more`` key is "[u]sed to force disable/enable search more button".
However, it only applies on fields where the records returned by the name_search are more than the node's limit.

This commit fixes this behaviour, forcing the "Search More..." option to be *always* disabled/enabled if ``search_more`` key is set either locally (in the node's options) or globally (via the ir.config_parameter record).
Moreover, it establishes a clear line of priority for checking whether "Search More..." should be displayed/hidden (local node has the highest priority, global config mid priority, limit check as fallback).

Closes #1994

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -222,70 +222,76 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
                         });
                     }
 
-                    // Search more... if more results that max
-                    var can_search_more =
-                            self.nodeOptions &&
-                            self.is_option_set(self.nodeOptions.search_more),
-                        search_more_undef =
-                            _.isUndefined(self.nodeOptions.search_more) &&
-                            _.isUndefined(
-                                self.ir_options["web_m2x_options.search_more"]
-                            ),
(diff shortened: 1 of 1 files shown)


yankinmax (approved):

simahawk (approved):
LG

PR #2013, opened 2021-09-02
[12.0][FIX] web_timeline UTC timezone and + 1 hour when adding events

Martronic-SA:
Date times must be sent as UTC to Odoo. and why adding 1 hour? just leave the javascript timeline make the job.


pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

PR #2023, opened 2021-09-10
[12.0][IMP] web_m2x_options: Disable hardcoded "Search More..." limit when no search value is set.

ypapouin:
To improve `name_get` performance, there is an existing hard-coded limit.
But when no search value is set, this `name_get` query is useless since only the domain will limit the result.

With this PR, when the field input is empty and when the user clicks on "Search More...", we let the Form doing its `web/dataset/search_read` query without locking the domain on `'[id', 'in', (...)]`. And since the `name_get` is not called, it should be faster.


ypapouin:
@CarlosRoca13, can you review this PR ?

Albertocanal (approved):

PR #2024, opened 2021-09-10
[12.0][ADD] web_widget_one2many_product_picker_sale_stock_available_info_popup

Tardo:
Adds support for 'sale_stock_available_info_popup' in the product picker

cc @tecnativa TT31744


web_widget_one2many_product_picker_sale_stock_available_info_popup/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+Install and use it. No configuration required.
```
legalsylvain: This file is not necessary.

pedrobaeza (commented):
Why do we need all of this JS stuff and can't be done only extending original XML view?

chienandalu:
I'd say that the way the base ('web_widget_one2many_product_picker_sale_stock' ) is implemented makes it necessary. It's a declaration of the fields that `sale_stock_available_info_popup` renders. I agree that a reuse of the widget without need of further glue module would be ideal, but I don't know if there are technical reasons that disallowed it...

Tardo:
It is made this way because there is no "product_picker_sale". This module can be used in "form" and "formPWA" without having to make any changes for the specific view.

pedrobaeza:
OK, let's continue for now this way and we'll refactor in v13. Can you check Travis?

Tardo:
Changes done

pedrobaeza (approved):

chienandalu (approved):
Tested :+1: 

Thanks!

PR #2025, opened 2021-09-11
[12.0] web_widget_digitized_signature: Allow employee user to set their own signature in preferences view

petrus-v:
Before this pull request, connect as demo/demo user you won't be able to set its digitized signature in the preferences form (click on user name on the top right corner > preferences)

Using this PR each user is able to set it own signature without administrator access right !


legalsylvain (approved):
Make sense

ivantodorovich (approved):

PR #2026, opened 2021-09-13
[12.0][FIX]web_widget_digitized_signature:Respect custom width-height set f…

diggy128:
…or widget, add border

Although widget allowed to set size in view (width, height) it was hardcoded to 550x150.
Allow setting width and height for widget and fallback to previous default of 550x150
Set border on signature widget to explicitly show widget limits


legalsylvain (approved):
thanks. LGTM. 

Code review / No test.

PR #1905, opened 2021-04-28
[IMP] web_widget_bokeh_chart: Upgrade library version

etobella:
@LoisRForgeFlow 
This intends to upgrade the library.

Tested with https://github.com/OCA/reporting-engine/pull/506

PR #2027, opened 2021-09-14
[FIX] web_widget_bokeh_chart: remove custom-bokeh.scss

albariera:
When adding a bokeh chart inside a page element, I have faced the problem that the visualization is not good when trying to plot a bokeh row or colum, as you can see in the next image: 
mal
Inspecting version 14.0, we have realized that if we remove the custom-bokeh.scss, this problem is resolved, as you can see here: 
bien


etobella:
@LoisRForgeFlow This scss was removed on https://github.com/OCA/web/pull/1905 for 14.0. WDYT?

We have tested the change on DDMRP and it seemed to work fine.

JordiBForgeFlow:
:+1: Works for us!

JordiBForgeFlow (approved):

etobella (approved):

PR #2037, opened 2021-09-26
[ADD] github stale action: Work on oldest issues/prs first

hbrunn:
the action from #2034 ran for the first time, that's nice. But with the amount of open issues we have here (and in a couple of other repos), this runs into limits: https://github.com/OCA/web/runs/3712831044?check_suite_focus=true#step:2:1310

With this proposal, the action looks at oldest issues/prs first, and this way we hopefully don't have to increase the amount of operations, as I can't judge how many this will be over all of OCA.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.2.0
+_commit: v1.2.1
 _src_path: gh:oca/oca-addons-repo-template
 dependency_installation_mode: PIP
 generate_requirements_txt: true
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -13,6 +13,7 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           # General settings.
+          ascending: true
           remove-stale-when-updated: true
           # Pull Requests settings.
           # 120+30 day stale policy for PRs
@@ -50,6 +51,7 @@ jobs:
         uses: actions/stale@v4
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+          ascending: true
           only-labels: "needs more information"
           exempt-issue-labels: "no stale"
           days-before-stale: 15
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Shouldn't this be proposed on the template? cc @sbidoul @Yajo

hbrunn:
I think we should use one repo for testing some stuff, and then move what we learn for the template. Maybe even set it to running daily here so that we can see the results faster?

pedrobaeza:
OK, I thought you have the solution clear, but if experimentation is needed, then let's go ahead with just one repo.

yajo (approved):
Some comments about this.

1. I attach the logs, so when they get garbage-collected this can be used for reference: [logs_478.zip](https://github.com/OCA/web/files/7234682/logs_478.zip)
2. Very old stuff will get processed anyways on later runs, right? In case the 30 operations are exhausted, do we really want to exhaust them with old stuff by default? I mean: newer stuff is more likely going to get a reaction... Well, I agree with this change, but I just wanted to raise this question.
4. FYI we can do template prereleases and those can be applied selectively with `copier --prerelease update`.

hbrunn:
1. that could come handy, thanks
2. my assumption is that old issues are more likely to be closed - from how I read https://github.com/actions/stale/blob/main/src/classes/issues-processor.ts#L534, we'd be stuck with the first 100 issues forever if we don't increase the amount of operations, right?
3. nice!

pedrobaeza (commented):
I agree on stalling oldest ones first.

Shouldn't this contain the improvements on the action for not repeating twice the same operation?

yajo:
> Shouldn't this contain the improvements on the action for not repeating twice the same operation?

I just released https://github.com/OCA/oca-addons-repo-template/releases/tag/v1.2.1, you can update from copier and rebase this patch if you want.

pedrobaeza (approved):

hbrunn:
I did the update right now, so this now also contains the improved template

yajo:
OK, let's see how it goes. If it works fine, we should add it to the template itself.

yajo:
I just checked the action from this weekend and this works really nice: https://github.com/OCA/web/runs/3782134552?check_suite_focus=true

PR #2041, opened 2021-10-04
web responsive error xml templates device

mt-software-de:


pedrobaeza:
Can you explain the steps to reproduce the error in OCA runbot?

mt-software-de:
No i can not explain how to reproduce it in OCA runbot, because i dont know how your runbot is working, is there an howto for it?
The problem is, that "device" is not defined as an variable the device is callable via env.device.

pedrobaeza:
I'm afraid the problem you are having is only on your side then, as the module is working without problems in a clean environment. OCA runbot can be started from the button "Runbot | Try me" that is on the README.

mt-software-de:
ok i will check this. thanks

SplashS:
@mt-software-de `device` here is defined via deviceContext wrapper.
it provides the automatic rerendering on browser viewport size change.
The way you provide may work but without this automatic rerendering.
Can you describe the problem you've got?

mt-software-de:
sorry for the late answer im very short of time right now. 
i will check why we are facing this error, but for now i will close this pull request

PR #2046, opened 2021-10-07
[13.0][FIX] web_time_range_menu_custom: Mix custom and 'no custom' periods

Tardo:
Apply correct periods when mix "custom" with "no custom" periods.

Steps to reproduce:
- Go to pivot view
- Select a custom time range
- Select a predefined comparison period (Ex. Previous Period)

Now you see that the comparison period is not being applied.

After this PR the comparison period is applied correctly.

cc @tecnativa TT29249

### DIFF
--- a/web_time_range_menu_custom/static/src/js/control_panel_model.js
+++ b/web_time_range_menu_custom/static/src/js/control_panel_model.js
@@ -52,6 +52,21 @@ odoo.define("web_time_range_menu_custom.ControlPanelModel", function(require) {
                             context.timeRangeMenuData.timeRange
                         );
                     }
+                    if (filter.comparisonTimeRangeId !== "custom_comparison_period") {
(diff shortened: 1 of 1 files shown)


web_time_range_menu_custom/static/src/js/control_panel_model.js
```diff
@@ -52,6 +52,22 @@ odoo.define("web_time_range_menu_custom.ControlPanelModel", function(require) {
                             context.timeRangeMenuData.timeRange
                         );
                     }
+
```
pedrobaeza: I'm not very used to JS guidelines, but empty lines is a common practice?

pedrobaeza (approved):
Please indicate steps to reproduce the problem in UI.

Tardo:
First comment was updated and the empty line was removed.

pedrobaeza:
Please put the steps also in the commit message (and the TT).

chienandalu (approved):

PR #2047, opened 2021-10-09
[15.0][MIG] web_responsive

lijoantony:
Migration from 14.0 to 15.0

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Web Responsive",
     "summary": "Responsive web client, community-supported",
-    "version": "14.0.1.0.2",
+    "version": "15.0.1.0.0",
     "category": "Website",
     "website": "https://github.com/OCA/web",
     "author": "LasLabs, Tecnativa, " "Odoo Community Association (OCA)",
@@ -15,7 +15,7 @@
     "depends": ["web", "mail"],
     "development_status": "Production/Stable",
     "maintainers": ["Yajo", "Tardo"],
-    "data": ["views/assets.xml", "views/res_users.xml", "views/web.xml"],
+    "data": ["views/res_users.xml", "views/web.xml"],
     "qweb": [
         "static/src/xml/apps.xml",
         "static/src/xml/form_buttons.xml",
@@ -26,5 +26,11 @@
         "static/src/xml/control_panel.xml",
         "static/src/xml/search_panel.xml",
     ],
+    "assets": {
+        "web.assets_backend": [
+            "web_responsive/static/src/css/**/*",
+            "web_responsive/static/src/js/**/*",
+        ],
+    },
     "sequence": 1,
 }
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Branches for 15.0 are still not created, so you have to wait a bit. Then you need to follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0, and due to the huge refactoring on OWL, I doubt these few changes are enough.

legalsylvain:
> due to the huge refactoring on OWL, I doubt these few changes are enough.

As far as i unterstood the gery talk regarding owl, v15 seems compatible with old and new framework. Maybe we could talk regarding OCA policy ?
Do we have to migrate in new framework or allow old syntax ? ( Like for V8, when new API was introduced)
Not sure my point is relevant ?

pedrobaeza:
@legalsylvain It's not a question of choice, but if things like navbar has been converted to OWL, then the JS behind https://github.com/OCA/web/blob/14.0/web_responsive/static/src/js/web_responsive.js is not working anymore, and it should be converted to be compatible. You can plug to the legacy system, or the new one, but some work should be done, as it's not going to work as is for sure (I hope to be wrong, but our first investigations pointed to this).

legalsylvain:
Yes you're right ! I face to the same situation with PoS. If Pos is in owl, oca module should be in owl. (Afaik)
My question was more when we have the choice. Do we accept new "legacy syntax" in v15, or do we force contributors to develop in owl (if possible) ?
No idea regarding the answer thought, but better to ask when there are no modules in v15 branch.

pedrobaeza:
I think we should leave each contributor to decide. The rule is that the module works in the current version. Anticipating one version the conversion shouldn't be mandatory, and maybe it's even counter productive, as the compatibility later may be higher depending on which part you are touching.

SplashS:
I mostly migrated web_responsive to v15.
There will be a lot of changes.
Some investigations:
1. In odoo webclient only views are still not converted to owl.
2. Control panel exists in two separated variants, both in owl. For new owl views (pivot and graph) and for other old views. We should patch them both.
3. There is new hotkey service and now it’s simple to patch it.
4. Kanban, list, form and calendar views can mostly reuse code from web_responsive v14. They will be migrated to owl in v16.
5. Fuzzy search has rewritten from scratch inside odoo and has limitations.
6. Navbar has completely rewritten in owl.
7. New discuss features are ready for mobile from the box.

So I agree with @pedrobaeza what we should choose technology (owl or widgets) based on technology of patched component.

nabilragab:
> @legalsylvain It's not a question of choice, but if things like navbar has been converted to OWL, then the JS behind https://github.com/OCA/web/blob/14.0/web_responsive/static/src/js/web_responsive.js is not working anymore, and it should be converted to be compatible. You can plug to the legacy system, or the new one, but some work should be done, as it's not going to work as is for sure (I hope to be wrong, but our first investigations pointed to this).

yes it's not going to work, I have made an attempt with the web_refresher which is rendering on the navbar but I couldn't get it to render, could you give some pointers on how to develop the migrations correctly?

@SplashS would appreciate a pointer.

legalsylvain:
2 blocking points. Could you : 
- make a PR against new 15.0 branch. (it is created now)
- keep history, (via cherry pick)

Thanks !

pedrobaeza:
Superseded by #2058

PR #2036, opened 2021-09-24
[13.0][FIX] web_drop_target: Don't collide with odoo drag and drop implementation

Tardo:
+ Old:
dg_old

+ New:
dg_new

cc @tecnativa TT32093

PR #2052, opened 2021-10-15
[ADD] web_drop_target_mail: init

oerp-odoo:
Module to correctly integrate mail composed message with drag and drop
functionality.

If only web_drop_target module is used for composed message, it will not
only break existing functionality (images drag and drop stop working),
but it will also create "ghost" attachments on for
`mail.compose.message` object, which are not used anywhere and not
visible even by using search (because Odoo filters out attachments that
have res_id=0).

This module makes sure that dragged images are put as content on html
widget and other type of files as attachments.

Because we can't directly relate attachments with transient model, we
use built in attachment field to temporary store it there, which
is later (after message is posted) resolved by standard functionality
by relating attachments to parent record.

web_drop_target improvements/fixes:

* _handle_drop_items: override on FormController was overwriting event
    object with index, which would then be propagated further.
* introduced more control with file types. Now its possible to either
    define types or subtypes (e.g. all types of images) to either
    include (allow) or exclude from drag and drop.
* _attachment_fname: attribute can be used to specify attachment field
    name that would be used to store attachments data (only tested with FieldMany2ManyBinaryMultiFiles widget)
* _use_drag_events: attribute to enable/disable drag events if its
    interfering with other similar events. Enabled by default.

Added improvements are backwards compatible (it won't change existing
functionality if no options are changed)

Video, to show how it works.

[drag-n-drop-compose-msg.zip](https://github.com/OCA/web/files/7351700/drag-n-drop-compose-msg.zip)


pedrobaeza:
Hi, we fixed this in the base module in 13.0 in #2036

oerp-odoo:
I see. Are you planning to back port it to 12.0 too? As I still need this for 12.0.

pedrobaeza:
I'm afraid not, as we are going forward, but you can take it and do it.

oerp-odoo:
Does your implementation work correctly with composed message (pop up)? In this implementation, it is made sure to attach images in `body`, but other files are attached as if they were manually uploaded via `Attach a File` button.

pedrobaeza:
On v13, we rely on native drag and drop capabilities of Odoo, so no extra layer. The PR only assures that the new one doesn't intercept. But in v12 there's no drag and drop for that things, so there can be room for your improvement in this version.

PR #2056, opened 2021-10-20
[MIG] web_refresher: Migration to 15.0

nabilragab:
There is a minor issue I seem to understand, theoretically should work, but it doesn't render the refresher next to the pager components, the mistake I am doing is probably related to the core JS framework changes, can someone give a hint on how to  render the button :/


web_refresher/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "Web Refresher",
+    "version": "15.0.1.0.0",
+    'category': 'Hidden',
+    "author": "Compassion Switzerland, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/web",
+    'web.assets_qweb': [
```
SplashS, line 8: You should put "web.assets_qweb" inside another dict. It should looks like "assets": { "web.assets_qweb": [...], },

web_refresher/static/src/legacy/xml/web_refresher.xml
```diff
@@ -0,0 +1,28 @@
...
+<!-- This is now considered part of the legacy js components and is expected to change in a future release  -->
+<templates id="template" xml:space="preserve">
+
+    <t
+        t-name="web_refresher.Pager"
+        t-inherit="web.Pager"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+        <xpath expr="//nav[hasclass('o_pager')]//span[1]" position="after">
```
SplashS, line 13: Please prefer simple xpaths. in this case I suggest //span[hasclass('o_pager_counter')]

SplashS (changes requested):
Hello! I think I give you enough feedback to make it works.

In v15 there are two implementations of ControlPanel.
In this PR module only legacy one has been patched.
Fortunately new implementation is used only for pivot and graph views. And web_refresher module doesn't support them anyway.

pedrobaeza:
Already merged through other PR.

PR #2062, opened 2021-10-26
[13.0][ADD] web_widget_one2many_tree_line_duplicate

Tardo:
clone

cc @tecnativa TT32393


CarlosRoca13 (approved):
Tested locally

carlosdauden (approved):

web_widget_one2many_tree_line_duplicate/README.rst
```diff
@@ -0,0 +1,96 @@
...
+   :local:
+
+Usage
+=====
+
+This module works on all one2many tree views.
+
+**Available Options**
+
+- allow_clone > Add the icon to clone the line (default: true)
```
pedrobaeza: The default should be the contrary: False.

pedrobaeza (changes requested):
Why a lot of onchanges are executed when cloning the line? The clone should happen without executing such onchanges. Just "clone" the existing data in the other line.

Selección_004

Tardo:
@pedrobaeza The main cause to do it in this way is because the Odoo framework can't difference between manual values (values passed as argument) and 'onchange' values.. When I do all changes in the "CREATE", the onchange values overwrite the manual values. So, for example, if you change the price in the line... the 'onchange' overwrite it to put the price defined in the product.
Other causes are restrictions made by other "known/unknown" modules. Warning messages or other types of implementations.

Tardo:
Changes done... now clones without launch an onchange per field.

kittiu:
> Changes done... now clones without launch an onchange per field.

Yes, this is an important behavior.

kittiu:
I think this can be more useful to allow selecting only some columns to duplicate.

Tardo:
@kittiu It does not fall within our use case. I have added it to the roadmap in case anyone is interested in doing it.

This PR is ready to review/merge.

pedrobaeza (approved):
Tested in customer instance

legalsylvain (approved):
tested on runbot. works like a charm.

thanks for this module.

PR #2063, opened 2021-10-26
[13.0][IMP] web_disable_export_group: Differenciate between the XLSX export and export all

etobella:
With this change, we have now two groups, one for the excel and one for the Export from Actions:

image

A JS test has been added

### DIFF
--- a/web_disable_export_group/migrations/13.0.2.1.0/post-migration.py
+++ b/web_disable_export_group/migrations/13.0.2.1.0/post-migration.py
@@ -1,0 +1,10 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    group = env.ref("web_disable_export_group.group_export_data")
+    new_group = env.ref("web_disable_export_group.group_export_xlsx_data")
+    groups = env["res.groups"].search([("implied_ids", "=", group.id)])
+    groups.write({"implied_ids": [(4, new_group.id)]})
+    group.users.write({"groups_id": [(4, new_group.id)]})
--- a/web_disable_export_group/security/groups.xml
+++ b/web_disable_export_group/security/groups.xml
@@ -2,8 +2,21 @@
 <!-- Copyright 2016 Onestein
      License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).-->
 <odoo>
+    <record model="ir.module.category" id="category_export_data">
+        <field name="name">Export</field>
+    </record>
+
+    <record id="group_export_xlsx_data" model="res.groups">
+        <field name="name">Direct Export</field>
+        <field name="category_id" ref="category_export_data" />
+        <field
+            name="users"
+            eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
+        />
+    </record>
     <record id="group_export_data" model="res.groups">
-        <field name="name">Export Data</field>
+        <field name="name">Advanced Export</field>
+        <field name="category_id" ref="category_export_data" />
         <field
             name="users"
             eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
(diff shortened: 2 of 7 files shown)


pedrobaeza:
Please add migration script for belonging to both groups those users that previously were on the only group.

etobella:
It will be done automatically, as the Export Group implies Export XLSX group (At least it was cleaner IMO)

pedrobaeza:
Uhm, but maybe you want one and not the other in the same hierarchy level. I think they also deserves now an specific `ir.module.category` ("Export" for example).

etobella:
Ok! Let me check it, but no problem :smile:

etobella:
image
@pedrobaeza Done!

web_disable_export_group/security/groups.xml
```diff
@@ -2,8 +2,21 @@
...
+    <record id="group_export_xlsx_data" model="res.groups">
+        <field name="name">Export XLSX Data</field>
+        <field name="category_id" ref="category_export_data" />
+        <field
+            name="users"
+            eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
+        />
+    </record>
     <record id="group_export_data" model="res.groups">
         <field name="name">Export Data</field>
```
pedrobaeza:
```suggestion
        <field name="name">Direct Export</field>
```

web_disable_export_group/security/groups.xml
```diff
@@ -2,8 +2,21 @@
 <!-- Copyright 2016 Onestein
      License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).-->
 <odoo>
+    <record model="ir.module.category" id="category_export_data">
+        <field name="name">Export</field>
+    </record>
+
+    <record id="group_export_xlsx_data" model="res.groups">
+        <field name="name">Export XLSX Data</field>
```
pedrobaeza:
```suggestion
        <field name="name">Advanced Export</field>
```

web_disable_export_group/migrations/13.0.2.1.0/post-migration.py
```diff
@@ -0,0 +1,10 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    group = env.ref("web_disable_export_group.group_export_data")
+    new_group = env.ref("web_disable_export_group.group_export_xlsx_data")
+    groups = env["res.groups"].search([("implied_ids", "=", group.id)])
+    groups.write({"implied_ids": [(4, new_group.id)]})
```
pedrobaeza, line 9: Why doing this?

etobella: For example that could be necessary if you are using rules, as the group will be ignored

web_disable_export_group/security/groups.xml
etobella: Maybe this should be renamed Advanced Export and XLSX  renamed to Direct Export. Isn't it?

pedrobaeza: I don't know which is each one, but Advanced/Direct, avoiding the XLSX, as both exports by default to that format.

web_disable_export_group/migrations/13.0.2.1.0/post-migration.py
pedrobaeza: Rules? This group is for a very specific thing, and you shouldn't have children groups, or if you have them, then do the direct assignation of users, not with implied.

etobella: Sorry, I ment roles: https://github.com/OCA/server-backend/tree/13.0/base_user_role

In this case, the groups assigned automatically to the user are ignored. You need to check which groups have the old group assigned and then apply the XLSX too.

pedrobaeza (approved):

olgamarcocb (approved):
:+1:

pedrobaeza:
Please fw-port it to v14:

/ocabot merge nobump

Issue #2057, opened 2021-10-21
[13.0][BUG] web_drop_target changes made a regression in mail_drop_target


LoisRForgeFlow:
Since this commit 1c0c2609dd94f23106432a189a7fb09f5ee05e6e the dependent module [mail_drop_target](https://github.com/OCA/social/tree/13.0/mail_drop_target) stopped working as expected. Before the emails dragged and dropped into the chatter were added to the chatter conversation, now they are attached as regular files.

Before that change:
![Peek 2021-10-21 09-31](https://user-images.githubusercontent.com/23449160/138231897-758dff77-889d-4aa6-9fcb-6b71a1f09095.gif)

After that change:
![Peek 2021-10-21 09-26](https://user-images.githubusercontent.com/23449160/138231911-2eeca1ca-d769-4ecb-a9c8-8b80389a4e7a.gif)

cc @Tardo

pedrobaeza:
I have to say that I have never used that module, and now that I see it in action, the feature that is adding is barely a sacrilege, as it's deforming the intention of the chatter. Anyways, we should repair it for those that use it, but I strongly recommend to not do it.

LoisRForgeFlow:
You might be right but as you also are saying, the feature was already there and this has been a regression.

Thanks for the quick reaction as always!

LoisRForgeFlow:
Closed via #2065

PR #2065, opened 2021-10-28
[FIX] web_drop_target: Fix check files and set the main element as parent

etobella:
This commit implements two changes: 

The first one is to use element as drop zone if no overlay zone is defined. We cannot find o_content otherwise.

The second one is to fix the check if a file exists. This is important because if the filter was added, the _.empty failed

It solves https://github.com/OCA/web/issues/2057

### DIFF
--- a/web_drop_target/static/src/js/web_drop_target.js
+++ b/web_drop_target/static/src/js/web_drop_target.js
@@ -16,8 +16,8 @@ odoo.define("web_drop_target", function(require) {
         // Add the mime types you want to support here, leave empty for
         // All types. For more control, override _get_drop_items in your class
         _drop_allowed_types: [],
-        // Determine the zone where can drop files
-        _drop_zone_selector: ".o_content",
+        // Determine the zone where can drop files, if not defined, we use the element
+        _drop_zone_selector: undefined,
 
         /**
          * @override
@@ -82,7 +82,9 @@ odoo.define("web_drop_target", function(require) {
          */
         _add_overlay: function() {
             if (!this._drop_overlay || this._drop_overlay.length === 0) {
-                this.$drop_zone = this.$(this._drop_zone_selector).last();
+                if (this._drop_zone_selector)
+                    this.$drop_zone = this.$(this._drop_zone_selector).last();
+                else this.$drop_zone = this.$el;
                 // Element that represents the zone where you can drop files
                 // TODO: This name is preserved for not breaking the compatibility with other modules,
                 // but should be changed in new versions to follow the standard ($name)
@@ -138,7 +140,7 @@ odoo.define("web_drop_target", function(require) {
          */
         _onBodyFileDragover: function(ev) {
             ev.preventDefault();
-            if (!_.isEmpty(this._get_drop_items(ev))) {
+            if (_.isEmpty(this._get_drop_items(ev))) {
                 const drop_zone_offset = this.$drop_zone.offset();
                 const overlay_css = {
                     top: drop_zone_offset.top,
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Tested locally, LGTM

LoisRForgeFlow:
cc @pedrobaeza @Tardo could you have a look?

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -138,7 +140,7 @@ odoo.define("web_drop_target", function(require) {
          */
         _onBodyFileDragover: function(ev) {
             ev.preventDefault();
-            if (!_.isEmpty(this._get_drop_items(ev))) {
+            if (this._get_drop_items(ev).length === 0) {
```
pedrobaeza: Should this check before if is set?

etobella: The problem is, that when it checks data, the _.filter inside _get_drop_items returns an array, on the other case, it is an object. On the second case, _.isEmpty works perfectly, but fails on the first one.

pedrobaeza: Shouldn't we normalize then the return of `_get_drop_items`?

etobella: Yes, and probably use our proposal... Let me explain, previously, the items were processed and were consistent.

OnFileDragOver should have no drop_items, but as it was not working, it was added a negative. The right solution would be remote the negative and process data on the same way (probably returning to the old commit logic...)

pedrobaeza: I let them @Tardo and you to agree this.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -138,7 +140,7 @@ odoo.define("web_drop_target", function(require) {
          */
         _onBodyFileDragover: function(ev) {
             ev.preventDefault();
-            if (!_.isEmpty(this._get_drop_items(ev))) {
+            if (_.isEmpty(this._get_drop_items(ev))) {
```
etobella, line 143: _get_drop_items should be empty on DragOver, as the files have not been droped.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -16,8 +16,8 @@ odoo.define("web_drop_target", function(require) {
         // Add the mime types you want to support here, leave empty for
         // All types. For more control, override _get_drop_items in your class
         _drop_allowed_types: [],
-        // Determine the zone where can drop files
-        _drop_zone_selector: ".o_content",
+        // Determine the zone where can drop files, if not defined, we use the element
+        _drop_zone_selector: undefined,
```
Tardo, line 20: Fallback to $el seems ok to me... but this change can cause problems with current implementations no?. For example, i think that DMS will need be updated.

Tardo (approved):
LGTM :+1: Thanks for the fix. I missed converting FileList to Array :/

web_drop_target/static/src/js/web_drop_target.js
etobella: Well, with the current implementation, DMS is not working :cry: 

Tested with the changes and without. Does it work for you (maybe a problem from my side)?

Tardo: http://3510273-13-0-257ac3.runbot2-3.odoo-community.org/ here works ...

etobella: Then it is my fault... I will check again :sob:

etobella: I had a wrong branch in local. I restored the right code and it worked

It is functional without any changes. Would be great to apply a small patch, but it is not really necessary at first (nothing breaks :smile)

pedrobaeza:
Is this needed in v14?

etobella:
Probably, I will review between today and monday :wink:

PR #2067, opened 2021-11-02
[MIG] web_widget_x2many_2d_matrix: Migration to 15.0

AyoubZahid:
Migrated to v15


pedrobaeza:
Thanks for the contribution.

PR #2082, opened 2021-11-16
[13.0][IMP] web_widget_one2many_product_picker_sale_stock: Adapt to work with product_picker

Tardo:
cc @tecnativa TT32978


pedrobaeza:
Please leave only one commit

PR #2083, opened 2021-11-16
[14.0][MIG] `web_ckeditor4` (renamed to `web_widget_ckeditor`)

### LINKED ISSUE #2019: [RFC] widget to display html fields in Odoo 12 (web_ckeditor4 ?)
Hi all, 

In old versions of odoo, I used to use the module ``web_ckeditor4`` ([last version 10.0](https://github.com/OCA/web/tree/10.0/web_ckeditor4)) that allows to use the [ckeditor4](https://github.com/ckeditor/ckeditor4) library for html fields.

I didn't found in Odoo 12.0 (and later) that module, and no work around it.

Is there a new way to display html fields ? the default odoo widget (``widget="html"``) is very poor.

CC : @quentinDupont. Ref ticket : apps/deck/#/board/144/card/1843

ivantodorovich:
Migrated `web_ckeditor4` (from 10.0).

It's a complete refactor based on [CKEditor5](https://ckeditor.com/ckeditor-5/), basically built from scratch


legalsylvain:
Hi @ivantodorovich. 
- First thanks for porting this very usefull module ! 
- I just tested on runbot and it works like a charm. 
- a minor problem found : when I try to move an image, there is a "Drop your files here" that appears, et that doesn't disappear. (you have to reload the page). Did you experimented this issue ? [Test](http://3516675-2083-78d8d5.runbot2-2.odoo-community.org/web#id=1&action=305&active_id=1&model=project.task&view_type=form&cids=1&menu_id=203).

just a quite important question : As far as I remember, in the previous web_ckeditor4 implementation, all the fields was automatically display with ckeditor widget. In the new implementation, it requires to overload each field. I see two problem to that implementation : 
- very less end-use frendly. (people downloading module on OCA / odoo store with technical knowledge can not use it).
- it ask to overload manually by code, or to change by the UI each view.

Have you considered to make the new display by default ?

ivantodorovich:
Thanks @legalsylvain  for the quick review!

> just a quite important question : As far as I remember, in the previous web_ckeditor4 implementation, all the fields was automatically display with ckeditor widget. In the new implementation, it requires to overload each field. I see two problem to that implementation :
>
> very less end-use frendly. (people downloading module on OCA / odoo store with technical knowledge can not use it).
it ask to overload manually by code, or to change by the UI each view.
Have you considered to make the new display by default ?

I considered it and honestly I wasn't sure what the right approach is.
I didn't like the intrusiveness of replacing the original `html` widget system-wide.
However I think you're right, it's easier for non-technical people. 

I've just pushed some changes to make it enabled by default, and configurable through the `web_widget_ckeditor.replace_html_widget` config parameter


> a minor problem found : when I try to move an image, there is a "Drop your files here" that appears, et that doesn't disappear. (you have to reload the page). Did you experimented this issue ? Test.

AFAICS this "Drop your files here" is coming from [the web_drop_target module](https://github.com/OCA/web/blob/50413d7459778106b4f3c76367cf6faee85cf50f/web_drop_target/static/src/xml/widgets.xml#L11). It could either be an issue with that module, or just incompatibilities with CKEditor.

But I tested without `web_drop_target` and this issue doesn't exist

web_widget_ckeditor/static/src/js/field_ckeditor.js
```diff
@@ -0,0 +1,341 @@
...
+             */
+            _onChange: function () {
+                this._doDebouncedAction.apply(this, arguments);
+            },
+        }
+    );
+
+    field_registry.add("ckeditor", FieldHtmlCKEditor);
+
+    // Replace core's html widget with ckeditor, depending on the ir.config_parameter
```
simahawk: IMO this would be easier by loading another asset view on demand: use the flag `active` to disable it.
You avoid issuing another call too. If you want to keep this, I'd recommend doing one single read on the top together w/ the settings.

simahawk (approved):
LG, code review only.

legalsylvain:
> AFAICS this "Drop your files here" is coming from the web_drop_target module. It could either be an issue with that module, or just incompatibilities with CKEditor.

Thanks ! i regreped Odoo but I didn't found that sentence. maybe just add a comment in the ROADMAP.rst section, talking regarding incompatibility.

>  I didn't like the intrusiveness of replacing the original html widget system-wide.

Well, I used this module during years. (before my migration to 12.0) In fact, I never saw this module as intrusive, mainly because the widget ckeditor is better than the native widget, and allow most possibilities.

> I've just pushed some changes to make it enabled by default, and configurable through the web_widget_ckeditor.replace_html_widget config parameter

awsome ! i didn't dare to ask such option, because I thought it was hard to implement.

web_widget_ckeditor/static/src/js/field_ckeditor.js
ivantodorovich: Thanks!! Indeed I like this approach. I've just pushed the changes

ivantodorovich:
> Thanks ! i regreped Odoo but I didn't found that sentence. maybe just add a comment in the ROADMAP.rst section, talking regarding incompatibility.

Added to ROADMAP :)

tarteo (approved):
Thanks!! LGTM

tommysnello:
i have tried to make the back port to version 12.0 without success, someone has advices to give me for the accomplish?

SnitCom:
hi there, is there ckeditor for documentation style?

PR #2084, opened 2021-11-17
[12.0][add] web_session_allow_public

fredzamoabg:
add web_session_allow_public from https://github.com/OCA/web/pull/402


tafaRU:
@fredzamoabg I'm not clear what this module is intended for. Could you please provide more info?
I'd also ask for some feedback from @OCA/web-maintainers.

simahawk:
Not sure it's still the case for v12 but here's the historical issue on v11: https://github.com/odoo/odoo/issues/24289

simahawk:
regarding the name and the position, it's probably more suitable for OCA/server-auth as `auth_public_session` or something similar.

fredzamoabg:
> regarding the name and the position, it's probably more suitable for OCA/server-auth as `auth_public_session` or something similar.

moved to https://github.com/OCA/server-auth/pull/313.

PR #2087, opened 2021-11-19
[FIX] web_m2x_options: Expose objects for inheritance

### LINKED ISSUE #2086: web_m2x_options > form.js must return Objects otherwhise no more inherits are allowed.
If `odoo.define('web_m2x_options.web_m2x_options', function (require)` does not return the inherited Objects no more inheritance are allowed

## Module
[]12.0 web_m2x_options

## Describe the bug
If `odoo.define('web_m2x_options.web_m2x_options', function (require)` does not return the inherited Objects no more inheritance are allowed

## To Reproduce
Try to inherit one of the field function of "FieldMany2ManyTags", "FieldMany2One" after installing web_m2x_options module.

cscocca:
Fixes #2086

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -373,4 +373,13 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
             }
         },
     });
+
+    return{
+        M2ODialog: M2ODialog,
+        FieldMany2One: FieldMany2One,
+        FieldMany2ManyTags: FieldMany2ManyTags,
+        FieldOne2Many: FieldOne2Many,
+        FormFieldMany2ManyTags: FormFieldMany2ManyTags,
+        FormView: FormView,
+    };
 });


legalsylvain (approved):
Thanks.

cscocca:
Thank you all for the module!

web_m2x_options/static/src/js/form.js
```diff
@@ -373,4 +373,13 @@ odoo.define('web_m2x_options.web_m2x_options', function (require) {
...
     });
+
+    return{
+        M2ODialog: M2ODialog,
+        FieldMany2One: FieldMany2One,
+        FieldMany2ManyTags: FieldMany2ManyTags,
+        FieldOne2Many: FieldOne2Many,
+        FormFieldMany2ManyTags: FormFieldMany2ManyTags,
+        FormView: FormView,
+    };
```
Tardo, line 384: :fire: "include" doesn't need to be explicitly exported.
```suggestion
    return M2ODialog;
```

Tardo (changes requested):
Thanks!

web_m2x_options/static/src/js/form.js
cscocca: I tried to not export "include" objects... but inherited element does not work. 
It's only work if I export them. Maybe I made anything wrong?

Tardo: The require system will wait for the dependencies to be satisfied before invoking the code. So, what you should do is to request the module that manipulates the object first and then request the module that gives you the object reference. This will also depend on the order in which your code is loaded, that's why the module that modifies the object is requested, to make sure that it has been loaded.
In code:
```javascript
odoo.define("my_module", function(require) {
    "use strict";

    require("web_m2x_options.web_m2x_options");
    const FieldMany2One = require('web.relational_fields').FieldMany2One;
    
    FieldMany2One.include({
         _onInputFocusout: function () {
              console.log("Hey!");
         },
    });
});
```

cscocca: Ok I will try your solution monday. 
Thanks

cscocca: Hi, I Just tested your suggestion and it work fine. But It's also true if "include" Objects are returned no more require is needed. 
`odoo.define("my_module", function(require) {
    "use strict";

    const FieldMany2One = require("web_m2x_options.web_m2x_options");

//  Returning "Include" Object this line not needed anymore. Just one require is enough.
//    const FieldMany2One = require('web.relational_fields').FieldMany2One;
    
    FieldMany2One.include({
         _onInputFocusout: function () {
              console.log("Hey!");
         },
    });
});`

Both Solutions works.. I wait for answer

PR #2088, opened 2021-11-19
[13.0][FIX] web_pwa_oca: Guess properly mimetype

pedrobaeza:
It seems since https://github.com/odoo/odoo/commit/3bd0327d6d1734736986bef543349a69af3e2414 that the mimetype detection is not working properly.

Let's resave SVG for having a compatible one + add a pointer when the mimetype doesn't conform.

@Tecnativa


pedrobaeza:
Green, so merging directly, as GitHub is a bit unstable.

PR #2066, opened 2021-10-30
[14.0] [IMP] web_drop_target

etobella:
Apply improvements from 13.0

PR #2090, opened 2021-11-22
[FIX] web_drop_target: Hide it when it is not the current action or has no chatter

etobella:
Fixes https://github.com/OCA/web/pull/2066#issuecomment-955672542 on 13.0

### DIFF
--- a/web_drop_target/static/src/js/web_drop_target.js
+++ b/web_drop_target/static/src/js/web_drop_target.js
@@ -5,7 +5,7 @@
 
 odoo.define("web_drop_target", function(require) {
     "use strict";
-
+    const ActionManager = require("web.ActionManager");
     const FormController = require("web.FormController");
     const core = require("web.core");
     const qweb = core.qweb;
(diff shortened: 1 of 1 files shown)


web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -140,7 +140,15 @@ odoo.define("web_drop_target", function(require) {
...
         _onBodyFileDragover: function(ev) {
             ev.preventDefault();
-            if (_.isEmpty(this._get_drop_items(ev))) {
+            const actionManager = this.findAncestor(function(ancestor) {
+                return ancestor instanceof ActionManager;
+            });
+            const controller = actionManager.currentDialogController;
+            if (
+                _.isEmpty(this._get_drop_items(ev)) &&
+                this.renderer.chatter &&
```
Tardo: Won't this make it not work if you don't have a chatter? for example, DMS.

etobella: Ouch!!! Thanks, I think I have a better idea now :wink:

victoralmau:
Can you review it again @Tardo? When this is ready we can unlock other PRs that depend on it. Thank you!

victoralmau:
@etobella can you review it according to similar changes https://github.com/OCA/web/pull/2066 in v14?

etobella:
@victoralmau it is the same commit we find in https://github.com/OCA/web/pull/2066

The other commits are already present on commit history of web_drop_target https://github.com/OCA/web/commits/13.0/web_drop_target

victoralmau (approved):

Tardo (approved):

PR #2098, opened 2021-11-29
[IMP/FIX] web_widget_domain_editor_dialog: work in dialog

chienandalu:
- IMP: Now it's possible to work with the domain editor when the widget
has de in_dialog option.
- FIX: When a boolean opertator was present in the domain, the editor dialog
would raise an error as the default filter couldn't be interpretated.

cc @Tecnativa TT32827

please review @Tardo


pedrobaeza (approved):
It seems all the content of this dialog is not translated to the user language, probably due to a missing context propagation, but I think this is not related to this PR

CarlosRoca13 (approved):

MiquelRForgeFlow:
Please, forward port to v14 :)

PR #2106, opened 2021-12-01
[11.0] allow multi selection on one2may fields too

Guenzn:
Currently the multi selection is limited to many2many fields, but also on one2many fields a multiselection is usefull

### DIFF
--- a/web_widget_many2many_tags_multi_selection/README.rst
+++ b/web_widget_many2many_tags_multi_selection/README.rst
@@ -38,6 +38,7 @@ Contributors
 * Sylvain Calador <sylvain.calador@akretion.com>
 * Jamin Shah <jaiminshah2009@gmail.com>
 * Maxence Groine <mgroine@fiefmanage.ch>
+* Guenter Selbert <guenter.selbert@sewisoft.de>
 
 Do not contact contributors directly about support or help with technical issues.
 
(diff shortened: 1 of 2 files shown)


elvise:
@Guenzn can you rebase ?

Guenzn:
> @Guenzn can you rebase ?

done

elvise:
@Guenzn i trying to test it but from my perspective it doesn’t work. 
Can you clarify how can use it ?

Guenzn:
> @Guenzn i trying to test it but from my perspective it doesn’t work. Can you clarify how can use it ?

click at many2many/one2many dropdown on "show more" and select all checkboxes you want to add.
e.g. my use case was from custom module where i have to select a set of accounts at a transaction types form view for intrastat declaration. I'm currently don't know a use case at odoo core but if you would change the widget of partners child_ids to many2many_tags you could reproduce that. same would be a product category

elvise:
@Guenzn thanks for your reply! 
I testing it from sale order, click on sale order line (o2m), click on search more (field product), Get the popup but there aren’t check box there 

IMG_1138

Guenzn:
No, product_id at salorderline is many2one and not one2many. There it doesn‘t make sense to choose more than one object.

Guenzn:
No, product_id at salorderline is many2one and not one2many. There it doesn‘t make sense to choose more than one object.

PR #2117, opened 2021-12-23
[15.0][MIG] web_widget_url_advanced

JasminSForgeFlow:
Standard Migration
@ForgeFlow


web_widget_url_advanced/static/src/js/url_widget.js
```diff
@@ -0,0 +1,49 @@
...
+         * @override
+         * @private
+         */
+        _renderReadonly: function () {
+            // Base widget uses `this.attrs.text` instead of `this.value` when available.
+            this.attrs.text = this._get_text();
+            this._super.apply(this, arguments);
+            var prefix = this.attrs.prefix_name || this.attrs.options.prefix_name;
+            if (prefix) {
+                this.$el.html(
```
simahawk, line 44: is this double call to `.html` correct?

web_widget_url_advanced/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2018 Simone Orsi - Camptocamp SA
+# License LGPLv3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
+{
+    "name": "Web URL widget advanced",
+    "summary": "This module extends URL widget "
+    "for displaying anchors with custom labels.",
+    "category": "Web",
+    "version": "15.0.1.0.1",
```
simahawk:
```suggestion
    "version": "15.0.1.0.0",
```

simahawk (changes requested):

web_widget_url_advanced/static/src/js/url_widget.js
JasminSForgeFlow: Yes, it is.

inner html is for replace value and return html data. while outer will set that replaced data

simahawk: sorry but I don't get your point. The `attr` function changes an attribute, which is what we need here.
Hence, I don't get why you want to:

1. load the full html w/ a call to `.html()`
2. parse it
3. set the attr in the same way
4. call again `html()` to replace the full element

This sounds much more expensive than before and for no good reason apparently.
Of course, I might be wrong... can you elaborate a bit please? :innocent:

JasminSForgeFlow: Hi @simahawk,

Actually, in Odoo 15 things changed, so only replacing attrs won't work.
only replace attrs will replace in string and return new string but won't affect in actual content.
So in order to update actual content need to replace the whole HTML.

Thanks

simahawk: Ok, thanks for the explanation.

AFAIS what changed is that `this.$el` is not the link anymore. The [link is a child element](https://github.com/odoo/odoo/blob/15301a1ddaf00e51688d8134b7cb19baf386577c/addons/web/static/src/legacy/js/fields/basic_fields.js#L1849).

Hence, instead of calling `this.$el.attr` you should probably call it on the inner elem.
Something like: `this.$el.find("a").attr(...)`.
Have you tried this already?

JasminSForgeFlow: Hi,
Yes, I had tried that, it wasn't working.

Thanks

LoisRForgeFlow:
@simahawk could you update your review?

web_widget_url_advanced/static/src/js/url_widget.js
simahawk: ok, I surrender :smile: 
Can we at least add a TODO to find a better way? Might imply to propose a change to Odoo core too... :pray:

simahawk (approved):
you can squash the last commit while adding the TODO :wink:

JasminSForgeFlow:
> you can squash the last commit while adding the TODO wink

Hi  @simahawk. 

Added TODO message and squash last commit.

Thanks

LoisRForgeFlow (approved):
LGTM

PR #2121, opened 2022-01-07
[MIG] web_view_calendar_list: Migration to 14.0

olgamarcocb:
HELP: I do not understand why "CalendarRenderer.extend" is acting as an include.


etobella (changes requested):
On Runboat I created a Calendar List for Maintenance Request with the following format:

```
<calendar_list date_start="schedule_date" date_delay="duration" color="user_id" event_limit="5">
                <field name="user_id" filters="1"/>
                <field name="priority"/>
                <field name="maintenance_type"/>
            </calendar_list>
```
I was unable to test it (it seems to happen the same error on the standard calendar view)

olgamarcocb:
I do not understand why "CalendarRenderer.extend" is acting as an include. Any thoughts?

olgamarcocb:
Include fixed with the change of @Martronic-SA Thanks!

etobella (approved):
Tested locally and it works

albariera (approved):
LGTM :+1:

PR #2125, opened 2022-01-13
web_notify: bump dev status

sbidoul:
It exists since many version and is actively maintained.


sbidoul:
I suppose nobody will object to this.

/ocabot merge patch

PR #1777, opened 2021-01-08
[IMP] web_tree_dynamic_colored_field: New background color property

ypapouin:
I had to customize the cell background color.
This PR add a background color property customizable using a field.

image

PR #2130, opened 2022-01-19
[12.0][FIX] web_tree_dynamic_colored_field: all colors attributes must be parsed to work properly

ypapouin:
The current implementation merged from https://github.com/OCA/web/pull/1950 does not browse all possible values of the `colors` attribute: You cannot set a background color AND a font color at the same time.
This PR fix this behavior.

It also add a demo:
image

Note that this fix is based on my initial https://github.com/OCA/web/pull/1777

@legalsylvain, @hbrunn, can you please review this PR ?


legalsylvain (approved):
Makes sense. never used both indeed. 
thanks ! 
(code review / no test).

PR #2138, opened 2022-02-02
[FIX] 14.0 tree many2one clickable context eval

PierrickBrun:
On some fields there is a context that can make the action fail and display an error like this one:

Capture d’écran de 2022-02-02 12-48-18

For instance, this field cannot be opened with the module https://github.com/odoo/odoo/blob/14.0/addons/purchase/views/purchase_views.xml#L233 

This is a temporary fix as it would be better to evaluate the context before passing it, but at least there is no error displayed to the user.


kevinkhao:
duplicate : https://github.com/OCA/web/pull/2152/files

PR #2139, opened 2022-02-03
[15.0] [MIG] web_drop_target

JasminSForgeFlow:
Standard Migration
@ForgeFlow


theangryangel:
Confirm functionally working on an out of box 15.0 Enterprise install (was literally about to start looking at what needed to be done to port this, so thank you <3)

victoralmau (changes requested):
First of all thanks for the work.
Some details that I think should be resolved:
- When trying to drag a file, it allows you to do it to an area that in my opinion should not appear.
web_drag_drop-1

- There is no area available to drag files in the modal window of a message (I seem to remember that in v14 this worked fine).
web_drag_drop-2

JasminSForgeFlow:
> First of all thanks for the work. Some details that I think should be resolved:
> 
> * When trying to drag a file, it allows you to do it to an area that in my opinion should not appear.
>   web_drag_drop-1
> * There is no area available to drag files in the modal window of a message (I seem to remember that in v14 this worked fine).
>   web_drag_drop-2

Hi @victoralmau,

I have checked in v13 and v14 for point 2 - drop file in email popup. There is an option come to drop a file, but when you drop the file it doesn't attach there, seems it doesn't affect anywhere.

Thanks

etobella (approved):
It worked fine for me on a local installation, so I think we can proceed with merge @victoralmau  WDYT?

victoralmau (approved):

PR #2142, opened 2022-02-07
[14.0][FIX] web_widget_ckeditor: `ir.config_paramter` permissions and typo

ivantodorovich:
JS fails if user doesn't have admin rights, as `ir.config_parameter` can't be read otherwise.

Also, a typo ~`console.warning`~ `console.warn`


legalsylvain (approved):
makes sense. thanks !

PR #2143, opened 2022-02-09
[10.0][IMP] web_export_view: support for buttons

AaronHForgeFlow:
* sometimes button appear in a list view depending on a condition,
  when that happens it is useful for users. E.g. The button is an
  icon that opens the invoices directly form the sale order list
  and it only shows if it has invoices, then is useful to see
  what orders has invoices.

cc @ForgeFlow


JoanMForgeFlow (approved):
Code seems okey!
Looks good

PR #2146, opened 2022-02-16
[12.0][ADD] web_widget_mail_list_activity: Backport from odoo 14

ypapouin:
Add `widget="list_activity"` introduced in Odoo 14.0

sshot-443


flotho (changes requested):
amazing backport!
Yet I have some issue with runboat https://runboat.odoo-community.org/api/v1/builds/b08b70748-f752-435b-90d4-d70c3681845b/log
could you check please

ypapouin:
@sbidoul, can you take a look on this runboat issue ?

sbidoul:
The init log (https://runboat.odoo-community.org/api/v1/builds/b08b70748-f752-435b-90d4-d70c3681845b/init-log) shows the initialization did not complete correctly. But there is no error message... weird.

ypapouin:
@sbidoul , after « manual bisection » I found that the issue comes from `web_widget_mpld3_chart` module and its `external_dependencies` to `mpld3`.
I don't understand why the 0.5.7 version is used whereas `mpld3==0.3` is set in `requirements.txt`
```
Collecting mpld3
  Downloading mpld3-0.5.7-py3-none-any.whl (201 kB)
```

sbidoul:
@ypapouin thanks for the investigation. 

runboat (and our GitHub actions CI) do not use requirements.txt but per addons dependencies.
I attempted to pin mpld3 in #2171 but this does not seem to resolve the issue.

PR #2157, opened 2022-03-02
[13.0][FIX] web_company_color: Recalculated color when we do not want

CarlosRoca13:
If we change something with base.document.layout wizard, this will launch the colors computing even without changing the logo. This procedure will change the value of the colors if they were set by hand.

This change forces the user to press a button to recompute the colors and avoid possible undesired changes.

cc @Tecnativa TT34902

please @pedrobaeza @Tardo review this

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -74,33 +74,35 @@ class ResCompany(models.Model):
                 "color_navbar_bg_hover",
                 "color_navbar_text",
             )
-            if "logo" in values:
-                if values["logo"]:
-                    _r, _g, _b = image_to_rgb(convert_to_image(values["logo"]))
-                    # Make color 10% darker
-                    _h, _l, _s = rgb_to_hls(_r, _g, _b)
-                    _l = max(0, _l - 0.1)
-                    _rd, _gd, _bd = hls_to_rgb(_h, _l, _s)
-                    # Calc. optimal text color (b/w)
-                    # Grayscale human vision perception (Rec. 709 values)
-                    _a = 1 - (0.2126 * _r + 0.7152 * _g + 0.0722 * _b)
-                    values.update(
-                        {
-                            "color_navbar_bg": n_rgb_to_hex(_r, _g, _b),
-                            "color_navbar_bg_hover": n_rgb_to_hex(_rd, _gd, _bd),
-                            "color_navbar_text": "#000" if _a < 0.5 else "#fff",
-                        }
-                    )
-                else:
-                    values.update(self.default_get(fields_to_check))
-
             result = super().write(values)
-
             if any([field in values for field in fields_to_check]):
                 self.scss_create_or_update_attachment()
         else:
             result = super().write(values)
         return result
+
+    def button_compute_color(self):
+        self.ensure_one()
+        values = self.default_get(
+            ["color_navbar_bg", "color_navbar_bg_hover", "color_navbar_text"]
+        )
+        if self.logo:
+            _r, _g, _b = image_to_rgb(convert_to_image(self.logo))
+            # Make color 10% darker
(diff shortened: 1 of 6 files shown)


Tardo (commented):
Looks good.. but for some reason its not working on runboat... Can you confirm that it is something related to runboat?

web_company_color/models/res_company.py
```diff
@@ -74,34 +74,35 @@ def write(self, values):
...
             result = super().write(values)
-
             if any([field in values for field in fields_to_check]):
                 self.scss_create_or_update_attachment()
         else:
             result = super().write(values)
         return result
 
+    def button_compute_color(self):
+        values = self.default_get(
```
pedrobaeza, line 86:
```suggestion
        self.ensure_one()
        values = self.default_get(
```

web_company_color/readme/CONTRIBUTORS.rst
```diff
@@ -2,3 +2,7 @@
 * Lois Rilo <lois.rilo@forgefloww.com> (ForgeFlow)
 * Simone Orsi <simone.orsi@camptocamp.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza: Group this one as well

pedrobaeza:
> Looks good.. but for some reason its not working on runboat... Can you confirm that it is something related to runboat?

Can you be more specific? What is not working?

Tardo:
> > Looks good.. but for some reason its not working on runboat... Can you confirm that it is something related to runboat?
> 
> Can you be more specific? What is not working?

The color fields update but when I refresh the page I don't see those colors applied to the navbar. In runboat 14.0 I see it updates, but in this PR and 13.0 runboat seems not to update the css.

web_company_color/models/res_company.py
```diff
@@ -74,34 +74,35 @@ def write(self, values):
...
-                            "color_navbar_bg_hover": n_rgb_to_hex(_rd, _gd, _bd),
-                            "color_navbar_text": "#000" if _a < 0.5 else "#fff",
-                        }
-                    )
-                else:
-                    values.update(self.default_get(fields_to_check))
-
             result = super().write(values)
-
             if any([field in values for field in fields_to_check]):
```
pedrobaeza, line 78: These 2 lines should be moved to the new method.

pedrobaeza (changes requested):
Problem seen

web_company_color/models/res_company.py
CarlosRoca13: I don't think that's the problem, locally is working well, btw the new method is launching a write with the values computed

Could be that the context on runboat has the key ignore_company_color?
https://github.com/OCA/web/pull/2157/files#diff-9b4dca5f4bdbc5450fc14eeba219185f4894189e94d026978af5a65841c5f85eR71

pedrobaeza: Yes, you are right that it's called anyway, as after the write method is there.

pedrobaeza: I don't see the problem right now.

CarlosRoca13: I test it on runboat and it's working

@Tardo, did you test it with the correct company?

CarlosRoca13:
Other changes done :+1:

pedrobaeza:
I don't understand what is happening:

Selección_041

CarlosRoca13:
> I don't understand what is happening:
> 
> Selección_041

That was a test done by me :sweat_smile: 

Try it again please

CarlosRoca13:
BTW you are changing it on "My Company (Chicago)" but you are at "My Company (San Francisco)"

pedrobaeza (approved):

pedrobaeza:
Oh, true. Merging as working:

/ocabot merge minor

Please fw-port it to v14

PR #2158, opened 2022-03-03
[14.0][FIX] web_company_color: Recalculated color when we do not want

CarlosRoca13:
If we change something with base.document.layout wizard, this will launch the colors computing even without changing the logo. This procedure will change the value of the colors if they were set by hand.

This change forces the user to press a button to recompute the colors and avoid possible undesired changes.

cc @Tecnativa TT34902

fw-port of https://github.com/OCA/web/pull/2157

ping @pedrobaeza

### DIFF
--- a/web_company_color/i18n/es.po
+++ b/web_company_color/i18n/es.po
@@ -4,84 +4,72 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 14.0\n"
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as already merged in previous version:

/ocabot merge minor

PR #2163, opened 2022-03-10
[IMP] web_tree_image_tooltip : add the possibility to define an alternative image field 

legalsylvain:
add the possibility to define an alternative image field  for the tooltip image to reduce the bandwidth used.

## Rational

when adding ``<field name="image" widget="image" />`` in the tree view, it will increase a lot the size of the ``search_read`` call result.
If we add ``<field name="image_small" widget="image" />`` in the tree view, the load will be faster, but the image displayed in the tooltip will be with a low quality.

With that PR, we allow a new syntax ``options="{'tooltip_image': 'image_field'}"`` to display another image in the tooltip.

### 1. display tree view of product.product with demo data 

image

Size : 8.28 ko

### 2. Add ``<field name="image" widget="image" />`` 

image

Size : 1.24 Mo (x150)

### 3. add ``<field name="image_small" widget="image" options="{'tooltip_image': 'image'}"/>``

(allowed by this PR) 

image

Size : 73.57 ko (x9)

CC : @nikul-serpentcs (original author in V11 / V12), @phuctranfxvn (migration to 13), @newtratip (migration to 14)

Thanks for your review.


ChrisOForgeFlow (approved):
LGTM

AaronHForgeFlow (approved):

PR #2110, opened 2021-12-13
[15.0][MIG] web_widget_x2many_2d_matrix

ChrisOForgeFlow:
Standard Migration
@ForgeFlow

PR #2164, opened 2022-03-11
[FIX] web_widget_x2many_2d_matrix: make the possibility to have alternative values for headers 

legalsylvain:
In the current implementation, ``field_label_y_axis`` and ``field_label_x_axis`` can be defined, but there are not used in the javascript algorithm, and so defining those values doesn't work.
this patch fixes the problem.

### DIFF
--- a/web_widget_x2many_2d_matrix_example/models/res_users.py
+++ b/web_widget_x2many_2d_matrix_example/models/res_users.py
@@ -1,0 +1,13 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models, api, fields
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    matrix_display_name = fields.Char(compute="_compute_matrix_display_name")
+
+    @api.depends("name", "email")
+    def _compute_matrix_display_name(self):
+        for user in self:
+            user.matrix_display_name = "%s (%s)" % (user.name, user.email)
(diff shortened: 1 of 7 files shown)


web_widget_x2many_2d_matrix_example/models/res_users.py
```diff
@@ -0,0 +1,13 @@
...
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    matrix_display_name = fields.Char(compute="_compute_matrix_display_name")
+
+    @api.depends("name", "email")
+    def _compute_matrix_display_name(self):
+        for user in self:
+            user.matrix_display_name = "%s (%s)" % (user.name, user.email)
```
StefanRijnhart: web_widget_x2many_2d_matrix_example/models/res_users.py:13:75: W292 no newline at end of file

legalsylvain:
Hi 
@StefanRijnhart : thanks for your review. I fixed pre-commit stuff. All CI is now green.

@ChrisOForgeFlow.
- Could you take a look on this one ? 
- as you ported [here](https://github.com/OCA/web/pull/2110) the module in V15, could you say me if the bug is present in V15 and the bug fix should be ported ?

General note : this bugfix is working in production since 2 monthes without problem.

StefanRijnhart (approved):
I see the usage of this mechanism was lost in the migration to 11.0: https://github.com/OCA/web/commit/69cc921ab36e4dee42c68d749f3ce259bc5b5e5f

ChrisOForgeFlow (approved):
I see that in the 15 it continues with that bug, it will be necessary to do the fwd of this, the change seems good to me

PR #2171, opened 2022-03-23
[12.0] web_widget_mpld3_chart: fix mpld3 dependency

sbidoul:
Coming from https://github.com/OCA/web/pull/2146#issuecomment-1075182792

This is an attempt to fix an installation issue on runboat.
The reason for the pin is unknown at the moment, unfortunately.

@ypapouin

### DIFF
--- a/setup/web_widget_mpld3_chart/setup.py
+++ b/setup/web_widget_mpld3_chart/setup.py
@@ -2,5 +2,11 @@ import setuptools
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        "external_dependencies_override": {
+            "python": {
+                "mpld3": "mpld3==0.3",
+            },
+        },
+    },
 )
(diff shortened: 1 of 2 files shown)


ypapouin:
Does the runboat init log capture all stdout and stderr ?

sbidoul:
> Does the runboat init log capture all stdout and stderr ?

It should, as it's [basically](https://github.com/sbidoul/runboat/blob/2a8450b462c71bb9a40d390cee7dfc27b4a392bb/src/runboat/k8s.py#L281-L286) `kubectl logs`.

ypapouin:
Is there a way to blacklist this module from runboat in the 12.0 branch to disable its automatic install ?
This way we could go forward as all current PR targeted to 12.0 are not testable.

sbidoul:
@ypapouin there is no easy way to disable a module on runboat. Modules are supposed to at least install properly...
But if it fails on runboat it should fail on GitHub actions too, so let's try this: #2207

PR #2172, opened 2022-03-23
[15.0][MIG] web_dialog_size: migration

Mantux11:
Migration to `15.0` with some changes because of `OWL`. If you have any better ideas or suggestions in JS code, it will be appreciated.


web_dialog_size/static/src/js/web_dialog_size_owl.esm.js
```diff
@@ -0,0 +1,62 @@
+/** @odoo-module **/
```
legalsylvain: Hi. Why do you put ``esm`` in the name of the file ? 
also, I think that putting ``owl`` is unnecessary. Don' you think ?

pedrobaeza: `esm` is for being friendly with OWL linting. Please check the reference at https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

Mantux11: @legalsylvain Talking about esm files, @pedrobaeza explained here, thanks. And without that, pre-commit is also returning errors, so it seems because of lintering. 

For another your question, did you mean why file name is having owl tag in name or why I used owl in here?

legalsylvain: Thanks for the esm explanation.

Regarding owl, my question was : why put owl in the name of the file.

Mantux11: I put it because there is two files: one for older code, and for owl, it's like to separate where is old code, and where is the new. It's not necessary, I could rename without owl.

baimont (approved):
functional test

theangryangel (approved):
Functionally works well.

The only thing I've spotted that I'm not sure if it's a bug or intentional is clicking the expand/contract triggers an onchange and read calls, causing a small delay. 

At least in the 12.0 version this didn't happen, so I'd guess it's owl doing something on the click? 

But I'll be upfront, I hadn't spent too long looking at this.

yajo:
/ocabot merge patch

Thanks!

@moduon MT-1187

PR #2182, opened 2022-04-03
[12.0][ADD] web_widget_uom_view_precision

yibudak:
This module provides a widget for quantity fields. It formats the value according to the UOM of the product.

#1279

image


web_widget_uom_view_precision/models/uom_view_precision_backend.py
```diff
@@ -0,0 +1,19 @@
+from odoo import api, models
+
+
+class UomViewPrecisionBackend(models.AbstractModel):
+
+    _name = 'uom.view.precision.backend'
+    _description = 'Uom View Precision Backend'
+
+    @api.model
+    def get_uom_view_precisions(self):
```
legalsylvain, line 10: I don't see the interest of this abstract class. why not just move this function in the ``uom_uom.py`` file ?

web_widget_uom_view_precision/readme/USAGE.rst
```diff
@@ -0,0 +1,22 @@
...
+        <form string="View name">
+            ...
+            <field name="quantity_field" widget="uom"/>
+            ...
+        </form>
+    </field>
+    ...
+
+
+You can also specify the uom field with::
```
legalsylvain: could you specify what is the default, if no value is specified in ``uom_field`` ?

legalsylvain (commented):
great module. Thanks ! 
some minor remarks in line. otherwise, LGTM. (code review / no test).

web_widget_uom_view_precision/readme/USAGE.rst
yibudak: the default is "uom_id", i clarified the description.

bosd:
@yibudak What is the status of this one?

yibudak:
> @yibudak What is the status of this one?

Well, this module works flawlessly but the thing is you need to add attributes on every fields. Example:

`                <field name="qty_producing" position="attributes">
                    <attribute name="widget">uom</attribute>
                    <attribute name="options">{'uom_field': 'product_uom_id'}</attribute>
                </field>`

Also I've sent a CLA document but still no response after months. Am I doing something wrong?

bosd:
Oops, you've included a bit too much files in the latest commits.
Maybe better to undo those?

bosd:
Can you fix pre-commit?

PR #2190, opened 2022-04-13
[13.0][IMP] web_widget_one2many_tree_line_duplicate: Clone all fields

Tardo:
cc @tecnativa TT35922

### DIFF
--- a/web_widget_one2many_tree_line_duplicate/static/src/js/basic_model.js
+++ b/web_widget_one2many_tree_line_duplicate/static/src/js/basic_model.js
@@ -5,6 +5,7 @@ odoo.define("web_widget_one2many_tree_line_duplicate.BasicModel", function(requi
     "use strict";
 
     const BasicModel = require("web.BasicModel");
+    const rpc = require("web.rpc");
 
     function dateToServer(date) {
         return date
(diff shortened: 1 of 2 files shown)


web_widget_one2many_tree_line_duplicate/static/src/js/basic_model.js
```diff
@@ -55,62 +57,79 @@ odoo.define("web_widget_one2many_tree_line_duplicate.BasicModel", function(requi
...
 
-            if (
-                command.position === "bottom" &&
-                list.orderedResIDs &&
-                list.orderedResIDs.length >= list.limit
-            ) {
-                list.tempLimitIncrement = (list.tempLimitIncrement || 0) + 1;
-                list.limit += 1;
+            let read_data = Promise.resolve();
+            if (!this.isNew(command.id)) {
```
pedrobaeza: This makes it fail when you click on the clone button on a new line not yet saved, and I don't understand why can't you get the information from JS side. The problem is that the previous code doesn't fetch all data that you see in the UI. It's true that if the record is saved, you can first do this, and the update with UI existing information (it may be changed meanwhile).

pedrobaeza (changes requested):

web_widget_one2many_tree_line_duplicate/static/src/js/basic_model.js
Tardo: Fixed. The client does not know the data of the other views until the views are loaded. If a record is cloned from which the form has not been opened, those fields will not be copied, that is why it is necessary to call "copy_data", as well as to obtain the rest of the fields that are not in a view.
Changes have also been applied to copy the fields of all the loaded views and not only of the "list" type as until now. These values have more weight than those obtained by "copy_data".

pedrobaeza (approved):

pedrobaeza:
Fast-tracking as it was is not functional:

/ocabot merge patch

PR #2192, opened 2022-04-15
[14.0][IMP] web_ir_actions_act_view_reload: add model ir_actions

nguyenhk99:


leemannd (approved):

web_ir_actions_act_view_reload/models/ir_actions.py
```diff
@@ -0,0 +1,17 @@
+from odoo import fields, models
+
+
+class IrActionsActWindowClose(models.Model):
+    _name = "ir.actions.act_view_reload"
```
grindtildeath: You should use same name for class and model.

grindtildeath (commented):
@HoKhanhNguyen99 Can you please explain a bit why this is needed?

StephaneMangin (approved):

web_ir_actions_act_view_reload/models/ir_actions.py
StefanRijnhart: This shows up as an `Outdated` diff, but the class is still called `IrActionsActWindowClose` and the model name is still `ir.actions.act_view_reload`. I would have expected to see them aligned at this point.

PR #2198, opened 2022-04-25
Update from template

etobella:
From Travis to Github Actions


etobella:
Due to some problems with chrome-headless, I had to split the tests for web_widget_model_viewer as it was affecting other modules.

@OCA/web-maintainers

PR #2206, opened 2022-05-01
[14.0][ADD] web_widget_x2many_2d_matrix, no input as string

kittiu:
Do not allow character in the input field.

no_char

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js
@@ -22,6 +22,10 @@ odoo.define("web_widget_x2many_2d_matrix.X2Many2dMatrixRenderer", function (requ
     // X2Many2dMatrixRenderer is heavily inspired by Odoo's ListRenderer
     // and is reusing portions of code from list_renderer.js
     var X2Many2dMatrixRenderer = BasicRenderer.extend({
+        events: {
+            "keyup input": "_onKeyUp",
+        },
+
         /**
          * @override
          */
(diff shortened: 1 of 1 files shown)


kittiu:
@OCA/web-maintainers 
Because sometime user may accidentally include string and when saved it just disappear without warning.
But my JS is very limited, not sure this is the correct way to work with, please feel free to suggest. Thanks!

tarteo (approved):
Thanks, LGTM :+1:

pedrobaeza:
Not sure this limitation is OK, as you value field may be of any type. I think you should use the widget of the field type (or even allow to customize it).

legalsylvain:
> Not sure this limitation is OK, as you value field may be of any type. I think you should use the widget of the field type (or even allow to customize it).

Indeed, AFAIK, you can "matrix" any fields, including "char".

>  when saved it just disappear without warning.

Well, maybe that point should be improved ? adding a blocking message if data is not correct.

hbrunn (changes requested):
you'll need to check if the field type is one of the numerical ones, as this widget also supports char fields.

I also think the check needs to be more complex, because any non-US way of formatting numbers won't pass through `isNaN`.

In the end, I think this is a job for https://github.com/OCA/OCB/blob/14.0/addons/web/static/src/js/fields/abstract_field.js#L305 where you can check if all your input is valid, and set error classes otherwise

pedrobaeza:
I don't have too much knowledge in JS, but populating the field widget isn't possible and you don't have to worry about these things?

tarteo:
If all field types are supproted it should be done by the field class (of the specific field type)

kittiu:
Hello Thanks to all pointers. In fact, I also not satisfied by what I did (so far it is ok only on my project).
I would love to make it on the field widget level, just that my JS is very poor. With the given pointer I will try again.

May be, I will check if it is possible to add field type as parameter of matrix widget. Thanks again.

PR #2207, opened 2022-05-04
[12.0] apply OCA repo template and use GitHub actions

sbidoul:
Next try following https://github.com/OCA/web/pull/2171


sbidoul:
So this is green (GitHub actions tests and runboat).

The difference is that we don't pin mpld3, which was probably the culprit. Now I don't know if the pin was important for that addon, but at least this unlocks the situation.

sbidoul:
@OCA/web-maintainers feel free to merge if this looks ok for you.

legalsylvain (approved):
Thanks !

etobella (approved):

etobella:
Thanks!!!
/ocabot merge patch

ypapouin:
Thank you !

PR #2208, opened 2022-05-05
[13.0][FIX] web_m2x_options: Scroll dance

Tardo:
Fixes an issue that causes the screen scroll to "dance" when selecting an element.

cc @tecnativa TT36313


MiquelRForgeFlow:
Testing.

pedrobaeza (approved):
Let's patch this old one:

/ocabot merge patch

PR #2209, opened 2022-05-06
[14.0][IMP] web_company_color: Forward port #1903 from 12.0

HekkiMelody:
Forward port of https://github.com/OCA/web/pull/1903 
commit e18209c

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -44,6 +44,47 @@ class ResCompany(models.Model):
             }
           }
         }
+        a[href],
+        a[tabindex],
+        .btn-link,
+        .o_external_button {
+          color: %(color_link_text)s !important;
+        }
+        a:hover,
+        .btn-link:hover {
+          color: %(color_link_text_hover)s !important;
+        }
+        .btn-primary:not(.disabled),
+        .ui-autocomplete .ui-menu-item.ui-state-focus a {
+          color: %(color_button_text)s !important;
+          background-color: %(color_button_bg)s !important;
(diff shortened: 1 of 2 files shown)


web_company_color/models/res_company.py
```diff
@@ -73,7 +124,32 @@ def write(self, values):
                 "color_navbar_bg",
                 "color_navbar_bg_hover",
                 "color_navbar_text",
+                "color_button_bg",
+                "color_button_bg_hover",
+                "color_button_text",
+                "color_link_text",
+                "color_link_text_hover",
             )
+            if "logo" in values:
```
StefanRijnhart: You seem to be undoing https://github.com/OCA/web/pull/2158/files. Is that on purpose?

HekkiMelody: Ah, good catch. I didn't really follow this module's history and just ported the commit trying to preserve as much of it as possible.

Thank you :pray: 
Fixing it now.

HekkiMelody:
Added a commit to fix #2248

StefanRijnhart (approved):

StefanRijnhart:
@hbrunn your improvement is being forward ported! Care to review?

hbrunn (approved):

LoisRForgeFlow (commented):
I'm late to this, but I just hit into this change. I see 2 problems:

1. Starting with the field in blank you need to set all fields in order for it to have an effect. (you cannot just set the navbar colors for instance).
2. When computing colors from logo, only the nabvar ones are filled.

PR #2210, opened 2022-05-06
[13.0][FIX] web_dialog_size: resize columns from one2many tree

chienandalu:
For dialogs containing one2many trees we'd expect their columns to resize. We have to trigger the propper upstream method to achieve it.

cc @Tecnativa TT36323

please review @tardo @pedrobaeza


pedrobaeza:
Shouldn't it triggered after changing class?

chienandalu:
> Shouldn't it triggered after changing class?

It didn't make a practical difference as it's a deferred call. Anyway it makes more sense in the code indeed

Tardo (approved):

pedrobaeza (approved):

pedrobaeza:
Should this be fw-port it to v14?

PR #2212, opened 2022-05-09
[14.0][FIX] web_dialog_size: resize columns from one2many tree

chienandalu:
Forward of
- #2210 

cc @Tecnativa TT36323

ping @pedrobaeza @tardo


pedrobaeza:
I'm still having the problem on a v13, so not sure this fix is enough.

pedrobaeza:
In this instance, the start maximized option is set.

cesar-tecnativa (approved):

pedrobaeza:
@chienandalu can you please check my last comment?

PR #2219, opened 2022-05-13
[14.0][MIG] web_widget_many2many_tags_multi_selection

JuanyDForgeflow:
Standard Migration to 14.0

In a many2many_tags widget when a lot of entries should be selected it's fastidious to select 80% of them. Then you may click on 'search more', but impossible to select several attributes at once.

This module adds a checkbox to this list so multiple entries can be selected at once.


pedrobaeza:
This is something native since some time AFAIK.

JuanyDForgeflow:
@pedrobaeza AFAIK you have to select the records one by one

JuanyDForgeflow:
@pedrobaeza I checked and you're right, I'll proceed to close this pr

PR #2201, opened 2022-04-26
[LIC] web_drop_target: Relicense to LGPL

etobella:
Foillowing comment from https://github.com/OCA/dms/pull/170#issuecomment-1108761836
@hbrunn @marcelsavegnago

PR #2221, opened 2022-05-18
[LIC] web_drop_target: Relicense to LGPL

etobella:
#2201 #2202


etobella:
Fast tracking, as it was approved already
/ocabot merge patch

PR #2176, opened 2022-03-31
[MIG][15.0] web_notify

em230418:
Porting notes:
1. Removed notify_default, since it does not pass validation in [1]
2. Modifications to notificaation widgets are removed in prefer of original widget implementation
3. Odoo improved notification usage in [2], so no need to add extra channels that depend on user's id
    
[1] https://github.com/odoo/odoo/blob/21be7ca0c9eef18184d06755ff83337be1e4752e/addons/web/static/src/core/notifications/notification.js#L49
[2] https://github.com/odoo/odoo/commit/543af27c7d6836ffac9e80ff8490b6ddbd849221

PR #2224, opened 2022-06-07
[MIG] web_notify in v15

stvn-vaidehi:


pedrobaeza:
Thanks for the contribution.

And please also check tests, that are red.

Note that there's another attempt in #2176

PR #2077, opened 2021-11-12
[14.0][IMP] web_responsive: Add chatter position view_users_form

andreampiovesana:
Allow to change chatter position also in base.view_users_form

PR #2229, opened 2022-06-12
[14.0][IMP] web_responsive: Add chatter position view_users_form

dreispt:
Allow to change chatter position in user form view


dreispt:
Supersedes #2077

PR #2230, opened 2022-06-13
[14.0][ADD] web_widget_statusbar_visible_from_record

StefanRijnhart:
This technical module allows you as a developer to determine which states are visible in the `statusbar` widget per record per field.


StefanRijnhart:
~~I'm running into https://github.com/acsone/setuptools-odoo/issues/55 when generating the setup files. Having a closer look.~~ Fixed now.

MrGigSolutions (approved):
A more flexible solution than using the visible_status. I tried it and it works nicely.

PR #2231, opened 2022-06-13
[13.0][IMP] web_pwa_oca: Added css classes to control the visibility of view fields

Tardo:
- Added css classes to control the visibility of view fields in standalone mode
- Improve "pull refresh" feature
- Improve kanban usage in standalone mode


pedrobaeza:
There are incorrect merge commits. It seems you do git pull instead of git rebase

PR #2232, opened 2022-06-16
[15.0][FIX]-web_responsive module: Go to wrong action menu when access from Root Screen Search Bar when user is set with Home Action Menu

hieulucky111:
# How to reproduce

- Install module web_responsive
- Set Home Action for user in the tab Preferences
- Refresh the page
- Type menu on the Search Bar
- Click on the menu we want
- Redirect us to wrong menu


jcdrubay:
There is no link to runboat test available?

jcdrubay:
Tested in our test environment, it works nicely.

jcdrubay (approved):

Tardo (approved):
Thanks!

PR #2245, opened 2022-07-04
Reduce checkbox size form 24px to 12px

mohamedhagag:
Just a UI enhancement as it looks too big for checkbox

image

image

### DIFF
--- a/web_responsive/static/src/legacy/scss/web_responsive.scss
+++ b/web_responsive/static/src/legacy/scss/web_responsive.scss
@@ -432,14 +432,14 @@ html .o_web_client .o_action_manager .o_action {
 
         .custom-control-label {
             &::after {
-                width: 24px;
-                height: 24px;
+                width: 12px;
+                height: 12px;
             }
             &::before {
                 outline: none !important;
                 border: 1px solid #4c4c4c;
-                width: 24px;
-                height: 24px;
+                width: 12px;
+                height: 12px;
             }
         }
     }


pedrobaeza:
That was done on purpose, so we don't want to go back on that. This is friendly with touch devices where our fingers can't mark such small checks. It's just you have to get used to. Closing as won't change.

mohamedhagag:
@pedrobaeza Sure, You are right in your PoV, 

But we can add media for small screens in CSS to show bigger checkbox on mobile and touch devices !

Would you accept such solution ? @pedrobaeza

pedrobaeza:
I don't, as I prefer also these big checkboxes in normal screens. Let's see if others think the contrary.

mohamedhagag:
I was sure you will not, although it's a simple logic 
Best wishes for the community & OCA

yajo:
> But we can add media for small screens in CSS to show bigger checkbox on mobile and touch devices !

This seems a good solution. Would you update the code please?

pedrobaeza:
Our customer feedback is that they want big checkboxes even in desktop. It's difficult to click on such small checkboxes on certain environments.

pedrobaeza (changes requested):
Unless this is configurable by an option, keeping current preference as default, this is not mergeable.

Tardo:
Since having these larger checkboxes doesn't require reserving more space, I don't think it's wrong to have them always large.

yajo:
I agree they are quite ugly though. So it makes sense for somebody to want to customize them. Let's see what the OP says.

pedrobaeza:
What `OP` means? This can't be removed, so if you want it, it should be configurable.

sergio-teruel (changes requested):
I do not agree with this change.
In responsive mode wwith  small checkbox is complicated to click on it

yajo:
> What `OP` means?

Original Poster. @mohamedhagag in this context. I'm just letting him decide if he wants to do it your way, or if he prefers to close the PR.

PR #2251, opened 2022-07-13
[15.0][MIG] web_calendar_slot_duration: Migration to 15.0

CRogos:
- Updated version numbers 
- Moved assets to manifest file

I've tested the code with v15 successfully, but there is a reproducible issue when migrating from v14 to v15.
There is an Server Error 500 which is related to the asset "/web_calendar_slot_duration/static/src/js/calendar_model.js". 
The Odoo Enterprise Support already confirmed, that they have to add an additional migration script to the upgrade process. 
But I wonder why this is necessary and whether there is better way, working for all users?


CRogos:
Still working with the Odoo Enterprise support on the asset upgrade issue, but this also happens on all other 3rd party modules with assets. During the upgrading of the database, the "old" asset is added to the ir_asset table. After an uninstallation of the upgraded module, this entry remains in table ir_asset and causes an Error 500, because the asset file is not available anymore. 
Therefore I think  this is an upgrade process issue at Odoo enterprise and not an issue of this migrated module.

@pedrobaeza do you know of any similar problem regarding migrating assets from v14 to v15?

Another question.  Should these be squashed?
image

pedrobaeza:
Not sure what you mean about the assets. On the image, the commits you mention can be squash with the 14.0 migration commit.

CRogos:
@pedrobaeza thanks for the git history feedback, I will change it, but maybe not before Monday.

I mean the change from xml-file to __manifest__.py and the internal handling in odoo. In odoo there is a new table ir_asset, where these kind of resources are added. 
v15: __manifest__.py
image
v14: assets.xml
image

pedrobaeza:
Yes, in this version you should load the assets through the manifest, but you don't have to worry about how it's done internally. Just change it in your migrated module. Prefix the new asset file with `esm.` for pre-commit compatibility.

CRogos:
OK all changes done. Looking for reviewers!

CRogos:
@hparfr , @yajo, @victoralmau  as veteran of this module, may I ask you for a quick review?

yajo (approved):

victoralmau (approved):

CRogos:
@pedrobaeza could you have a look why it is not finishing?

CRogos:
@pedrobaeza could you have another look, the [Action ](https://github.com/OCA/web/actions) has finished a while ago successful.

PR #2252, opened 2022-07-15
[10.0][MIG] help_popup

AaronHForgeFlow:
Supersedes https://github.com/OCA/web/pull/624/


AaronHForgeFlow:
You may think I have no idea what I am doing, and you will be probably right. But for some reason that module was not working well with web_responsive and now it is seems to be fine. But yes:
noidea

Also, travis config seems need to be fixed :S

PR #2254, opened 2022-07-15
[ADD] web_button_no_save

SirTakobi:


eLBati (approved):
Thanks

SirTakobi:
@OCA/web-maintainers can somebody have a look? For beta modules, one review should be enough. Thanks!

legalsylvain (commented):
Thanks for your contribution.

Question : Could you say use cases in the readme ? (I dont understand in which case this module can be usefull) 

Thanks.

SirTakobi:
> Thanks for your contribution.
> 
> Question : Could you say use cases in the readme ? (I dont understand in which case this module can be usefull)
> 
> Thanks.

Thank you for having a look :smile:, I have added a use-case.
A bit of history: the save-record-on-button-click is an unexpected behavior for new developers and users of Odoo, there are some questions about this in the forum [^1], in StackOverflow [^2] and also among Odoo issues [^3].

[^1]: https://www.odoo.com/forum/help-1/prevent-save-model-when-clicking-on-a-button-that-raises-a-wizard-138815
        https://www.odoo.com/forum/help-1/disable-autosave-record-on-button-click-32953
[^2]: https://stackoverflow.com/questions/31963214/odoo-prevent-button-from-closing-wizard
[^3]: https://github.com/odoo/odoo/issues/26527

simahawk (approved):

simahawk:
> @OCA/web-maintainers can somebody have a look? For beta modules, one review should be enough. Thanks!

I don't see a dev status in the manifest. Could you add it?

SirTakobi:
> > @OCA/web-maintainers can somebody have a look? For beta modules, one review should be enough. Thanks!
> 
> I don't see a dev status in the manifest. Could you add it?

Thanks for the review!
dev status added

etobella (approved):
Confirm will not work with this change... however it will never work, as confirm expect a save.... as you can see here:

https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/js/views/form/form_controller.js#L463-L500

I think we can proceed with merge, or do you want to fix the confirm?

SirTakobi:
> Confirm will not work with this change... however it will never work, as confirm expect a save.... as you can see here:
> 
> https://github.com/odoo/odoo/blob/12.0/addons/web/static/src/js/views/form/form_controller.js#L463-L500
> 
> I think we can proceed with merge, or do you want to fix the confirm?

Thanks for the review!
I didn't think of that, but I'd rather have this merged as is.
Just as reference, I added the issue in the **Known issues** of the module.

etobella:
Something happened with docker... Let's try again :smile: 
/ocabot merge nobump

PR #2257, opened 2022-07-19
[13.0][FIX] web_responsive: mentions overlapping

chienandalu:
When we have the chatter in the right side, mentions tend to overlap with other screen elements, so the dialog is not fully visible. With this change, we move the mentions dialog to the bottom of the composer window to avoid such problem.

Before:
image

After:
image

In future versions is already fixed to behave like this

cc @Tecnativa TT38108

please review @pedrobaeza @CarlosRoca13


pedrobaeza (approved):

CarlosRoca13 (approved):
Tested on runbot :+1:

PR #2260, opened 2022-07-26
[15.0][ADD] web_chatter_position

JuMiSanAr:
This module allows to choose between the "normal" position of the chatter a a new "sided" position.

Peek 2022-07-26 09-17

This module comes from: https://apps.odoo.com/apps/modules/15.0/ob_chatter_position/

### DIFF
--- a/web_chatter_position/__manifest__.py
+++ b/web_chatter_position/__manifest__.py
@@ -1,0 +1,27 @@
+# Copyright 2022 Hynsys Technologies
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Chatter Position",
+    "summary": "Add an option to change the chatter position",
+    "version": "15.0.1.0.0",
+    "author": "Hynsys Technologies, Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    "license": "LGPL-3",
+    "category": "Extra Tools",
+    "images": ["static/description/images/web_chatter_position.png"],
+    "depends": ["web", "mail"],
+    "data": ["views/res_users.xml", "views/web.xml"],
+    "assets": {
+        "web.assets_backend": [
+            "/web_chatter_position/static/src/scss/chatter_position.scss",
+            "/web_chatter_position/static/src/scss/attachment_viewer.scss",
+            "/web_chatter_position/static/src/js/form_chatter_position.js",
+        ],
+        "web.assets_qweb": [
+            "/web_chatter_position/static/src/xml/form_buttons.xml",
+        ],
+    },
+    "installable": True,
+    "auto_install": False,
+}
--- a/web_chatter_position/readme/CONTRIBUTORS.rst
+++ b/web_chatter_position/readme/CONTRIBUTORS.rst
@@ -1,0 +1,2 @@
+* Hynsys Technologies <hynsystechnologies@gmail.com>
+* Juan Miguel Sánchez Arce <juan.sanchez@camptocamp.com>
--- a/web_chatter_position/readme/DESCRIPTION.rst
+++ b/web_chatter_position/readme/DESCRIPTION.rst
@@ -1,0 +1,4 @@
+Configurable Chatter Position.
+Change Chatter Position in User Preferences.
+Change Chatter Position on the fly.
+Supports Both Community & Enterprise Edition.
--- a/setup/web_chatter_position/odoo/addons/web_chatter_position
+++ b/setup/web_chatter_position/odoo/addons/web_chatter_position
@@ -1,0 +1,1 @@
+../../../../web_chatter_position
--- a/setup/web_chatter_position/setup.py
+++ b/setup/web_chatter_position/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/web_chatter_position/README.rst
+++ b/web_chatter_position/README.rst
@@ -1,0 +1,1 @@
+Wait for the bot to create a proper README.
--- a/web_chatter_position/__init__.py
+++ b/web_chatter_position/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/web_chatter_position/models/__init__.py
+++ b/web_chatter_position/models/__init__.py
@@ -1,0 +1,1 @@
+from . import res_users
--- a/web_chatter_position/models/res_users.py
+++ b/web_chatter_position/models/res_users.py
@@ -1,0 +1,23 @@
+# Copyright 2022 Hynsys Technologies
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    chatter_position = fields.Selection(
+        [("bottom", "Bottom"), ("sided", "Sided")],
+        default="sided",
+    )
+
+    #  Override so that the user can change the chatter_position field
+
+    @property
+    def SELF_READABLE_FIELDS(self):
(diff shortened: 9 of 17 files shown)


web_chatter_position/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "Chatter Position",
+    "summary": "Chatter Position",
+    "version": "15.0.1.0.2",
+    "author": "Odoo Being, Odoo Community Association (OCA), camptocamp",
```
simahawk: OCA must stay at the end

web_chatter_position/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "website": "https://github.com/OCA/web",
+    "license": "LGPL-3",
```
simahawk, line 7: for the license ask Odoo Being

web_chatter_position/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "category": "Extra Tools",
+    "images": ["static/description/images/web_chatter_position.png"],
+    "support": "odoobeing@gmail.com",
```
simahawk: drop this support key

web_chatter_position/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Odoo Being <odoobeing@gmail.com>
+* Camptocamp <info@camptocamp.com>
```
simahawk: your name here, not info

web_chatter_position/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+{
```
simahawk, line 4: add copyright to  py and js files

simahawk (changes requested):
link in the PR description where the module comes from maybe

web_chatter_position/__manifest__.py
simahawk: version 1.0.0

JuMiSanAr:
> link in the PR description where the module comes from maybe

All updated.

ajaniszewska-dev:
@JuMiSanAr  pre commit fails

web_chatter_position/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2022 Odoobeing SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Chatter Position",
+    "summary": "Add an option to change the chatter position",
+    "version": "15.0.1.0.0",
+    "author": "Odoo Being, camptocamp, Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "Odoo Being, Camptocamp, Odoo Community Association (OCA)",
```

simahawk:
Tested on runboat. The field is duplicated in user settings:

image

Also, "normal" does not seem meaninful. Maybe "bottom"? Or "default"?

luc-demeyer:
I agree with comment of @simahawk  and I suggest 'bottom' in stead of 'normal'.

ivantodorovich:
The same feature is implemented in `web_responsive` module: https://github.com/OCA/web/blob/15.0/web_responsive/models/res_users.py

(probably why the field is duplicatedo on runboat)

The issue with `web_responsive` is that is not compatible with Odoo Enterprise. I think the feature should be removed from `web_responsive` if we want to isolate it in a separate module.

simahawk:
> The same feature is implemented in `web_responsive` module: https://github.com/OCA/web/blob/15.0/web_responsive/models/res_users.py
> 
> (probably why the field is duplicatedo on runboat)
> 
> The issue with `web_responsive` is that is not compatible with Odoo Enterprise. I think the feature should be removed from `web_responsive` if we want to isolate it in a separate module.

Good point. @JuMiSanAr pls have a look.

pedrobaeza:
I don't think we should extract in this version the feature, and in v16 it's supposed that enterprise interface will be moved to community, so only this new module will be needed

simahawk:
then we should probably let them co-exist and exclude each other + update ROADMAP on both

pedrobaeza:
I think just the README mention is enough, as exclude them in the manifest causes troubles in CIs.

youring:
Maybe a better place to put the icon is far right beside the "Next" pager button? @JuMiSanAr Thanks for moving this into OCA.
put_icon_far_right

web_chatter_position/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Hynsys Technologies <hynsystechnologies@gmail.com>
+* Camptocamp <juan.sanchez@camptocamp.com>
```
simahawk:
```suggestion
* Juan Sanchez <juan.sanchez@camptocamp.com>
```

web_chatter_position/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,4 @@
+Configurable Chatter Position.
+Change Chatter Position from User Preferences.
+Change Chatter Position on fly.
```
simahawk:
```suggestion
Change Chatter Position on the fly.
```

simahawk: you must mention the conflict w/ web_responsive

JuMiSanAr:
> Maybe a better place to put the icon is far right beside the "Next" pager button? @JuMiSanAr Thanks for moving this into OCA. put_icon_far_right

Hello @youring,

Thanks for your input!

Since that's only a stylistic change and this PR is just intended to integrate the module into the OCA, I'd postpone it for a possible new version. Hope that's ok.

web_chatter_position/static/description/index.html
```diff
@@ -0,0 +1,139 @@
...
+<div class="oe_slogan" style="margin-top:10px !important;">
+    <div>
+        <a  class="btn btn-primary btn-lg mt8"
+        style="color: #FFFFFF !important;border-radius: 0;" href="mailto:hynsystechnologies@gmail.comm"><i
+        class="fa fa-envelope"></i> Email </a>
+    </div>
+    <br>
+</div>
+</section>
+</div>
```
MihranThalhath: these are not needed i guess

web_chatter_position/static/description/index.html
```diff
@@ -0,0 +1,138 @@
...
+<h3 class="oe_slogan" style="margin-top:20px;" >Need Any Help Or Have Any Feature Requests?</h3>
+<div class="oe_slogan" style="margin-top:10px !important;">
+    <div>
+        <a  class="btn btn-primary btn-lg mt8"
+        style="color: #FFFFFF !important;border-radius: 0;" href="mailto:hynsystechnologies@gmail.comm"><i
+        class="fa fa-envelope"></i> Email </a>
+    </div>
+    <br>
+</div>
+
```
MihranThalhath: Since the module is maintained by OCA, the index and the images can be removed (or changed) I guess

simahawk: This file is going to be re-generated. You can delete it.

MihranThalhath (approved):
LGTM 👍🏼

JuMiSanAr:
> LGTM 👍🏼

Thanks for the reviews and the approval!

web_chatter_position/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+{
+    "name": "Chatter Position",
+    "summary": "Add an option to change the chatter position",
+    "version": "15.0.1.0.0",
+    "author": "Hynsys Technologies, Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    "license": "LGPL-3",
+    "category": "Extra Tools",
+    "images": ["static/description/images/web_chatter_position.png"],
+    "depends": ["web", "mail"],
```
i-vyshnevska, line 13: you can have yourself here as a maintainer

simahawk (approved):

Issue #2266, opened 2022-08-01
web_timeline: lines are always sorted by alphabetical order, ignoring _order of the class


alexis-via:
In web_timeline on v14, the lines are always sorted by alphabetical order of the "display_name" field ; the _order of the class is ignored.

My code analysis:
The problem is not in the JS lib that is used by web_timeline. The JS lib supports ordering via the groupOrder parameter, cf https://visjs.github.io/vis-timeline/docs/timeline/
The groupOrder argument is used in the code here https://github.com/OCA/web/blob/14.0/web_timeline/static/src/js/timeline_view.js#L124 : we pass it the method group_order : https://github.com/OCA/web/blob/14.0/web_timeline/static/src/js/timeline_view.js#L143
This methods compare the property "content" of the group.
This property "content" corresponds to the "display_name", cf https://github.com/OCA/web/blob/14.0/web_timeline/static/src/js/timeline_renderer.js#L464

Now we need to find a way to improve the module to use _order of the group_by field for sorting the lines. Any help/ideas are welcomed!

alexis-via:
I had an idea to solve the problem: use the fact that the group_by made by odoo are correctly sorted by the _order of the group_by field and generate an "order" property in addition to "content" and "id", and then use "groupOrder" = "order" to sort on the "order" property. 
For that, I generated the "order" property in the "split_group" method https://github.com/OCA/web/blob/14.0/web_timeline/static/src/js/timeline_renderer.js#L374 But, according to my tests, the loop inside the split_group method is not executed in the _order of the group_by field, so it doesn't work :-(

I don't know if my idea is bad or if my implementation is wrong. It's still not very clear to me how the data is gathered by odoo. I don't see any reference to "read_group" in the code, I only see a "search_read". So I guess the data is taken by a search_read and the group_by logic is made in the JS code and maybe the search_read doesn't have any specific order...

Any other idea is welcomed.

alexis-via:
I'm making big progress on this issue:
[...]

PR #2268, opened 2022-08-02
[14.0][FIX] web_timeline: order group_by field (and update JS lib)

alexis-via:
Fix bug #2266: before this fix, the lines of the timeline view were always sorted by alphabetical order, ignoring the native order of the model on which the group_by field points to.

Fix duplicate entries in the 'fields' argument when the 'colors' parameters uses the same field in multiple conditions.


alexis-via:
I took the opportunity of this PR to update the JS lib to a newer version. But it's not linked to the bug ; the bug fix is in the first commit.

chienandalu (commented):
@CarlosRoca13 can you check it?

CarlosRoca13 (approved):
I will add the changes to the migration to 15.0

Thanks!

pedrobaeza (changes requested):
I have manually added a timeline view in project and use `date_deadline` as starting time for making it work, and I found some weird behavior:

Peek 03-08-2022 18-28

CarlosRoca13:
Hmm I did not found that problems... I have done the test with project_timeline in local machine 
Peek 04-08-2022 06-42

pedrobaeza:
I see that you are not moving any task to "Unassigned". Can you try?

CarlosRoca13:
unassigned

No problems found...

pedrobaeza (approved):

pedrobaeza:
OK, then it should be the hot change I did on runbot. Let's merge it:

/ocabot merge major

chienandalu:
Could it be a browser specific compatibility issue?

pedrobaeza:
I was doing the tests on Firefox. I will test it again when merged.

PR #2240, opened 2022-06-28
[15.0][MIG] web_ir_actions_act_view_reload: Migration to 15.0

manuelcalerosolis:
This is the migration of ir_actions_act_view_reload to 15.0

PR #2272, opened 2022-08-09
[15.0][MIG] web_ir_actions_act_view_reload

ryanc-me:
Fortunately, Odoo 15.0 is still using the old-style controllers/renderers, so this wasn't too difficult once we can access the legacy `widget`.

The 16.0 migration for this module might be interesting!


ryanc-me:
Note, previous attempt at #2053

ryanc-me:
Note, previous attempt at #2240

pedrobaeza (dismissed):
Thanks for the contribution.

ryanc-me:
Thanks Pedro.

I see that the fork branch should be named `15.0-mig-$MODULE` - I can't change that without creating a new PR. Will the current state work for you?

pedrobaeza:
Don't worry about the branch name in this case, as such name is only a transient one that is not transferred to any place once the pull request is merged. It's better to not use `15.0` one, as you can have several pull requests for the same branch, though, but for following ones ( that I hope to be a lot :smile: )

ryanc-me:
Thanks @pedrobaeza, for sure there will be some more PRs to come!

etobella (changes requested):
Something does not work:

I followed this steps:

- Access runboat
- Create a server action on any model with the code `action={'type': 'ir.actions.act_view_reload'}`
- Create the Contextual Action
- Refresh
- Access the model and press the button
- The following error is raised:

```
Error: The ActionManager service can't handle actions of type ir.actions.act_view_reload
    at doAction (http://oca-web-15-0-pr2272-3d7bde7f5331.runboat.odoo-community.org/web/assets/debug/web.assets_backend.js:25001:23) (/web/static/src/webclient/actions/action_service.js:1138)
```

web_ir_actions_act_view_reload/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "category": "Web",
+    "version": "15.0.1.0.0",
+    "license": "LGPL-3",
+    "author": "Modoolar, CorporateHub, Ryan Cole, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    "depends": ["web"],
+    "installable": True,
+    "assets": {
+        "web.assets_backend": [
+            "web_ir_actions_act_view_reload/static/src/*",
```
wildi1: In my case the js was not loaded this way. Changing line 18 to the following worked for me:
`"web_ir_actions_act_view_reload/static/src/**/*.esm.js",`

ryanc-me: @wildi1 Thanks - good catch. I guess the JS was cached in my browser during tests. Fixed now 👍

wildi1: @ryanc-me :+1:

CarlosRoca13 (approved):
Tested @etobella use case and tested if is added the action by code.

It seems to be working fine

Thanks!

ryanc-me:
Thanks @CarlosRoca13!

Ping @etobella, it looks like the merge is blocked because of a change request. Could you please take a look?

StefanRijnhart:
@etobella I followed your scenario and I did not encounter this problem. Running the action correctly triggered the method that this module registers for the action type.

image

StefanRijnhart (approved):

etobella (approved):
Thanks!!! Now it works perfectly
:smile: 

/ocabot merge nobump

PR #2278, opened 2022-08-18
13.0 mig web export view

vancouver29:
[13.0][MIG]web_export_view


web_export_view/controllers/controllers.py
```diff
@@ -15,22 +15,22 @@
 class ExcelExportView(ExcelExport):
     def __getattribute__(self, name):
         if name == "fmt":
-            raise AttributeError()
-        return super(ExcelExportView, self).__getattribute__(name)
+            return super(ExcelExportView, self).__getattribute__(name)
+        raise AttributeError()
```
yweng8111, line 19: Why do you change origin function like this?

simahawk:
Please check https://github.com/OCA/web/issues/1388

image

web_export_view/controllers/controllers.py
vancouver29: @yweng8111 It is just for the purpose of improving Test coverage. The logic stays untouched.

vancouver29:
> Please check #1388
> 
> image

@simahawk, I agree that this export function already exists in v13. But with this `Export All`, we can only export all the records instead of the selected ones. Since not every user wants to export all the records ...

PR #2282, opened 2022-09-02
15.0 mig web widget digitized signature

imlopes:


web_widget_digitized_signature/models/res_users.py
```diff
@@ -0,0 +1,41 @@
...
+
+from odoo import api, fields, models
+
+
+class ResUsers(models.Model):
+    _name = "res.users"
+    _inherit = ["res.users", "mail.thread"]
+
+    digital_signature = fields.Image(
+        oldname="signature_image",
```
legalsylvain, line 15:
```suggestion
```

web_widget_digitized_signature/models/res_users.py
```diff
@@ -0,0 +1,41 @@
...
+        copy=False,
+        attachment=True,
+        max_width=1024,
+        max_height=1024,
+    )
+
+    @api.model
+    def create(self, vals):
+        res = super(ResUsers, self).create(vals)
```
legalsylvain, line 24:
```suggestion
        res = super().create(vals)
```

web_widget_digitized_signature/models/res_users.py
```diff
@@ -0,0 +1,41 @@
...
+        res._track_signature(vals, "digital_signature")
+        return res
+
+    def write(self, vals):
+        self._track_signature(vals, "digital_signature")
+        return super(ResUsers, self).write(vals)
```
legalsylvain, line 30:
```suggestion
        return super().write(vals)
```

legalsylvain (changes requested):
Hi. Signature widget has been introduced in V13 in odoo community.
See difference between : 
https://github.com/odoo/odoo/tree/12.0/addons/web/static/src/js/fields
and https://github.com/odoo/odoo/tree/13.0/addons/web/static/src/js/fields

So, I don't see the reason to port this module. ``widget="signature"`` will work without the module installed. Could explain why you ported this module ? Maybe I missed something.

Thanks !

Issue #2284, opened 2022-09-05
[V12] Web_export_view no funciona con la ultima actualización del navegador Chrome


inprbau:
**Module**
 Web Export Current View (web_export_view)

**Describe the bug**
With the lastest update of Chrome Navigator (V 105.0.5195.102) when the excel is generated its empty. Only appear the column's name.

**To Reproduce**
All the versions of this modul are affected: V12, V11, V10

**Steps to reproduce the behavior:**  
 I try to reproduce at the runbot of the modul, but all of the instances give the same error: 
![image](https://user-images.githubusercontent.com/50661813/188428355-5cbfb0cc-f964-4517-bcd7-eb6c1ed9ef7a.png)
The error is: 
![image](https://user-images.githubusercontent.com/50661813/188428615-740450c7-6faf-464f-a118-d7cf86b6b4f6.png)

1. Go to a app and show the tree list (V12 community):
2. Select two or more lines and press the botton
![image](https://user-images.githubusercontent.com/50661813/188429336-ca7cf2da-8e36-4098-9eb2-5253ed6fc4d2.png)

3.  The excel archive is empty
![image](https://user-images.githubusercontent.com/50661813/188429501-e7064fe2-7b94-4020-abc1-7964252f12a8.png)

Expected behavior
The excel contains the information of the view

Additional context
This behavior is the same in all the versions of Odoo app and excel. The problem is the version of chrome: 105.0.5195.102

JonathanStein:
Changing ":checkbox" to "[type=checkbox]" in https://github.com/OCA/web/blob/12.0/web_export_view/static/src/js/web_export_view.js#L76 seems to fix the problem.

inprbau:
Thanks @JonathanStein 
The same bug appear with microsoft edge (update to v. 105.0.1343.27). Don't appears with Mozilla Firefox, the lastet version is 104.0.1
With this change will be solved?

pedroortega-punt:
@JonathanStein
If a pull request is made, will chrome and edge versions lower than 105.0.1 stop working?

JonathanStein:
Based on a single test with Chrome 104: The Excel export still works when ":checkbox" is changed to "[type=checkbox]".

alv97:
Can apply this to V11?

pedrobaeza:
You can cherry-pick the commit and make the pull request. Closing as solved in the commented version.

PR #2285, opened 2022-09-06
Fix error with Chrome 105

JonathanStein:
See https://github.com/OCA/web/issues/2284


ioans73 (approved):
LGTM 👍🏻

ap-wtioit (approved):
Thanks for fixing this

PR #2292, opened 2022-09-13
[15.0][FIX] web_advanced_search: Prevent error in many2one fields

victoralmau:
Prevent error in many2one fields

**Before**
antes

**After**
despues

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa


ivantodorovich (approved):

chienandalu (approved):
Thanks!

PR #2293, opened 2022-09-14
[15.0][FIX] web_timeline: utc, create and update

jguenat:
When creating a new record with the timeline view, the dates where not correctly saved as UTC. fixes this oca/project#842

After saving the new record, the timeline view is emptied. I fixed this by removing the renderer and reload.
It seems to work but I'm not sure about this modification.
Any feedback @CarlosRoca13 ?

### DIFF
--- a/web_timeline/static/src/js/timeline_controller.js
+++ b/web_timeline/static/src/js/timeline_controller.js
@@ -58,7 +58,7 @@ odoo.define("web_timeline.TimelineController", function (require) {
 
             let fields = this.renderer.fieldNames;
             fields = _.uniq(fields.concat(n_group_bys));
-            return $.when(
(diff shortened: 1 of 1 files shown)


jguenat:
Following odoo/odoo#102621 an "Uncaught Promise > updateProm.finally is not a function" error occurs each time the timeline view is updated.
I added a commit to fix this.
@cesar-tecnativa 
@CarlosRoca13 
@pedrobaeza 
I ping you because your were active in the V15 migration.
Reviews much appreciated :) I can review other non-JS pr if you need it

pedrobaeza (approved):

CarlosRoca13 (approved):
Thanks :+1:

Issue #2007, opened 2021-08-17
[14.0][BUG] web_access_rule_buttons: Error opening full message composer in chatters


gaikaz:
**Steps to reproduce:**

1. Install the module `web_access_rule_buttons`
2. Install any module that has a "chatter" section. For example "Contacts" (`contacts`)
3. Open any contact
4. In the chatter section, click "Send Message" and then expand the little text field to the "Full Composer"

**Current behavior:**

It opens the full composer but also gives the error seen in screenshot and video below.

**Expected behavior:**

No error ( ¯\\\_(ツ)_/¯ duh... ) 

**Screenshot:**

![error](https://user-images.githubusercontent.com/17097810/129798746-0e036cba-7d6f-4840-b049-d63604162431.png)

**Video evidence from OCA runbot:**

https://user-images.githubusercontent.com/17097810/129798782-4ae28280-cd33-4f81-b08c-bfff78391d78.mp4

PR #1752, opened 2020-12-03
14.0 [mig] web_access_rule_buttons

dsolanki-initos:

PR #2294, opened 2022-09-14
[15.0][MIG] web_access_rule_buttons

ryanc-me:
JS migrations were simple as `web.FormController` is still non-ESM legacy mode.

However, the `check_access_rule_all` checks were convoluted. The logic was:
 1. Check if user has access, but ignore the result (??)
 2. Check if the record is transient, empty, or if the user is admin: **allow edit**
 3. Otherwise: **deny edit**

I've changed the logic so:
 1. Check if the record is transient or empty: **allow edit**
 2. Check if the user has access: **allow edit**
 3. Otherwise: **deny edit**

As part of review, can someone double-check the previous code ([here](https://github.com/OCA/web/commit/1dceb13f8fb0b0e8ba3a83c2da2916150efd62e1#diff-f633608e95636af811c976cb57d688ba97a89fd965788b71a95e44bf5d1e5df4))

There is some discussion around the changes in #2007 and #1752.


web_access_rule_buttons/static/src/js/form_controller.js
```diff
@@ -0,0 +1,32 @@
+/* Copyright 2016 Camptocamp SA
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+odoo.define("web_access_rule_buttons.main", function (require) {
+    "use strict";
+    var FormController = require("web.FormController");
+    FormController.include({
+        async _update(state, params) {
+            return this._super(state, params).then(this.show_hide_buttons(state));
+        },
```
ivantodorovich: I suggest rewriting this in modern js

```suggestion
        async _update(state, params) {
            const res = await this._super.apply(this, arguments);
            await this.show_hide_buttons(state);
            return res;
        },
```

web_access_rule_buttons/static/src/js/form_controller.js
```diff
@@ -0,0 +1,32 @@
...
+            return self
+                ._rpc({
+                    model: this.modelName,
+                    method: "check_access_rule_all",
+                    args: [[state.data.id], ["write"]],
+                })
+                .then(function (accesses) {
+                    self.show_hide_edit_button(accesses.write);
+                });
+        },
```
ivantodorovich:
```suggestion
        async show_hide_buttons(state) {
            const access = await this._rpc({
                model: this.modelName,
                method: "check_access_rule_all",
                args: [[state.data.id], ["write"]],            
            })
            this.show_hide_edit_button(access.write);
        },
```

web_access_rule_buttons/static/src/js/form_controller.js
```diff
@@ -0,0 +1,32 @@
...
+        show_hide_edit_button: function (access) {
+            if (this.$buttons) {
+                var button = this.$buttons.find(".o_form_button_edit");
+                if (button) {
+                    button.prop("disabled", !access);
+                }
+            }
+        },
```
ivantodorovich:
```suggestion
        show_hide_edit_button(access) {
            if (this.$buttons) {
                const button = this.$buttons.find(".o_form_button_edit");
                if (button) {
                    button.prop("disabled", !access);
                }
            }
        },
```

web_access_rule_buttons/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
```
ivantodorovich:
```suggestion
```

License is not needed on `__init__.py` files. Moreover it's inconsistent with the module license (AGPL, not LGPL)

web_access_rule_buttons/tests/test_access_rule_buttons.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2019 Onestein BV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.tests.common import TransactionCase
+
+
+class TestAccessRuleButtons(TransactionCase):
+    def setUp(self):
```
ivantodorovich: Use `setUpClass` instead.
Otherwise it runs before each unit test and it's not performant

web_access_rule_buttons/__manifest__.py
```diff
@@ -4,16 +4,18 @@
 {
     "name": "Web Access Rules Buttons",
     "summary": "Disable Edit button if access rules prevent this action",
-    "version": "14.0.1.0.0",
-    "author": "Camptocamp, Onestein, Odoo Community Association (OCA)",
+    "version": "15.0.1.0.0",
+    "author": "Camptocamp, Onestein, Ryan Cole, Odoo Community Association (OCA)",
```
ivantodorovich: Unfortunately a migration does not qualify as author (unless a big refactor takes place).
You can add yourself as a contributor in the `readme/CONTRIBUTORS.rst` file

web_access_rule_buttons/models/models.py
```diff
@@ -22,19 +22,16 @@ def check_access_rule_all(self, operations=None):
...
-            try:
-                self.check_access_rule(operation)
-            except exceptions.AccessError:
-                result[operation] = False
-            if (
-                self.is_transient()
-                or self.ids
-                and self.env.user.has_group("base.user_admin")
-            ):
+            if self.is_transient() or not self.ids:
```
ivantodorovich, line 25: The previous version had `and self.env.user.has_group("base.user_admin")`, could you comment on its removal?

ryanc-me:
@ivantodorovich thanks for the feedback! Most of those changes are done (I also rebased on oca/web).

Regarding the `models.py` logic changes, I did leave some comments in the PR description.

See the original [`models.py` file here](https://github.com/OCA/web/blob/1dceb13f8fb0b0e8ba3a83c2da2916150efd62e1/web_access_rule_buttons/models/models.py)

User access was checked on L25-L28, then ignored and unconditionally overwritten on L29-L39. The final logic was:
 * Allow access if the model is transient
 * Allow access if `ids` exists AND the user is in `base.user_admin` group
 * Deny access in all other cases
 * No checks on user read/write/etc access

Here is the [14.0 migration PR](https://github.com/OCA/web/pull/1752). Maybe I'm wrong and there is some `try-except-if-else` block in Python?

Either way the addition of `base.user_admin` wasn't commented on in the 14.0 PR, and I think it was actually wrong, perhaps the author mistook the purpose of the module, or it was an attempt to fix the issue mentioned in the PR.

ivantodorovich (approved):
I see, makes sense.

Thanks a lot for your work 👍🏻

ryanc-me:
Thanks Ivan!

BTW, do you know why the two tests are failing?

They were succeeding before, the only change I made was `setUp` -> `setUpClass`. The error is a bit confusing.

ivantodorovich:
Hello @ryanc-me !

Our CI is failing with this:

```
2022-10-28 23:48:43,570 248 ERROR odoo odoo.addons.web_domain_field.tests.test_qunit: FAIL: TestQunit.test_qunit
  File "/__w/web/web/web_domain_field/tests/test_qunit.py", line 11, in test_qunit
    self.browser_js(
  File "/opt/odoo/odoo/tests/common.py", line 1608, in browser_js
    self.fail('%s\n%s' % (message, error))
AssertionError: Some js test failed
QUnit test failed: web_domain_field > one2many: field as domain attribute value [message: "Promise rejected during "one2many: field as domain attribute value": Unimplemented route: /web/dataset/call_kw/res.partner/check_access_rule_all", expected: "undefined"]
QUnit test failed: web_domain_field > one2many: field as domain attribute value [message: "Expected 2 assertions, but 1 were run", expected: "undefined"]
2 / 2 tests failed.
```

From the message, we can see what's failing are actually the QUnit tests from the module `web_domain_field`. However, it is related to the module on this PR, as it's failing due to an `Unimplemented route: /web/dataset/call_kw/res.partner/check_access_rule_all`

The reason for this is because the QUnit test suite runs in isolation from the server, so all RPC calls are mocked.
Even though this module doesn't have QUnit tests of its own, it's modifying the `FormController` adding an extra server request. When the FormController is used during the QUnit tests run, and tries to do that request with the mocked server, it fails.

The solution is to teach the mocked server to mock this request. Simply do something similar to this:

- https://github.com/odoo/odoo/blob/a92e0ad633985db3cb5ef8cf4d0e97dfb4580d77/addons/google_drive/static/tests/mock_server.js
- https://github.com/odoo/odoo/blob/a92e0ad633985db3cb5ef8cf4d0e97dfb4580d77/addons/google_drive/__manifest__.py#L32-L34

In your case, I think, the mocked response should return
```js
{read: true, write: true, ... }
```

ryanc-me:
@ivantodorovich thank you detailed assistance, looks like it's all passing now!

I've just rebased on oca/web and squashed :)

I guess now we just need 1 more review?

ivantodorovich:
awersome ;) thanks a lot!

PR #2296, opened 2022-09-19
[14.0][IMP] web_disable_export_group: Differenciate between the XLSX export and export all

victoralmau:
FWP from 13.0: https://github.com/OCA/web/pull/2063

With this change, we have now two groups, one for the excel and one for the Export from Actions

Please @chienandalu and @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/web_disable_export_group/README.rst
+++ b/web_disable_export_group/README.rst
@@ -25,11 +25,9 @@ Web Disable Export Group
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-In the standard Odoo the UI option 'Export' that is present in the 'Action' menu
-and the 'Export All' button of any list view is always enabled (for every user).
+The standard export group prevents both options: 'Direct Export (xlsx)' and 'Export All'.
 
-This module makes the option 'Export' and 'Export All' enabled only for the users
-that belong to the Export Data group.
+This module adds a new group for the 'Direct Export (xlsx)' feature, leaving the standard one for only the 'Export All' feature.
 
 Admin user can always use the export option.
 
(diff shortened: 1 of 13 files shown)


pedrobaeza:
On v14, there's a standard group called `Access to export feature`, so you have to deal with it for having the same features. In fact, I don't know why this module has been migrated to 14.0 in its previous form.

victoralmau:
> Access to export feature

Ok, reviewing the "Access to export feature" permission in v14, I think this module should not add any extra permissions, and use only the "Access to export feature" permission. Right?

victoralmau:
I am really thinking that this module is NOT needed in v14. What is the process in this case?

pedrobaeza:
Well, we are going to lose the possibility to differentiate between one exportation and the other. Please check the initial motivation of this split to see if it's something that can be removed.

victoralmau:
2 groups for export was added in v13 only: https://github.com/OCA/web/pull/2063 maybe @etobella can explain the reason for this separation and if it is still necessary.

etobella:
Two groups were created in order to split the permissions. On 13, you needed to have the export permissions in order to export on excel (before the change). After checking on 14 core I see they did the first approach. A single group for both permissions. I think it is not the best approach.... About this module we have two options, adapt it only for the XLSX behavior or remove it, but I would prefer to keep it.

etobella:
The same approach exists on 15 and 16 on odoo CE

victoralmau:
Changes done. Actually the module was not working properly until now and many changes should have been done in the migration to v14.

chienandalu (approved):
Looks great. Thanks!

victoralmau:
Ping @pedrobaeza

pedrobaeza (changes requested):
README should reflect the new situation:

- The standard export group prevents both options.
- This module adds a new group for the X feature, leaving the standard one for only the Y feature (not sure which one is each).

For our customers, we will initially uninstall it on v14 migration, not needing such differentiation.

pedrobaeza (approved):
To be merged when unrelated CI problem is fixed.

pedrobaeza:
Please rebase to see if everything is OK now.

victoralmau:
IMO now, it's ready to merge.

PR #2297, opened 2022-09-20
[FIX] web_responsive : apply same z-index value for the menu as odoo CE

legalsylvain:
Rational : 
- the odoo menu in the community edition has ``z-index : 1000``. see : https://github.com/odoo/odoo/blob/12.0/addons/web/static/lib/bootstrap/css/bootstrap.css#L2991
- however, the ``web_responsive`` that introduce a new pretty responsive menu has a ``z-index : 100``.

So,
- the value is not consistent
- this value conflict with some implementation. For exemple, if we try to integrate the leaflet.js (https://github.com/OCA/geospatial/pull/310) that use a lot of high z-index.

Thanks for your review.

### DIFF
--- a/web_responsive/static/src/css/web_responsive.scss
+++ b/web_responsive/static/src/css/web_responsive.scss
@@ -12,7 +12,7 @@ $chatter_zone_width: 35%;
     max-height: calc(100vh - #{$o-navbar-height});
     position: fixed;
     width: 100vw;
-    z-index: 100;
+    z-index: 1000;
     // Inline style will override our `top`, so we need !important here
     top: $o-navbar-height !important;
     transform: none !important;


pedrobaeza (approved):
I don't see any reason for not putting that.

rvalyi (approved):

pedrobaeza:
Please check if it's the same on future versions:

/ocabot merge patch

legalsylvain:
> Please check if it's the same on future versions

I checked V15 but the code is totally différent. Not sure if this patch will be required. 
On V16 the module should be obsolète or to refactor entirely if @fpodoo respect his roadmap. (Make responsive UI open-source)

https://mobile.twitter.com/fpodoo/status/1544606815786262528

legalsylvain:
Note : the PR to refactor responsivity in Odoo CE is here : https://github.com/odoo/odoo/pull/100580

legalsylvain:
CC : @OCA/web-maintainers

pedrobaeza:
Seeing the associated runbot, it doesn't seem very... responsive

Selección_003
Selección_004

and it doesn't include the app switcher, so the OCA module will still be required.

legalsylvain:
Seeing the associated runbot, it doesn't seem very... responsive

> it's tiring...

Thanks @pedrobaeza for your quick review on odoo PR ! Hope this will be fixed on the stable V16...

PR #2228, opened 2022-06-10
[15.0][FIX]kanban: valid attribute 'records_draggable'

MackZhong:
Kanban View has an attribute `records_draggable`, means 'whether it should be possible to drag records when kanban is grouped'.
But code at line 97 and line 179 assume the option `recordsDraggable` dur to the value of `config.device.isMobile`, thus the attribute `records_draggable` has no effect on destop platform.
Sorry for terrible English.

PR #2302, opened 2022-09-25
[15.0][FIX] web_responsive: fix records_draggable attribute

SplashS:
Kanban View has an attribute 'records_draggable', means 'whether it should be possible to drag records when kanban is grouped'.
It's support was broken on desktop.
This PR supersedes #2228


pedrobaeza (approved):

pedrobaeza:
Let's fix this one:

/ocabot merge patch

PR #2304, opened 2022-09-26
[15.0][FIX] web_m2x_options M2ODialog not fully functional

theangryangel:
Prior to this patch:

* M2ODialog does not function correctly due to upstream removal of `search_create_popup` custom_event (https://github.com/odoo/odoo/commit/1d4d2a68315dcc07acf225d449b8d2e1fd3fb5a8). The Create and Edit button just does "nothing".
* M2ODialog offers both Create and Create and Edit, even if the options are disabled

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -29,40 +29,43 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
         init: function (parent, name, value) {
             this.name = name;
             this.value = value;
+
+            const buttons = [];
+
+            if (parent._canCreate() && parent._canShowQuickCreate()) {
+                buttons.push({
+                    text: _t("Create"),
+                    classes: "btn-primary",
+                    close: true,
+                    click: function () {
+                        this.trigger_up("quick_create", {value: this.value});
+                    },
+                });
+            }
+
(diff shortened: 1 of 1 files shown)


CRogos (changes requested):
The Code is working and looks good.
 
I am not sure if it is really useful to also remove the "Create and Edit" button in the popup with the same configuration keys. 
image

We removed the "Create..." from the dropdown because it was used accidently to many times, but using the button would be ok for us. 

I did another fix #2318 addressing another issue, which will have some conflict on the _search changes.

Shall we add another config key for the popup and merge both PR before merging it into 15.0?

theangryangel:
> We removed the "Create..." from the dropdown because it was used accidently to many times, but using the button would be ok for us.

Happy for it to be another option, but the client who raised this correctly pointed out that if you're banning create on the dropdown then it almost certainly makes no sense to be in the popup, because you'll still end up with a badly configured product :)

However, I do understand that not everyone will have this opinion, so I'm happy to make it a separate option, if that's what people want? :)

> Shall we add another config key for the popup and merge both PR before merging it into 15.0?

Makes sense, but I'm not sure if I have the time to do this for at least a few days (probably longer). Do you have time to try and pull these changes into yours, and I close this PR? What works for you? :)

theangryangel:
Closing as superseded by [#2901](https://github.com/OCA/web/pull/2901)

PR #2305, opened 2022-09-26
[13.0][IMP] web_refresher: Remove refresh button from x2Many fields.

sergio-teruel:
cc @Tecnativa

I think that with one refresh button in control panel It's enough.
If we diplay one button for every x2Many field in the view is a lot disturb information.

ping @chienandalu @carlosdauden @malvaprior

### DIFF
--- a/web_refresher/README.rst
+++ b/web_refresher/README.rst
@@ -60,6 +60,7 @@ Contributors
 
   * João Marques
   * Alexandre D. Díaz
+  * Sergio Teruel
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


carlosdauden (approved):

chienandalu (approved):

pedrobaeza:
Isn't this taking the same space?

Selección_017

sergio-teruel:
This  behaviour is due to oder module

pedrobaeza:
Ok, 
/ocabot merge patch

Please fw-port it to upper versions

PR #2312, opened 2022-10-05
[14.0] Add web_fix_modules_load

simahawk:
Odoo loads translations and module info using their names.
When you have a lot of modules installed (eg: 500+)
this can lead to very big GET requests (more than 12k) which, in most of the cases,
will be blocked by the web server (eg: nginx) because they are too big.

This module tries to fix this by using modules' ids instead of names
reducing dramatically the size of such requests.

FTR so far we worked around the problem by increasing nginx `large_client_header_buffers` size https://github.com/camptocamp/docker-odoo-nginx/blob/master/9.0/templates/nginx.conf.tmpl#L43

Relates to https://github.com/odoo/odoo/issues/71579

Ref: BSRD-551

### DIFF
--- a/web_fix_modules_load/controllers/main.py
+++ b/web_fix_modules_load/controllers/main.py
@@ -1,0 +1,48 @@
+# Copyright 2022 Camptocamp SA
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import http
+
+from odoo.addons.web.controllers.main import WebClient
+
+
+class WebClientPatched(WebClient):
+    """Handle conversions of modules' ids to their names."""
+
+    def _get_mod_names(self, mods):
+        """Retrieve module names from their IDs in any form."""
+        mod_names = mods
+        model = http.request.env["ir.module.module"].sudo()
+        if isinstance(mods, str):
+            mods = [int(x.strip()) for x in mods.split(",") if x.strip().isdigit()]
+        if mods and isinstance(mods[0], int):
+            mod_names = model.browse(mods).mapped("name")
+        return mod_names
+
+    @http.route()
+    def qweb(self, unique, mods=None, db=None):
+        # Here `mods` comes as 1,2,3,4 string
+        mods = self._get_mod_names(mods)
+        return super().qweb(unique, mods=mods, db=db)
+
+    @http.route()
+    def bootstrap_translations(self, mods):
+        mods = self._get_mod_names(mods)
+        return super().bootstrap_translations(mods)
+
+    @http.route()
+    def csslist(self, mods=None):
+        mods = self._get_mod_names(mods)
+        return super().csslist(mods=mods)
+
+    @http.route()
+    def jslist(self, mods=None):
+        mods = self._get_mod_names(mods)
+        return super().jslist(mods=mods)
+
+    @http.route()
+    def translations(self, unique, mods=None, lang=None):
+        mods = self._get_mod_names(mods)
+        mods = ",".join(mods)
+        return super().translations(unique, mods, lang)
--- a/web_fix_modules_load/static/src/js/session.js
+++ b/web_fix_modules_load/static/src/js/session.js
@@ -1,0 +1,127 @@
+odoo.define("web_load_translations_fix.Session", function (require) {
+    "use strict";
+
+    var Session = require("web.Session");
+    var core = require("web.core");
+    var _t = core._t;
+
+    /**
+     * Override session manager to change how modules' are loaded: use ids instead of names.
+     *
+     */
+    Session.include({
+        _modules_info: function () {
+            return odoo._modules_info;
+        },
+        _module_ids: function (names) {
+            const mod_names = names ? names : this.module_list;
+            const info = this._modules_info();
+            const ids = [];
+            _.each(mod_names, function (name) {
+                if (info[name]) {
+                    ids.push(info[name].id);
+                }
+            });
+            console.debug("Session: load module names", mod_names);
+            return ids;
+        },
+        /**
+         * Full override due to no available hook.
+         * The whole code is taken as-is from Odoo core (comments included).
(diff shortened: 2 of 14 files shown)


TDu (approved):
Nice idea to switch from names to ids to get around the problem.
As always would be good to have some tests, but...

web_fix_modules_load/static/src/js/session.js
```diff
@@ -0,0 +1,127 @@
...
+            var prom = this.session_reload();
+
+            if (this.is_frontend) {
+                return prom.then(function () {
+                    return self.load_translations();
+                });
+            }
+
+            return prom.then(function () {
+                var promise = self.load_qweb(self._module_ids().join(","));
```
sebalix, line 69: `load_qweb` has no parameter in 15.0, this will require some changes at migration, I don't know if it's a big one.

web_fix_modules_load/static/src/js/session.js
```diff
@@ -0,0 +1,127 @@
...
+                    return promise.then(function () {
+                        return self.load_modules();
+                    });
+                }
+                return Promise.all([
+                    promise,
+                    self
+                        .rpc("/web/webclient/bootstrap_translations", {
+                            mods: self._module_ids(),
+                        })
```
sebalix, line 80: Are we forced to send module IDs here? `bootstrap_translations` is a `POST` endpoint (so the module IDs are in the payload, not the header), we should not have the issue there?

web_fix_modules_load/static/src/js/session.js
```diff
@@ -0,0 +1,127 @@
...
+            var loaded = Promise.resolve(self.load_translations());
+            var locale = "/web/webclient/locale/" + self.user_context.lang || "en_US";
+            var file_list = [locale];
+            if (to_load.length) {
+                loaded = Promise.all([
+                    loaded,
+                    self
+                        .rpc("/web/webclient/csslist", {
+                            mods: self._module_ids(to_load),
+                        })
```
sebalix, line 108: Same question here.

sebalix (commented):
LG, one question: was it more complicated to move these `GET` parameters to `POST`?

web_fix_modules_load/static/src/js/session.js
simahawk: good point, I'll add a comment

web_fix_modules_load/static/src/js/session.js
simahawk: I'll check, thanks

web_fix_modules_load/static/src/js/session.js
simahawk: same answer :)

web_fix_modules_load/static/src/js/session.js
simahawk: Thinking about it twice: I'd say they should all behave in the same way.

sebalix: Your choice!

sebalix (approved):

web_fix_modules_load/static/src/js/session.js
```diff
@@ -0,0 +1,127 @@
+odoo.define("web_load_translations_fix.Session", function (require) {
```
ivantodorovich, line 1: nitpicking:

```suggestion
odoo.define("web_fix_modules_load.Session", function (require) {
```

web_fix_modules_load/static/src/js/session.js
```diff
@@ -0,0 +1,127 @@
...
+        _module_ids: function (names) {
+            const mod_names = names ? names : this.module_list;
+            const info = this._modules_info();
+            const ids = [];
+            _.each(mod_names, function (name) {
+                if (info[name]) {
+                    ids.push(info[name].id);
+                }
+            });
+            console.debug("Session: load module names", mod_names);
```
ivantodorovich, line 25: Is this intentional? Otherwise I'd just remove it

```suggestion
```

web_fix_modules_load/controllers/main.py
```diff
@@ -17,7 +17,7 @@ def _get_mod_names(self, mods):
         if isinstance(mods, str):
             mods = [int(x.strip()) for x in mods.split(",") if x.strip().isdigit()]
         if mods and isinstance(mods[0], int):
-            mod_names = model.browse(mods).mapped("name")
+            mod_names = ",".join(model.browse(mods).mapped("name"))
```
simahawk: @TDu are you sure about this change? We are supposed to get a list here. If you need it in `translations` just join there.

TDu: With the module uninstalled, I see the module list being a comma separated string in the controllers. That is why I added this change.
But now I do not really know why both solutions seems to work the same :shrug:

TDu: I finally removed this change

francesco-ooops:
@simahawk is this good enough to set to ready for review? thanks!

simahawk:
In prod since ~months~ more than 1 year...

web_fix_modules_load/static/src/js/session.js
```diff
@@ -0,0 +1,127 @@
...
+        },
+        _module_ids: function (names) {
+            const mod_names = names ? names : this.module_list;
+            const info = this._modules_info();
+            const ids = [];
+            _.each(mod_names, function (name) {
+                if (info[name]) {
+                    ids.push(info[name].id);
+                }
+            });
```
ivantodorovich, line 24: nitpicking: vanilla js is more powerful nowadays, making tools like this a bit pointless. The trend is to favour vanilla js code, which is usually easier to read

```suggestion
            const ids = mod_names.filter(name => info[name]).map(name => info[name].id);
```

ivantodorovich (approved):
non-blocking nitpicking comments: approved 

Thanks!!

francesco-ooops:
@simahawk anything to change or can it be merged?

simahawk:
Rebased and squashed

PR #2317, opened 2022-10-12
[16.0][ADD] new module web_constrasted_theme to identify better buttons and fields in Odoo CE

legalsylvain:
### Description

This module extend the Odoo Community Edition ``web`` module to improve visibility of form view.

**Rational:**
In Odoo V16, the design is very pure. That's great, but it generates some problem for users :

* buttons and fields are not identifiable. (we can not know exactly where there are until you hover over them with the cursor)

* there is no indication for the required fields until trying to save (or exit the screen)

In a way, this module restores the form display of version 15, but preserving the "save on the fly new feature".

**Without this module**

product_template_form_without_module

**With this module**

product_template_form_with_module


legalsylvain:
@sebastienbeau : this one is ready to review.

celm1990:
Hi @legalsylvain only curiosity, I think  is better make a pull request to odoo's repo instead of create New module, o any reason to not does

legalsylvain:
Hi @celm1990. 
Thanks for your review.
I don't get your point : 
- Odoo changed the design on purpose. there is no chance that my pull request will be accepted.
- I think it's a matter of taste. Some may find this module useful, while others prefer the default UI. A module is suitable.

don't you think ?

legalsylvain:
@pedrobaeza : could you take a time for the naming of this module. i'm not sure of this one.

thanks !

pedrobaeza:
Difficult name... It can be `web_theme_high_contrast`.

mileo:
> web_theme_high_contrast

The main problem it's CE don't have any UX anymore. Why don't we make a theme that really improves CE's UX? Something like  web_theme_oca, that can have other dependencies like web_responsive, this module and others.

legalsylvain:
> The main problem it's CE don't have any UX anymore.

Sorry I don't get it, the UX for CE and EE looks the same. At least in form & tree views. the module proposed, for instance is installable and should works in in EE.

If you want to talk about a modulel named ``web_theme_oca`` could you open a dedicated issue to elaborate ? 

Thanks !

mileo (approved):
LGTM

image

youring:
IMHO, high contrast is usually used to describe color, not a border. 
I suggest **web_theme_classic_form** or simply **web_theme_classic**, inspired from _windows classic theme_, which can carry more css styles, layouts, or even features that worth to be ported from the old version.

rvalyi:
> IMHO, high contrast is usually used to describe color, not a border. 
> I suggest **web_theme_classic_form** or simply **web_theme_classic**, inspired from _windows classic theme_, which can carry more css styles, layouts, or even features that worth to be ported from the old version.

Not a big deal but I agree, web_theme_classic_form would be a better name.

pedrobaeza:
I also agree. I had no ideas when proposed the other name.

legalsylvain:
Hi thanks for all your proposal. ``web_theme_classic`` looks great ! 

I did the changes in the next days.

Note : ``web_theme_classic_form`` doesn't match, because it's not only about form view. (In fact I made some changes in tree editable view, in search view, etc...)

quentinDupont (approved):
Tested and it works and looks like very well :+1:

andreampiovesana (approved):
LGTM

houssine78 (approved):

houssine78:
@pedrobaeza This PR is ready for merging, could you please merge it ? Thanks !

PR #2318, opened 2022-10-14
[15.0][FIX] web_m2x_options did not work with widgets

CRogos:
There was a refactoring of the odoo JS code from v13->v14 which was not taken into account into the migration of the web_m2x_options.

This causes that the widget option did not work properly when web_m2x_options is installed. The red part is triggered by this widget and is missing. 
<field name="task_id" options="{'no_create_edit': True}" widget="task_with_hours"/>
image

I've updated the search function from the [odoo/odoo](https://github.com/odoo/odoo/blob/15.0/addons/web/static/src/legacy/js/fields/relational_fields.js#L565) and refactored the implementation.

This was also an issue in the v14 of web_m2x_options so a backport might be useful when this change is merged.

### DIFF
--- a/web_m2x_options/models/ir_config_parameter.py
+++ b/web_m2x_options/models/ir_config_parameter.py
@@ -12,6 +12,7 @@ class IrConfigParameter(models.Model):
             "web_m2x_options.limit",
             "web_m2x_options.search_more",
             "web_m2x_options.m2o_dialog",
+            "web_m2x_options.m2o_dialog_remove_buttons",
             "web_m2x_options.field_limit_entries",
         ]
         values = self.sudo().search_read([["key", "in", opts]], ["key", "value"])
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -11,6 +11,8 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
         relational_fields = require("web.relational_fields"),
         ir_options = require("web_m2x_options.ir_options");
 
+    const {escape} = owl.utils;
+    const {sprintf} = require("web.utils");
     var _t = core._t,
         FieldMany2ManyTags = relational_fields.FieldMany2ManyTags,
         FieldMany2One = relational_fields.FieldMany2One,
@@ -18,9 +20,15 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
         FormFieldMany2ManyTags = relational_fields.FormFieldMany2ManyTags;
 
     function is_option_set(option) {
-        if (_.isUndefined(option)) return false;
-        if (typeof option === "string") return option === "true" || option === "True";
-        if (typeof option === "boolean") return option;
+        if (_.isUndefined(option)) {
+            return false;
+        }
+        if (typeof option === "string") {
+            return option === "true" || option === "True";
+        }
+        if (typeof option === "boolean") {
+            return option;
+        }
         return false;
     }
 
@@ -29,40 +37,43 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
         init: function (parent, name, value) {
             this.name = name;
             this.value = value;
+
+            const buttons = [];
+
+            if (parent._canCreate() && parent._canShowDialogQuickCreate()) {
+                buttons.push({
+                    text: _t("Create"),
+                    classes: "btn-primary",
+                    close: true,
+                    click: function () {
+                        this.trigger_up("quick_create", {value: this.value});
+                    },
+                });
+            }
+
+            if (parent._canCreate() && parent._canShowDialogCreateAndEdit()) {
+                buttons.push({
+                    text: _t("Create and edit"),
+                    classes: "btn-primary",
+                    close: true,
+                    click: function () {
+                        this.trigger_up("search_create_popup", {
+                            view_type: "form",
+                            value: this.value,
+                        });
+                    },
+                });
+            }
+
+            buttons.push({
+                text: _t("Cancel"),
+                close: true,
+            });
+
             this._super(parent, {
                 title: _.str.sprintf(_t("Create a %s"), this.name),
                 size: "medium",
-                buttons: [
-                    {
-                        text: _t("Create"),
-                        classes: "btn-primary",
-                        click: function () {
-                            if (this.$("input").val()) {
-                                this.trigger_up("quick_create", {
-                                    value: this.$("input").val(),
-                                });
-                                this.close(true);
-                            } else {
-                                this.$("input").focus();
-                            }
-                        },
-                    },
-                    {
-                        text: _t("Create and edit"),
-                        classes: "btn-primary",
-                        close: true,
-                        click: function () {
-                            this.trigger_up("search_create_popup", {
-                                view_type: "form",
-                                value: this.$("input").val(),
-                            });
-                        },
-                    },
-                    {
-                        text: _t("Cancel"),
-                        close: true,
-                    },
-                ],
+                buttons: buttons,
             });
         },
         start: function () {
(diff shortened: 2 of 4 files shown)


CRogos:
Picked fix of #2304 and added configuration to sync (Create/Create+Edit) buttons in Dialog with dropdown (Create/Create+Edit) values.

I can squash the changes when reviewed.
@theangryangel could you have a look if this solution is working for you?

CRogos:
@nguyenhk99 , @goliveirab I saw you recently active on this module. Could please you review this fix?

CRogos:
Happy new year @pedrobaeza , could you have a look on this bug fixes. We are using it now for a while and it is working fine.

pedrobaeza:
I can't help too much on the JS side. Maybe @Tardo now that he is active again can help.

njeudy (approved):
Great, I just try and review code ! thanks

CRogos:
@sonhd91 could you please review this fix before migrating to v16?

CRogos:
@sanchonuria, @etobella, @kluna1998 could you also have a look on this fix. Same issue on v15 as https://github.com/OCA/web/pull/2409

etobella:
Well, it is not the same issue. The one fixed on 14.0 was related to M2ODialog, you try to fix the view and search. Isn't it?

CRogos:
@etobella, you are right. nevertheless would you be so kind and review this change? The bugfix is in our production since 4 months now, but here it is still not merged in master. 

I would like to port this module to Odoo 16, but first everything should be fixed.

CRogos:
@pedrobaeza do you have any advice how this PR can be processed faster? 
This PR is now open for 5 months and you might imagine that this is very frustrating.

pedrobaeza:
Yeah, I know, but the flows are there for ensuring certain stability in the changes that get the main repository. You can review other PRs, and ask in exchange that they review yours to speed up this process. You can also become a maintainer of the module (with enough merits), to exceptionally auto-merge these patches, making sure that the change is legit.

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +101,250 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
     });
 
     FieldMany2One.include({
+        custom_events: _.extend({}, FieldMany2One.prototype.custom_events, {
+            search_create_popup: "_onSearchCreatePopup",
+        }),
+
+        _canCreate: function () {
+            const self = this;
```
Tardo: 🔥 use 'this' directly

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +101,250 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
...
+            return self.can_create && !self.nodeOptions.no_create;
+        },
+
+        _canShowQuickCreate: function () {
+            const self = this;
```
Tardo: 🔥 use 'this' directly

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +101,250 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
...
+            if (!_.isUndefined(self.nodeOptions.no_quick_create))
+                return !self.nodeOptions.no_quick_create;
+
+            if (!m2x_create_undef) return m2x_create;
+
+            return true;
+        },
+
+        _canShowCreateAndEdit: function () {
+            const self = this;
```
Tardo: 🔥  use 'this' directly

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +101,250 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
...
+                    _.isUndefined(self.nodeOptions.create_edit) &&
+                    _.isUndefined(self.nodeOptions.no_create_edit),
+                m2x_create_edit_undef = _.isUndefined(
+                    ir_options["web_m2x_options.create_edit"]
+                ),
+                m2x_create_edit = is_option_set(
+                    ir_options["web_m2x_options.create_edit"]
+                );
+
+            if (!create_edit_undef) return create_edit;
```
Tardo: Please, use brackets

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +101,250 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
...
+
+            if (!m2x_create_edit_undef) return m2x_create_edit;
```
Tardo: Please, use brackets

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +101,250 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
...
+        },
+
+        _canFilterDialogRemoveButtons: function () {
+            var m2o_dialog_remove_buttons_undef = _.isUndefined(
+                    ir_options["web_m2x_options.m2o_dialog_remove_buttons"]
+                ),
+                m2o_dialog_remove_buttons = is_option_set(
+                    ir_options["web_m2x_options.m2o_dialog_remove_buttons"]
+                );
+            if (!m2o_dialog_remove_buttons_undef) return m2o_dialog_remove_buttons;
```
Tardo: Please, use brackets

Tardo (changes requested):
Thanks for the fix! 
For my part, only very minor changes are necessary.

I can't test it due to that runboat is not available. :/

CRogos:
@Tardo thank you for your feedback. I've updated the code and runboat is now also available.

Tardo (changes requested):
With the new changes the create/edit options are never shown:
noex

It should show:
m2x_act

CRogos:
> With the new changes the create/edit options are never shown: noex
> 
> It should show: m2x_act

Did you test on Runboat? 
In the System Parameters the config was set to 1 instead of True/False. I assume someone configured it manually?

image

Could you retest and maybe install it on a clean database?

Tardo:
@CRogos It was me who put 1 instead of 'True' :S. In the master branch this configuration is not needed.
Am I missing something?

CRogos:
> @CRogos It was me who put 1 instead of 'True' :S. In the master branch this configuration is not needed. Am I missing something?

No you can delete these keys. Than the behaviour should be the Odoo default. But putting invalid/unexpected values in the configuration might cause an unexpected behavior. ;)

Or am I wrong and 1 was an valid value before my change?

web_m2x_options/models/ir_config_parameter.py
```diff
@@ -12,6 +12,7 @@ def get_web_m2x_options(self):
             "web_m2x_options.limit",
             "web_m2x_options.search_more",
             "web_m2x_options.m2o_dialog",
+            "web_m2x_options.m2o_dialog_remove_buttons",
```
Tardo, line 15: Please, add documentation for this new parameter.

web_m2x_options/static/src/js/form.js
```diff
@@ -482,7 +476,7 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
             this._super(attrs);
             var limit = ir_options["web_m2x_options.field_limit_entries"];
             if (!_.isUndefined(limit)) {
-                attrs.limit = parseInt(limit);
+                attrs.limit = parseInt(limit, 10);
```
Tardo: I prefer the use of "unary plus" as it is more concise with the operation we want to do here.
```suggestion
                attrs.limit = +limit;
```

web_m2x_options/static/src/js/form.js
```diff
@@ -96,272 +107,255 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
...
+                    return;
+                }
+
                 new M2ODialog(this, this.string, this.$input.val()).open();
             }
         },
 
-        _search: function (search_val) {
-            var self = this;
+        _search: async function (searchValue = "") {
```
Tardo: From my tests, this kind of statement doesn't work very well with inheritance. I'm not sure if it's still happening. I would prefer a more classic use of promises (new Promise...).

Tardo (changes requested):
The tests I have done on runbot are correct :)

web_m2x_options/static/src/js/form.js
CRogos: Unary plus is not possible, because it is automatically replaced by pre-commit with:
Number(limit)

CRogos:
@Tardo this was a bigger refactoring as expected. Please re-test everything carefully.

Here are the changes compared with the previous PR: 
https://github.com/OCA/web/compare/adb8aeb..ca83e4950de4d97e115792c3ada856bf112b15c8

CRogos:
Not sure what to do with this PR. The module has now be migrated to v16 without these changes. As I will also switch to v16 soon, I will not do any effort into this issue anymore. 
Very frustrating that the issue could not be solved/merged in 6 months!

PR #2321, opened 2022-10-18
[MIG] web_refresher: Migration to 16.0

hugosantosred:


web_refresher/__init__.py
```diff
@@ -0,0 +1,9 @@
+##############################################################################
+#
+#    Copyright (C) 2019 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Samuel Fringeli <samuel.fringeli@me.com>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
```
legalsylvain: This empty file doesn't require that copyright text. you can let empty.

web_refresher/__manifest__.py
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Web Refresher",
+    "version": "16.0.1.0.0",
+    "author": "Compassion Switzerland, Odoo Community Association (OCA)",
```
legalsylvain, line 4: adopt an addon ?

legalsylvain (approved):
Hi. thanks for porting this module !
2 non blocking point inline.
Otherwise, LGTM. Runboat test. works as expected. Thanks.

hugosantosred:
Thanks for the quick review @legalsylvain!

Changes done

web_refresher/__manifest__.py
```diff
@@ -1,11 +1,11 @@
 {
     "name": "Web Refresher",
-    "version": "15.0.1.0.0",
-    "author": "Compassion Switzerland, Odoo Community Association (OCA)",
+    "version": "16.0.1.0.0",
+    "author": "Compassion Switzerland, FactorLibre, Odoo Community Association (OCA)",
```
pedrobaeza: This migration doesn't qualify for being co-authors of the module. You should add yourself to contributors instead.

pedrobaeza (changes requested):
Functional review OK, but check the comment inline.

web_refresher/__manifest__.py
hugosantosred: ok! Thanks for the review!, changes are already done

pedrobaeza (approved):

pedrobaeza:
This patch is missing from versions 15 and 16: #2255

Maybe some of you can include it.

PR #2324, opened 2022-10-26
[12.0][FIX] web_disable_browser_autocomplete: Qweb include path

JonathanStein:
The module tries to include qweb template from 'static/src/web', but the file is actually in 'static/src/xml'


tarteo (approved):
Thanks!

PR #2330, opened 2022-10-31
[14.0] [IMP] ckeditor: Make todoList available

lieblinger:
This PR adds a converter for TodoList. 

The default schema for TodoLists from CKEditor is not supported by Odoo, cause <input> elements get stripped away.

My PR adds functions to convert the TODOList schema back and forwards (Odoo <-> CKEditor). By default this PR saves Odoo's schema to the database and converts it back to the CKEditor schema for the frontend. This allows to keep TODOList working, even if someone switches back to Odoo's default editor.

What is not working: With Odoo's editor it was possible to check boxes in ReadOnlyView. This is not implemented by me and also not really needed in my case.

### DIFF
--- a/web_widget_ckeditor/static/src/js/field_ckeditor.js
+++ b/web_widget_ckeditor/static/src/js/field_ckeditor.js
@@ -138,6 +138,16 @@ odoo.define("web_widget_ckeditor.field_ckeditor", function (require) {
                 return this.$target.val();
             },
             /**
+             * @override
+             */
+            _setValue: function (value, options) {
+                let $obj = $('<div>').html(value)
+                if ($obj.find("ul.todo-list").length) {
+                    value = this._todoListCKEditor2Odoo($obj);
+                }
+                return this._super(value, options);
+            },
+            /**
              * Gets the CKEditor toolbar items configuration.
              * If not found, returns the default configuration.
              */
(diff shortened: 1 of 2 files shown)


ivantodorovich:
Hello, thanks a lot for this!

I tested it and it works 🎉 

I would rather keep the same Odoo core style, if possible. The scss you added don't really look like the "standard" to-do list, do you think that's possible?:

image


Also, please check pre-commit, as it's failing. ([for more information](https://odoo-community.org/blog/news-updates-1/black-isort-pre-commit-97))

tarteo (approved):
Awesome thanks!
image

ivantodorovich:
Hello @lieblinger 

I see you're struggling with pre-commit.

I recommend you to install it locally, as it will fix most of the "issues" automatically for you, and at the very least save you lots of time by running the checks in your terminal.

You can install it following the official docs: https://pre-commit.com/#install
Then, with a shell on your git clone, run `pre-commit install`. That will install the pre-commit hooks into your local clone.
The hooks will run before every new commit, but since you have already unstyled commits, you might need to run `pre-commit run -a` to scan the whole repo

bofiltd (approved):

bosd:
@lieblinger Can you squash the last 3 commits?

dreispt:
@lieblinger Plase squash your commits

PR #2332, opened 2022-11-09
[15.0][DCK] Updated from template

CRogos:
@pedrobaeza

### DIFF
--- a/web_m2x_options/__init__.py
+++ b/web_m2x_options/__init__.py
@@ -1,2 +1,1 @@
 from . import models
-from . import tests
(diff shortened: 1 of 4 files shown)


web_m2x_options/__init__.py
```diff
@@ -1,2 +1 @@
-from . import models
-from . import tests
+from . import models
```
pedrobaeza: Please add a trailing empty line

CRogos:
done.
Funny that now the test coverage is failing, because we deleted one line. Could we disarm this check and allow reductions by 0.5 or 1%. I had this problem now several times, when cleaning up code which is covered by unit tests.

pedrobaeza (approved):

pedrobaeza:
Yeah, I have the feeling that the check is not correct, but not matter, as that one is not taking into account on the bot merge command. It's just aesthetic one.

PR #2337, opened 2022-11-14
[14.0][FIX] Updated from template

victoralmau:
Updated from template

Ping @pedrobaeza 

@Tecnativa


pedrobaeza:
`[DCK]` doesn't make sense in GitHub. It should be `[FIX]`.

ap-wtioit (approved):

pedrobaeza:
Second test round is red.

ap-wtioit:
Tests fail in web_access_rule_buttons which is loaded right at the beginning of the test run, so the failure seems unrelated.

pedrobaeza:
Yeah, but we need to have it green to merge it...

victoralmau:
Some of the changes applied in `web_access_rule_buttons` are those made in https://github.com/OCA/web/pull/2294

pedrobaeza (approved):

web_access_rule_buttons/models/models.py
```diff
@@ -22,19 +22,16 @@ def check_access_rule_all(self, operations=None):
...
-            try:
-                self.check_access_rule(operation)
-            except exceptions.AccessError:
-                result[operation] = False
-            if (
-                self.is_transient()
-                or self.ids
-                and self.env.user.has_group("base.user_admin")
-            ):
+            if self.is_transient() or not self.ids:
```
ap-wtioit, line 25: i feel like changing the logic of another module deserves more attention than just a commit message that says fixed tests.

unfortunately i cannot just remove my approval for just the test fix commit: https://stackoverflow.com/a/44747392/925125

Do you know where the ignoring AccessErrors for `self.ids and self.env.user.has_group("base.user_admin")` came from (what the intention was)? (As this logic was removed in the fix tests commit)

ap-wtioit (changes requested):

web_access_rule_buttons/models/models.py
victoralmau: Sorry, this change is best explained at https://github.com/OCA/web/pull/2294.

ap-wtioit: Well, in my opinion it would have been better to explain that in the commit message, as those will be forever with the commit (unless someone uses rebase commits).

And if we start accepting commits within the OCA that copy contributions of other contributors without crediting them we are the same as Odoo but i guess that's how we are now.

I approved the initial commits but not the uncredited use of other peoples work.

web_access_rule_buttons/tests/test_access_rule_buttons.py
```diff
@@ -1,25 +1,38 @@
...
 
-class TestAccessRuleButtons(TransactionCase):
+class TestAccessRuleButtons(common.TransactionCase):
     def setUp(self):
-        super(TestAccessRuleButtons, self).setUp()
-
-        self.curr_obj = self.env["res.currency"]
+        super().setUp()
         self.curr_record = self.env.ref("base.USD")
+        new_test_user(self.env, login="test-user", groups="base.group_system")
```
simahawk, line 10: this is creating a new user at every test run :cry:

web_access_rule_buttons/models/models.py
simahawk: From my POV this is a small change that does not really imply "stealing work" from someone else. But I definitely understand your point. 

Yet, in both work, there's no way to reuse the same commit to address the same problem.

Because it has been poorly named and not separated.
The change is not to fix tests, is to fix the code and then adapt tests. Or vice-versa? :man_shrugging: 
Also, we'll have a conflict when trying to port this changes fwd unless the author of that PR does not cherry-pick this commit in his PR and discard his changes on this file.

pedrobaeza:
It was not the intention to hijack anything, but to fix the situation to be able to have this branch green to continue with other PRs not related to this module. Indeed, it would be preferable to commit with `--author`, but it was a forgot. As this is not latest branch, this mistake won't be propagated, as 15.0 contains the proper attribution, so I think it's not a big problem, but if @ivantodorovich, which is the original author, considers that the attribution should be preserved for such bits of code, this can be reverted and re-issue with such attribution.

PR #2341, opened 2022-11-22
[15.0][MIG] web_pivot_computed_measure: Migration to 15.0

CarlosRoca13:
cc @Tecnativa TT36578

Refactored to use OWL

git_computed_measure

ping @sergio-teruel @chienandalu


sergio-teruel (approved):
Functional review

CarlosRoca13:
I have added a new improvement to avoid searching for values using jquery but using the new feature that allows data-binding using state variables on js and t-model on xml

CarlosRoca13:
> I have added a new improvement to avoid searching for values using jquery but using the new feature that allows data-binding using state variables on js and t-model on xml

Removed, because when testing it with tours @chienandalu and me found some incompatibility and it made us suspect that it is not fully implemented on 15.0, so we preffer to make the change when trying to migrate to 16.0

pedrobaeza (approved):
/ocabot merge nobump

PR #2344, opened 2022-11-24
[FIX] web_widget_x2many_2d_matrix: make the possibility to have alternative values for headers of rows and columns working

vincent-hatakeyama:
This is a port of #2164 

As indicated in https://github.com/OCA/web/pull/2164#pullrequestreview-1037290313, it affects the version for Odoo 15.


legalsylvain (approved):
Thanks for porting the patch.

PR #2347, opened 2022-11-29
[15.0][IMP] web_ir_actions_act_view_reload: Change IrActionsActViewReload to AbstractModel

victoralmau:
Change IrActionsActViewReload to AbstractModel.

Please @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/web_ir_actions_act_view_reload/models/ir_actions_act_view_reload.py
+++ b/web_ir_actions_act_view_reload/models/ir_actions_act_view_reload.py
@@ -1,7 +1,9 @@
 from odoo import models
 
 
-class IrActionsActViewReload(models.Model):
+class IrActionsActViewReload(models.AbstractModel):
     _name = "ir.actions.act_view_reload"
-    _inherit = "ir.actions.actions"
     _description = "View Reload"
+
+    def _get_readable_fields(self):
+        return self.env["ir.actions.actions"]._get_readable_fields()


pedrobaeza:
I don't get why a model is needed. @ryanc-me Can you explain why?

victoralmau:
Any news about this?

chienandalu (commented):
I'd say for this case it could be handled with an `AbstractModel`. Somenthing like would have the same efffect without the need to create a table and all the things that go along with it:

```
class IrActionsActViewReload(models.AbstractModel):
    _name = "ir.actions.act_view_reload"
    _description = "View Reload"
```

victoralmau:
> I'd say for this case it could be handled with an `AbstractModel`. Somenthing like would have the same efffect without the need to create a table and all the things that go along with it:
> 
> ```
> class IrActionsActViewReload(models.AbstractModel):
>     _name = "ir.actions.act_view_reload"
>     _description = "View Reload"
> ```

Although it is a good idea, it is necessary to inherit from `ir.actions.actions`, and in doing so we have the following error.

_The abstract model 'ir.actions.act_view_reload' cannot inherit from the non-abstract model 'ir.actions.actions'._

pedrobaeza:
And adding an ACL this way is never a good idea.

chienandalu:
> Although it is a good idea, it is necessary to inherit from `ir.actions.actions`, and in doing so we have the following error.
> 
> _The abstract model 'ir.actions.act_view_reload' cannot inherit from the non-abstract model 'ir.actions.actions'._

Maybe you can fool it not inheriting from `ir.actions.actions`  but adding a `_get_readable_fields` method:

```
    def _get_readable_fields(self):
        return self.env['ir.actions.actions']._get_readable_fields()
```

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2349, opened 2022-11-29
web_widget_image_webcam: Migration to 16.0

emlafza:


web_widget_image_webcam/static/src/js/webcam_widget.js
```diff
@@ -0,0 +1,140 @@
...
+    var core = require("web.core");
+    var rpc = require("web.rpc");
+    var Dialog = require("web.Dialog");
+    const {patch} = require("@web/core/utils/patch");
+    const {ImageField} = require("@web/views/fields/image/image_field");
+
+    var _t = core._t;
+    var QWeb = core.qweb;
+
+    patch(ImageField.prototype, "web_widget_image_webcam", {
```
legalsylvain, line 23: Naive question : why do you replace "include" by "patch"? 
Is this syntax overloadable ? 

thanks !

emlafza: Hi @legalsylvain, I just wanted to use this 'patch' utility provided by Odoo described here https://www.odoo.com/documentation/16.0/developer/reference/frontend/patching_code.html#

about the rest, I suppose that functions added to ImageFIeld can be overloaded following the same procedure.

legalsylvain: @ivantodorovich any point of view regarding the topic patch vs include/extend ?

ivantodorovich: The change seems unnecessary whilst the object to patch extends from `OdooClass` (`web.Class`)
But this is super old code and odoo is moving out from this approach to a pure js class AFAIK, so we'll probably see the use of `patch` more and more

web_widget_image_webcam/static/src/js/webcam_widget.js
```diff
@@ -9,129 +9,124 @@ odoo.define("web_widget_image_webcam.webcam_widget", function (require) {
...
+                method: "get_webcam_flash_fallback_mode_config",
+            }).then(function (default_flash_fallback_mode) {
+                self.flash_fallback_mode = default_flash_fallback_mode;
+            });
+        },
 
+        _setWebcamParams() {
             // ::webcamjs:: < https://github.com/jhuckaby/webcamjs >
             // Webcam: Set Custom Parameters
+            var self = this;
```
ivantodorovich: This is totally unnecessary.
The `var self = this` trick is old, and useless in modern js. It was used to keep a reference for `this` within callback functions defined within the method.. so it's even not needed here, as there aren't any callbacks in this method.

Simply read `flash_fallback_mode` from `this`

web_widget_image_webcam/static/src/js/webcam_widget.js
```diff
@@ -9,129 +9,124 @@ odoo.define("web_widget_image_webcam.webcam_widget", function (require) {
...
+        setup() {
+            var self = this;
+            self._super.apply(self, arguments);
+            self.flash_fallback_mode = false;
+            rpc.query({
+                model: "ir.config_parameter",
+                method: "get_webcam_flash_fallback_mode_config",
+            }).then(function (default_flash_fallback_mode) {
+                self.flash_fallback_mode = default_flash_fallback_mode;
+            });
```
ivantodorovich: I'd put this rpc query outside the setup method. Simply put it in the global scope.
Because we don't need to read this value from the server for each field widget instance, we only need to read it once.

You can find an example here: https://github.com/OCA/web/blob/0253aa72c554cbaf68080725dd091ce74162bf9a/web_widget_ckeditor/static/src/js/field_ckeditor.js#L20

Anyway, not blocking as it was like this before the migration

ivantodorovich (approved):

web_widget_image_webcam/static/src/js/webcam_widget.js
emlafza: Hi @ivantodorovich ,

thank you for your review.

I'm not sure about your suggestion; I mean, the global variables should be set after browser reload, so when assets are loaded. if parameter is changed global variable will not be updated, there is something I'm missing about that?

web_widget_image_webcam/static/src/js/webcam_widget.js
emlafza: sure thanks! I'll fix that

web_widget_image_webcam/static/src/js/webcam_widget.js
ivantodorovich: Well, reading this setting requires 1 HTTP request, so it's better not to abuse it.
If you don't expect the setting to change often (and you don't on this case), then a browser reload is more than acceptable

emlafza: ok thanks. based on this I'll fix that too.

web_widget_image_webcam/static/src/js/webcam_widget.js
```diff
@@ -0,0 +1,141 @@
...
+
+    const getWebcamFlashFallbackModeConfig = rpc.query({
+        model: "ir.config_parameter",
+        method: "get_webcam_flash_fallback_mode_config",
+    });
+
+    patch(ImageField.prototype, "web_widget_image_webcam", {
+        setup() {
+            this._super.apply(this, arguments);
+            this.flash_fallback_mode = this._initWebcamFlashFallbackModeConfig();
```
ivantodorovich:
```suggestion
            this.flash_fallback_mode = this._initWebcamFlashFallbackModeConfig();
```

Since you're calling an async function, you'd have to await it in order to get the actual value.
As it is right now, `this.flash_fallback_mode` will be assigned a `Promise`.

You can only await within `async` functions, so it can't be done during `setup`.
Maybe the right place would be to override the `willStart` hook

```js
async willStart() {
    this.flash_fallback_mode = await this._initWebcamFlashFallbackModeConfig();
}
```

or just do this and you could remove the `_initWebcamFlashFallbackModeConfig` method

```js
async willStart() {
    this.flash_fallback_mode = await getWebcamFlashFallbackModeConfig;
}
```


Alternatively, you could do everything on `_setWebcamParams` if you make that method `async`, because it seems that's the only place where it's used

ivantodorovich (changes requested):

web_widget_image_webcam/static/src/js/webcam_widget.js
emlafza: thanks for the explanation

web_widget_image_webcam/static/src/js/webcam_widget.js
```diff
@@ -0,0 +1,132 @@
...
+                swfURL: "/web_widget_image_webcam/static/src/lib/webcam.swf",
+            });
+        },
+
+        onWebcamClicked() {
+            var self = this,
+                WebCamDialog = $(QWeb.render("WebCamDialog")),
+                img_data = false;
+
+            self._setWebcamParams();
```
ivantodorovich, line 45:
```suggestion
        async onWebcamClicked() {
            var self = this,
                WebCamDialog = $(QWeb.render("WebCamDialog")),
                img_data = false;

            await self._setWebcamParams();
```

Now, since `_setWebcamParams` is an async function, you'd probably want to await it too.
Otherwise the flow of execution will continue and the params would not be set.

In real life, this will probably never happen because the click will surely go after the read config promise has already been resolved (user action is probably slower than http response execution)

But if we want to be peeky, you should definitely await this.

If you do this, then it would be good to refactor this method a little bit to make use of the `await` keyword. Things you could achieve with it:
- Remove the `var self = this` hack (not needed in modern js -- if callbacks are defined as arrow functions)
- Convert callbacks to arrow functions
- Convert `then()` to `await`
- Possibly move the Dialog's click buttons callbacks to separate methods (ie: `onDialogTakeSnapshotButtonClicked`, `onDialogSaveButtonClicked`). This would make the function much shorter, and easy to maintain.


Anyway, this is all nitpicking.

ivantodorovich (approved):

legalsylvain (approved):
Functional review, LGTM Thanks !

PR #2350, opened 2022-11-29
[15.0][IMP] web_widget_many2one_simple: Add a demo view to better test the functionality.

victoralmau:
Add a demo view to better test the functionality.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT40367

### DIFF
--- a/web_widget_many2one_simple/README.rst
+++ b/web_widget_many2one_simple/README.rst
@@ -64,6 +64,13 @@ Usage
 You need to declare the usage of the new widget as follows:
 ``<field name="xxx" widget="many2one_simple" options="...">``
 
+If you have a database with demo data, you can test this widget following these instructions:
+
+#. Go to Contacts.
+#. Create or edit a record.
+#. Set "Industry" field (in Sales & Purchase tab).
+#. Only if you set the correct industry name ("Administrative" for example), it will be defined. Otherwise, it will try to create one with that name.
+
 Example:
 ~~~~~~~~
 
--- a/web_widget_many2one_simple/__manifest__.py
+++ b/web_widget_many2one_simple/__manifest__.py
@@ -10,6 +10,9 @@
     "depends": ["web"],
     "installable": True,
     "maintainers": ["Tardo"],
+    "demo": [
+        "demo/res_partner_view.xml",
+    ],
     "assets": {
         "web.assets_backend": [
             "web_widget_many2one_simple/static/src/css/many2one_simple.scss",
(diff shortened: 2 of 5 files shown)


web_widget_many2one_simple/README.rst
```diff
@@ -61,6 +61,9 @@ Options:
 Usage
 =====
 
+Go to Contacts > Create or edit a record and set "Industry" field (in Sales & Purchase tab).
```
pedrobaeza: Put this after the other paragraph indicating that using demo data, the widget can be used this way.

web_widget_many2one_simple/README.rst
```diff
@@ -61,6 +61,9 @@ Options:
...
+Only if you set the correct indstry name (Administrative for example) will it be defined, otherwise it will try to create one with that name.
```
jcdrubay:
```suggestion
Only if you set the correct industry name (Administrative for example) will it be defined, otherwise it will try to create one with that name.
```

pedrobaeza:
```suggestion
Only if you set the correct industry name ("Administrative" for example), it will be defined. Otherwise, it will try to create one with that name.
```

chienandalu (approved):
Thanks! :+1:

web_widget_many2one_simple/README.rst
```diff
@@ -64,6 +64,9 @@ Usage
 You need to declare the usage of the new widget as follows:
 ``<field name="xxx" widget="many2one_simple" options="...">``
 
+Go to Contacts > Create or edit a record and set "Industry" field (in Sales & Purchase tab).
```
pedrobaeza: Still missing the explanation about demo data:

```suggestion
If you have a database with demo data, you can test this widget following these instructions:

#. Go to Contacts.
#. Create or edit a record.
#. Set "Industry" field (in Sales & Purchase tab).
...
```

web_widget_many2one_simple/README.rst
```diff
@@ -64,6 +64,12 @@ Usage
 You need to declare the usage of the new widget as follows:
 ``<field name="xxx" widget="many2one_simple" options="...">``
 
+If you have a database with demo data, you can test this widget following these instructions:
+#. Go to Contacts.
```
pedrobaeza, line 69: Incorrect syntax: missing empty line. This is not how I put the suggestion...

pedrobaeza (approved):

PR #2356, opened 2022-12-06
[MIG] web_domain_field: Migration to 16.0

imarulbala:


web_domain_field/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2017 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Web Domain Field",
+    "summary": """
+        Use computed field as domain""",
+    "version": "16.0.1.0.1",
```
StefanRijnhart: As per guideline (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#tasks-to-do-in-the-migration)
```suggestion
    "version": "16.0.1.0.0",
```

StefanRijnhart (approved):
Thanks! Tested and code reviewed. I noticed that https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/legacy/js/core/py_utils.js had not changed compared to Odoo 15, so it seems correct that this module's overwrites are not changed either.

pedrobaeza:
FYI, we can play with pyjs support in domain expressions and don't require this module. Check this example:

https://github.com/OCA/rma/pull/326

It's a matter of taste sometimes, but I tend to avoid extra dependencies at maximum.

StefanRijnhart:
@pedrobaeza interesting! It would be good to put an example like that in this module's README, under _Do you really need this module?_ @imarulbala

nicolascol:
Hi, I need a runboat to test this PR. Could someone create one?

FernandoRomera (approved):
Test in local.
Ok.

pedrobaeza:
Right now is not possible to test it on runboat, as there's a problem that should be fixed ASAP, but it depends on a voluntary effort by the OCA infrastructure maintainers, so it would be done when possible.

FernandoRomera:
@pedrobaeza 
Can you merge this PR, please?.

pedrobaeza:
It lacks the clarification that Stefan mentions.

StefanRijnhart:
@pedrobaeza I pushed such a clarification myself now. For the record, I don't think Odoo's py.js is the same as http://pyjs.org but it does not matter for the argument of course.

StefanRijnhart:
And welcome back Runboat!

PR #2361, opened 2022-12-13
[15.0][FIX] web_widget_x2many_2d_matrix: fix totals option

vincent-hatakeyama:
Also refactored to use underscore method rather than redefining a similar one.

Note that there was an issue with the original code:
```javascript
                node.show_row_totals ||
                    this.is_aggregatable(field_defs[this.field_value])
                    ? "1"
                    : ""
```

Due to precedence of the operators, `node.show_row_totals || this.is_aggregatable(field_defs[this.field_value])` is evaluated before returning "1". Then it would be converted to a boolean. If the option is set in the view that code would ignore it completely and would display row totals.

The issue was introduced in a5e6acc3a64626cf99b3d31f2f295be16f41b27b / #1373.

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
+++ b/web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
@@ -39,8 +39,8 @@ odoo.define("web_widget_x2many_2d_matrix.widget", function (require) {
             this.field_y_axis = node.field_y_axis || this.field_y_axis;
             this.field_label_x_axis = node.field_label_x_axis || this.field_x_axis;
             this.field_label_y_axis = node.field_label_y_axis || this.field_y_axis;
-            this.x_axis_clickable = this.parse_boolean(node.x_axis_clickable || "1");
-            this.y_axis_clickable = this.parse_boolean(node.y_axis_clickable || "1");
+            this.x_axis_clickable = _.str.toBool(node.x_axis_clickable || "1");
+            this.y_axis_clickable = _.str.toBool(node.y_axis_clickable || "1");
             this.field_value = node.field_value || this.field_value;
             // TODO: is this really needed? Holger?
             for (var property in node) {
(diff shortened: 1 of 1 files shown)


web_widget_x2many_2d_matrix/static/src/js/widget_x2many_2d_matrix.js
```diff
@@ -39,8 +39,8 @@ odoo.define("web_widget_x2many_2d_matrix.widget", function (require) {
             this.field_y_axis = node.field_y_axis || this.field_y_axis;
             this.field_label_x_axis = node.field_label_x_axis || this.field_x_axis;
             this.field_label_y_axis = node.field_label_y_axis || this.field_y_axis;
-            this.x_axis_clickable = this.parse_boolean(node.x_axis_clickable || "1");
-            this.y_axis_clickable = this.parse_boolean(node.y_axis_clickable || "1");
+            this.x_axis_clickable = _.str.toBool(node.x_axis_clickable || "1");
```
legalsylvain, line 42: @hpar any suggestion regarding the use of _ ?

vincent-hatakeyama: It is used in the web module, for example:

https://github.com/odoo/odoo/blob/15.0/addons/web/static/src/legacy/js/views/form/form_controller.js#L632
```python
            ev.data.callback(_.str.toBool(isFocused));
```

legalsylvain:
@ChrisOForgeFlow : could you take a look on this one ?

norlinhenrik:
I have tested and I confirm that this PR fixes the `show_row_totals` and `show_column_totals` options.

legalsylvain:
Thanks.

/ocabot merge patch

@vincent-hatakeyama. Could you forward this patch upstream if required. (on 16.0 version : https://github.com/OCA/web/tree/16.0/web_widget_x2many_2d_matrix)

PR #2363, opened 2022-12-14
[15.0][IMP] web_widget_open_tab: Support char field type.

amh-mw:


amh-mw:
Initial investigation showed that the field widget doesn't actually care what field you use for the link, so the supportedTypes: integer restriction is unnecessary.

https://github.com/OCA/web/blob/e0c4ba1166ec3ec5c697970b596dcafa600bfa1c/web_widget_open_tab/static/src/js/widget.js#L26

It was straightforward to display the record name, which is much nicer from a user experience perspective.

Screenshot 2022-12-14 at 9 40 11 AM

amh-mw:
My reluctance to merge this stems from the fact that I think this functionality would be great across most field types, but the current implementation extends AbstractField, not any of the specific field types. I don't have the bandwidth right now to convert this to some sort of mixin.

PR #2368, opened 2022-12-27
[FIX] web_timeline group_bys default array if undefined

KKamaa:
So there is a bug when you open a timeline record then go back to the timeline view, the **n_group_bys** is undefined and pops up an error. So I added a default definition of empty array if the timeline's **n_group_bys** is undefined. @pedrobaeza @thomaspaulb


pedrobaeza (approved):

thomaspaulb (approved):

CarlosRoca13 (approved):

pedrobaeza:
Should this be fixed in v15 as well?

PR #2299, opened 2022-09-23
[14.0][FIX] web_drop_target: check controller is not null

Saran440:
fix it from issue https://github.com/OCA/web/issues/2298

PR #2369, opened 2022-12-27
[FIX] drag_drop element cannot be seen, if check for undefined value

KKamaa:
There was a merge fix [PR](https://github.com/OCA/web/pull/2299) it lead to an issue where the drag drop panel could not be accessed any more. I did a fix here things are back to normal, I couldn't regenerate the controller issue, in case anyone finds it let me know what you did to get it, but for now the module works well unless anything pops up. @tarteo @pedrobaeza another bug which I found and hopefully fixed.

### DIFF
--- a/web_drop_target/static/src/js/web_drop_target.js
+++ b/web_drop_target/static/src/js/web_drop_target.js
@@ -151,18 +151,9 @@ odoo.define("web_drop_target", function (require) {
          * @private
          * @param {MouseEvent} ev
          */
-        _onBodyFileDragover: function (ev) {
+        _onBodyFileDragover: function(ev) {
             ev.preventDefault();
-            const actionManager = this.findAncestor(function (ancestor) {
-                return ancestor instanceof ActionManager;
-            });
-            const controller = actionManager.currentDialogController;
-            if (
-                _.isEmpty(this._get_drop_items(ev)) &&
-                this._checkDragOver() &&
-                (controller === undefined ||
-                    (controller && controller.jsID === this.controllerID))
-            ) {
+            if (_.isEmpty(this._get_drop_items(ev)) && this._checkDragOver()) {
                 const drop_zone_offset = this.$drop_zone.offset();
                 const overlay_css = {
                     top: drop_zone_offset.top,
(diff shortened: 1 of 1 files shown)


KKamaa:
@tarteo @thomaspaulb hey this fix also needs to be added.

KKamaa:
@pedrobaeza this fix also needs to be checked.

pedrobaeza:
What means "drag drop panel could not be accessed any more"? Please give steps to reproduce the problem. And the patch seems very big and removing things in a first sight, but I'm not JS expert.

Maybe @etobella can help here.

KKamaa:
> What means "drag drop panel could not be accessed any more"? Please give steps to reproduce the problem. And the patch seems very big and removing things in a first sight, but I'm not JS expert.
> 
> Maybe @etobella can help here.

So the issue is that the pop up showing the drag drop doesn't show when you drag. This fix was to return the functionality of the pop up allowing a drag drop attachment to work.

web_drop_target/static/src/js/web_drop_target.js
```diff
@@ -153,16 +152,7 @@ odoo.define("web_drop_target", function (require) {
          */
         _onBodyFileDragover: function (ev) {
             ev.preventDefault();
-            const actionManager = this.findAncestor(function (ancestor) {
-                return ancestor instanceof ActionManager;
-            });
-            const controller = actionManager.currentDialogController;
```
thomaspaulb, line 159: @etobella We need your input here. If I compare Runboat of current 14.0 with Runboat of this PR, and in both I install CRM and open the page of any random `res.partner`, then purely functionally, on 14.0 the dropzone does not appear on drag and this PR fixes that. Upon Javascript debugging, I see that in the 14.0 case, `actionManager.currentDialogController` is `null`. So this whole code does not get executed.

@KKamaa claims this is since #2299 but to me it seems that it's introduced in #2066 instead. Since you're the one that proposed that, could you give your opinion on removing these checks in the way that was done here? Maybe there's a simpler solution to the problem?

thomaspaulb: (in other words, functional review OK, but I'm also a bit confused about the code change)

etobella (approved):
Tested on runbot and everything seems to work as expected, so, IMO, :+1:

thomaspaulb (approved):

PR #2371, opened 2022-12-28
[14.0][FIX] web_responsive: slashes in app names

chienandalu:
When the app name had a slash ("/") on it we couldn't properly find it by name. This alternative doesn't depend on such weak rule. We'll find out the root app element for our submenu from the data already available in the widget.

cc @Tecnativa TT40970

ping @pedrobaeza


pedrobaeza (approved):
Tested on runboat and working.

Let's fast-track it to fix the problem:

/ocabot merge patch

Is this present in v15?

chienandalu:
> Is this present in v15?

No as it relies in OWL and the menu events are handled differently.

PR #2379, opened 2023-01-04
[15.0][IMP] web_widget_x2many_2d_matrix: Include currency for formatting monetary values in row/column sums

vincent-hatakeyama:


pedrobaeza:
This is not `[ADD]`, but `[IMP]`, as you are not adding a new module, but adding a feature on an existing one.

vincent-hatakeyama:
> This is not `[ADD]`, but `[IMP]`, as you are not adding a new module, but adding a feature on an existing one.

Fixed.

legalsylvain:
@ChrisOForgeFlow : could you take a look on this one ?

norlinhenrik:
LGTM
Regarding `show_row_totals `and `show_column_totals `README says: In case of monetary values, it is expected that the currency is the same for all cells; otherwise set this to False.
image

norlinhenrik:
FYI When I apply this PR to version 16.0, I get an OwlError.
`Caused by: TypeError: Cannot read properties of undefined (reading 'fields')`

norlinhenrik (approved):

PR #2396, opened 2023-01-27
[15.0][FIX] web_responsive: only change width of main control panel

### LINKED ISSUE #2395: [15.0] web_responsive: arrows unavailable in search more view
## Module

The bug is in web_responsive. When uninstalled, there is no issue anymore.

## Describe the bug

When using the *Search More…* entry in a many2one field, the arrows are not visible at some resolution or zoom level.

## To Reproduce

**Affected versions**: 15.0

I did not test in previous version.

Steps to reproduce the behavior in the runboat:
1. In quotations go to S00002
2. In the Customer field select *Search More…*
3. The button to go to the next or previous page are hidden behind the list.
   Depending on the size of the window, it can sometimes be visible. The issue also appears when switching languages when the text is longer than the English version. In my case, the issue is visible right away.

Screenshot of the view in runboat

**Expected behavior**

I expect the button to be visible, as they are in runbot following the same steps.

Odoo 15.0 runbot screenshot

**Additional context**

The screenshot are from Firefox, but I also see the issue in Chromium. My screen resolution is 1920×1080.
A colleague has the issue with Chrome too.

vincent-hatakeyama:
The Search More panel is incorrectly displayed due to the reduced screen so only change width on the main screen.

Fixes #2395

### DIFF
--- a/web_responsive/static/src/components/control_panel/control_panel.scss
+++ b/web_responsive/static/src/components/control_panel/control_panel.scss
@@ -19,12 +19,12 @@
     }
     // For FULL HD devices
     @media (min-width: 1900px) {
-        .o_cp_top_left,
-        .o_cp_bottom_left {
+        .o_action_manager & .o_cp_top_left,
+        .o_action_manager & .o_cp_bottom_left {
             width: 60%;
         }
-        .o_cp_top_right,
-        .o_cp_bottom_right {
+        .o_action_manager & .o_cp_top_right,
+        .o_action_manager & .o_cp_bottom_right {
             width: 40%;
         }
     }


pedrobaeza:
I don't think removing it is the right solution, but restricting it to the main screen, not the popup ones, which is the problem. That media query is done for very big displays (but as said, the main view).

web_responsive/static/src/components/control_panel/control_panel.scss
```diff
@@ -18,16 +18,6 @@
         }
     }
     // For FULL HD devices
-    @media (min-width: 1900px) {
-        .o_cp_top_left,
```
SplashS, line 22: As @pedrobaeza wrote it's better to avoid using this rule in popup rather than removing it at all.
You can change this part of code like this:
```scss
            .o_action_manager & .o_cp_top_left,
            .o_action_manager & .o_cp_bottom_left {
                width: 40%;
            }
            .o_action_manager & .o_cp_top_right,
            .o_action_manager & .o_cp_bottom_right {
                width: 60%;
            }
```

vincent-hatakeyama: I’ve tested the solution, it looks like it works. I’ve kept the original width (the comment below was inverting them, I do not think that was intentional).

SplashS (approved):
Tested on runboat.
Works as expected. Only main screen has shrunk width of search bar.

PR #2399, opened 2023-01-31
[14.0][FIX] #3039 support_branding ACL bug

KKamaa:
A bypass fix for users with low ACL security settings

### DIFF
--- a/support_branding/models/res_company.py
+++ b/support_branding/models/res_company.py
@@ -2,30 +2,23 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, models
-from odoo.exceptions import AccessError, UserError
+from odoo.exceptions import UserError, AccessError
 
 
 class ResCompany(models.Model):
     _inherit = "res.company"
 
     @api.model
-    def get_ir_config_param_data(self, key):
-        if not self.env.user.has_group("base.group_user"):
-            raise AccessError(
-                _(
-                    "You are not allowed to access this "
-                    "functionality, please contact Admin for "
-                    "more support"
-                )
-            )
-        try:
-            self.env.cr.execute(
-                "select value from ir_config_parameter where " "key=(%s);", (key,)
-            )
-            res = self.env.cr.fetchone()
-        except Exception as e:
-            raise UserError(_("Error: %s" % e))
-        else:
-            if res:
-                return "%s" % res
-            return ""
+    def get_support_branding_config_param_data(self):
+        if not self.env.user.has_group('base.group_user'):
+            raise AccessError(_('You are not allowed to access this '
+                                'functionality, please contact Admin for '
+                                'more support'))
+        self.env.cr.execute(
+           "select key, value from ir_config_parameter where key ilike "
+           "'%support_%';"
+        )
+        res = self.env.cr.dictfetchall()
+        if any(res):
+            support_vals = {x['key']: x['value'] for x in res}
+            return support_vals
--- a/support_branding/models/res_company.py
+++ b/support_branding/models/res_company.py
@@ -1,0 +1,29 @@
+# Copyright 2023 Sunflower IT
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+
+from odoo import models, api
+
+_logger = logging.getLogger(__name__)
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    @api.model
+    def get_ir_config_param_data(self, key):
+        try:
+            self.env.cr.execute("select value from ir_config_parameter where "
+                                "key='%s';" % key)
+            res = self.env.cr.fetchone()
+        except Exception as e:
+            _logger.error('\n\n ERROR: %s \n\n', e)
+            return ''
+        else:
+            if res:
+                return '%s' % res
+            return ''
+
+
+
--- a/support_branding/models/res_company.py
+++ b/support_branding/models/res_company.py
@@ -20,7 +20,7 @@ class ResCompany(models.Model):
             )
         self.env.cr.execute(
             "select key, value from ir_config_parameter where key ilike "
-            "'%support_%';"
+            "'support_%';"
         )
         res = self.env.cr.dictfetchall()
         if any(res):
--- a/support_branding/models/res_company.py
+++ b/support_branding/models/res_company.py
@@ -3,27 +3,25 @@
 
 import logging
 
-from odoo import models, api
+from odoo import api, models
 
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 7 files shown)


KKamaa:
@thomaspaulb did the fix here though just bypassed by adding a query through a function in res_company model

support_branding/static/src/js/res_config_edition.js
```diff
@@ -7,22 +7,22 @@ odoo.define("support_branding.ResConfigEdition", function (require) {
         willStart: function () {
             var self = this;
             var def_1 = this._rpc({
-                model: "ir.config_parameter",
-                method: "get_param",
+                model: "res.company",
+                method: "_get_support_branding_vals",
```
thomaspaulb: You're calling `_get_support_branding_vals`, but where is it defined? I see only a definition of the more general `get_ir_config_param_data`, with which you can get any key

support_branding/models/res_company.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    @api.model
+    def get_ir_config_param_data(self, key):
```
thomaspaulb: Anyone who can read the company record, can now call this function, so it's not very safe - `ir.config.parameter` can contain all kinds of stuff: API keys, passwords, ... we don't want all portal users to be able to read those.

What I meant with "add these settings to a proper res.config.settings screen and then get them by reading from the company record instead", was, that we define the `support_company_*` fields on `res.company`. But that was just one idea out of many possibilities. Maybe it's enough if you just define `get_support_branding_data` instead, which gives out a dictionary with _only_ the support branding keys, of which it is OK if everyone reads them. Or maybe just backend users is enough:

```
if not user.has_group('base.group_user'):
    raise AccessError(...)
```

thomaspaulb (changes requested):

support_branding/static/src/js/res_config_edition.js
KKamaa: fixed this, it should be `get_ir_config_param_data`

support_branding/models/res_company.py
KKamaa: @thomaspaulb added the above and a unit test for it testing with portal user demo

thomaspaulb: Still not secure. Now, anyone who's an internal user, can read all key/value pairs in ir.config.parameter. You need to make it into a function that returns only the support_branding keys.

KKamaa: @thomaspaulb I switched to key/value pair based on search for support_branding values as they are stored using the prefix `support_` so query to `key ilike %support_%` returns the values then I use them later.

KKamaa:
@tarteo @hbrunn maybe a review here

support_branding/models/res_company.py
```diff
@@ -0,0 +1,28 @@
...
+                _(
+                    "You are not allowed to access this "
+                    "functionality, please contact Admin for "
+                    "more support"
+                )
+            )
+        self.env.cr.execute(
+            "select key, value from ir_config_parameter where key ilike "
+            "'%support_%';"
+        )
```
hbrunn, line 23: why don't you just do a `sudo().search()` here? I think it would also be good to search for `['|', ('company_id', '=', False), ('company_id', 'in', company_ids)]` with order `company_id nulls last` to support different branding per company

hbrunn (changes requested):

thomaspaulb (approved):

support_branding/models/res_company.py
KKamaa: Not sure why we would use **company_id** because its not included in the ```ir_config_parameter``` model. @thomaspaulb @hbrunn unless I'm missing a bigger picture here with the requested changes.

ddejong-therp:
Something is still not working. When you click "Test Support Branding", the background color is not applied to the popup anymore. Could it be that the `support_branding_color` is not loading correctly?

support_branding/models/res_company.py
ddejong-therp: From what I understand, currently the code uses the same config parameters for each company, like it used to do anyway. Making the module with multi-company support would not be part of the fix I guess, but it would be cool.

If you would do that, maybe what could be done is use queries like this:
```
self.env.cr.execute(
    "select key, value from ir_config_parameter where key ilike "
    "'%support_%i_%';" % company_id
)
```
But then I suppose you need to move the view away from the general settings and maybe into the company form, then you should have a company id. Or maybe leave it and use `self.env.context['allowed_company_ids']` somehow?

ddejong-therp: From what I understand, currently the code uses the same config parameters for each company, like it used to do anyway. Making the module with multi-company support would not be part of the fix I guess, but it would be cool.

If you would do that, maybe what could be done is use queries like this:
```
self.env.cr.execute(
    "select key, value from ir_config_parameter where key ilike "
    "'%support_%i_%';" % company_id
)
```
But then I suppose you need to move the view away from the general settings and maybe into the company form, then you should have a company id, at which point you could even move the parameters to the `res.company` model itself as fields. Or maybe leave it and use an id from `self.env.context['allowed_company_ids']` somehow, but I believe that would be confusing to the end user.

support_branding/models/res_company.py
```diff
@@ -0,0 +1,28 @@
...
+            raise AccessError(
+                _(
+                    "You are not allowed to access this "
+                    "functionality, please contact Admin for "
+                    "more support"
+                )
+            )
+        self.env.cr.execute(
+            "select key, value from ir_config_parameter where key ilike "
+            "'%support_%';"
```
ddejong-therp: Also, the parameters are saved like so: `support_%`, so the `%` in the beginning _could_ be removed, but it is not a big deal. However, perhaps it would have been more neat if it was saved as `support_branding.%`, but that is another issue by itself and it would require a migration script.

support_branding/models/res_company.py
thomaspaulb: In general it seems to me a too obscure feature, since this module is targeted at explaining users where to direct support queries, and for 90% of cases, even in a multi-company instance, that whole instance is supported by just one Odoo partner.

hbrunn: fine with me

bamidev:
alright, LGTM. The mutli-company feature is not part of this fix anyway, whether it was a good idea or not.

bamidev (approved):
alright, LGTM. The mutli-company feature is not part of this fix anyway, whether it was a good idea or not.

hbrunn (approved):

hbrunn:
@ddejong-therp if you you don't mind I'd say
/ocabot merge patch

PR #2401, opened 2023-02-02
[FIX] web_responsive: Selection of menu fails

etobella:
Without this change the selected menu was not properly showed:

Old behavior:

Peek 2023-02-02 14-08

### DIFF
--- a/web_responsive/static/src/js/web_responsive.js
+++ b/web_responsive/static/src/js/web_responsive.js
@@ -247,7 +247,7 @@ odoo.define("web_responsive", function (require) {
                     continue;
                 } else {
                     const root = this._isInMenuTree(node.children, id);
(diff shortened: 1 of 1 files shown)


Tardo:
Can you give me some more specific steps? The changes apparently look correct but I can't replicate the bug in runboat. Thanks!

etobella:
If you write the text it will not change the menu. It doesn't happen always. For example if you write Sale and select Sale it will not open properly. If you write technical and select a menu of technical it will not open too.

pedrobaeza:
Please check if this compatible with #2371

etobella:
It seems so

The main change introduced on #2371 is maintained. It just fixes a small bug :smile:

pedrobaeza (approved):

PR #2407, opened 2023-02-09
[15.0][ADD] web_widget_remote_measure: New module

chienandalu:
cc @Tecnativa TT38745

ping @carlosdauden 

Example of behavior in the oncoming `stock_remote_device`:

remote-device-example


pedrobaeza:
Being web client implementation, I think it should be called `web_widget_remote_measure` or similar.

chienandalu:
Ok, changed :)

web_widget_remote_measure/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+To configure your remote devices:
+
+#. Go to *Settings > Technichal > Resource > Remote devices*
```
legalsylvain:
```suggestion
#. Go to *Settings > Technical > Resource > Remote devices*
```

web_widget_remote_measure/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module allows to input data from remote decives in your network. Currently, only
```
legalsylvain:
```suggestion
This module allows to input data from remote devices in your network. Currently, only
```

web_widget_remote_measure/readme/USAGE.rst
```diff
@@ -0,0 +1,15 @@
+The remote device has to be in the users network so their web clients can reach them.
+
+In order to test a device you can:
+
+#. Go to *Settings > Technichal > Resource > Remote devices*
```
legalsylvain:
```suggestion
#. Go to *Settings > Technical > Resource > Remote devices*
```

pedrobaeza:
I wonder the overlap with https://odoo-community.org/shop/remote-base-3251?search=remote#attr=21539 and indeed seeing this code, it's including in the same module the base definition with the widget.

pedrobaeza:
This is also a very specific "remote" thing, so maybe including `websocket` reference or something can be convenient.

chienandalu:
> I wonder the overlap with https://odoo-community.org/shop/remote-base-3251?search=remote#attr=21539 and indeed seeing this code, it's including in the same module the base definition with the widget.

I'd say that module's purpose is to store the accesses to the Odoo server

chienandalu:
> This is also a very specific "remote" thing, so maybe including `websocket` reference or something can be convenient.

It can be extended to a Webservices interface in the future if needed

pedrobaeza:
Then `remote_webservice` prefix?

chienandalu:
But it could be either webservices or websockets

pedrobaeza:
OK, I don't insist then.

web_widget_remote_measure/readme/CONFIGURE.rst
chienandalu: Thanks :)

rousseldenis:
And why not envisaging connecting devices like IOT (that could be connected to POS) and more specifically to pywebdriver ?

That allows you to implement protocols there. With the advantage of being available for all flows you want.

chienandalu:
> And why not envisaging connecting devices like IOT (that could be connected to POS) and more specifically to pywebdriver ?
> That allows you to implement protocols there. With the advantage of being available for all flows you want.

Connection to any remote measure device (IoT or the device itself via web services) would be possible if implemented but we want to keep the infrastructure required to implement this in the customer's setup as simple as possible.

For our current approach, adding an IoT (or pywebdriver) in the middle is an extra point of potential failure that can complicate our maintenance.

rousseldenis:
> > And why not envisaging connecting devices like IOT (that could be connected to POS) and more specifically to pywebdriver ?
> > That allows you to implement protocols there. With the advantage of being available for all flows you want.
> 
> Connection to any remote measure device (IoT or the device itself via web services) would be possible if implemented but we want to keep the infrastructure required to implement this in the customer's setup as simple as possible.
> 
> For our current approach, adding an IoT (or pywebdriver) in the middle is an extra point of potential failure that can complicate our maintenance.

So, implementing a particular protocol should maybe go in another module

bosd:
Let us know when you're ready

carlosdauden (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #2409, opened 2023-02-16
[FIX] web_m2x_options: M2Dialog not working properly

sanchonuria:
When opening the M2Dialog, the value was not propagated,

As you can see in the example, the value is `MY LOT NAME` but it is not inherited.
Peek 2023-02-16 12-39

This change fixes the bug


kluna1998 (approved):

etobella (approved):
LGTM Tested on runbot

Issue #2410, opened 2023-02-16
[15.0] web_responsive module Blank page


bvanjeelharia:
## Module
web_responsive

## Describe the bug
A blank page opens up on pressing Alt + A (shortcut for home page) if there is no app opened before.

## To Reproduce
**Affected versions**:15.0

1.Login to an odoo instance with web_responsive module installed
2.Once the home page(backend) is loaded, press Alt+A 
3.A blank page would open up

**Expected behavior**
Nothing should happen if there's no app opened prior to pressing Alt+A on home page.

pedrobaeza:
Is this reproducible on runboat?

bvanjeelharia:
@pedrobaeza , yes its reproducible on runboat

pedrobaeza:
OK, seen. It's the same as clicking on the AppSwitcher icon. Which behavior do you expect? I don't see this as something bad, as we don't need to pre-load for example the discuss app if you are not going to navigate to it. I see that you say that is fixed for 16.0. How did you fix it?

bvanjeelharia:
I expect the system not to take me to a blank page.For 16.0, I have made changes so that the system doesn't open up the blank page. Instead it stays on the home page.

pedrobaeza:
OK, then please apply the same fix in a new PR to 15.0. Another controversial feature is to leave the AppSwitcher opened when landing on Odoo after login. I don't like too much that option.

bvanjeelharia:
Ok, I will apply the same fix in a PR.Once I apply this change the AppSwitcher wouldn't be opened when landing on Odoo after login

PR #2411, opened 2023-02-16
[15.0][FIX] Fix problem of blank page when using web_responsive

bvanjeelharia:
This PR should fix the error of blank page mentioned in https://github.com/OCA/web/issues/2410


pedrobaeza:
I have tested this PR, and it's still showing the AppSwitcher, but no icon shown. Maybe you misunderstood me when I refer to AppSwitcher. For me, that's not the button, but the app selector screen. I think you call it "Home screen". For me, the home screen is the one you select as initial action on the user.

And this is important, as we are losing this option since this version, as only if you know which is the action (and if there's a menu for it), you can access such initial action. Do you understand me?

bvanjeelharia:
Yes @pedrobaeza , I got your point. I have made the changes likewise, please review it

pedrobaeza (approved):
Great! Tested on runboat and working.

chienandalu (approved):

PR #2412, opened 2023-02-17
[16.0][MIG] web_notify: Migration to 16.0

baimont:
https://github.com/OCA/web/issues/2309

Tested locally using the Test web notify tab on users


FrancoMaxime (approved):
LGTM

sbejaoui (approved):
Functional test.

lmignon (approved):

alexis-via:
While testing web_notify in v16 with the module currency_old_rate_notify, I found 2 problems:
1) in v14, it was working well when "message" contained HTML. In v16, the HTML code of "message" is displayed as text !
notify_html_bug

2) if target is not set, then the message is displayed to self.env.user and not to the user on which the method was called !!!

In the method _notify_channel(), you have:
```
if not target:
            target = self.env.user.partner_id
```
it should be:
```
if not target:
            target = self.partner_id
```
If you don't accept this change, you need to make it very clear in the documentation that, if you don't use the target argument, the message is NOT sent to the user on which you called the method, but to the current user!!!

baimont:
> 2. if target is not set, then the message is displayed to self.env.user and not to the user on which the method was called !!!
> 
> In the method _notify_channel(), you have:
> 
> ```
> if not target:
>             target = self.env.user.partner_id
> ```
> 
> it should be:
> 
> ```
> if not target:
>             target = self.partner_id
> ```
> 
> If you don't accept this change, you need to make it very clear in the documentation that, if you don't use the target argument, the message is NOT sent to the user on which you called the method, but to the current user!!!

Done

baimont:
> 1. in v14, it was working well when "message" contained HTML. In v16, the HTML code of "message" is displayed as text !
>    notify_html_bug

Concerning this, I believe it is due to the t-out attribute here https://github.com/odoo/odoo/blob/d594069767cb0f48c829a25f33abc6da620e7235/addons/web/static/src/core/notifications/notification.xml#L12

But as I am not familiar with owl I'm not sure what's a good way to fix this. If anyone has an idea, you're welcome fixing it with a new commit.

nguyenminhchien (approved):
lgtm

nguyenminhchien:
> > 1. in v14, it was working well when "message" contained HTML. In v16, the HTML code of "message" is displayed as text !
> >    notify_html_bug
> 
> Concerning this, I believe it is due to the t-out attribute here https://github.com/odoo/odoo/blob/d594069767cb0f48c829a25f33abc6da620e7235/addons/web/static/src/core/notifications/notification.xml#L12
> 
> But as I am not familiar with owl I'm not sure what's a good way to fix this. If anyone has an idea, you're welcome fixing it with a new commit.

Hello @baimont i just created this [PR](https://github.com/acsone/web/pull/15/files) to help you show a html formatted message.
Please review it.
image

baimont:
> > > 1. in v14, it was working well when "message" contained HTML. In v16, the HTML code of "message" is displayed as text !
> > >    notify_html_bug
> > 
> > 
> > Concerning this, I believe it is due to the t-out attribute here https://github.com/odoo/odoo/blob/d594069767cb0f48c829a25f33abc6da620e7235/addons/web/static/src/core/notifications/notification.xml#L12
> > But as I am not familiar with owl I'm not sure what's a good way to fix this. If anyone has an idea, you're welcome fixing it with a new commit.
> 
> Hello @baimont i just created this [PR](https://github.com/acsone/web/pull/15/files) to help you show a html formatted message. Please review it. image

Thanks very much @nguyenminhchien ! Tested and approved

marielejeune:
Hi, 
Thanks for the work.
There is an issue: notifications are not popping as soon as we use workers.

web_notify/views/res_users_demo.xml
```diff
@@ -0,0 +1,62 @@
...
+<!-- Copyright 2018 Aitor Bouzas <aitor.bouzas@adaptivecity.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="view_users_form_simple_modif_inherit" model="ir.ui.view">
+        <field name="name">res.users.form.simple.view.modif.inherit</field>
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook/page[1]" position="after">
+                <page string="Test web notify" name="test_web_notify">
```
sbidoul, line 11: Not related to the migration, but... it would be nice to show this page only in developer mode?

sbidoul: Ah no, sorry this view is demo data. Interesting approach.

sbidoul:
> There is an issue: notifications are not popping as soon as we use workers.

After further investigation with @marielejeune we found the issue was not related to workers but to Chrome blocking web socket due to an invalid certificate. So this should be good to go.

/ocabot merge nobump

pedrobaeza:
@sbidoul remember that you enabled a runboat certificate (changing http to https) for these cases :smile:

PR #2414, opened 2023-02-21
[14.0][IMP] web_ir_actions_act_view_reload: add missing model

kluna1998:
When adding the type on an action, an error was raised. With this change, it should work again.

Ported the change from 15.0, not cherry-picked, as other changes were not necessary

@etobella


sanchonuria (approved):

etobella (approved):

etobella:
Thanks for fixing the bug
/ocabot merge patch

PR #2426, opened 2023-03-03
[14.0][ADD] web_tooltip_field

cuongnmtm:
Enable the management of the help messages per field.

### DIFF
--- a/web_tooltip_field/__init__.py
+++ b/web_tooltip_field/__init__.py
@@ -1,0 +1,4 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
+from .patch import post_load
--- a/web_tooltip_field/__manifest__.py
+++ b/web_tooltip_field/__manifest__.py
@@ -1,0 +1,20 @@
+# Copyright 2023 Komit - Cuong Nguyen Mtm
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Web Tooltip Field",
+    "version": "14.0.1.0.0",
+    "summary": "Showing a Field ToolTip in Odoo",
+    "category": "Tools",
+    "website": "https://github.com/OCA/web",
+    "author": "KOMIT, Odoo Community Association (OCA)",
+    "maintainers": ["cuongnmtm"],
+    "license": "AGPL-3",
+    "depends": ["base", "web"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/ir_model_fields_help_tooltip_view.xml",
+    ],
+    "post_load": "post_load",
+    "application": False,
+    "installable": True,
+}
--- a/web_tooltip_field/models/ir_model_fields_help_tooltip.py
+++ b/web_tooltip_field/models/ir_model_fields_help_tooltip.py
@@ -1,0 +1,24 @@
+# Copyright 2023 Komit - Cuong Nguyen Mtm
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class IrModelFieldsHelpTooltip(models.Model):
+    _name = "ir.model.fields.help.tooltip"
+    _description = "Help Tooltip for Fields"
+    _rec_name = "field_id"
+
+    field_id = fields.Many2one(
+        "ir.model.fields", required=True, ondelete="cascade", index=True
+    )
+    model = fields.Char(related="field_id.model", store=True)
+    help = fields.Text(translate=True)
+
+    _sql_constraints = [
+        (
+            "field_id_uniq",
+            "unique(field_id)",
+            "Help Tooltip for Field already exists!",
+        ),
+    ]
--- a/setup/web_tooltip_field/odoo/addons/web_tooltip_field
+++ b/setup/web_tooltip_field/odoo/addons/web_tooltip_field
@@ -1,0 +1,1 @@
+../../../../web_tooltip_field
--- a/setup/web_tooltip_field/setup.py
+++ b/setup/web_tooltip_field/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/web_tooltip_field/models/__init__.py
+++ b/web_tooltip_field/models/__init__.py
@@ -1,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import ir_model_fields_help_tooltip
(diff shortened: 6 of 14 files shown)


web_tooltip_field/models/fields.py
```diff
@@ -0,0 +1,21 @@
+from odoo.fields import Field
```
hieulucky111: should be put the file outside of models. Maybe `tools` folder.

web_tooltip_field/models/fields.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def _description_help(self, env):
+    help_tooltip = env["ir.model.fields.help.tooltip"].search(
+        [
+            ("field_id.name", "=", self.name),
+            ("field_id.model", "=", self.base_field.model_name),
+        ],
+        limit=1,
```
hieulucky111: i think we have to create a constraint.

web_tooltip_field/models/fields.py
```diff
@@ -0,0 +1,21 @@
...
+    )
+    if help_tooltip and help_tooltip.help:
```
hieulucky111: you can put `help != ''` in search condition as well.

web_tooltip_field/models/fields.py
```diff
@@ -0,0 +1,21 @@
...
+        return help_tooltip.help
+    if self.help and env.lang:
+        model_name = self.base_field.model_name
```
hieulucky111: put this variable on top of function, so we can re-use in search condition.

web_tooltip_field/models/ir_model_fields_help_tooltip.py
```diff
@@ -0,0 +1,13 @@
...
+
+class IrModelFieldsHelpTooltip(models.Model):
+    _name = "ir.model.fields.help.tooltip"
+    _description = "Help Tooltip for Fields"
+    _rec_name = "field_id"
+
+    field_id = fields.Many2one(
+        "ir.model.fields", required=True, ondelete="cascade", index=True
+    )
+    model_id = fields.Many2one("ir.model", related="field_id.model_id", store=True)
```
hieulucky111: i think should be `model`, so we can use in search condition above.

web_tooltip_field/models/ir_model_fields_help_tooltip.py
```diff
@@ -0,0 +1,13 @@
...
+    help = fields.Text()
```
hieulucky111: should we have unique constraint for field.

hieulucky111 (changes requested):

hieulucky111 (approved):
Code Review
LGTM.

Leonidas-VII (commented):
I think all tests will be passed after we fix this warning https://github.com/OCA/web/pull/2426/checks?check_run_id=12747072097
(missing automatic test)

web_tooltip_field/fields.py
```diff
@@ -0,0 +1,25 @@
...
+
+def _description_help(self, env):
+    model_name = self.base_field.model_name
+    help_tooltip = env["ir.model.fields.help.tooltip"].search(
+        [
+            ("field_id.name", "=", self.name),
+            ("model", "=", model_name),
+            ("help", "not in", [False, ""]),
+        ],
+        limit=1,
```
DucTruongKomit, line 15: Maybe we should set order="id", no?

DucTruongKomit (approved):
Look gud to me for the reviewing code!

web_tooltip_field/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import models
+from . import fields
+from . import tests
```
petrus-v: This import is not needs.

If you keep it all tests module will be loaded on production, this can be undesirable in some circumstance (importing test libraries...)

```suggestion
```

web_tooltip_field/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "website": "https://github.com/OCA/web",
+    "author": "KOMIT, Odoo Community Association (OCA)",
+    "maintainers": ["cuongnmtm"],
+    "license": "AGPL-3",
+    "depends": ["base", "web"],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/ir_model_fields_help_tooltip_view.xml",
+    ],
+    "application": True,
```
petrus-v: I'm not sure we can considre this module as an application, an application in odoo world is mainly a new entry in the main menu I think.

```suggestion
```

web_tooltip_field/fields.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+    if help_tooltip:
+        return help_tooltip.help
+    if self.help and env.lang:
+        field_help = env["ir.translation"].get_field_help(model_name)
+        return field_help.get(self.name) or self.help
+    return self.help
+
+
+Field._description_help = _description_help
```
petrus-v: Could you make this monkey patch in a post load hooks as mentioned in OCA's conventions https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id11

Also I'm wondering if we could reuse the parent method something such as:

```suggestion
_former_field_description_help = Field._description_help


def _description_help(self, env):
    model_name = self.base_field.model_name
    help_tooltip = env["ir.model.fields.help.tooltip"].search(
        [
            ("field_id.name", "=", self.name),
            ("model", "=", model_name),
            ("help", "not in", [False, ""]),
        ],
        limit=1,
    )
    if help_tooltip:
        return help_tooltip.help
    return _former_field_description_help(self, env)


Field._description_help = _description_help
```

web_tooltip_field/models/ir_model_fields_help_tooltip.py
```diff
@@ -0,0 +1,24 @@
...
+class IrModelFieldsHelpTooltip(models.Model):
+    _name = "ir.model.fields.help.tooltip"
+    _description = "Help Tooltip for Fields"
+    _rec_name = "field_id"
+
+    field_id = fields.Many2one(
+        "ir.model.fields", required=True, ondelete="cascade", index=True
+    )
+    model = fields.Char(related="field_id.model", store=True)
+    help = fields.Text()
```
petrus-v: Should we allow to translate this field ?

```suggestion
    help = fields.Text(translate=True)
```

petrus-v (changes requested):
Nice idea, I'm sure this module will be useful !

web_tooltip_field/fields.py
petrus-v: Here you can find an example: https://github.com/OCA/server-tools/blob/16.0/module_change_auto_install/__manifest__.py#L15

cuongnmtm:
> Nice idea, I'm sure this module will be useful !

Thank you, @petrus-v , for your excellent suggestions! I have learned a lot from them.

bofiltd (approved):

petrus-v (approved):

PR #2429, opened 2023-03-06
[14.0][FIX] web_timeline: css and styles not charged when rendering timeline

CarlosRoca13:
cc @Tecnativa TT40299

ping @victoralmau @pedrobaeza

PR #2430, opened 2023-03-06
[15.0][FIX] web_timeline: css and styles not charged when rendering timeline

CarlosRoca13:
On library version update a new feature was added at Timeline creation level to allow modifying the xss. By keeping it active the css classes are not showed

As the view is on backend, has no sense to keep it active. So we deactivate the feature to allow show the timeline correctly.

cc @Tecnativa TT40299

FW-PORT of https://github.com/OCA/web/pull/2429

ping @victoralmau @pedrobaeza


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2442, opened 2023-03-09
[16.0][FIX] web_responsive: Show View Type Always

### LINKED ISSUE #2437: [16.0][BUG] web_responsive: selection of view type get under tree header
## Module
web_responsive

## Describe the bug
When reducing window size (for instance because I want to get 2 windows side-by-side on my desktop), the buttons to change view type get on a second line which is hidden by tree view header :
image

## To Reproduce

1. Install web_responsive
2. Get to a tree view (CRM or sales for instance)
3. Reduce size of the window, at some point the buttons are hidden on a second line
 
Note that if you reduce more they appear again... when search bar gets right next to window title on the left of the screen : 
image

**Expected behavior**
Buttons should be kept visible (or maybe reduced but should no be hidden by tree view)

bvanjeelharia:
Fixes https://github.com/OCA/web/issues/2437

### DIFF
--- a/web_responsive/static/src/components/control_panel/control_panel.scss
+++ b/web_responsive/static/src/components/control_panel/control_panel.scss
@@ -34,7 +34,7 @@
         }
     }
     .o_cp_bottom_right {
-        height: 30px;
+        height: 10%;
     }
     // Mobile Control panel (breadcrumbs, search box, buttons...)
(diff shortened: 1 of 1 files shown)


remi-filament (approved):
Tested OK on un Runboat, thanks !

PR #2145, opened 2022-02-14
[MIG] web_tree_dynamic_colored_field: Migration to 15.0

amh-mw:

PR #2443, opened 2023-03-10
[FIX][12.0] web_tree_dynamic_colored_field : return first color that match with valid expression.

legalsylvain:
trivial patch.

For the time being, if there are many key / value (color / expression) all the colors corresponding to valid expressions will be set. So the last color will be put. 

**Without that patch** : ``{'bg_color': 'red: True; green: True;'}`` will return 'green'
**With that patch** : ``{'bg_color': 'red: True; green: True;'}`` will return 'red'

CC : @amh-mw as you migrated the module in V15 recently here https://github.com/OCA/web/pull/2145
CC : @damdam-s : original author


amh-mw (approved):
LGTM

quentinDupont (approved):

legalsylvain:
/ocabot merge patch
thanks for your reviews !

PR #2446, opened 2023-03-13
[16.0][MIG] web_m2x_options: Migration to 16.0

anothingguy:
- in V16, `no_color_picker` won't be migrated, use native option: `no_edit_color` instead)
- in V16, `no_open_edit` won't be migrated: `no_open_edit` has the same behavior of `no_open` in v15 in term of functionality and the code also confirmed that. I'm not sure why we should maintain option `no_open_edit` so I didn't migrated this option to v16


AungKoKoLin1997 (commented):
Functional Review: It works as expected except "no_open_edit". May be I am missing something when I test. I set up "no_open_edit" to true but it still opens in edit mode. May be because of odoo16 default behavior is in edit mode?

anothingguy:
Hi @AungKoKoLin1997 , thanks for your review.
I double checked, `no_open_edit` has the same behavior as `no_open` in v15 in term of functionality and the code also confirmed that. I'm not sure why we should maintain option `no_open_edit` so I didn't migrate this option to v16.
Could you double check `no_open_edit` in v15. Or maybe I'm wrong somewhere ?
- Also `no_open_edit` is not mentioned officially in the module document

web_m2x_options/static/src/components/relational_utils.esm.js
```diff
@@ -0,0 +1,224 @@
...
+        // returned.
+        if (!_.isUndefined(this.ir_options["web_m2x_options.limit"])) {
+            this.props.searchLimit = parseInt(
+                this.ir_options["web_m2x_options.limit"],
+                10
+            );
+        }
+
+        if (typeof this.props.nodeOptions.limit === "number") {
+            this.props.searchLimit = this.props.nodeOptions.limit;
```
jesusVMayor, line 46: The limit option seems to work fine in many2one fields, but not in many2many_tags, this.props.nodeOptions is empty in this widget.
To reproduce in runbot:
- Remove web_m2x_options.limit parameter
- edit sale.order form view adding limit in tag_is field
` <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'color', 'no_create_edit': True, 'limit': 10}"/>`

anothingguy: yes, confirmed this issue, I will update soon, thanks

anothingguy: hi @jesusVMayor, I fixed that issue. Could you take a look it when you have a chance. Thanks in advance

jesusVMayor: Yeah, thanks for the fix, works fine.

jesusVMayor (approved):
LGTM

web_m2x_options/README.rst
```diff
@@ -0,0 +1,206 @@
...
+
+``no_open_edit`` *boolean* (Default: value of ``no_open`` which is ``False`` if not set)
+
+  Causes a many2one not to offer to click through in edit mode, but well in read mode
+
+``open`` *boolean* (Default: ``False``)
+
+  Makes many2many_tags and one2many rows buttons that open the linked resource
+
+``no_color_picker`` *boolean* (Default: ``False``) (obselete: use native option: ``no_edit_color`` instead)
```
StephaneMangin, line 88:
```suggestion
``no_color_picker`` *boolean* (Default: ``False``) (obsolete: use native option: ``no_edit_color`` instead)
```

StephaneMangin (approved):

web_m2x_options/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2015 0k.io
+# Copyright 2016 ACSONE SA/NV
+# Copyright 2017 Tecnativa
+# Copyright 2020 initOS GmbH.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "web_m2x_options",
+    "version": "16.0.1.1.0",
```
CRogos, line 9: Shouldn't the version be changed to 16.0.1.0.0?

CRogos (commented):
@gurneyalex @pedrobaeza 

I am a little bit wondering how this could be merged?
- Migration guidelines not followed 
(Commits not correct)
image

- changes of pending bugfix PR not included
https://github.com/OCA/web/pull/2318
(Button of dialog cannot be configured)
image

At least the widgets issue is fixed and the JS refactoring looks great.

pedrobaeza:
Well, reviewers not always check all these things, and @gurneyalex probably checked only the approvals without being very picky. They are not very critical things though.

PR #2447, opened 2023-03-13
[16.0][mig] web_widget_image_download: Migration to 16.0

dsolanki-initos:


web_environment_ribbon/models/web_environment_ribbon_backend.py
```diff
@@ -5,7 +5,6 @@
 
 
 class WebEnvironmentRibbonBackend(models.AbstractModel):
-
```
legalsylvain, line 8: please avoid to change other modules.

legalsylvain (commented):
could you rebase ? I like to test it on runboat.

thanks !

quentinDupont (changes requested):
Functionnal review OK but there's still @legalsylvain comment that need to be adress

legalsylvain:
obsolete PR. module merged.

PR #2450, opened 2023-03-14
[15.0][MIG] web_help: Backport to 15.0

tarteo:
Had to do some stuff to make it work also for the legacy frameworks.


tarteo:
Hmm, tests  fail because it uses web.test_env in test instead of normal env / legacy env. Not sure how to properly fix this (without creating weird code).

AaronHForgeFlow:
I know this is kind of stale, but functionally speaking, it works ok.

PR #2455, opened 2023-03-16
[14.0][IMP]web_company_color: generate all colors from the company's …

PicchiSeba:
…logo.

This PR aims to add automatic color generation for all of the company palette colors.

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -151,13 +151,27 @@ class ResCompany(models.Model):
             # Calc. optimal text color (b/w)
             # Grayscale human vision perception (Rec. 709 values)
             _a = 1 - (0.2126 * _r + 0.7152 * _g + 0.0722 * _b)
+            color_navbar_bg = n_rgb_to_hex(_r, _g, _b)
+
+            _s += 0.25 if _s < 0.5 else -0.25
+            color_button_bg = n_rgb_to_hex(*hls_to_rgb(_h, _l, _s))
+            color_button_bg_hover = n_rgb_to_hex(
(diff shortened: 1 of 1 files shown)


web_company_color/models/res_company.py
```diff
@@ -151,13 +151,24 @@ def button_compute_color(self):
...
+            )
+
             values.update(
                 {
-                    "color_navbar_bg": n_rgb_to_hex(_r, _g, _b),
+                    "color_navbar_bg": color_navbar_bg,
                     "color_navbar_bg_hover": n_rgb_to_hex(_rd, _gd, _bd),
                     "color_navbar_text": "#000" if _a < 0.5 else "#fff",
+                    "color_button_bg": button_color_bg,
+                    "color_button_bg_hover": "#fff" if _a < 0.5 else "#000",
```
Tardo: Perhaps the best thing to do here is to make the button a little brighter based on the color of the button.... ¿?

PicchiSeba: That's right.
After some tries I opted for a 20% increase in saturation.
For now I hardcoded a maximum _s value of 80% because I don't really know how we should handle high saturation colors.

PicchiSeba: I'm open to other proposals as well

Tardo (approved):
Ok for me.
It may be interesting to switch to using the 'colorsys' library and some color scheme such as triadic... But that would be for a future PR.

HekkiMelody (approved):

pedrobaeza:
Please fw-port it to upper versions:

/ocabot merge major

PR #2460, opened 2023-03-20
[16.0][FIX] web_responsive: Translation Button Position

### LINKED ISSUE #2459: [16.0] web_responsive: wrong place of translation button for name fields
## Module
web_responsive

## Describe the bug
If you enable translation for name fields (which are having oe_title class in the form), the translation button appears on the left side beneath the title instead of top right

## To Reproduce
**Affected versions**: `16.0.1.0.4`

1. activate translation for field which is wrapped in `oe_title` class div on the form

image

**Expected behavior**
To have the translation button of the field inside the title on top right, but it's below left

image

**Additional context**
I think that the solution would be to filter out spans with class `o_field_translate` on the [web_responsive.scss#L171](https://github.com/OCA/web/blob/16.0/web_responsive/static/src/legacy/scss/web_responsive.scss#L171)

```scss
// Support for long title (with ellipsis)
.oe_title {
    .o_field_widget:not(.oe_inline) {
        display: block;
        span:not(.o_field_translate) {
            display: block;
            max-width: 100%;
            text-overflow: ellipsis;
            white-space: nowrap;
            overflow: hidden;
            width: initial;
            &:active {
                white-space: normal;
            }
        }
    }
}
```

bvanjeelharia:
Fixes https://github.com/OCA/web/issues/2459

### DIFF
--- a/web_responsive/static/src/legacy/scss/web_responsive.scss
+++ b/web_responsive/static/src/legacy/scss/web_responsive.scss
@@ -168,7 +168,7 @@ html .o_web_client .o_action_manager .o_action {
         .oe_title {
             .o_field_widget:not(.oe_inline) {
                 display: block;
-                span {
+                span:not(.o_field_translate) {
                     display: block;
                     max-width: 100%;
                     text-overflow: ellipsis;


Tardo (approved):
Thanks. Tested on runboat.

pedrobaeza (approved):
/ocabot merge patch

PR #2461, opened 2023-03-20
[15.0][IMP] web_widget_numeric_step: Add border to buttons shown

CarlosRoca13:
cc @Tecnativa 

ping @sergio-teruel @pedrobaeza

### DIFF
--- a/web_widget_numeric_step/static/src/xml/numeric_step.xml
+++ b/web_widget_numeric_step/static/src/xml/numeric_step.xml
@@ -10,7 +10,7 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
             <div class="input-group widget_numeric_step">
                 <div class="input-group-prepend">
                     <button
-                        class="fa fa-minus btn btn-default btn_numeric_step"
+                        class="fa fa-minus btn btn-secondary btn_numeric_step"
                         aria-label="Minus"
                         title="Minus"
                         type="button"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please put images of the before/after.

CarlosRoca13:
Before this change:
image

After this change:
image

pedrobaeza:
I'm not sure it's best with border, and in fact, v16 goes against this in its design. Why do you want this?

CarlosRoca13:
I think that using the border makes it clearer the limits of the buttons and a mobile user will know more clearly where to click.

pedrobaeza:
On a non precise environment like mobiles, where the "cursor" is your finger, the important thing is the focus center, and this one is clearly shown by the + / - icon, so I still think it's better the old approach, as putting a lot of borders makes you vision to need more concentration on the elements. That's why the "clean" approach that UIs are taking (can be called also minimalist) try to save the maximum all the superfluous UI elements.

legalsylvain:
> I'm not sure it's best with border, and in fact, v16 goes against this in its design. Why do you want this?

Agree with @pedrobaeza on that topic. The general design of Odoo 16 is "no border". I don't think it's the good idea if each ``web`` module implement design that does'nt match with the general one.

however, if you don't like this direction (as I do !), you can use this OCA module : https://github.com/OCA/web/tree/16.0/web_theme_classic

regards.

sergio-teruel (approved):

CarlosRoca13:
Ok, so let's close this PR

Thanks for the reviews

PR #2462, opened 2023-03-21
[16.0][MIG] web_no_bubble

astirpe:


legalsylvain:
Is this module required in v16 ?
I dont see bubble when i use it... Or did i missed something ?

astirpe:
@legalsylvain I locally reproduced the "bubbles" this way:

- create a fresh db (**without demo data!**)
- install hr_expense
- login as admin
- open the expenses menu (My Expenses)

legalsylvain:
Ow ! Just tested. Thanks ! Indeed, no bubble in demo mode. I don't understand why... Do you have an idea ? 
Anyway :+1: This module should be ported.

astirpe:
@legalsylvain I think it's on purpose: no bubble in demo mode. As far as I remember, it was always like that. :see_no_evil:

tarteo (commented):
The bubble disapears but I still get the tooltip. Is this as intended?
image

Weird place for the tooltip:
image

astirpe:
@Tarteo thanks! I guess the tooltip text should be hidden as well. With commit https://github.com/OCA/web/pull/2462/commits/25a791c872477471869b063839e37b43c880f55f also the tooltip text is hidden, would you check if it's ok?

tarteo (approved):
Works perfectly now, thanks!

liebana:
Functional tested / no code review.

LGTM

liebana (approved):

PR #2465, opened 2023-03-23
[15.0][UPD] web_ir_actions_act_view_reload: Added ACL to avoid Odoo warning

JonathanStein:
Added ir.model.access.csv with an access rule to avoid Odoo's warning when models does not have an ACL


chienandalu (changes requested):
Duplicated: https://github.com/OCA/web/pull/2347

JonathanStein:
Yes, the change to an abstract model in PR #2347 eliminates the need for an ACL

PR #2467, opened 2023-03-24
[15.0][IMP] web_company_color: generate all colors from the company's…

PicchiSeba:
… logo.

As suggested in #2455 I'm submitting the same modification in the more recent versions.

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -44,6 +44,47 @@ class ResCompany(models.Model):
             }
           }
         }
+        a[href],
+        a[tabindex],
+        .btn-link,
+        .o_external_button {
+          color: %(color_link_text)s !important;
+        }
+        a:hover,
+        .btn-link:hover {
+          color: %(color_link_text_hover)s !important;
+        }
+        .btn-primary:not(.disabled),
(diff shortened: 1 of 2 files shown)


PicchiSeba:
Tests were failing because of some missing fields. so I cherry-picked the relative commits from branch 14.0 and force-pushed.
Hopefully this will fix some if not all of the previously failing tests

tarteo:
Hey thanks for your contribution! I have a question. Does this also supposed to change the background color of the statusbar buttons?

image

PicchiSeba:
Can't say for sure for Odoo 15 as I'm using v14, I was just asked to fw-port to upper versions too.

But testing in v14 i can assert that these colors do not apply a background to the statusbar buttons (I tested the status bar in the Helpdesk module, assuming it is the same component everywhere. Correct me if I'm wrong), and it remains white.

What changes in the statusbar is the text color.
Screenshot from 2023-03-24 14-22-56

In my test, text color was changed to #94a181, the same assigned to `color_link_text`

We could have another commit adding that feature, tho.

minhthie:
I still see that the status bars have different color. The same happens for the buttons
image

PicchiSeba:
I guess something's different between v14 and v15. I'm setting this PR as Draft and I'll investigate when i can

PR #2475, opened 2023-04-01
[MID] web_widget_json_graph: migration to v16.0

therbin404:
The js file was rewrited with the new javascript module templates


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0.

PR #2055, opened 2021-10-18
[MIG] web_environment_ribbon: Migration to 15.0

ap-wtioit:
replaces #2049, now based on branch 15.0 and following https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

Info @wt-io-it

PR #2479, opened 2023-04-05
[14.0][IMP] web_environment_ribbon: hide the ribbon when navigation is in use

HekkiMelody:
Backport of fce9051d581cc3cbae86313c73e00ea4f65c62b4

### DIFF
--- a/web_environment_ribbon/static/src/css/ribbon.css
+++ b/web_environment_ribbon/static/src/css/ribbon.css
@@ -1,5 +1,6 @@
 /* Copyright 2015 Francesco OpenCode Apruzzese <cescoap@gmail.com>
    Copyright 2017 Thomas Binsfeld <thomas.binsfeld@acsone.eu>
+   Copyright 2021 Andreas Perhab <a.perhab@wtioit.at>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't like this option, and more to backport it to a version that already works the other way.

legalsylvain:
> I don't like this option,

Could you elaborate ? #2311 introduces that change on last 16 release.

ap-wtioit (approved):
Tested on Firefox and Chrome, works in 14.0 as well.

@pedrobaeza and i had a little discussion about this in #2055, starting here https://github.com/OCA/web/pull/2055#issuecomment-1021935032, i still think if there is a problem with some code a separate issue should be opened to make it easier to work on a solution and not spreading the discussion on multiple PRs

pedrobaeza (changes requested):
As discussed.

I didn't have time to redact such "issue" asked by Andreas, but what it can't be is to backport something debatable.

HekkiMelody:
I thought it was a nice QoL improvement but I didn't realize it was a controversial change and I can definitely live without it. I do not mind closing the PR.

PR #2431, opened 2023-03-07
[16.0][mig] web_widget_url_advanced: Migration to 16.0

dsolanki-initos:

PR #2486, opened 2023-04-20
[16.0] [MIG] web_widget_url_translatable: Migration to 16.0

rven:


simahawk:
Question: wouldn't be better to merge this feature into https://github.com/OCA/web/tree/15.0/web_widget_url_advanced?

rven:
> Question: wouldn't be better to merge this feature into https://github.com/OCA/web/tree/15.0/web_widget_url_advanced?

Could be an option! But the web_widget_url_advanced is also in PR. So difficult to add my changes into that one.
Could this be an option for v17?

I also notice that the https://github.com/OCA/web/pull/2431 is not converted to OWL!

rven:
Hallo?

simahawk (approved):

simahawk:
LG just go ahead as this on 16 and plan for 18? You can add this to the roadmap I'd say.
Nitpicking: there's a fixup commit that I would squash.

PR #2497, opened 2023-05-09
[15.0][FIX] web_pivot_computed_measure: Error when using filter created with favorites

CarlosRoca13:
Before this changes, when trying to select a user filter that includes a computed measurement, an error was thrown as it did not know where to retrieve the previously computed measurement from.

With the applied changes we are saving the computed measurements in the context. This way we can retrieve the filter when loading it.

cc @Tecnativa TT43005

ping @sergio-teruel @chienandalu


chienandalu (approved):

CarlosRoca13:
An incompatibility with web_time_range_menu_custom has been found. Wait for fix to merge

sergio-teruel (approved):

CarlosRoca13:
The problems found should be solved, so it can be merged

PR #2506, opened 2023-05-17
[16.0][FIX] web_domain_field: if you have a domain saved in a binary variable, parsing that text will fail.

RodrigoBM:
If you have a domain saved in a binary variable, parsing that text will fail.

https://github.com/OCA/web/assets/6359121/a6cb04e2-acbe-4a80-99f8-eb5a14e51678

You can see the example of this field https://github.com/OCA/OCB/blob/16.0/addons/account/models/account_tax.py#L1308

### DIFF
--- a/web_domain_field/static/lib/js/pyeval.js
+++ b/web_domain_field/static/lib/js/pyeval.js
@@ -153,9 +153,17 @@ odoo.define('web.domain_field', function (require) {
             if (_.isString(domain)) {
                 // Modified part or the original method
                 if (domain in evaluation_context) {
-                    result_domain.push.apply(
-                        result_domain, $.parseJSON(evaluation_context[domain]));
-                    return;
+                    var fail_parse_domain = false;
+                    try {
+                        var domain_parse = $.parseJSON(evaluation_context[domain]);
+                        console.warn("`web_domain_field is deprecated. If you want to use this functionality you can assign a unserialised domain to a fields.Binary");
+                    } catch (e) {
+                        fail_parse_domain = true;
+                    }
+                    if (!fail_parse_domain) {
+                        result_domain.push.apply(result_domain, domain_parse);
+                        return;
+                    }
                 }
                 // End of modifications
 
--- a/web_domain_field/readme/DESCRIPTION.rst
+++ b/web_domain_field/readme/DESCRIPTION.rst
@@ -1,3 +1,8 @@
+.. warning::
+    This module is deprecated.
+    If you want to use this functionality you can assign a unserialised
+    domain to a fields.Binary, `example <https://github.com/OCA/OCB/blob/16.0/addons/account/models/account_tax.py#L1308>`_
+
 This technical addon allows developers to specify a field domain in a view
 using the value of another field in that view, rather than as a static
 XML attribute.


RodrigoBM:
cc: @FernandoRomera @StefanRijnhart @pedrobaeza @imarulbala

StefanRijnhart:
Wow, that's a great find! Apparently, the functionality is already present as an undocumented feature in Odoo 16. I just tested this view and it actually works. Even the Odoo devs themselves are surprised, see https://github.com/odoo/odoo/pull/113027#discussion_r1142417362.

Can you add a comment over your fix saying that it adds compatibility with the implementation from core Odoo (assigning an unserialized domain to a fields.Binary) and add a big fat warning (https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#warnings) to the module's documentation saying the module is deprecated?

sbidoul:
@lmignon @rousseldenis

RodrigoBM:
> Wow, that's a great find! Apparently, the functionality is already present as an undocumented feature in Odoo 16. I just tested this view and it actually works. Even the Odoo devs themselves are surprised, see [odoo/odoo#113027 (comment)](https://github.com/odoo/odoo/pull/113027#discussion_r1142417362).
> 
> Can you add a comment over your fix saying that it adds compatibility with the implementation from core Odoo (assigning an unserialized domain to a fields.Binary) and add a big fat warning (https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#warnings) to the module's documentation saying the module is deprecated?

Its ok?

web_domain_field/static/lib/js/pyeval.js
```diff
@@ -153,9 +153,17 @@ odoo.define('web.domain_field', function (require) {
             if (_.isString(domain)) {
                 // Modified part or the original method
                 if (domain in evaluation_context) {
-                    result_domain.push.apply(
-                        result_domain, $.parseJSON(evaluation_context[domain]));
-                    return;
+                    var fail_parse_domain = false;
+                    try {
+                        var domain_parse = $.parseJSON(evaluation_context[domain]);
+                        console.warn("This module is deprecated. If you want to use this functionality you can assign a unserialised domain to a fields.Binary");
```
StefanRijnhart: Please add the module name in the console warning.

```suggestion
                        console.warn("web_domain_field is deprecated. If you want to use this functionality you can assign a unserialised domain to a fields.Binary");
```

StefanRijnhart:
@RodrigoBM works fine! Just a small comment.

RodrigoBM:
> @RodrigoBM works fine! Just a small comment.

Done

rousseldenis:
Great !

@StefanRijnhart Are we sure this is a wanted functionality by Odoo ? To make sure it wouldn't be removed in a further version

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
@rousseldenis the feature is currently in use in Odoo, in Odoo 16 and in master. Given how enthousiastic the Odoo developers responded when they found out it worked, I don't think it will be removed anytime soon :smile: And in the other case, we can always reïntroduce this module in a future series, right?

lmignon (approved):

ged-odoo:
Well, I just found out about this. As far as I can tell, there is no good reason for the binary field used as domain to stop working (obviously not a promise though!). It is unlikely that we change the way binary fields work, and I created an internal task to add a test to make sure the JS code keep working.

pedrobaeza:
Thanks for that, Géry!

PR #2507, opened 2023-05-17
[MIG] web_tree_image_tooltip: Migration to 16.0 

siemenv:


eversatis:
Why this app migration was given up? It is a very useful function.

siemenv:
I had no time to finish this migration.

CarlosRoca13:
> Why this app migration was given up? It is a very useful function.

@eversatis The module is not needed anymore see: https://github.com/OCA/OpenUpgrade/pull/4518

PR #2509, opened 2023-05-19
[ADD] web_model_read_only

ilyasProgrammer:
Module disables modification operations (write, create, delete) on selected model to selected users groups.

### DIFF
--- a/web_model_read_only/models/models.py
+++ b/web_model_read_only/models/models.py
@@ -1,0 +1,30 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, exceptions, fields, models
+
+
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    readonly_group_ids = fields.Many2many(
+        "res.groups", relation="ir_model_group_readonly_rel"
+    )
+
+
+class BaseModel(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def check_access_rights(self, operation, raise_exception=True):
+        if operation != "read":
+            model = self.env["ir.model"]._get(self._name)
+            if (
+                model.readonly_group_ids
+                and self.env.user.groups_id & model.readonly_group_ids
+            ):
+                if raise_exception:
+                    raise exceptions.AccessError(
+                        _("You are not allowed to modify this record.")
+                    )
+                return False
+        return super(BaseModel, self).check_access_rights(operation, raise_exception)
--- a/web_model_read_only/tests/test_models.py
+++ b/web_model_read_only/tests/test_models.py
@@ -8,18 +8,20 @@ class TestIrModel(SavepointCase):
         super(TestIrModel, cls).setUpClass()
         cls.readonly_group = cls.env.ref("base.group_user")
         cls.group_manager = cls.env.ref("base.group_erp_manager")
-        cls.ir_model = cls.env["ir.model"].create(
-            {"name": "x_test_model", "model": "x_test.model"}
-        )
-        User = cls.env["res.users"]
-        cls.manager = User.create(
+        cls.group_partner_manager = cls.env.ref("base.group_partner_manager")
+        cls.partner_model = cls.env["ir.model"]._get("res.partner")
+        res_users_obj = cls.env["res.users"]
+        cls.manager = res_users_obj.create(
             {
                 "name": "Manager",
                 "login": "Manager",
-                "groups_id": cls.group_manager,
+                "groups_id": [
+                    (4, cls.group_manager.id),
+                    (4, cls.group_partner_manager.id),
+                ],
             }
         )
-        cls.user_readonly = User.create(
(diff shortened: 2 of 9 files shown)


yostashiro:
@ilyasProgrammer Not sure your business requirement but could https://github.com/OCA/server-tools/tree/14.0/base_model_restrict_update cover your use case?

francesco-ooops:
Hi @yostashiro, thanks for the heads up! I'm a functional and I was testing your module, if I understand correctly: 

- you can set one or more models to be edited only by "permitted" users

- "permitted" users can edit all restricted models?

I ask as I tried to test on runbot but couldn't find either "Grant Update Permit" or "Readonly User" in user form.

yostashiro:
> * you can set one or more models to be edited only by "permitted" users
> * "permitted" users can edit all restricted models?

@francesco-ooops Yes for both.

> I ask as I tried to test on runbot but couldn't find either "Grant Update Permit" or "Readonly User" in user form.

I realize that the button presentation in the user form has been broken and created a PR to fix it here: https://github.com/OCA/server-tools/pull/2626

francesco-ooops:
Hi @yostashiro , 

your solution works as described but has two limitations IMHO: 

- I cannot allow User A to edit Model 1 but not Model 2, and User B to edit Model 2 but not Model 1

- If I have 10 users allowed to edit Model 1 and I need to restrict one of those users from editing it, I need to set 9 users as "permitted", which is a bit counterintuitive

On the other hand, I like that the flag "read only" sets virtually the whole DB in "guest mode", although the "read only" implemented only restricts user from saving records, but allows for example to browse related objects (eg: read-only user in SO cannot change customer, but can check all customer records)

I think the solution proposed in this module is different enough to allow both modules to be published in OCA :)

francesco-ooops (approved):
Functional review ok!

francesco-ooops:
@yostashiro what do you think?

web_model_read_only/models/models.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+
+class BaseModel(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def check_access_rights(self, operation, raise_exception=True):
+        if operation != "read":
+            model = self.env["ir.model"].search([("model", "=", self._name)])
```
geomer198: You could use `_get` function get model record by model name.
For example:
```
self.env['ir.model']._get(self._name)
```

geomer198 (commented):
Please add test coverage for functional.

yostashiro:
@francesco-ooops I see that the design concept of this module is somewhat similar to that of https://github.com/OCA/server-ux/tree/14.0/base_menu_visibility_restriction - normally Odoo makes things more permissive for groups in the upper node, but the design of this module works in the opposite way.  Therefore, I suppose that you are naturally expected to define new groups to be assigned to users and models, specifically for the purpose of access restriction.

I think the function can be useful as long as the usage is properly documented.

I was thinking, however, it may still be good to conflate the proposed function into base_model_restrict_update.  Maybe we can think about it for future versions.

Shouldn't this module be put in https://github.com/OCA/server-tools instead?

geomer198 (changes requested):
Please add tests for functional.

web_model_read_only/tests/test_models.py
```diff
@@ -0,0 +1,82 @@
...
+from odoo.tests.common import SavepointCase
+
+
+class TestIrModel(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestIrModel, cls).setUpClass()
+        cls.readonly_group = cls.env.ref("base.group_user")
+        cls.group_manager = cls.env.ref("base.group_erp_manager")
+        cls.ir_model = cls.env["ir.model"].create(
```
geomer198: For what you create new model? Could you use existing model for example `res.partner`?

web_model_read_only/tests/test_models.py
```diff
@@ -0,0 +1,82 @@
...
+            {"name": "x_test_model", "model": "x_test.model"}
+        )
+        User = cls.env["res.users"]
```
geomer198: Better change naming. For this situation and similar use: `res_users obj`. `obj` suffix used for the showing what variable contain empty model object.

web_model_read_only/tests/test_models.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+        cls.user_readonly = User.create(
+            {
+                "name": "User readonly",
+                "login": "user_readonly",
+                "groups_id": cls.readonly_group,
+            }
+        )
+
+    def test_check_access_rights(self):
```
geomer198: Please split this function on the separate cases.
And will make more tests for the coverage all cases.

geomer198 (changes requested):
Please fix comments.

geomer198 (commented):
For all assertion check need will add message.

francesco-ooops:
@ilyasProgrammer can you close this one?

PR #2513, opened 2023-05-23
[16.0] [FIX] web_advanced_search - Prevent closure dropdown

rven:
An error selecting many2one fields in the Custom filter.

When select to add a field "is equal to" and adding an extra condition like the first one, the filter disappears!


hbrunn (approved):

augusto-weiss:
LGTM

ivantodorovich (approved):
LGTM. Tested functionally.
Thanks!

PR #2515, opened 2023-05-24
[ADD] new module web_view_rtree

huguesdk:


web_view_rtree/readme/ROADMAP.rst
```diff
@@ -0,0 +1,20 @@
+This module is in an alpha state. It is still being developed and should be
+considered as experimental.
```
legalsylvain:
```suggestion
```

``    "development_status": "Alpha",`` is enough. It will add an explicit warning in the README.rst file. It will be removed also automatically, when the status will change.

legalsylvain (commented):
Nice R&D ! 
Could you add a screenshot of the view mentionned in the DESCRIPTION.rst file ? 
(``.. figure:: static/description/rtree.png``)

web_view_rtree/readme/ROADMAP.rst
huguesdk: thanks for your suggestions! done.

legalsylvain (commented):
thanks for the screenshot and nice new view ! 
I already see the product_category_rtree_view module being developed!

legalsylvain:
Very general question. (maybe not in the scope of your work). What about to overload the tree view, and change the display when a "group by" is selected, improving the native view.

**Pro :** 
- The feature could so be enabled on all the existing views, without developping new module.
- the view could benefit from all the other tree view improvements.

**Limitation :** 
- the feature should be enabled only if there is only one "group by field" and if "the group by field" has a "parent_id" field.

**Cons :** 
- maybe very hard to develop or impossible ;-)

youring:
Neating feature that I've been long dreamed of... Thanks @huguesdk. 
I have tested your project_task_rtree_view module. 
One suggestion is to keep the rtree view expand status while go back to the rtree view(by click the breadcrumb)  after drill down to the each record.

huguesdk:
@legalsylvain very nice idea! i didn’t think of it, i’ve only seen this view as a separate view, but your idea makes complete sense. however, i think it would probably be quite hard to do without breaking things. also, to allow for the same features as the rtree view in the regular tree view, there would need to be a way to group by fields that are not on the model itself. the rtree view allows to navigate the relationships across several models. for example, for the task rtree, the top parent could be the `user_id` of the project, even though it cannot be accessed directly from a task.

huguesdk:
@youring thanks, i hope it will meet your expectations. ☺️ thanks for the suggestion, it is already in [the roadmap](https://github.com/coopiteasy/web/tree/16.0-add-web_view_rtree/web_view_rtree#known-issues-roadmap).

i’m glad you found the `project_task_rtree_view` module. it is just a tiny demo module to easily try out the view. just curious: how did you find it?

legalsylvain:
> i’m glad you found the project_task_rtree_view module. it is just a tiny demo module to easily try out the view. just curious: how did you find it? 

I didn't found ``project_task_rtree_view`` module. I was talking about ``product_category_rtree_view`` that could be great to develop.

huguesdk:
> > i’m glad you found the project_task_rtree_view module. it is just a tiny demo module to easily try out the view. just curious: how did you find it?
> 
> I didn't found `project_task_rtree_view` module. I was talking about `product_category_rtree_view` that could be great to develop.

@legalsylvain i was asking that question to @youring. ☺️ fyi, that other module is [here](https://github.com/coopiteasy/addons/tree/16.0-add-project_task_rtree_view/project_task_rtree_view).

youring:
> just curious: how did you find it?

Haha...by the clue of your added screenshot, I thought maybe you had pr submitted on OCA/project, but not found. Then I tried look around your own repos by latest activities, and found the 16.0-add-project_task_rtree_view branch.

Thanks for the roadmap. And more suggestions:
Under the usage of rtree scenarios, in most of the cases while after viewing the child items we tend to go back the the parent rtree view to click the next records, at this point, I prefer:
1. to know which row i just clicked, esp. after a long record digest. Maybe change style of background color of last viewed row would be enough.
2. to click next/another record in pop-up window ( also called dialog window or likely wizard window), do some review or edit/update/save, and return to rtree view.
3. auto refresh the rtree view if update happens in 2.

BTW, I tried "edit=bottom", seems not working.

huguesdk:
@youring thanks for the suggestions! fyi, as a first step, the view was developed to be read-only, hence some of the limitations.

bealdav:
Hi it seems a really nice module here thanks.

Is there a lot of job to release this module.

Thanks a lot

huguesdk:
hi @bealdav. [the roadmap](https://github.com/coopiteasy/web/tree/3eee82fed038c73c1756c92ce4a48b18512af06c/web_view_rtree#known-issues-roadmap) explains the main issues that should ideally be resolved before it is production-ready. also, it is currently a read-only view, it does not allow to edit or add new records. feel free to test it, report new issues, and maybe even propose changes by creating pull requests against [this branch](https://github.com/coopiteasy/web/tree/16.0-add-web_view_rtree).

huguesdk:
@youring the rtree view now remembers its state when navigating away and back to it. feel free to report any issue regarding this.

youring:
> the rtree view now remembers its state when navigating away and back to it.

Yes, by clicking the breadcrumb, it works. But not work by clicking browser backward arrow, that would be more compliant with ergonomics ;)

Still tested locally using your project_task_rtree_view module. 
1. It seems there is some css issue in the column selector drop-down menu. I can not select the Customer column. If i click that position, it will actually open the first project.
image

2. The sub row (child items) is now click-able, and seems edit-able, but not save-able. Maybe we should make it un-clickable since rtree views are currently read only.
image

3. After uninstall web_view_rtree(with project_task_rtree_view), go back to original tree view (Project > Tasks), it pops up User Error
> No default view of type 'rtree' could be found !

Have to upgrade project module to reload odoo original view.

luc-demeyer:
@huguesdk 
I downloaded the module and it looks promising, let's hope we get it merged one of these days.
I tried to use it for a custom development  and for which I want to pass a context to the rtree view (for use in computed fields).
But as opposed to the standard tree view the context that I pass via the action is removed in the rtree view.
Is this on purpose or something that still needs to be fixed ?

huguesdk:
@youring many thanks for your feedback. i totally agree that the back button should work. if you have any information about how this should be fixed, feel free to share it.

@luc-demeyer the context not being passed is most probably a bug. if you can help fixing this, it would be very welcome.

any help is welcome: please create pull requests on the [`16.0-add-web_view_rtree` branch in the coopiteasy/web repository](https://github.com/coopiteasy/web/tree/16.0-add-web_view_rtree).

huguesdk:
closing this for now. the code has been moved to [this branch](https://github.com/IMIO/oca-web/tree/16.0-add-web_view_rtree), and @IMIO will open a new pull request when the code will be mature enough.

PR #2523, opened 2023-05-30
[15.0][IMP] web_widget_open_tab

AungKoKoLin1997:
This PR adds open-tab field in tree views by selecting "Add Open Tab Field" field in the ir.model record. When you do this, the open-tab field is added right after the name field in the tree if the field exists, otherwise at the beginning of the tree.
Back port of https://github.com/OCA/web/pull/2572
@qrtl

### DIFF
--- a/web_widget_open_tab/models/ir_ui_view.py
+++ b/web_widget_open_tab/models/ir_ui_view.py
@@ -1,0 +1,32 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from lxml import etree
+
+from odoo import api, models
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        res = super(Base, self).fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar, submenu=submenu
+        )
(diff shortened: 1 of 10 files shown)


web_widget_open_tab/models/ir_ui_view.py
```diff
@@ -0,0 +1,32 @@
...
+        res = super(Base, self).fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar, submenu=submenu
+        )
+        arch = etree.fromstring(res["arch"])
+        model = self.env["ir.model"]._get(self._name)
+        if view_type == "tree" and model.add_open_tab_field:
+            id_elem = """<field name="id" widget="open_tab" nolabel="1"/>"""
+            id_elem = etree.fromstring(id_elem)
+            tree = arch.xpath("//tree")[0]
+            name_field = tree.xpath('./field[@name="name"]')
```
yostashiro: Wonder if we should consider display_name field as well. In that case, we may prepare a method to return the name field.

e.g.
```
    @api.model
    def _get_name_field(self, tree):
        if tree.xpath('./field[@name="display_name"]'):
            return tree.xpath('./field[@name="display_name"]')
        return tree.xpath('./field[@name="name"]')
```

yostashiro (approved):
Code review and functional test in runboat. 👍 

image
image

rinaldifirdaus (approved):
Functional Test: Work As Expected!

AungKoKoLin1997:
@oca/web-maintainers Is it possible to review this PR? This is the back port of merged PR for 16.0

PR #2528, opened 2023-06-09
[16.0] web_drop_target: Migration to 16.0

cjallais:
Migration to 16.0


CoraVanga:
Hi @cjallais , could you please check again

I tried to drag and drop the file to the form view but it seems that it's not work

image

jcdrubay:
Instead of this in the README: 

> #. drag a file from your local computer onto an Odoo form view

It seems that the current version of the module behaves like this:

> #. drag a file from your local computer onto the Attachment area of the open chatter in Odoo form view


But this change make it harder to override for other behaviors. I would like to drag & drop anywhere in the form view to upload the file in a binary field displayed in the form view instead of adding as attachment.

Such behavior would be nice for instance to upload a Proof of Payment in the "Register Payment" overlay window
image

yibudak:
@cjallais What's the status of this PR?

luisDIXMIT:
Watching this PR, I think this module is no longer necessary in 16.0 https://github.com/OCA/OpenUpgrade/pull/4134

PR #2534, opened 2023-06-17
[16.0][IMP] web_company_color: Forward port #2209 from 14.0

HekkiMelody:
Cherry pick of d59008f837740f2d14231cea527c366a81a7799f and 8c8ffe752855c5eadd6cbc724a37b19e4c348348

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -44,6 +44,47 @@ class ResCompany(models.Model):
             }
           }
         }
+        a[href],
+        a[tabindex],
+        .btn-link,
+        .o_external_button {
+          color: %(color_link_text)s !important;
+        }
+        a:hover,
+        .btn-link:hover {
+          color: %(color_link_text_hover)s !important;
+        }
+        .btn-primary:not(.disabled),
+        .ui-autocomplete .ui-menu-item.ui-state-focus a {
+          color: %(color_button_text)s !important;
+          background-color: %(color_button_bg)s !important;
+          border-color: %(color_button_bg)s !important;
+        }
+        .btn-primary:hover:not(.disabled),
+        .ui-autocomplete .ui-menu-item.ui-state-focus a:hover {
+          color: %(color_button_text)s !important;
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

CasVissers-360ERP (approved):

aleuffre:
For completeness' sake, I feel like I should mention that the module doesn't work as cleanly as it used to in v14 and below. See also the Migration to v15.

https://github.com/OCA/web/pull/2467

HekkiMelody:
For completeness' sake, I feel like I should mention that the module doesn't work as cleanly as it used to in v14 and below. See also the forward port of the same commits v15, that suffers from similar issues.

https://github.com/OCA/web/pull/2467

LoisRForgeFlow:
Hi all,

it would be great if you could drop your opinion on this issue https://github.com/OCA/web/issues/2721.

Thanks in advance.

web_company_color/models/res_company.py
```diff
@@ -118,6 +174,14 @@ def _scss_get_sanitized_values(self):
...
                 ),
                 "color_navbar_text": (values.get("color_navbar_text") or "#FFF"),
+                "color_button_bg": values.get("color_button_bg") or "$primary",
+                "color_button_bg_hover": values.get("color_button_bg_hover")
+                or 'darken(theme-color("primary"), 10%)',
+                "color_button_text": values.get("color_button_text") or "#FFF",
+                "color_link_text": values.get("color_link_text")
+                or 'theme-color("primary")',
+                "color_link_text_hover": values.get("color_link_text_hover")
+                or 'darken(theme-color("primary"), 10%)',
```
hbrunn, line 184: you might want to double check if those fallbacks are still the default values in this version, otherwise there'll be weird effects when not all values are set by the user

anajuaristi:
Hello
Testing this great module in V16, I can say that today, it's needed to fill any color field on company form to make it work but setting them properly it works really fine. 
Thank you all!

PR #2536, opened 2023-06-20
[UPD] web_dialog_size: reduce rpc call to get config

hailangvn:
After installing web_dialog_size and setup its parameter, there's a slight delay where we can see the modal at its normal size and then after a rpc to ir.config_parameter:get_web_dialog_size_config() the modal is maximized.

This commit is to set a session value then use it to set initial dialog size upfront without delay.

### DIFF
--- a/web_dialog_size/models/__init__.py
+++ b/web_dialog_size/models/__init__.py
@@ -1,1 +1,2 @@
 from . import ir_config_parameter
+from . import ir_http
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):
Very needed patch!

StefanRijnhart (approved):
Code review, functional test.

pedrobaeza:
Oops, I have aborted my merge.

PR #2538, opened 2023-06-21
[15.0][MIG] web_disable_export_group: Migration to 15.0

chienandalu:
- Refactor to make the `web_disable_export_group.group_export_xlsx_data` as a standalone permission

cc @Tecnativa TT43531

please review @victoralmau @etobella @pedrobaeza @sergio-teruel


etobella (approved):
LGTM :+1:

pedrobaeza:
Having checked only "Direct export (XLSX)":

Selección_033

and if I checked the other, both are checked automatically, so no granularity at the end.

chienandalu:
> Having checked only "Direct export (XLSX)" [error]:

Fixed. I had manually added the ACL in my local :facepalm: 

> and if I checked the other, both are checked automatically, so no granularity at the end.

It makes sense as:

- When you have full export access, why not having the limited view export feature (which is the same under the hood)
- But you may want to limit that some users can only export that screen results.
- And you for sure want to avoid that some other users can't export data at all.

Something funny I realized doing tests: the pivot export has a completely different mechanism and it isn't limited in any manner :S (should we put it behind the xls export group?)

victoralmau (approved):

PR #2544, opened 2023-06-27
[16.0][UPD] web_theme_classic Make placeholders more visible

remi-filament:
This PR proposes to have placeholder in fields more visible, since these were hardly visible since v16...

Previous display : 
image

New display : 
image

### DIFF
--- a/web_theme_classic/__manifest__.py
+++ b/web_theme_classic/__manifest__.py
@@ -18,6 +18,9 @@
         "web.assets_backend": [
             "/web_theme_classic/static/src/scss/web_theme_classic.scss",
         ],
+        "web._assets_primary_variables": [
+            "/web_theme_classic/static/src/scss/primary_variables.scss",
+        ],
     },
     "installable": True,
     "application": True,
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Makes sense ! thanks for your contribution !

benj-filament (approved):

Tardo (changes requested):
This should be done using `web._assets_primary_variables` asset bundle to modify the `$input-placeholder-color` variable.

remi-filament:
> This should be done using `web._assets_primary_variables` asset bundle to modify the `$input-placeholder-color` variable.

Thanks @Tardo for looking into this, it is indeed much simpler and cleaner, done !

Tardo:
Now I see it a little darker than in the screenshots. I don't know why :/...

remi-filament:
I kept the same value, so there should be no difference, maybe this is related to browser ?

Tardo:
Yes... Firefox seems to apply a color fading, while chromium does not :S

Tardo:
Useful info: https://stackoverflow.com/a/19871806

So, you should modify 'web theme_classic.scss' and add those rules with opacity 1.0 :/

** For me this should be fixed directly in Odoo.

remi-filament:
Thanks for your findings. I see this is now fixed in upstream (https://github.com/odoo/odoo/commit/aa7ab0a92c16a2e911ad39b73d2535572f892e75) so this PR is not needed anymore AFAICT

PR #2546, opened 2023-06-28
[16.0][IMP] web_help: Improve helpers

tarteo:
Add some helper functions from version 15.0

### DIFF
--- a/web_help/static/src/helpers.esm.js
+++ b/web_help/static/src/helpers.esm.js
@@ -1,16 +1,40 @@
 /** @odoo-module **/
 import {registry} from "@web/core/registry";
+import {Component} from "@odoo/owl";
 
-export function findTrip(model, viewType) {
+export async function findTrip(model, viewType) {
     const trips = registry.category("trips").getAll();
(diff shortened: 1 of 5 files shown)


web_help/static/src/helpers.esm.js
```diff
@@ -1,16 +1,40 @@
...
         return matchedTrips[0].Trip;
     }
     return null;
 }
+
+export function wait(ms) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export async function waitUntilAvailable(selector, ms = 50) {
```
Tardo, line 24: A bit "hard" to traverse the DOM tree every 50ms by default. Perhaps you could use MutationObserver, but the implementation would be more complex and probably not worth the effort.

Tardo (approved):
A small visual glitch with the icon. It can be fixed by moving the icon inside '.o_cp_bottom_right':
icon_help

tarteo:
> A small visual glitch with the icon. It can be fixed by moving the icon inside '.o_cp_bottom_right': icon_help

@Tardo Thanks fixed it

PR #2548, opened 2023-06-28
[16.0][MIG] web_ir_actions_act_window_page

hbrunn:
This addon allows a developer to return the following action types:

`{'type': 'ir.actions.act_window.page.next'}`

or:

`{'type': 'ir.actions.act_window.page.prev'}`

which trigger the form's controller to page into the requested direction on the client
side.

A use case could be the case of a validation flow. As a developer, you set up a tree
view with a domain on records to be validated. The user opens the first record in a form
view and validates the record. The validation method returns the 'next' action type so
that the browser window of the user is presented with the next record in the form view.


tarteo (approved):
Thanks! Tested locally works perfectly

image

StefanRijnhart:
Please mention the version numbers in your migration commits.
image

web_ir_actions_act_window_page/__init__.py
```diff
@@ -1,19 +1,3 @@
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2013 Therp BV (<http://therp.nl>).
```
StefanRijnhart: Maybe keep the original credits?

StefanRijnhart: Oh I do still see them elsewhere, so that's OK.

StefanRijnhart (changes requested):
This is beautifully adapted to the new upstream code. The use of the bus is revealing.
Please fix the commit messages.

web_ir_actions_act_window_page/static/src/web_ir_actions_act_window_page.esm.js
```diff
@@ -0,0 +1,43 @@
...
+    executeWindowActionPage(params, 1)
+);
+actionHandlersRegistry.add("ir.actions.act_window.page.list", async (params) =>
+    executeWindowActionList(params)
+);
+
+patch(Pager.prototype, "navigate event listener", {
+    setup() {
+        this._super.apply();
+        const handleNavigate = (ev) => this._handleNavigate(ev);
```
ged-odoo, line 34: random unsolicited code review. I hope it's not rude. You can import `useBus` from `"@web/core/utils/hooks"`. It allow you to register when mounted and it will take care of unregistering when unmounted:
```js
  useBus(this.env.bus, "pager:navigate", handleNavigate);
```

hbrunn: on the contrary, very much appreciated and sorry for the long delay! Took your suggestion in 3a0e116feac

StefanRijnhart (approved):

PR #2551, opened 2023-07-04
[13.0][IMP] web_widget_one2many_product_picker: Allow set first active group instead of set All by default

CarlosRoca13:
cc @Tecnativa TT41603

ping @carlosdauden @pedrobaeza


pedrobaeza:
AFAIK, there was a mechanism for selecting the active tab. Anyway, here you are just selecting the next one, isn't it?

carlosdauden (approved):

CarlosRoca13:
> AFAIK, there was a mechanism for selecting the active tab. Anyway, here you are just selecting the next one, isn't it?

I tried before this changes to set as active another one, but just seems to be selected, because the _activeSearchGroup was allways set as the first element. And the first element is allways the **All** (Even though the button you have set as active is shown as selected)

pedrobaeza (approved):
Seen directly by chat.

/ocabot merge patch

PR #2553, opened 2023-07-05
[ADD] web_field_tooltip : new module to manage tooltips

baimont:
Hello,

Here I propose a new module to manage tooltips for fields on any model.

This module is directly inspired from https://github.com/OCA/web/pull/1460 

You will be able to Manage Tooltips for a model on the form view, which will allow you to display a tooltip text next to the field of your choice.

image

image


web_field_tooltip/models/ir_model_fields_tooltip.py
```diff
@@ -0,0 +1,59 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class IrModelFieldsTooltip(models.Model):
+
+    _name = "ir.model.fields.tooltip"
+    _description = "Field Tooltip"
+    _rec_name = ""
```
benwillig: I'm not sure this is needed

web_field_tooltip/static/src/js/web_field_tooltip.js
```diff
@@ -0,0 +1,131 @@
...
+         * @override
+         */
+        _renderTagLabel: function(node) {
+            var self = this;
+            var $result = this._super.apply(this, arguments);
+            var fieldName = node.tag === "label" ? node.attrs.for : node.attrs.name;
+
+            // Check if there's any tooltip that must be rendered for the given view
+
+            rpc.query({
```
benwillig: I don't think it's a good approach to make the query here, as it will create one query per field each time we render it (even using the pagination and even if there is no tooltip for the current model). I think the tooltip per fields should be retrieved only one time, when requesting the views to be displayed

web_field_tooltip/models/ir_model_fields_tooltip.py
baimont: right, thanks

web_field_tooltip/static/src/js/web_field_tooltip.js
baimont: you're right. i'll try to improve this

baimont: here's a solution that doesn't satisfy me fully: https://github.com/OCA/web/pull/2553/commits/5fbb022a00dc317f57d400b5d1fad156ed6120df

problem is if you don't reload the page and go back directly to form view using breadcrumbs, you'll have old tooltips (new tooltips won't be there and deleted tooltips will still be there)

benwillig: In my opinion this is not a problem as there are a lot of things that are not refreshed without doing a full reload. Can you just store the tooltips on the FormRenderer instead of ```$el``` ? Even if it works, it seems weird to me to store data on the DOM element. Otherwise I think we should inherit from load_views to load the tooltips, and then use it in the FormRenderer, but I'm not sure that we could inherit easily from odoo code to store the retrieved tooltips.

rousseldenis:
@benwillig @baimont Could you check and fix tests ?

baimont:
> @benwillig @baimont Could you check and fix tests ?

done, time to approve :-)

baimont:
ping @ThomasBinsfeld

benwillig (approved):

jdoutreloux (approved):

baimont:
@OCA/web-maintainers ready to merge :-)

FrancoMaxime (approved):
LGTM

PR #2554, opened 2023-07-05
[16.0][MIG] web_tree_duplicate

hbrunn:
With this module you can duplicate records directly from the tree view.


StefanRijnhart (commented):
I thought this would be a fun way to quickly create thousands of duplicate partners, but even though I click 'Select <total nr of> records' in the tree view, it is only the 80 visible records that are duplicated. Is this on purpose? 

image

After duplication, there are 80 records more:

image

hbrunn:
not on purpose, but I just saw this is consistent with how `delete` works: https://github.com/OCA/OCB/blob/16.0/addons/web/static/src/views/list/list_controller.js#L452

I'd be inclined to address this issue with a sentence in the readme that if you really want to duplicate all records, you have to increase the limit of your list view, would that be sufficient in your opinion?

StefanRijnhart:
Documenting the behaviour in the readme would be great!

hbrunn:
done

StefanRijnhart (approved):
Thanks! Care to squash that fixup commit?

houssine78 (approved):

trisdoan:
Hello @hbrunn, I see you made the duplicate option enabled by default, which is different from the original version. Is there any rationale behind this ?

Also, the docs confused me.

> Set `duplicate` to `false` to enable it or `true` to (explicitly) disable it.

It should be: Set `duplicate` to `false` disable it, right?

hbrunn:
the rationale is to behave consistently with how this works on forms.

Thanks for the heads-up about the readme, fixed in https://github.com/OCA/web/pull/2788

PR #2560, opened 2023-07-10
[14.0][ADD] web_model_read_only

solo4games:
Module disables modification operations (write, create, delete) on selected model to selected users groups.
supersedes #2509


tarteo:
Hey @solo4games, so if I understand correctly this is a new method to make models readonly for some groups.
I think this module is more suitable for https://github.com/oca/server-tools

francesco-ooops:
@tarteo thanks for the tip, do you think we should rename it "model_read_only" too?

francesco-ooops:
@solo4games you can close this one

PR #2562, opened 2023-07-12
[16.0][FIX] web_widget_bokeh_chart: Make bokeh charts work when inputs change

BernatPForgeFlow:


web_widget_bokeh_chart/static/src/js/web_widget_bokeh_chart.esm.js
```diff
@@ -21,9 +21,13 @@ class BokehChartWidget extends CharField {
...
     get json_value() {
-        var value = JSON.parse(this.props.value);
-        value.div = markup(value.div.trim());
-        return value;
+        try {
+            var value = JSON.parse(this.props.value);
+            value.div = markup(value.div.trim());
+            return value;
+        } catch (error) {
+            return {};
```
LoisRForgeFlow: maybe a message in the place of the graph would be better to understand that there is an issue. Something like "Could not display graph" or similar. WDYT?

LoisRForgeFlow (approved):
Thanks! It now works as expected

PR #2565, opened 2023-07-14
[15.0][FIX] web_responsive: increase z-index of attachment_viewer

CarlosRoca13:
Before this change, when the attachment_viewer was opened from the chat, it was displayed in front of the chat and could obscure part of the image.
image

After this change, this is no longer the case, as the attachment_viewer is displayed in front of the chat.
image

cc @Tecnativa TT44471

ping @pedrobaeza @pilarvargas-tecnativa


legalsylvain (approved):

pilarvargas-tecnativa (approved):
LGTM :)

PR #2569, opened 2023-07-18
[13.0][MIG] web_chatter_position

SergiCForgeFlow:


LoisRForgeFlow (approved):
Functional review in EE and CE (without web_responsive). Working as expected :+1:

JordiMForgeFlow:
It looks like the field appears twice in the user preferences view:
image

JoanSForgeFlow (approved):
LGTM!

JordiMForgeFlow (approved):
Code + functional review 👍🏼

LoisRForgeFlow:
> It looks like the field appears twice in the user preferences view: image

That is because the feature is also included in web_responsive, this module extract it to be able to have the feature without web_responsive or with web_enterprise

PR #2574, opened 2023-07-26
[FIX] web_widget_bokeh_chart : bad image path in description, break readme file

legalsylvain:
Trivial.

the image is not visible here https://github.com/OCA/web/blob/16.0/web_widget_bokeh_chart/README.rst

with that patch, it is.


LoisRForgeFlow (approved):
Thanks!

/ocabot merge patch

PR #2578, opened 2023-07-31
[15.0][FIX] web_search_with_and

JordiToledo:
When we search by the id of any record and not a string, it gives an error on the line: ? source.value.trim().
It is because the value of source.value is an integer. To fix it, we do the trim() operation just if source.value is an string.

Update: I haven't explained myself well. What i tried to say is when we search by a suggested value:
image

When we select "My Company", source.value is an integer, and this why we cannot trim. It is just this case, when we search for "my com", for example, we search by a string and we can trim.

### DIFF
--- a/web_search_with_and/static/src/js/search_bar.js
+++ b/web_search_with_and/static/src/js/search_bar.js
@@ -27,8 +27,8 @@ odoo.define("web_search_with_and/static/src/js/search_bar.js", function (require
                     filterId: source.filterId,
                     value:
                         "value" in source
-                            ? source.value.trim()
-                            : this._parseWithSource(labelValue, source).trim(),
+                            ? source.value
+                            : this._parseWithSource(labelValue.trim(), source),
                     label: labelValue.trim(),
                     operator: source.filterOperator || source.operator,
                     isShiftKey: this.isShiftKey,


web_search_with_and/static/src/js/search_bar.js
```diff
@@ -27,7 +27,7 @@ odoo.define("web_search_with_and/static/src/js/search_bar.js", function (require
                     filterId: source.filterId,
                     value:
                         "value" in source
-                            ? source.value.trim()
+                            ? source.value
```
StefanRijnhart, line 30: Tested on this PR's runboat: added a search field `id` on the res.users search view. Search for value '1'. Same error still occurs.

`source.value` is a string type in line 30, but cast as an integer by `_parseWithSource`. You might want to check the type of the return value of `_parseWithSource` before trimming.

image

JordiToledo: I moved the trim() function inside the _parseWithSource. With that modification, we will never try to trim and integer value and we keep trimming the strings.

StefanRijnhart:
Looking good now! Please squash the commits.

JordiToledo:
Done.

StefanRijnhart (approved):
Thanks!

Review: code review and functional test.

PR #2591, opened 2023-08-17
[ADD] web_session_auto_close

IT-Ideas:
This commit adds the web_session_auto_close module which purpose is to is to automatically close the session when the last odoo tab is closed or when the computer gets in idle mode.

The `timeout` occurs 15 seconds after the tab has been closed or the computer got idle.


IT-Ideas:
@sbidoul here is the PR for the auto session timeout at tab closing. Could you have a look and let me know what you think of it?
Thanks!

sbejaoui (approved):
Code review + functional test

pierre-halleux (approved):

IT-Ideas:
Just rebased and corrected typo in `manifest.py`

PR #2571, opened 2023-07-24
[16.0] [FIX] web_chatter_position: fixes JS assuming "o_form_sheet_bg" is always found

BT-jmontero:
Fixes the issue reported here: https://github.com/OCA/web/issues/2464

PR #2595, opened 2023-08-21
[FIX] web_chatter_position: Fixes bug where incompatible apps crash

Florian-Laport:
In some (new) apps, the query selector `querySelector('.o_form_sheet_bg')` in `web_chatter_position` will not return anything. This causes a crash when you try to open the app in question.

You can replicate this by installing both the new 'Knowledge' app, as well as `web_chatter_position`, and trying to open the former when the current user's preferences have been set to chatter position 'bottom'.

Odoo avoids this problem by exiting early without a chatter if there no `sheet-bg`. This PR implements the exact same solution. See the Odoo source [here](https://github.com/odoo/odoo/blob/16.0/addons/mail/static/src/views/form/form_compiler.js#L171).

### DIFF
--- a/web_chatter_position/__manifest__.py
+++ b/web_chatter_position/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Chatter Position",
     "summary": "Add an option to change the chatter position",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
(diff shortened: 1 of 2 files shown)


Florian-Laport:
Apparently duplicate of #2571 as it fixes the same issue. Leaving it open because I think this one should be merged instead. May be closed otherwise.

Florian-Laport:
#2571 was merged so I'm closing this.

PR #2600, opened 2023-08-29
[16.0][FIX] web_timeline: Add record -> popup form

norlinhenrik:
`resId: null` caused an error when creating a record from timeline view: resId is not a number or boolean.
`resId: false` fixed the error.


pedrobaeza (approved):

etobella (approved):
/ocabot merge patch

PR #2601, opened 2023-08-30
[16.0][MIG] web_calendar_slot_duration: Migration to 16.0

norlinhenrik:


norlinhenrik:
This module is good.
Tests fail because of `web_domain_field`.

CRogos:
Are you sure this is still working? Either I do something wrong, or it is not working. 
I've also realized that the Odoo default calendar snap duration changed from 30 min to 15 min.

image
Have you tried something else than 0:15:00?

norlinhenrik:
So that's what this module is for. I agree, the action context does not change the behavior in the calendar view in 16.0.

I did a basic migration because it is a dependency for `calendar/resource_booking`.

pedrobaeza:
Superseded by #2646

PR #2602, opened 2023-08-31
[16.0][FIX] web_company_color: Changing just one color fixed

EvaSForgeFlow:
Before it didn't work if just a color was set because it couldn't find "color_navbar_bg_hover". Moreover, in odoo enterprise the navbar background color only worked for the main app menu, now it works for both main app menu and the navbar of the apps.

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -16,6 +16,7 @@ class ResCompany(models.Model):
 
     SCSS_TEMPLATE = """
         .o_main_navbar {
+          background: %(color_navbar_bg)s !important;
           background-color: %(color_navbar_bg)s !important;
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
Before the fix:


After the fix:


The problem was only in EE.

LoisRForgeFlow (approved):
Functional and code review :+1:

LoisRForgeFlow:
To help review I extend the explanation of the color_navbar_bg_hover change. The css variable `$o-navbar-inverse-link-hover-bg` has been dropped in v16, instead an on the fly `darken` is done based on the base color. So if no specific color is selected for the hover, the navbar color darkened 10% will be used.

etobella (approved):

PR #2605, opened 2023-09-04
[14.0][IMP] web_ir_actions_act_view_reload: add model ir_actions

Ricardoalso:
Rebase operated with branch 14.0

https://github.com/OCA/web/pull/2192

### DIFF
--- a/web_ir_actions_act_view_reload/models/ir_actions.py
+++ b/web_ir_actions_act_view_reload/models/ir_actions.py
@@ -1,0 +1,28 @@
+from odoo import fields, models
+
+
+class IrActionsActViewReload(models.Model):
+    """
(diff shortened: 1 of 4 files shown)


web_ir_actions_act_view_reload/models/ir_actions.py
```diff
@@ -0,0 +1,28 @@
+from odoo import fields, models
+
+
+class IrActionsActViewReload(models.Model):
```
chienandalu, line 4: Do using an `AbstractModel` have the same effect?

chienandalu: Follow the same technique as in https://github.com/OCA/web/pull/2347

PR #2608, opened 2023-09-06
[15.0][IMP] web_timeline: Add configuration options

zamberjo:
Added **horizontalScroll** and **verticalScroll** options.

SEE: https://visjs.github.io/vis-timeline/docs/timeline/


pedrobaeza:
This is not [ADD], but [IMP], as it's not adding a module, but a feature inside an existing module.

zamberjo:
> This is not [ADD], but [IMP], as it's not adding a module, but a feature inside an existing module.

You're right, I did it quickly and didn't think about it.

pedrobaeza (approved):

Jenniferrin (approved):
I have checked the functioning and it is correct.

PR #2610, opened 2023-09-08
[15.0][IMP] web_responsive: public message chatter color

chienandalu:
We give a color clue to the users so they can better tell if they're sending an internal note or a public message.

Peek 08-09-2023 10-10

Dislaimer: we use the `has()` pseudo class mainly because is simpler to implement and the support is quite extended right now. Even Firefox will catch up soon (now it has support behing a config flag).

So users with compatible browser will enjoy it, while others will just stay as they were.

cc @Tecnativa TT44953

please review @carolinafernandez-tecnativa @pedrobaeza

PR #2613, opened 2023-09-13
[16.0][IMP] web_responsive: public message chatter color

chienandalu:
fw of #2610 

We give a color clue to the users so they can better tell if they're sending an internal note or a public message.

Dislaimer: we use the `has()` pseudo class mainly because is simpler to implement and the support is quite extended right now. Even Firefox will catch up soon (now it has support behing a config flag).

So users with compatible browser will enjoy it, while others will just stay as they were.

cc @Tecnativa TT44953

please review @pedrobaeza @carolinafernandez-tecnativa


pedrobaeza (approved):

chienandalu:
ping @carolinafernandez-tecnativa

carolinafernandez-tecnativa (approved):

carolinafernandez-tecnativa (approved):

PR #2614, opened 2023-09-15
[14.0][ADD] web_assets_warmup

jbaudoux:
Ensure that assets are generated and stored in the DB when Odoo starts

If the assets from the database are not up-to-date, they are regenerated by
Odoo when we print a report, but to do so Odoo forces the commit, so if an
exception occurs after (or during) the report rendering, it let the database in
a broken state (picking have been validated in this case).

To prevent this issue, we need to ensure that the assets are well-generated
when Odoo starts, not when the report is printed.

cc @sebalix @mt-software-de


sebalix (approved):

simahawk (approved):

simahawk:
FTR I'm not sure if the name is really explicative. Something like `web_assets_warmup` might be more appropriated. Not a blocker tho.

web_generate_assets/hooks.py
```diff
@@ -0,0 +1,52 @@
...
+    dbname = os.environ.get("DB_NAME")
+    reg = registry(dbname)
+    with reg.cursor() as cr:
+        cron_module, cron_ref = "web_generate_assets", "cron_generate_assets"
+        query = """
+           SELECT model, res_id
+           FROM ir_model_data
+           WHERE module=%s
+           AND name=%s;
+        """
```
mt-software-de, line 25: Since we could have dedicated services for odoo web/odoo cronjobs, we should a a lock.
Otherwise we may get a "serialize access" error. 
```suggestion
        query = """
           SELECT model, res_id
           FROM ir_model_data
           WHERE module=%s
           AND name=%s
           FOR UPDATE;
        """
```

mt-software-de (changes requested):

web_generate_assets/hooks.py
mt-software-de: I don't think we should use a "FOR UPDATE" because it locks the complete ir_model_data table. :-)

web_generate_assets/hooks.py
```diff
@@ -0,0 +1,52 @@
...
+        args = (cron_module, cron_ref)
+        cr.execute(query, args)
+        row = cr.fetchone()
+        # post_load hook is called before the update of the module so the
+        # ir_cron record doesn't exist on first install
+        if row:
+            model, res_id = row
+            if model != "ir.cron":
+                return
```
mt-software-de, line 34:
```suggestion
import hashlib
import logging
import os
import struct

from odoo import fields, registry

logger = logging.getLogger(__name__)


def active_cron_assets():
    """Plan the next execution of the cron responsible to generate assets."""
    if os.environ.get("RUNNING_ENV") == "dev":
        return
    dbname = os.environ.get("DB_NAME")
    reg = registry(dbname)
    with reg.cursor() as cr:
        cron_module, cron_ref = "web_generate_assets", "cron_generate_assets"
        query = """
           SELECT model, res_id
           FROM ir_model_data
           WHERE module=%s
           AND name=%s;
        """
        args = (cron_module, cron_ref)
        cr.execute(query, args)
        row = cr.fetchone()
        # post_load hook is called before the update of the module so the
        # ir_cron record doesn't exist on first install
        if row:
            model, res_id = row
            if model != "ir.cron":
                return
            hasher = hashlib.sha1(str("web_generate_assets").encode())
            int_lock = struct.unpack("q", hasher.digest()[:8])
            cr.execute("SELECT pg_advisory_xact_lock(%s);", (int_lock,))
```

mt-software-de (changes requested):

web_generate_assets/hooks.py
mt-software-de: Since we could have dedicated odoo services for web or cronjobs, we should create a lock. Otherwise we could get a serialize access error.

mt-software-de (changes requested):
Since we could have dedicated services for web and cronjobs. We should add a lock. Otherwise we can get a "serialize access" error. 

https://github.com/camptocamp/web/pull/3

rvalyi:
> FTR I'm not sure if the name is really explicative. Something like `web_assets_warmup` might be more appropriated. Not a blocker tho.

+1 for web_assets_warmup

jbaudoux:
web_generate_assets_initially ?

rvalyi:
> web_generate_assets_initially ?

asset "cache warmup" is really the technical term web developers would expect in a web framework IMHO. See for instance https://www.google.com/search?client=ubuntu-chr&hs=mW0&sca_esv=573172220&q=web+asset+cache+%22warmup%22+rails&oq=web+asset+cache+%22warmup%22+rails&aqs=heirloom-srp..

bofiltd (approved):

pedrobaeza:
@bofiltd putting a lot of reviews in few minutes in the corresponding PRs is not trustful and can't be taken as a serious review, so please abstain for doing that.

jbaudoux:
Module renamed to **web_assets_warmup**

web_assets_warmup/hooks.py
```diff
@@ -0,0 +1,70 @@
...
+from odoo import fields, registry
+
+logger = logging.getLogger(__name__)
+
+
+def active_cron_assets():
+    """Plan the next execution of the cron responsible to generate assets."""
+    if os.environ.get("RUNNING_ENV") == "dev":
+        return
+    dbname = os.environ.get("DB_NAME")
```
simahawk, line 19: `DB_NAME` is not a core odoo var. The db name should be take from config.
Not blocking, you can fix this later as this one is waiting since quite some time.

PR #2616, opened 2023-09-16
[10.0][FIX] web_responsive: Filter properly menus on AppDrawer search

pedrobaeza:
Disclaimer: On a project we have recovered bounced from other partner, and meanwhile we migrate it to latest versions, we have found this problem, that we fix and put it back into the community for those that are sticky to this version.

The filtering that is done on the menus by upstream doesn't take into account the visibility of their ancestors to decide if the menu is visible or not, which is needed for the AppDrawer menu search.

@Tecnativa TT45078

### DIFF
--- a/web_responsive/__init__.py
+++ b/web_responsive/__init__.py
@@ -1,1 +1,2 @@
 # -*- coding: utf-8 -*-
+from . import models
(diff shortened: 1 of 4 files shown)


pedrobaeza:
As expected, the old CI is not working, so this will require a manual merge.

chienandalu (approved):

carolinafernandez-tecnativa (approved):

PR #2617, opened 2023-09-22
[ADD][16.0] web_apply_field_style module

bealdav:
Apply css of your choice to arbitrary fields

demo


web_apply_field_style/models/base.py
```diff
@@ -0,0 +1,28 @@
...
+    def _get_view(self, view_id=None, view_type="form", **options):
+        arch, view = super()._get_view(view_id, view_type, **options)
+        if view_type == "form":
+            self._update_css_class(arch)
+        return arch, view
+
+    def _update_css_class(self, arch):
+        css = self._get_field_styles()
+        if css.get(self._name):
+            for style in css[self._name]:
```
legalsylvain:
```suggestion
            for style in css.get(self._name, []):
```

fcayre: +1 for Sylvain's remark, and deserves a specific test (generate a form view of a model with no style and check the view does not crash).

Useful module by the way, thank you for sharing!

fcayre (changes requested):

web_apply_field_style/models/base.py
bealdav: Yes, finally added with extra tests for different invalids entries. Thanks to you both

fcayre (approved):

web_apply_field_style/models/base.py
```diff
@@ -0,0 +1,37 @@
...
+        css = self._get_field_styles()
+        if css.get(self._name):
+            for style in css.get(self._name, []):
+                if not isinstance(css[self._name][style], list):
+                    raise exceptions.ValidationError(
+                        f"'{css[self._name][style]}' must be a list of fields"
+                    )
+                for field_name in css[self._name][style]:
+                    for field in arch.xpath(f"//field[@name='{field_name}']"):
+                        field.attrib["class"] = style
```
bealdav: TODO : need to add class instead of replace existing ones

web_apply_field_style/models/base.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @api.model
+    def _get_view(self, view_id=None, view_type="form", **options):
+        arch, view = super()._get_view(view_id, view_type, **options)
+        if view_type == "form":
+            self._update_css_class(arch)
+        return arch, view
+
+    def _update_css_class(self, arch):
+        css = self._get_field_styles()
```
bealdav, line 22: need to check the whole dict format in an other method instead of complexify current method ?

web_apply_field_style/README.rst
```diff
@@ -0,0 +1,101 @@
...
+    class Base(models.AbstractModel):
+        _inherit = "base"
+
+        def _get_field_styles(self):
+            res = super()._get_field_styles()
+            res["product.product"] = {
+                "my-css-class1": ["field1", "field2"],
+                "my-css-class2": ["field3"],
+            }
+            return res
```
legalsylvain, line 60:
```suggestion
    class ProductProduct(models.Model):
        _inherit = "product.product"

        def _get_field_styles(self):
            return = {
                "field1": ["my-css-class1", "my-css-class2"],
                "field2": ["my-css-class1"],
            }
```
Not a blocking point, but I see this implementation more simple.
Of course, it requires some refactoring.

bealdav: Thanks for the suggestion.
On my side I would use that module in the global scope of a project. I would inherit this method at the level of my project.
Then it's easier for me to only call the method once instead of for every model

What is your use case to call the method for each model ?

legalsylvain: > What is your use case to call the method for each model ?

well, I'm not sure to use this module, but I guess, that the usage could be, when we develop a new module to add this function in the model we are creating / orverloading.
That case, overloading base doesn't make sense, and force to create a dedicated file.

but I understand your needs also.

Don't know !

bealdav:
On my side I only create 1 file for all models.

If somebody tells me that he really want use this module and want a method overloaded by model, I could consider to introduce it, no problem for me.
On the contrary if you consider that the current code is correct, completely tested and documented, maybe you could approve the PR. Thanks a lot.

legalsylvain (approved):

bealdav:
Thanks a lot for your reviews

PR #2625, opened 2023-09-27
[15.0][FIX] web_advanced_search: Fixed changes from code editor not propagated

coffeinisfun:
Fix for changes from code editor not propagated to upper section

web_advanced_search_fix

### DIFF
--- a/web_advanced_search/static/src/js/DomainSelector.esm.js
+++ b/web_advanced_search/static/src/js/DomainSelector.esm.js
@@ -1,0 +1,57 @@
+/** @odoo-module **/
+
+import {patch} from "@web/core/utils/patch";
+import {_t} from "web.core";
+import Domain from "web.Domain";
+import DomainSelector from "web.DomainSelector";
+import basic_fields from "web.basic_fields";
+/**
+ * The redraw in the Debug Field does not trigger correctly
+ * so we overwrite it with the v14 Version
+ *
+ */
+patch(DomainSelector.prototype, "web.DomainSelector", {
+    /**
+     * @override
+     */
+    _onDebugInputChange(e) {
+        if (!$(".o_add_advanced_search").length) {
+            return this._super(...arguments);
+        }
+        const rawDomain = e.currentTarget.value;
+        try {
+            Domain.prototype.stringToArray(rawDomain);
+        } catch (err) {
+            // If there is a syntax error, just ignore the change
+            this.displayNotification({
+                title: _t("Syntax error"),
+                message: _t("Domain not properly formed"),
+                type: "danger",
+            });
+            return;
+        }
+        this._redraw(Domain.prototype.stringToArray(rawDomain)).then(
+            function () {
+                this.trigger_up("domain_changed", {
+                    child: this,
+                    alreadyRedrawn: true,
+                });
+            }.bind(this)
+        );
+    },
+});
+
+patch(basic_fields.FieldDomain.prototype, "web.basic_fields", {
+    /**
+     * Odoo restricts re-rendering the domain from the debug editor for supposedly
+     * performance reasons. We didn't ever came up with those and in v17 it's supported
+     * in the new advanced search.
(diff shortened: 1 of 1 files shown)


chienandalu (commented):
This indeed seems to be a deliberate change: https://github.com/odoo/odoo/pull/79855

Anyway, I'd target a patch to Odoo.

I just tested removing this `debug` flag that is passed to the `domain_changed` custom event and everything works flawlessly: https://github.com/odoo/odoo/blob/540dcc714acacc57bd3e971c758eb7d29941398d/addons/web/static/src/legacy/js/widgets/domain_selector.js#L604

My guess would be that the cause for the original patch maybe no longer applies. Indeed this functionality is recoverd in the new search tool for v17 that hopefully will deprecate this OCA module :)

web_advanced_search/static/src/js/DomainSelector.esm.js
```diff
@@ -0,0 +1,43 @@
...
+            this._redraw(Domain.prototype.stringToArray(rawDomain)).then(
+                function () {
+                    this.trigger_up("domain_changed", {
+                        child: this,
+                        alreadyRedrawn: true,
+                    });
+                }.bind(this)
+            );
+        }
+    },
```
chienandalu: I'd limit this override to the case we're on the advanced search dialog:
```suggestion
    /**
     * Our event is losing itself on our custom dialog. Just trigger the redraw
     * immediately
     * @override
     */
    _onDebugInputChange (e) {
        if (!$(".o_add_advanced_search").length) {
            return this._super(...arguments);
        }
        const rawDomain = e.currentTarget.value;
        try {
            Domain.prototype.stringToArray(rawDomain);
        } catch (err) { // If there is a syntax error, just ignore the change
            this.displayNotification({ title: _t("Syntax error"), message: _t("Domain not properly formed"), type: 'danger' });
            return;
        }
        this._redraw(Domain.prototype.stringToArray(rawDomain)).then(
            function () {
                this.trigger_up("domain_changed", {
                    child: this,
                    alreadyRedrawn: true,
                });
            }.bind(this)
        );
    },
```

web_advanced_search/static/src/js/DomainSelector.esm.js
chienandalu, line 1: We can include this patch as well so the domain also re-renders when the field widget is being used in a form field

```js
patch(FieldDomain.prototype, "web.basic_fields", {
    /**
     * Odoo restricts re-rendering the domain from the debug editor for supposedly
     * performance reasons. We didn't ever came up with those and in v17 it's supported
     * in the new advanced search.
     * @override
     */
    _onDomainSelectorValueChange(event) {
        this._super(...arguments);
        // Deactivate all debug conditions that cripple the functionality
        this.debugEdition = false;
    },
});
```

chienandalu (changes requested):
Hi, @coffeinisfun I made a little research myself and got to the conclussion that your solution is quite aimed. Some suggestions to improve it

coffeinisfun:
> Hi, @coffeinisfun I made a little research myself and got to the conclussion that your solution is quite aimed. Some suggestions to improve it

thanks for the suggestion! Seems to be a more elegant solution. I just adjusted it

web_advanced_search/static/src/js/DomainSelector.esm.js
chienandalu: Can you include this as well so it's usable in forms as well?

coffeinisfun: > Can you include this as well so it's usable in forms as well?

Done

chienandalu (approved):
Thanks!

CarlosRoca13 (approved):
Thx :)

pedrobaeza (approved):
/ocabot merge patch

chienandalu:
@coffeinisfun Will you forward it to v16?

PR #2633, opened 2023-10-03
[14.0][FIX] web_field_required_invisible_manager: restrict fields dependency

ilyasProgrammer:
- Add dependency to compute of restriction fields (fields from domain)
- Search only for manual state  fields
- Unlink created fields

TODO
When one model inherits another (e.g. res.users inherits res.partner) and new custom field is created, then same field is created for inheriting model as well, which should be avoided or at least duplicated field should be also manual, but created as base. I could not catch it in the create method.


francesco-ooops (approved):
Functional ok!

@ilyasProgrammer do you want to add this to Roadmap file?

> TODO
When one model inherits another (e.g. res.users inherits res.partner) and new custom field is created, then same field is created for inheriting model as well, which should be avoided or at least duplicated field should be also manual, but created as base. I could not catch it in the create method.

PR #2635, opened 2023-10-03
[16][FIX] web_remember_tree_column_width : Column resizing with list_button column

thibaultrey:
Fix an error when retrieving column sizes for lists containing one or more list_button columns.
This error occurs when a list_button column is placed between other columns in the list. This type of column is excluded from the querySelector string used to retrieve column headers, which can cause the index to shift and the column size to change when the list is displayed.

### DIFF
--- a/web_remember_tree_column_width/static/src/js/list_renderer.esm.js
+++ b/web_remember_tree_column_width/static/src/js/list_renderer.esm.js
@@ -11,10 +11,15 @@ patch(ListRenderer.prototype, "web_remember_tree_column_width.ListRenderer", {
     computeColumnWidthsFromContent() {
         const columnWidths = this._super.apply(this, arguments);
         const table = this.tableRef.el;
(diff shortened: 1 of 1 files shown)


luisg123v:
Please add commit description.

cuongnmtm (approved):

florian-dacosta:
Hi @luisg123v I've added the description of the PR to the commit, which seems clear enough to me now, is it ok?

For all : 
To reproduce the problem easily (before this fix) the best is to find a tree view, with a button which is not at the end (after all the fields).

For instance, if you install the module "stock", and you go to : Inventory => Operations => Inventory adjustments
There is a button named "reservations" in the middle. If you make the size of a field next to this button, you'll see easily the weird behavior.

florian-dacosta (approved):
Functional tests

luisg123v (approved):
/ocabot merge patch

PR #2638, opened 2023-10-05
[15.0][FIX] web_advanced_search: search more filters

chienandalu:
Fix advanced filters in the popup window

To investigate: try to do a `bus.off` over our fake many2one component for that custom event

cc @Tecnativa TT44025

### DIFF
--- a/web_advanced_search/static/src/js/owl/dropdown.esm.js
+++ b/web_advanced_search/static/src/js/owl/dropdown.esm.js
@@ -1,0 +1,20 @@
+/** @odoo-module **/
+
+import {Dropdown} from "@web/core/dropdown/dropdown";
+import {patch} from "web.utils";
+
+patch(Dropdown.prototype, "web.Dropdown", {
+    /**
+     * Our many2one widget in the filter menus has a dropdown that propagates some
+     * custom events through the bus to the search more pop-up. This is not replicable
+     * in core but we can simply cut it here
+     * @override
+     */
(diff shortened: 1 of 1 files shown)


pedrobaeza:
A side effect of this patch is that being on the popup, now the dropdowns of this popup are not close when you click on other part of the popup. Can you check if you can avoid it for that case?

chienandalu:
> A side effect of this patch is that being on the popup, now the dropdowns of this popup are not close when you click on other part of the popup. Can you check if you can avoid it for that case?

Well, it's not exactly a side effect of this change but of the fake widget used for the filter selector that stops the propagation of the click events when the pop-up window is opened (we couldn't see that effect because the dialog was closing itself before that)

I'm trying to find the right point to hook myself and be able to reset those events so everyone has the right ones but it's being tricky

pedrobaeza (approved):
As seen, there's no better solution, so it's good enough. Another thing I have found testing this:

Peek 2023-10-09 18-25

The many2one selection disappears when the the dropdown goes out. This doesn't happen if you input a text with "Contains" operator.

chienandalu:
> The many2one selection disappears when the the dropdown goes out. This doesn't happen if you input a text with "Contains" operator.

It's the same with a selection field. Odoo resets the state of the dropdown. This is an animation from a standard runbot:

Peek 10-10-2023 10-27

CarlosRoca13 (approved):

PR #2641, opened 2023-10-18
[15.0][FIX] web_widget_domain_editor_dialog: keep context

chienandalu:
The dialog needs to keep the context so translations and other things work as expected.

cc @Tecnativa TT45543

please review @pedrobaeza @CarlosRoca13

PR #2643, opened 2023-10-18
[13.0][FIX] web_widget_domain_editor_dialog: keep context

chienandalu:
fw from https://github.com/OCA/web/pull/2641
The dialog needs to keep the context so translations and other things work as expected.

cc @Tecnativa TT45543

please review @pedrobaeza @CarlosRoca13


pedrobaeza:
Fast-tracking as already merged in 15.0 and the patch is the same:

/ocabot merge patch

PR #2644, opened 2023-10-19
[14.0] web_edit_user_filter: Fix selection of facet

grindtildeath:
When clicking to edit a filter in search view, the _process_filters function will receive the faceID argument as a string, which will be compared to facet.groupId that is an Integer.

Therefore we should use == operator when comparing this value to allow type coercion from JS.


grindtildeath:
ping @robyf70 @simahawk

robyf70 (approved):
You are right! Sorry for my fault!!!
LGTM

PR #2645, opened 2023-10-23
[FIX][16.0] web_chatter_position: special case of sided chatter pos in mass mailing module

daiduongnguyen-odoo:
- Because the chatter in mass mailing form view lie under a notebook tab, if user choose sided chatter, it will disappear
Video before the fix:

https://github.com/OCA/web/assets/56789189/219c1bcb-c596-42f1-bfff-3a92317dd0f6

Video after the fix:

https://github.com/OCA/web/assets/56789189/6804c1b5-3f6b-413b-8567-457e2b3def0c


MihranThalhath (approved):
LGTM!

flotho (approved):
LGTM

PR #2648, opened 2023-10-28
[ADD]  web_widget_model_disable_nbrecords_computation

wouitmil:
This module allows disabling the nbrecords computation on widget domain fields

You'll then get this:
Screenshot 2023-10-28 at 13 16 11

Instead of this:
Screenshot 2023-10-28 at 13 15 05

Once installed, the domain widget won't compute the number of records by default for domain fields.

The default behaviour can be changed in settings
The default behaviour can be overridden on specific fields by adding **{'autoFetch':true|false}** on domain fields **attrs** attribute.

Screenshot 2023-10-28 at 13 13 52


web_widget_model_disable_nbrecords_computation/static/description/index.html
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE html
+    PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <title>Field omain widget disable nbRecords computation</title>
```
florentx: Typo. And Line 19 same typo
```suggestion
    <title>Field domain widget disable nbRecords computation</title>
```

web_widget_model_disable_nbrecords_computation/static/description/index.html
```diff
@@ -0,0 +1,23 @@
...
+        /*
+:Author: Hadrien Huvelle (hadrien.huvelle@camptocamp.com)
+
+*/
+    </style>
+</head>
+
+<body>
+
+    <h1 class="title">Field omain widget disable nbRecords computation</h1>
```
florentx:
```suggestion
    <h1 class="title">Field domain widget disable nbRecords computation</h1>
```

guewen:
Hi @wouitmil, looks like extra commits slipped in

guewen (approved):

pedrobaeza:
FYI, in 16.0, now this computation is optimized, showing 10.000+ if the number is higher for not having delays there. Maybe you can backport that feature for continuity (or migrate to 16 :wink:). Just a suggestion though. Nothing blocking, and I'm already in upper versions.

web_widget_model_disable_nbrecords_computation/static/src/xml/base.xml
```diff
@@ -0,0 +1,39 @@
...
+                    <i
+                        class="fa fa-arrow-right"
+                        role="img"
+                        aria-label="Domain"
+                        title="Domain"
+                    />
+                    <t t-esc="nbRecords" /> record[s] </button>
+                <span t-elif="isValid &amp;&amp; nbRecords==-1">
+                    Fetch nb records disabled
+                </span>
```
yvaucher: I see in the screen captures that the hyperlink is not available on this new text. Any reason for that? Doesn't it makes sense to let the user check the result of the domain even if we don't provide a counter?


Otherwise, I don't expect the user to understand what this message mean and domain editor might be available to anyone. I would go for a more simplistic message.

By simply removing the counter?

```suggestion
            <button t-elif="isValid &amp;&amp; nbRecords==-1" class="btn btn-sm btn-secondary o_domain_show_selection_button" type="button">
                    Show record(s)
            </button>
```

web_widget_model_disable_nbrecords_computation/static/src/js/basic_model.js
```diff
@@ -0,0 +1,57 @@
...
+                            ? true
+                            : rec[0].field_domain_widget_default_auto_fetch_nbrecords;
+                    if (defaultAutoFetch) {
+                        if (
+                            fieldInfo.options.autoFetch != undefined &&
+                            !fieldInfo.options.autoFetch
+                        ) {
+                            return Promise.resolve({
+                                model: domainModel,
+                                nbRecords: -1,
```
yvaucher, line 38: IIRC `search` with limit argument is only available since v15 or v16 so I guess we cannot make an efficient search using the ORM as it was done is newer Odoo version (as mentionned by @pedrobaeza )

yvaucher (commented):
Would it be possible to keep the button to open related records?

Otherwise LGTM

PR #2660, opened 2023-11-07
[17.0] mig web_theme_classic

legalsylvain:
Quite simple migration from 16.0 to 17.0
- add more specific css rule, using ``.o_form_view``
- remove css rules for search view, because there are now native in Odoo.


legalsylvain:
Hi @sbidoul. So I made my first PR on 17.0 branch.
- CI (pre-commit + test) looks OK :ok: 
- generation of pyproject.toml file :ok: 
- not generation of setup folder :ok: 
- runboat : not available  :-1: 

Note : 
@yajo, @pedrobaeza, @sbidoul 
as previously said [here](https://github.com/OCA/oca-addons-repo-template/pull/190), I really think I'll dislike the fact that each commit comes with a modification of two files (``readme.rst`` and ``static/description/index.html``). This is generating useless extra diff, and make review harder. see extra noice here : https://github.com/OCA/web/pull/2660/commits/296da50cbc03cbcc9483df42c1220d673dc76bf2

sbidoul:
@legalsylvain runboat is up, now.

francesco-ooops:
@sbidoul shouldn't readme files be converted in md for v17?

cc @TumbaoJu @lfreeke

sbidoul:
Good point. Do you think we should automate that in pre-commit?

Le mer. 8 nov. 2023, 11:54, francesco-ooops ***@***.***> a
écrit :

francesco-ooops:
@sbidoul yes, let's make the change happen :)

web_theme_classic/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,21 @@
+This module extend the Odoo Community Edition ``web`` module to improve visibility
+of form view.
+
+**Rational:**
+In Odoo V16, the design is very pure. That's great, but it generates some problem for
```
francesco-ooops: @legalsylvain v17 ?

pedrobaeza: I would put `Since Odoo 16...`, which serves for all the versions, but IMO, trying to keep the old behavior is anti-ergonomic and visually uncomfortable (it was like that on previous versions, but only on edit-mode, and now it's always). You should adapt to the new aspect with its advantages (which overpasses the cons). Here it's my non-asked opinion...

web_theme_classic/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,21 @@
...
+
+In a way, this module restores the form display of version 15, but preserving the "save on the fly" new feature.
+
+**Without this module**
+
+.. figure:: ../static/description/product_template_form_without_module.png
+
+**With this module**
+
+.. figure:: ../static/description/product_template_form_with_module.png
```
francesco-ooops: Screenshots should be updated. Also I think the button part has been addressed already: 

image

sbidoul:
The PR to convert fragments to markdown in pre-commit is https://github.com/OCA/oca-addons-repo-template/pull/228

web_theme_classic/readme/DESCRIPTION.rst
legalsylvain: >  trying to keep the old behavior is anti-ergonomic

Well, in "new odoo CE Design (16+)", users can not know where to click to change a non required field because the border is totally hidden. That's in my opinion, anti-ergonomic.

but hey, maybe I'm just reactionary, and I'll get used to it in the end !
Anyway, this PR is just a PoC to test the new V17 environment. (CI / pre-commit / runboat).
I'm not going to use this version.

Here it's my non-asked answer... ;-)

guestisp:
AWESOME !!! And working properly on v17.

Two small suggestions: lighter background (#f4f4f4 ?) for mandatory fields and a little bit of left/right padding on fields (ie 4px?)

From this:
image

To this:
image

legalsylvain:
> @legalsylvain runboat is up, now.

Thanks ! 

> The PR to convert fragments to markdown in pre-commit is https://github.com/OCA/oca-addons-repo-template/pull/228

thanks @sbidoul. works great.

hi @guestisp : 
Thanks for your review.

> Two small suggestions: lighter background (#f4f4f4 ?) for mandatory fields 

I'm not sure. In a hand, I like your proposal, but in the other hand it is a matter of taste, and users have been accustomed to this color for many years. What do you think ? (I'm OK with moving in your direction.)

> and a little bit of left/right padding on fields (ie 4px?)

Indeed. good proposal. 

could you make a PR against my branch for the second point, (and eventually the first one ?)

Thanks !

guestisp:
> I'm not sure. In a hand, I like your proposal, but in the other hand it is a matter of taste, and users have been accustomed to this color for many years. What do you think ? (I'm OK with moving in your direction.)

The rationale is simple: 9 times to 10, grey fields are disabled. At first sight when i've installed this addon, i've thought that there grey fields was the disabled one in the form. But they are editable, and looking at the css class, they are required filed. So, a very light grey is better than a dark grey. But even better would be to change the color at all (light yellow? light green? light-something-but-not-grey?)

It's just my two cent as a fresh Odoo user, nothing more. A simple suggestion.

> could you make a PR against my branch for the second point, (and eventually the first one ?)

Yup!

guestisp:
what about white fields but with purple (odoo purple) borders for the required fields ?

francesco-ooops:
@guestisp what about making it configurable?

legalsylvain:
Hi 
- regarding the required fields, it looks like there is no consensus, and make it configurable add a very big layer of complexity. So, I propose to keep it as it, as for the moment, the module copy the old behaviour of Odoo < 15.0.

- For the left/right padding on fields, I think it's consensual.

guestisp:
> @guestisp what about making it configurable?

that's not easy......

guestisp:
there is an issue with dark mode. The grey fields on dark mode are .... terrible:
image

before doing a PR, do you know if there is a css class added somewere that i can use to detect the dark mode ? Because that gray has to be changed, in dark mode.

(if I saw properly, there isn't any class to use as selector, the css file is replaced totally, so, how can I check for dark mode inside the scss file?)

legalsylvain:
> before doing a PR, do you know if there is a css class added somewere that i can use to detect the dark mode ? Because that gray has to be changed, in dark mode.

I guess you talk about the EE module. If yes, I don't have access to EE, so I can't help you. But you can take a look on the code, if you can read it.

regards.

guestisp:
not sure if dark mode is only available in Enterprise. i have to check.

and i have to find a way to set the background color variable only if not already defined, based on official docs [1], the scss for dark mode is loaded first, followed by everything else, so if the background was set in dark mode scss, following files should not redefine the var

[1] https://www.odoo.com/documentation/17.0/developer/reference/user_interface/scss_inheritance.html

web_theme_classic/readme/DESCRIPTION.rst
legalsylvain: Done, thanks !

legalsylvain:
> not sure if dark mode is only available in Enterprise. i have to check.

Dark mode is developped by : 
- Odoo, (EE proprietary code), in the module ``web_enterprise``, available on v16 and v17.
- OCA, (Open source code), in the module ``web_dark_mode``, available for V16 only (for the time being). (https://github.com/OCA/web/tree/16.0/web_dark_mode, developed by @fkantelberg)

I guess both implementation are totally different.

Note : I propose to merge this PR as it, as the changes proposed are not trivial. Further improvement are possible after.

guestisp:
> * For the left/right padding on fields, I think it's consensual.

https://github.com/grap/web/pull/2

web_theme_classic/static/description/index.html
```diff
@@ -0,0 +1,446 @@
+<?xml version="1.0" encoding="utf-8"?>
```
remi-filament:
```suggestion

```
This line should be removed otherwise you get a warning error when installing the module :
`WARNING odoo17 py.warnings: /home/user/dev/odoo17/ocb/odoo/addons/base/models/ir_module.py:178: DeprecationWarning: XML declarations in HTML module descriptions are deprecated since Odoo 17, portal_odoo_debranding can just have a UTF8 description with not need for a declaration.`

Note : this was fixed in maintainer-tools with https://github.com/OCA/maintainer-tools/pull/596 back in December 23 so it should not be present anymore when pre-commit reissues the index.html

remi-filament (changes requested):
Thanks @legalsylvain for porting this module.
1 minor comment related to XML tag not necessary anymore in index.html raising a WARNING during installation.
Apart from this, 👍

legalsylvain:
/ocabot rebase

@remi-filament : thanks for your patch. Is it now ok ? 

thanks !

celm1990 (approved):
LGTM tested functionally

PR #2662, opened 2023-11-08
[16.0][FIX] web_dark_mode: Add fields to field lists

fkantelberg:
**Exception:**
- Access denied error because a non-admin user doesn't have access to their own HR fields

**Reproduction:**
- Database with hr and web_dark_mode
- Create new regular user (non-admin) with an employee linked
- Log in as the user and go to your own profile (top right corner > Profile/Preferences)


legalsylvain (approved):
Thanks.

pedrobaeza (approved):
/ocabot merge patch

PR #2663, opened 2023-11-09
[16.0] [ADD] web_export_view_legacy

odooNextev:
This module restores the original features introduced by web_export_view module (https://github.com/OCA/web/tree/12.0/web_export_view).
From 13.0 version it was obsoleted by the Export All button in Odoo standard, but the new feature doesn't allow to select records to export 
in current view and to print m2m in one cell as it was in the OCA module.

Installing this module you will be able to select rows to export in the current view and have m2m values concatenated in one cell.

BEFORE:

image

AFTER:

image

BEFORE:

image

AFTER

image


odooNextev:
Hi @pedrobaeza,
do you think this module can be commonly useful?
Our customers wanted to restore the old features of web_export_view module that hadn't been fully absorbed by the standard "export all" function, in particular the ability to select only the records you want to export and not every.
The other functionality in which the new function is different from the old module is in m2m export and our customers prefer the old one that concatenated values in one cell.
I hope the screenshot in the description can be useful to understand better.

pedrobaeza:
Well, my path would be to:

- Fix in odoo core the m2m exportation.
- Educate users to pre-filter data in the list and the export.

But if you want to keep the module, then:

- No need to rename it. There's no collapse with the old name.
- Keep commit history.

odooNextev:
> * Fix in odoo core the m2m exportation.

Odoo support told me that m2m are presented separately for making easier analysis and review.

> * Educate users to pre-filter data in the list and the export.

You're right, but I even understand users tell me it's quicker to select rows.

> But if you want to keep the module, then:
> 
> * No need to rename it. There's no collapse with the old name.
> * Keep commit history.

Thank you @pedrobaeza, I'll follow this way.

PR #2666, opened 2023-11-13
Update translation files

Jordi-Buitrago:


pedrobaeza:
Please use Weblate for that:

https://odoo-community.org/resources/translate

PR #2366, opened 2022-12-20
[16.0] [IMP] web_refresher: Add component to pager to avoid instance for all views

CarlosRoca13:
Forward-port with adaptation of #2243 

cc @Tecnativa TT30229

please review this @chienandalu @pedrobaeza

PR #2365, opened 2022-12-20
[15.0][IMP] web_refresher: Work as component of control panel

CarlosRoca13:
Forward-port with adaptation of #2243 

cc @Tecnativa TT30229

please review this @pedrobaeza @sergio-teruel

PR #2243, opened 2022-06-30
[13.0][IMP] web_refresher: Work as element of control panel

Tardo:
Now the button is not hidden if the pagination is not visible.

cc @tecnativa TT30229

PR #2669, opened 2023-11-15
[17.0][MIG] web_refresher: Migration to 17.0

TesaLazarenko:


pedrobaeza:
Tested on runbot and working.

**question:** Is there a reason to put the button on the center instead of next to the pager as always?

LazarusUA:
image
Hi, there were a few reasons for that, in previous versions, page refresh was not available on all views where it was most often needed, in the current version I attach this in not only pager, which allows you to get this functionality almost everywhere.
In the image of odoo 15, as you can see, the update functionality is not available, so placing it in the search component solves this problem

LazarusUA:
> image Hi, there were a few reasons for that, in previous versions, page refresh was not available on all views where it was most often needed, in the current version I attach this in not only pager, which allows you to get this functionality almost everywhere. In the image of odoo 15, as you can see, the update functionality is not available, so placing it in the search component solves this problem

TesaLazarenko:
> Tested on runbot and working.
> 
> **question:** Is there a reason to put the button on the center instead of next to the pager as always?

image

Hi, there were a few reasons for that, in previous versions, page refresh was not available on all views where it was most often needed, in the current version I attach this in not only pager, which allows you to get this functionality almost everywhere. In the image of odoo 15, as you can see, the update functionality is not available, so placing it in the search component solves this problem

pedrobaeza:
But we worked on that on all the versions:

Original in 13.0: https://github.com/OCA/web/pull/2243
Forward-port for 15.0: https://github.com/OCA/web/pull/2365
Forward-port for 16.0: https://github.com/OCA/web/pull/2366

so if it's not working completely in 16.0, it should be fixed, but putting in the middle seems a bit weird.

TesaLazarenko:
> But we worked on that on all the versions:
> 
> Original in 13.0: #2243 Forward-port for 15.0: #2365 Forward-port for 16.0: #2366
> 
> so if it's not working completely in 16.0, it should be fixed, but putting in the middle seems a bit weird.

If someone else can confirm that this position is strange or inconvenient, I'll move this to the pager as it was before, but I'm not sure that this functionality will be available on reports

TesaLazarenko:
> > But we worked on that on all the versions:
> > Original in 13.0: #2243 Forward-port for 15.0: #2365 Forward-port for 16.0: #2366
> > so if it's not working completely in 16.0, it should be fixed, but putting in the middle seems a bit weird.
> 
> If someone else can confirm that this position is strange or inconvenient, I'll move this to the pager as it was before, but I'm not sure that this functionality will be available on reports

now I'm working on the preliminary look, it looks like everything could be fine

TesaLazarenko:
image

web_refresher/static/src/js/refresher.esm.js
```diff
@@ -1,16 +1,64 @@
...
+        if (searchModel) {
+            this._searchModelRefresh();
+        } else if (pagerProps) {
+            await this._pagerRefresh();
+        }
+    }
+
+    async onClickRefresh() {
+        await this.refresh();
+        this.notification.add(_t("Refreshed"), {
```
chienandalu: No very convinced with this :) Is it really needed?

TesaLazarenko: > No very convinced with this :) Is it really needed?

Are you talking about t**his.notification.add** or the entire component?

chienandalu: Yes, just the notification of the refresh :)

TesaLazarenko: this is not necessary, but I noticed in my experience that if you do not display anything, it is not immediately clear whether the click has worked, in a situation where there were no new changes to update

TesaLazarenko: If you think it's unnecessary or distracting, I can remove it

chienandalu: Maybe a shorter notification close timeout would be less anoying

TesaLazarenko: ok, I'll think about a simplified version of the notification, as a last resort I'll just remove it.
I'll deal with it later.

Any other comments on other points?

TesaLazarenko: ezgif com-video-to-gif
What about this animation instead of notification (the red ring is added by the screen recorder)

> Maybe a shorter notification close timeout would be less anoying

jcdrubay (approved):
Test on runboat, it works very nicely. Nice :)

pedrobaeza (approved):
/ocabot merge nobump

PR #2670, opened 2023-11-15
[16.0][MIG] web_calendar_color_field: Migration to 16.0

yankinmax:


web_calendar_color_field/readme/USAGE.rst
```diff
@@ -1,13 +1,12 @@
-Set a color attribute value on any many2one calendar field set as filter and
-set the calendar's color attribute to use the same many2one field.
+Set a color attribute value on any many2many calendar field set as filter and
```
ivantodorovich, line 1: It doesn't really make sense on `many2many` fields. If you have multiple records, from which one do you read the color?

We should stick to `many2one`, where there's only 1 record, and so we can read its related color field.

--


Having said that, I understand the motivation: the goal is to be able to use this feature on the `calendar.event.categ_ids` field.

Though, on this case, the solution would be to create a computed field (ie: `main_categ_id`) that refers to the first (maybe?) record in `categ_ids`. This will be your main category, and it will be on a many2one field that you can use to read the color from it.

All this, however, should not be on this module. This module is all about adding the feature to the `calendar` view, not implementing it for the `calendar` module. That's specific and belongs to another module (ie: `calendar_event_type_color` ? )

yankinmax:
Closed in favor for:
- [ ] https://github.com/OCA/web/pull/2687

PR #2673, opened 2023-11-17
[14.0][IMP] web_notify

DavidJForgeFlow:
When trying to use the notify function that comes from a normal user with sudo flag enabled, it wasn't letting the user access it. Adding the su flag to the check should allow sudo users using this feature.

@ForgeFlow


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2675, opened 2023-11-20
[15.0][IMP] web_notify

DavidJForgeFlow:
When trying to use the notify function that comes from a normal user with sudo flag enabled, it wasn't letting the user access it. Adding the su flag to the check should allow sudo users using this feature.

Forward port of #2673 

@ForgeFlow


chienandalu (approved):

leemannd (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2679, opened 2023-11-21
[15.0][FIX] web_pivot_computed_measure: Use js eval to avoid different types operation error

CarlosRoca13:
Before making these changes, when a calculated measure was formed by two different data types (int and bool), the system would throw an error because the expression was evaluated in Python. After making these changes, the expression is evaluated using JavaScript, so operating with different data types is allowed.

cc @Tecnativa TT46088

### DIFF
--- a/web_pivot_computed_measure/static/src/helpers/utils.esm.js
+++ b/web_pivot_computed_measure/static/src/helpers/utils.esm.js
@@ -1,0 +1,27 @@
+/** @odoo-module **/
+/* Copyright 2022 Tecnativa - Carlos Roca
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html) */
+
+/**
+ * Helper function to eval text for a given object
+ *
+ * @param {String} text
+ * @param {Object} vals
+ * @returns {any}
+ */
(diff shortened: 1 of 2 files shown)


CarlosRoca13:
Example before changes:
Error_computed

Example after changes:
solved_computed

sergio-teruel (approved):

web_pivot_computed_measure/static/src/helpers/utils.esm.js
```diff
@@ -0,0 +1,27 @@
...
+export const evalOperation = (text, vals) => {
+    for (const variable in vals) {
+        if (vals.hasOwnProperty(variable)) {
+            const regex = new RegExp(variable, "g");
+            text = text.replace(regex, vals[variable]);
+        }
+    }
+    try {
+        // eslint-disable-next-line no-eval
+        const res = eval(text);
```
chienandalu: Did you evaluate the security of this?

CarlosRoca13:
I’ve changed the code to avoid using eval. Please review it :smile:

CarlosRoca13:
ping @chienandalu

chienandalu (approved):

PR #2683, opened 2023-11-24
[FIX][16.0] web_responsive: broken delete message confirm dialog UI

ngochuy97hp:
The error will mostly occur with Chrome and Chromium core browsers. Firefox doesn't error because it doesn't support `:has` pseudo class :D

Before:
Screenshot from 2023-11-24 11-54-31

After:
Screenshot from 2023-11-24 11-53-54

### DIFF
--- a/web_responsive/static/src/components/attachment_viewer/attachment_viewer.scss
+++ b/web_responsive/static/src/components/attachment_viewer/attachment_viewer.scss
@@ -6,7 +6,7 @@
 .o_web_client .o_DialogManager_dialog {
     /* Show sided viewer on large screens */
     @media (min-width: 1533px) {
-        &:not(:has(.o_AttachmentDeleteConfirm)) {
+        &:has(.o_AttachmentViewer) {
             position: static;
         }
         .o_AttachmentViewer_main {


pedrobaeza:
I have tried on a standard runboat before this patch and it's working well. Isn't the problem maybe due to the backend theme you are using?

ngochuy97hp:
> I have tried on a standard runboat before this patch and it's working well. Isn't the problem maybe due to the backend theme you are using?

Hello. I just tested on this same runbot, the delete message dialog is pushed to the bottom if I set the position static again. Looking at the code, it's clear that there's a removal case missing for o_DeleteMessageConfirm. If just fix it normally, it will become `&:not(:has(.o_AttachmentDeleteConfirm, .o_DeleteMessageConfirm))` but that's not general. Have you tested with version 16?

pedrobaeza:
Yes, on runboat 16 with Firefox:

Peek 2023-11-24 08-44

ngochuy97hp:
Firefox doesn't seem to have this bug. I'm currently using Chrome, please try testing with its latest version.

https://github.com/OCA/web/assets/66666640/85e2245d-6e47-42f7-8d2f-bf5ee55713ad

ngochuy97hp:
Actually `:has` doesn't work on firefox, that's why you don't get the error

ap-wtioit:
> Actually `:has` doesn't work on firefox, that's why you don't get the error

One more month: https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idc-p/43733/highlight/true#M25293

ngochuy97hp:
> > Actually `:has` doesn't work on firefox, that's why you don't get the error
> 
> One more month: https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idc-p/43733/highlight/true#M25293

Yes 
Screenshot from 2023-11-24 15-57-57

pedrobaeza:
But your patch still includes `has`.

ngochuy97hp:
> But your patch still includes `has`.

Still have to use this, it's fine with most browsers except firefox (from version 121 it will be ok). The purpose of this PR is just that I want the static position to only apply when viewing the attachment dialog without affecting other things, not an improvement to not use `:has`.

pedrobaeza:
OK, I have checked in both versions in Firefox works OK, so no block from my part.

houssine78 (approved):

ngochuy97hp:
Happy New Year everyone, Firefox version 121 has been released, it already works with this PR

pedrobaeza (approved):
/ocabot merge patch

PR #2688, opened 2023-12-03
[16.0][FIX] web_responsive: ensure proper element width

sgheller:
**Before:**

image
image

**After:**

image


web_responsive/static/src/components/attachment_viewer/attachment_viewer.scss
```diff
@@ -37,7 +37,7 @@
                 display: none;
             }
             .o_AttachmentViewer_viewIframe {
-                width: 95%;
+                width: 95% !important;
```
chienandalu, line 40: This is because the bootstrap class has precedence. Maybe it'd be better to remove it from the template.

PR #2689, opened 2023-12-11
[IMP] web_field_required_invisible_manager: consider rel. fields with own views

ilyasProgrammer:


francesco-ooops (approved):
Functional ok!

web_field_required_invisible_manager/tests/test_web_field_required_invisible_manager.py
```diff
@@ -65,9 +69,92 @@ def setUpClass(cls):
...
+        )
+        with Form(self.partner_model_id, view=partner_form_view) as partner_form:
+            with partner_form.custom_required_restriction_ids.new() as new_line:
+                self.assertIsNotNone(new_line.required_model_id)
+                new_line.field_id = self.partner_name_field_id
+            with partner_form.custom_readonly_restriction_ids.new() as new_line:
+                self.assertIsNotNone(new_line.readonly_model_id)
+                new_line.field_id = self.partner_name_field_id
+            with partner_form.custom_invisible_restriction_ids.new() as new_line:
+                self.assertIsNotNone(new_line.invisible_model_id)
```
HekkiMelody: **todo**: I think these should be `self.assertTrue`, no? Because a many2one will never return `None` even if it is empty.

HekkiMelody (changes requested):
Code review, only a minor issue.

HekkiMelody (approved):
Code review, LGTM

PR #2691, opened 2023-12-12
[IMP] web_widget_numeric_step: display always on mobile

yajo:
Mobile screens benefit most from this module's enhanced usability.

However, since the events used for displaying/hiding the +/- buttons depend on mouse, at the end you'll never see these buttons on mobile screens.

Thus, here's a new approach: smaller screens *always* display the buttons.
imagen

@moduon MT-4396


Shide (approved):
Seems reasonable to me and the changes for this benefit can't be lower.

EmilioPascual (approved):
LGTM, Thanks @yajo

PR #2692, opened 2023-12-12
[FIX] web_widget_numeric_step: disable buttons on readonly fields

yajo:
Before this patch, if a field was readonly, the user could still use the +/- buttons to alter its value. Not good!

@moduon MT-4396


EmilioPascual (approved):
LGTM

Shide (approved):

PR #2405, opened 2023-02-06
[16.0][MIG] web_responsive

bvanjeelharia:

PR #2695, opened 2023-12-15
[16.0][FIX] web_responsive: Show Attachment Preview By Default In Form view

bvanjeelharia:
This commit would allow the attachment previews to load by default wherever applicable like for Invoices.
Check the following screenshot for invoices before this change:
Before Change
Check the following screenshot for invoices after this change is merged:
After Change

### DIFF
--- a/web_responsive/__init__.py
+++ b/web_responsive/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -19,10 +19,9 @@
     "development_status": "Production/Stable",
     "maintainers": ["Yajo", "Tardo", "SplashS"],
     "excludes": ["web_enterprise"],
-    "data": ["views/web.xml"],
+    "data": ["views/web.xml", "views/res_users.xml"],
     "assets": {
         "web.assets_backend": [
-            "/web_responsive/static/src/views/form/form_controller.esm.js",
             "/web_responsive/static/src/legacy/scss/web_responsive.scss",
             "/web_responsive/static/src/legacy/js/web_responsive.js",
             "/web_responsive/static/src/components/ui_context.esm.js",
@@ -44,6 +43,7 @@
             "/web_responsive/static/src/components/attachment_viewer/attachment_viewer.esm.js",
             "/web_responsive/static/src/components/attachment_viewer/attachment_viewer.xml",
             "/web_responsive/static/src/views/form/form_controller.scss",
+            "/web_responsive/static/src/views/form/form_controller.esm.js",
         ],
         "web.assets_tests": [
             "/web_responsive/static/tests/test_patch.js",
--- a/web_responsive/models/__init__.py
+++ b/web_responsive/models/__init__.py
@@ -1,0 +1,1 @@
+from . import res_users
--- a/web_responsive/models/res_users.py
+++ b/web_responsive/models/res_users.py
@@ -1,0 +1,25 @@
+# Copyright 2023 Onestein - Anjeel Haria
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    allow_attachment_preview = fields.Selection(
+        [
+            ("yes", "Yes"),
+            ("no", "No"),
+        ],
+        default="no",
+        help="Allows attachments to be previewed in form view by default",
+    )
+
+    @property
+    def SELF_READABLE_FIELDS(self):
+        return super().SELF_READABLE_FIELDS + ["allow_attachment_preview"]
+
+    @property
+    def SELF_WRITEABLE_FIELDS(self):
+        return super().SELF_WRITEABLE_FIELDS + ["allow_attachment_preview"]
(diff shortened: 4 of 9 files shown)


legalsylvain (changes requested):
hi @bvanjeelharia. 

Thanks a lot for your contribution. That is a good improvment to allow to have the preview attachment by default, because it saves click to users.

However, there are some problem in your implementation.

- blocking point : the place of the chatter changed. It is now allways on the bottom. (should not). Also,  if user click on the invoice, there is a double display.

image

_runboat / firefox 120.0.1 (64 bits)_

- Performance : This should be an option per user (opt out by default) : This feature is great, but make the UI slower, because there are extra read call to the server.

Kind regards.

pedrobaeza:
Exactly, I see this feature very annoying, and should be at least configurable, but I don't really see why having the preview always (although not needed), and having to scroll down for the chatter is an improvement...

legalsylvain:
> Exactly, I see this feature very annoying, and should be at least configurable, 

:+1: 

> why having the preview always (although not needed)

i see this feature very important for accountant people. In my company, 
- There a a lot of user that generate invoices. (no interesting for them)
- There are also accountants that are doing supplier invoicing control. For them, this feature saves 2 click.

pedrobaeza:
The problem is that when you access an invoice, you don't want to see the attachment most of the times. Only for the initial encoding is needed, or some later checks. But the rest of the time, this preview is useless. And this only works when then first attachment is the correct one. I have for example one ISP vendor that attaches a PDF with the call list, and as second attachment, the invoice, so in this case, it's worst the solution than the illness.

Doing 2 clicks is not so bad then. One thing we can do is to unfold attachment by default (when there are any), so the preview is only 1 click away. I think Odoo does this on v17.

legalsylvain:
> I think Odoo does this on v17.

Just tested on v17 : 
-> the "just one click is NOT implemented"
-> but the V17 works EXACTLY like the current PR ;-) 

When user click on an invoice and open a form view AND IF len(attachments) 
---> the attachments[0] is displayed automatically in the right.
---> the chatter is in the bottom



image

bvanjeelharia:
> > I think Odoo does this on v17.
> 
> Just tested on v17 : -> the "just one click is NOT implemented" -> but the V17 works EXACTLY like the current PR ;-)
> 
> When user click on an invoice and open a form view AND IF len(attachments) ---> the attachments[0] is displayed automatically in the right. ---> the chatter is in the bottom
> 
> image

Yes, that's the same in v16 as well(also the double display). That is why I had made the change likewise

bvanjeelharia:
> > > I think Odoo does this on v17.
> > 
> > 
> > Just tested on v17 : -> the "just one click is NOT implemented" -> but the V17 works EXACTLY like the current PR ;-)
> > When user click on an invoice and open a form view AND IF len(attachments) ---> the attachments[0] is displayed automatically in the right. ---> the chatter is in the bottom
> > image
> 
> Yes, that's the same in v16 as well(also the double display). That is why I had made the change likewise.
Also, if there are multiple attachments , it allows us to go through each of it by next/previous buttons

pedrobaeza:
Well, as said, this was intentionally put this way on `web_responsive` for not acting like that, as such behavior is worst than the current (and in previous version one). It was explained in https://github.com/OCA/web/pull/2405#issuecomment-1435661180, and discussed, being implemented as it is, so this should be put as an option, and disabled by default for stability reasons.

legalsylvain:
> It was explained in https://github.com/OCA/web/pull/2405#issuecomment-1435661180, and discussed, being implemented as it is

Thanks ! I didn't know all that analysis.

> so this should be put as an option, and disabled by default for stability reasons.

:+1:

bvanjeelharia:
So, the current PR is fine , but only should be put as an option , right? @pedrobaeza

pedrobaeza:
Check the issue commented by Sylvain, but if you put it as an option disabled by default, then go ahead. I suppose you put the option selectable per user, but it should also attend the core option for display it per model.

bvanjeelharia:
> but it should also attend the core option for display it per model.

Yes, this preview only comes up when the attachment_preview div is added like how chatter_div is added on the form view.

> I suppose you put the option selectable per user

Yes, which would be disabled by default

bvanjeelharia:
@legalsylvain and @pedrobaeza , can you please review it once now?

pedrobaeza (commented):
Tested and working. The preference field is not shown, and needed to add it manually to the UI. I'm seeing that `web_chatter_position` (I needed to uninstall it for avoiding interferences) has an option called "Sided", that seems to be thought for recovering this option. It's not working, but don't you think it's better to add this code in such module and have it all controlled the same way?

bvanjeelharia:
> The preference field is not shown, and needed to add it manually to the UI

The Preference field would show up in My Profile section exactly how the chatter position options show up.


> I'm seeing that `web_chatter_position` (I needed to uninstall it for avoiding interferences) has an option called "Sided", that seems to be thought for recovering this option. It's not working, but don't you think it's better to add this code in such module and have it all controlled the same way?

The web_chatter_position module may or may not be installed by everyone who use the web_responsive module so this change(code) has to be added in web_responsive module I believe.

And the sided option in web_chatter_position module doesn't work even if you uninstall the web_responsive module and check in invoices as this feature (Attachment Preview) is actually independent of the chatter position. So if needed, changes should also be added in the web_chatter_position module to allow/disallow Attachment Previews based on the chatter position or handle it any other way.

PaulGoubert (approved):
lgtm

PR #2696, opened 2023-12-18
[BUILD] web_responsive: resign as maintainer

yajo:


yajo:
fast-tracking

/ocabot merge nobump

pedrobaeza:
Just for reference, `[BUILD]` is not a valid tag: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

yajo:
Sorry, I got confused with conventional commits, which I'm more used to lately. In any case, I don't see any commit tag that fits this kind of change.

pedrobaeza:
It's true that this doesn't fit in any of the existing one, but `[BUILD]` doesn't seem also the best one. We can propose to add one. Suggestions? I think in `[CONF]`, `[MANT]` or `[OCA]`.

yajo:
In the world of conventional commits you'd use `docs`, `chore` or `build`. The standard is based on the one pioneered by AngularJS, and they removed `chore` in 2017 because it was often abused (https://github.com/conventional-commits/conventionalcommits.org/issues/303).

I think we can stick with `[BUILD]` (after all, changes on maintainership affect to the development cycle, which is generally a good fit for `[BUILD]`) or `[CHORE]`.

pedrobaeza:
For me that's counter-intuitive, but if others see it natural, no special problem for me.

simahawk:
FWIW to me is just an "update". Nothing more. Don't get too complicated w/ these tags guys! :smile:

pedrobaeza:
`[UPD]` is also a non valid tag according current guidelines.

PR #2700, opened 2023-12-22
[FIX] web_widget_numeric_step: no input focus on touch screens

yajo:
If you're using a tablet and click on the ➕ or ➖ buttons created by this module, the result was that the numeric input related to that button got focused.

On tablets, that means that the on-screen keyboard pops up. This usually triggers a layout recalculation and becomes clunky. Besides, it's useless, because if you wanted to use the keyboard, you'd have clicked on the input by yourself, and not in one of those buttons.

After this change, when using a touch screen, if you click on the +/- buttons, you won't auto-focus on the input. Thus, the keyboard won't show up.

@moduon MT-4472


legalsylvain (approved):

rafaelbn (approved):
👍🏼 great! Thank you!

PR #2701, opened 2023-12-22
[ADD] web_touchscreen: enable kanban sub-form views for tablets

yajo:
This module prefers kanban sub-form views when using a tablet, or any other device with a touch screen, no matter its size.

@moduon MT-4472


fcvalgar (approved):
Functional review, 

image


LGTM, thank you @yajo

rafaelbn (approved):
👍🏼 let's go!

arturoovf:
Hi, I don't get what this module is supposed to do? I tried with my iPad with this module installed and I didn't see any difference, can you provide an example where to appreciate the changes?  @yajo Thanks!

Shide:
> Hi, I don't get what this module is supposed to do? I tried with my iPad with this module installed and I didn't see any difference, can you provide an example where to appreciate the changes? @yajo Thanks!

According to the comment in .js file:

> Create a new env that extends the original one but overrides the way Odoo considers this device small: it will be small if it has touch capabilities, not only if the screen is small. 
> In practice, this will make the inline subforms prefer the kanban mode if possible.

yajo:
Before:
image


After:
image

yajo:
I improved the readme in https://github.com/OCA/web/pull/2716

arturoovf:
> Before: image
> 
> After: image

Thank you!

PR #2703, opened 2023-12-26
[ADD] New module web_merge_notebook_tab

legalsylvain:
### Description

This module extend the Odoo Community Edition ``web`` module to improve display of Form views that contains notebooks.

Odoo is very modular, that is great, but for some Forms, Odoo designs the view with a lot of tabs in a notebook,
one per functional topic.

If many modules are installed, it generates a lot of tabs. Consequently:

* it force users to click a lot of time in all the tabs, to configure all the data.
* it is not possible for a user to see quickly, if an item is correctly configured.

It is specially problematic for product or partner form views.

This module allows to 'merge' many tabs into a single tab, to avoid such UX troubles.

Demo exemple with the form view of the ``res.groups`` model:

**Without the module**

without_module

**With the module**

with_module

### Configure

* Go to "Settings > Technical > User Interface > Merge Notebook tab Settings"

* Create a new setting:
    * Select the view you want to change
    * Set the description of the tab, that will be displayed to the user. (``string`` parameter)
    * Set the name of the tab, that will be used technically. (``name`` parameter)
    * set the names of the tabs you want to merge, as a python list.

image


fah-mili (approved):
Tested and it works exactly as described. LGTM

quentinDupont (approved):
Tested on runboat with several views, work as described :ok_hand: 

Idea : add sequence parameter when we configure a merge notebook

legalsylvain:
> Idea : add sequence parameter when we configure a merge notebook

could you elaborate ?

quentinDupont:
> > Idea : add sequence parameter when we configure a merge notebook
> 
> could you elaborate ?

When I tested the module, I merge three notebooks and the notebook arrive on last position. I thought to myself: how cool would it be to be able to choose the position of this new tab? :) 
Butt in fact, the order of the tabs are not managed with sequences, so...

legalsylvain:
> I thought to myself: how cool would it be to be able to choose the position of this new tab? :)

I get it ! Thanks.
 Indeed. For the time being, it "append" the new tab in notebook, as the last element, using https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.append.

if we add some field in the setting, like ``tab_name_append_after = fied.char()`` , we could replace the call of append by a call of https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.insert to insert the new tab after the defined tab.

I'll give it a try in the next day. if it's not easy, i'll add a text in the roadmap section.

legalsylvain:
> I'll give it a try in the next day. if it's not easy, i'll add a text in the roadmap section.

Two years later ...

hi @quentinDupont : I implemented the tab_name_position field. If set, the new tab will be put after the named tab. Otherwise, it will be at the first place.

Could you take a look ? 

thanks !

quentinDupont:
> > I'll give it a try in the next day. if it's not easy, i'll add a text in the roadmap section.
> 
> Two years later ...
> 
> hi @quentinDupont : I implemented the tab_name_position field. If set, the new tab will be put after the named tab. Otherwise, it will be at the first place.
> 
> Could you take a look ?
> 
> thanks !

So good !
Awesome module. 
I tested, with success, on runboat the `tab name position` !

You can add this in README, also you can add that user can choose tab name thanks to `Tab Names Available` →very great idea this field

legalsylvain:
hi @quentinDupont ! thanks for your review ! 
description updated.
Also inserted a roadmap.rst part to mention that for the time being, groups management are not done and could be done in a second time.

/ocabot merge nobump

Issue #2394, opened 2023-01-26
Uninstallation of web_ir_actions_act_multi, v14 ce, break database


JakubSs:
module: web_ir_actions_act_multi
version: 14.0

**Steps to reproduce**
- uninstalling module

**Current behavior**
damages database, after several tries of uninstallation it removes several system columns from database and database is not longer capable of working. ie table ir_act_server and columns like name, binding_model_id, binding_type, binding_view_types... same behaviour seen in table ir_actions. 

**Expected behavior**
removing module without damaging database

Thanks

jason156:
I have the same problem.

redstarafrica:
I have the same issue

Has anyone found a solution to this issue?

baguenth:
This still happens for 16.0 :O 
Just destroyed an database by uninstalling this module.

carlos-lopez-tecnativa:
@pedrobaeza could you please close this issue? It has been resolved in  https://github.com/OCA/web/pull/3119

PR #2704, opened 2023-12-27
[17.0] [MIG] web_ir_actions_act_multi

augusto-weiss:


gaikaz (approved):

lbisiach:
@augusto-weiss are you close to merge this pr ?

augusto-weiss:
Hi @lbisiach, sorry i lost the notification... this pr must have review

web_ir_actions_act_multi/models/ir_actions.py
```diff
@@ -0,0 +1,15 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo import fields, models
+
+
+class IrActionsActMulti(models.Model):
+    _name = "ir.actions.act_multi"
+    _description = "Action Mulit"
+    _inherit = "ir.actions.actions"
+    _table = "ir_actions"
```
chienandalu, line 10: This is pretty dangerous without an uninstall hook as seen in https://github.com/OCA/web/issues/2394

augusto-weiss: wow... thats LGTM... i will implement that fix

augusto-weiss:
@chienandalu i update the table name, so now it's posible to uninstall the module without have a crashed db

web_ir_actions_act_multi/models/ir_actions.py
```diff
@@ -5,9 +5,9 @@
 
 class IrActionsActMulti(models.Model):
     _name = "ir.actions.act_multi"
-    _description = "Action Mulit"
+    _description = "Action Multi"
     _inherit = "ir.actions.actions"
-    _table = "ir_actions"
+    _table = "ir_actions_act_multi"
```
gaikaz: Yes, but this change very likely needs migration scripts, so that upgrading from Odoo 16, data is correctly separated and cleaned up from the `ir_actions` table and added to `ir_actions_act_multi` accordingly.

chienandalu (commented):
The goal of using the same table is to avoid populating too many relations for models that aren't used that much (if any)

To avoid deleting that system relation you could override the `_drop_table` method and to ignore it when the model table is that one of the system.

https://github.com/odoo/odoo/blob/88d0d1e15b7f3efeab39017512585248fd8c1d94/odoo/addons/base/models/ir_model.py#L293-L313

Another option is to make the model an abstract one, although that would avoid that you could declare those actions as data (which maybe isn't done anyway...)

augusto-weiss:
@chienandalu IMO make the model Abstract is a better option... i am not sure about a case where users want to declare data

chienandalu:
> IMO make the model Abstract is a better option... i am not sure about a case where users want to declare data

I think so, but in any case, I think a migration script should be made to avoid dropping the table when the ir.model gets unlinked. I think OU handles this with its non destructive policy but maybe in other contexts it isn't

augusto-weiss:
Hello everyone @chienandalu @gaikaz @lbisiach @pedrobaeza 
I'm working on migration script for the issue . I pushed a new PR with a commit which fix it.
If you think its ok, i can include it here.
https://github.com/OCA/web/pull/2997/commits/a01cfad69a25fc55115c0f013c12d0c785023e25
Let me know your opinions

carlos-lopez-tecnativa (commented):
@augusto-weiss @chienandalu could you please update the status of this PR

augusto-weiss:
Hi @carlos-lopez-tecnativa i'm waiting for some feedback.
Last changes could be:
https://github.com/OCA/web/commit/a01cfad69a25fc55115c0f013c12d0c785023e25

chienandalu (commented):
Hi @augusto-weiss sorry I had to grok the issue again :sweat_smile: 

Your option of using another table is safe enough, although at the cost of an extra table that won't be used. Not big deal anyway :)

Another option would be just using the model as abstract to avoid the uninstall problems, as that would ignore dropping the table: https://github.com/odoo/odoo/blob/88d0d1e15b7f3efeab39017512585248fd8c1d94/odoo/addons/base/models/ir_model.py#L297

And the third one to create an uninstall hook, which would probably kinda complex as we'd need to avoid dropping the table somehow...

So add your changes for the final review :)

carlos-lopez-tecnativa:
> Hi @augusto-weiss sorry I had to grok the issue again 😅
> 
> Your option of using another table is safe enough, although at the cost of an extra table that won't be used. Not big deal anyway :)
> 
> Another option would be just using the model as abstract to avoid the uninstall problems, as that would ignore dropping the table: https://github.com/odoo/odoo/blob/88d0d1e15b7f3efeab39017512585248fd8c1d94/odoo/addons/base/models/ir_model.py#L297
> 
> And the third one to create an uninstall hook, which would probably kinda complex as we'd need to avoid dropping the table somehow...
> 
> So add your changes for the final review :)

ping @augusto-weiss

augusto-weiss:
Sorry @chienandalu  i lost the notification! Thank @carlos-lopez-tecnativa for ping me

carlos-lopez-tecnativa:
@augusto-weiss please include https://github.com/OCA/web/pull/3119 and squash any additional administrative commits.

carlos-lopez-tecnativa:
> @augusto-weiss please include #3119 and squash any additional administrative commits.

ping @augusto-weiss

carlos-lopez-tecnativa:
> @augusto-weiss please include #3119 and squash any additional administrative commits.

ping @augusto-weiss

augusto-weiss:
Hi @carlos-lopez-tecnativa thanks about that. I included it.
cc @chienandalu @gaikaz

carlos-lopez-tecnativa (approved):
I expected that my commit would be included, keeping the original author—just do a `cherry-pick`. But anyway, LGTM.

CC @chienandalu @pedrobaeza, could you review this, please?

chienandalu (approved):

PR #2707, opened 2023-12-29
[14.0][ADD] New module web_custom_modifier

abenzbiria:


abenzbiria:
@TumbaoJu

abenzbiria:
@francesco-ooops @gva-acsone

TumbaoJu:
> @francesco-ooops @gva-acsone

This is the module we presented on the OCA Days 2023!!

abenzbiria:
Hi @legalsylvain as discussed in OCA days , there is module for customize view modifiers

web_custom_modifier/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# © 2023 Numigi (tm) and all its contributors (https://bit.ly/numigiens)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Web Custom Modifier",
+    "version": "14.0.2.0.1",
```
francesco-ooops, line 6: I think it should start from OCA version (not internal version)

Also in module description :)

web_custom_modifier/static/description/icon.png
francesco-ooops, line 1: I think it needs the OCA icon

francesco-ooops:
Hi @abenzbiria , first of all seems like there's a lot of interesting features here, so thanks for the PR! We've found similar needs in our day to day work. 👍

It seems to me at least a feature is duplicated with https://github.com/OCA/web/tree/14.0/web_field_required_invisible_manager , with the difference that this module does not allow to set a condition on _when_ to make a field required

Anyway this is not a very easy module to test functionally 😅 maybe it could be worth adding tests first?

ps: also change commit message as per guidelines :)

web_custom_modifier/__manifest__.py
TumbaoJu: @francesco-ooops Right, it's our first try to push a module to OCA so, thank you for the review ;)

abenzbiria:
Hi @francesco-ooops

thank you for your quick feedback.

 what is the coverage rate required by this repo. We are around 50%.

In the meantime we will correct the points you mentioned.

I wish you a happy New Year

Bests

web_custom_modifier/__manifest__.py
legalsylvain: > I think it should start from OCA version (not internal version)

No.

If the module exists before being shared into the OCA, revert to previous version could generate problem on odoo instances with the module installed.

Keep actual version is OK here.

mymage:
Hi everyone, I came here from Discord out for curiosity. I found the module useful and I just have a trivial question: the .pot file for translation must be included on the PR, so it's missing here, or it will be generated at merge time? Thanks

web_custom_modifier/__manifest__.py
francesco-ooops: @legalsylvain another IMP for https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id7 maybe? I think it could be worth opening a discussion on this.

FWIW, we use different names for internal modules (eg: ooops_web_custom_filter) and remove the ooops_ identifier at the beginning when publishing in OCA. Not only it makes it easier to distinguish them, but it's safer in case someone else on oca publishes a module with the same name.

anyway, @pedrobaeza what's your take on version?

pedrobaeza: You don't need to start with `1.0.0` when you "move" a module to OCA. Another different story is when you migrate an existing OCA module, where the version is reset.

legalsylvain:
> Hi everyone, I came here from Discord out for curiosity. I found the module useful and I just have a trivial question: the .pot file for translation must be included on the PR, so it's missing here, or it will be generated at merge time? Thanks

Hi. when you share a new module, you can add the .pot file if you want. (with .po files also). 
Once the PR is merged, .pot file will be created / updated.
the .po files will be then created / updated via weblate.

regards.

web_custom_modifier/__manifest__.py
legalsylvain: > FWIW, we use different names for internal modules (eg: ooops_web_custom_filter) and remove the ooops_ identifier at the beginning when publishing in OCA. Not only it makes it easier to distinguish them, but it's safer in case someone else on oca publishes a module with the same name.

The prefix technique is valid to distinguish OCA / non OCA modules and avoid conflicts, indeed. That's the advantage.
But the drawbacks are numerous. If you installed "oops_module_xxx" modules, you have to handle migration when you want to switch to the OCA "module_xxx". in some cases, it can be boring. and this may prevent people from sharing their module in the OCA, which is not desirable.

Most of the contributors have so a "incubator" repositories, without prefix : 
- https://github.com/grap/grap-odoo-incubator
- https://github.com/akretion/ak-odoo-incubator
- https://github.com/acsone/acsone-addons
- https://github.com/Tecnativa/extra-addons
- https://github.com/camptocamp/c2c-rd-addons

...


> @legalsylvain another IMP for https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id7 maybe? I think it could be worth opening a discussion on this.

Feel free to open an issue on that repo !

francesco-ooops: @legalsylvain thank you for sharing!

abenzbiria:
Hi @francesco-ooops  , @legalsylvain 

I added the OCA logo ,and adjust the  commit messages. 
Please share your comments with me  if  I need to fix other things .
The codecov still not OK , shoud I add other tests units ?

Thank you

francesco-ooops:
I think (unless there is a very specific reason) it's not good practice to add a module with a feature already present in another OCA module, so as per https://github.com/OCA/web/pull/2707#issuecomment-1872236212 would it be ok to remove the "set field required" part and use https://github.com/OCA/web/tree/14.0/web_field_required_invisible_manager instead?

abenzbiria:
Hi , 
The only reason is to manage all modifiers from the same menu.
The module was developed mainly for version 12.0 , that's why we added required/invisible and readonly modifiers.
I ll remove those modfiers from the module , test  it and let you know.

legalsylvain:
Hi @francesco-ooops, 

>  The only reason is to manage all modifiers from the same menu.

That's a good reason to keep the module as it, without removing features, IMO. If not it will make the module less interesting, in terms of user experience. At some point, various implementatison can live together in the OCA.

In these cases, I usually quote the "competing" modules in the readme. For exemple, the module mobile_app_inventory that was an alternative implementation of stock_scanner. 
see : https://github.com/OCA/stock-logistics-barcode/tree/8.0/mobile_app_inventory#similar-projects

francesco-ooops:
@legalsylvain fine by me, as long as 

- I can install both modules if I need features from both and they work correctly
- The same treatment is applied for new modules with replicated features which are more "convenient" to use in the new version

but would appreciate also feedback from others on this :)

francesco-ooops:
On the PR side: @abenzbiria I think when you have modules which change things on such a technical level, having features covered by tests is the best insurance both for functionals using the module per se (you can find unexpected bugs), both to ensure module features are maintained when new modules are added to same repo :)

polchampion:
@abenzbiria Is there a functional doc of the module? I tested a couple of options but some use cases are unclear to me.

bealdav:
Just tested

image

It works as expected.

Really thanks

bealdav (approved):

abenzbiria:
Hi @polchampion 

We added some use cases in the readme , you can also check our talk at OCA days  https://www.youtube.com/watch?v=RDEGpGGoIY0 (  4H:35 min)

@francesco-ooops  , @legalsylvain  we are now at 100% by file but the average still 92,97 % . Do you know why we have this  : https://app.codecov.io/gh/OCA/web/pull/2707


Thank you

web_custom_modifier/__manifest__.py
simahawk: To wrap it up: is perfectly fine. What is important is to mention this in the PR to not.... start such discussions :wink:

web_custom_modifier/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# © 2023 Numigi (tm) and all its contributors (https://bit.ly/numigiens)
```
simahawk: pls use the full word for "Copyright" rather than the symbol (as per template)

web_custom_modifier/models/common.py
```diff
@@ -0,0 +1,26 @@
+# © 2023 - today Numigi (tm) and all its contributors (https://bit.ly/numigiens)
```
simahawk, line 1: Could you name this file `utils.py` and move it to the root of the module?
Not a blocker.

web_custom_modifier/models/common.py
```diff
@@ -0,0 +1,26 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from typing import List, Mapping
+
+
+def set_custom_modifiers_on_fields(modifiers: List[dict], fields: Mapping[str, dict]):
```
simahawk, line 7: Question: why typing is used only in this case?

web_custom_modifier/models/ir_ui_view.py
```diff
@@ -0,0 +1,85 @@
...
+        modifiers = self.env["web.custom.modifier"].get(model_name)
+        node_with_custom_modifiers = _add_custom_modifiers_to_view_arch(modifiers, node)
+        set_custom_modifiers_on_fields(modifiers, name_manager.available_fields)
+        self.clear_caches()  # Clear the cache in order to recompute _get_active_rules
+        return super().postprocess(
+            node_with_custom_modifiers, current_node_path, editable, name_manager
+        )
+
+
+def _add_custom_modifiers_to_view_arch(modifiers, arch):
```
simahawk: these seem utils as well, no?

web_custom_modifier/models/ir_ui_view.py
```diff
@@ -0,0 +1,85 @@
...
+    """Add custom modifiers to the given view architecture."""
+    if not modifiers:
```
simahawk: no need to do this  check if modifiers is always a list (empty in case of no modifiers)

web_custom_modifier/models/ir_ui_view.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _add_custom_modifier_to_view_tree(modifier, tree):
+    """Add a custom modifier to the given view architecture."""
+    xpath_expr = (
+        "//field[@name='{field_name}'] | //modifier[@for='{field_name}']".format(
+            field_name=modifier["reference"]
+        )
+        if modifier["type_"] == "field"
+        else modifier["reference"]
+    )
```
simahawk:
```suggestion
    xpath_expr = modifier["reference"]
    if modifier["type_"] == "field" :
        xpath_expr = (
            "//field[@name='{field_name}'] | //modifier[@for='{field_name}']".format(
                field_name=modifier["reference"]
            ) 
        )
```
likely more readable?

web_custom_modifier/models/web_custom_modifier.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+
+class WebCustomModifierWithCachedModifiers(models.Model):
+    """Add a cache for getting the modifiers.
+
+    The system cache is invalidated when any modifier record is added / modified / deleted.
+    """
+
+    _inherit = "web.custom.modifier"
```
simahawk: is this inherit necessary?

web_custom_modifier/models/web_custom_modifier.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @api.model
+    def create(self, vals):
+        new_record = super().create(vals)
+        modules.registry.Registry(self.env.cr.dbname).clear_caches()
+        return new_record
+
+    def write(self, vals):
+        super().write(vals)
+        modules.registry.Registry(self.env.cr.dbname).clear_caches()
```
simahawk: Why invalidating ALL models caches?
Which cache do you want to wipe? The `_get_cache` one? If yes, just clear its cache w/ `self._get_cache.clear_cache()`.

web_custom_modifier/models/web_custom_modifier.py
```diff
@@ -0,0 +1,98 @@
...
+        return {
+            "models": self.mapped("model_ids.model"),
+            "key": self.key,
+            "type_": self.type_,
+            "modifier": self.modifier,
+            "reference": self.reference,
+            "excluded_group_ids": self.excluded_group_ids.ids,
+        }
+
+    def get(self, model):
```
simahawk, line 87: the cache should be by model IMO

web_custom_modifier/readme/USAGE.md
```diff
@@ -0,0 +1,91 @@
+## Access to the module
```
simahawk, line 1: nice docs, merci! :)

web_custom_modifier/tests/test_view_rendering.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def test_get_rendered_view_tree(self):
+        arch = self.env["res.partner"].fields_view_get(view_id=self.view.id)["arch"]
+        return etree.fromstring(arch)
+
+    def test_field_modifier(self, modifier="invisible"):
+        self.email_modifier.modifier = modifier
+        tree = self.test_get_rendered_view_tree()
+        el = tree.xpath("//field[@name='email']")[0]
+        assert _extract_modifier_value(el, modifier) is True
```
simahawk: any reason  not to use `self.assertTrue` ?

web_custom_modifier/tests/test_view_rendering.py
```diff
@@ -0,0 +1,168 @@
...
+            {
+                "model_ids": [(4, cls.env.ref("base.model_res_partner").id)],
+                "type_": "field",
+                "reference": "name",
+                "modifier": "optional",
+                "key": "show",
+            }
+        )
+
+    def test_get_rendered_view_tree(self):
```
simahawk:
```suggestion
    def _get_rendered_view_tree(self):
```
this is not a test

web_custom_modifier/tests/test_view_rendering.py
```diff
@@ -0,0 +1,168 @@
...
+        model_view = self.env.ref("base.view_model_form")
+        arch = self.env["ir.model"].fields_view_get(view_id=model_view.id)["fields"][
+            "field_id"
+        ]["views"]["tree"]["arch"]
+        tree = etree.fromstring(arch)
+        el = tree.xpath("//tree")[0]
+        assert el.attrib.get("limit") == "20"
+
+    def test_unlink(self):
+        self.assertTrue(self.email_modifier.unlink())
```
simahawk: this is not testing anything AFAIS

web_custom_modifier/views/web_custom_modifier.xml
```diff
@@ -0,0 +1,77 @@
...
+                        name="group_by_modifier"
+                        string="Modifier"
+                        context="{'group_by': 'modifier'}"
+                    />
+                </group>
+            </search>
+        </field>
+    </record>
+
+    <record id="custom_modifier_action" model="ir.actions.act_window">
```
simahawk, line 63: no specific form view?

simahawk (changes requested):

abenzbiria:
Hello @simahawk ,

Thanks for reviewing the code. We have corrected all your suggestions.

I hope the module meets OCA expectations.

I remain available if more information is needed.

@francesco-ooops   ,  @legalsylvain  , @TumbaoJu

legalsylvain (changes requested):
Hi. Could you remove .deb file ? Thanks !

abenzbiria:
> Hi. Could you remove .deb file ? Thanks !

Thanks @legalsylvain , it's done.

PaulGoubert (approved):
LGTM

legalsylvain (approved):
/ocabot merge nobump

PR #2713, opened 2024-01-08
[16.0][FIX] web_timeline: Wrap fields_get arg into a list

houzefa-abba:
By contract the first argument of the `fields_get` method is supposed to be a list. Before this fix, `web_timeline` would call `fields_get` with a string instead of a list.

Fortunately in case only 1 field is being grouped, this worked as Odoo does an `x in y` comparison [in its `fields_get` implementation](https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L2852), which does pass in a simple `"project_id" in "project_id"` case.

But that call remains invalid and can break when `fields_get` has been tweaked by other modules.

Here is Odoo code that does this right (with `[[` instead of just `[`): https://github.com/odoo/odoo/blob/16.0/addons/website/static/src/js/editor/snippets.options.js#L53

I checked this with [OCA's `project_timeline`](https://github.com/OCA/project/tree/16.0/project_timeline) module:
Screenshot at 2024-01-08 18-15-15

### DIFF
--- a/web_timeline/readme/CONTRIBUTORS.rst
+++ b/web_timeline/readme/CONTRIBUTORS.rst
@@ -10,6 +10,12 @@
   * Pedro M. Baeza
   * Alexandre Díaz
   * César A. Sánchez
+
 * `Onestein <https://www.onestein.nl>`_:
-    * Dennis Sluijk <d.sluijk@onestein.nl>
-    * Anjeel Haria
+
+  * Dennis Sluijk <d.sluijk@onestein.nl>
+  * Anjeel Haria
+
+* `XCG Consulting <https://xcg-consulting.fr>`_:
+
+  * Houzéfa Abbasbhay
--- a/web_timeline/static/src/js/timeline_renderer.js
+++ b/web_timeline/static/src/js/timeline_renderer.js
@@ -399,7 +399,7 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
                             await this._rpc({
                                 model: this.modelName,
                                 method: "fields_get",
-                                args: [grouped_field],
+                                args: [[grouped_field]],
                                 context: this.getSession().user_context,
                             }).then(async (fields) => {
                                 if (fields[grouped_field].type === "many2many") {


SirAionTech (approved):
Thanks for the PR!
I only reviewed the code and looks good to me.
Also big thanks for the detailed explanation :smile:

web_timeline/readme/CONTRIBUTORS.rst
```diff
@@ -13,3 +13,6 @@
 * `Onestein <https://www.onestein.nl>`_:
     * Dennis Sluijk <d.sluijk@onestein.nl>
     * Anjeel Haria
+* `XCG Consulting <https://xcg-consulting.fr>`_:
```
pedrobaeza, line 19: Incorrect RST format (extra empty line needed):

```suggestion

* `XCG Consulting <https://xcg-consulting.fr>`_:
```

houzefa-abba: [added](https://github.com/OCA/web/compare/0b200aae387ad4579fce253b6e54525ed56f5c10..a47659fda80aaa3e8f5137c5abe6c7b3105ba998)

I wasn't sure because there are 2 other companies in that list where the empty line is missing - we should add them too I guess

pedrobaeza: Tecnativa has the empty line. About Onestein, it's working somehow due to the extra 2 spaces in the following lines, but you can see in the GitHub rendering that it's shown weirdly. You can homogenize everything as it should be.

houzefa-abba: [done](https://github.com/OCA/web/compare/a82d7fb8b4317876867ff8fd91c1b422fa55766a..bc577b6656cfe747251b522d96aa027bb73dd4c7), should look better now

web_timeline/readme/CONTRIBUTORS.rst
```diff
@@ -10,6 +10,12 @@
   * Pedro M. Baeza
   * Alexandre Díaz
   * César A. Sánchez
+
```
pedrobaeza, line 13: Just for future references, this line is not needed below each of the company names. But it's OK this way as well.

pedrobaeza (approved):
/ocabot merge patch

PR #2718, opened 2024-01-11
[FIX] web_widget_numeric_step: restore tab navigation

yajo:
Before this patch, when navigating inputs hitting <kbd>TAB</kbd>, you were stopped before any numeric step input.

Now, you can navigate as usual and use Odoo without a 🐁.

@moduon MT-4657


Gelojr (approved):
LGTM Thanks @yajo https://www.loom.com/share/794e1f4a2cc2415b9a38ac7e861c02da?sid=d6ff6cb8-8bce-4fa4-9e35-895d3b5b291a

rafaelbn (approved):
👍🏼

PR #2722, opened 2024-01-16
[14.0][ADD] web addline large btn

KKamaa:
@ddejong-therp @legalsylvain @francesco-ooops a check on this


francesco-ooops:
@KKamaa from a functional standpoint, file "USAGE" could really need an example with steps for testing (eg: sale > click here > do this > ...) and a screenshot of result :)

legalsylvain:
+1 for a screenshot of the résult.

Is it possible to avoid extra configuration and add default value available for all the "add a line" Button in all odoo ?

KKamaa:
> @KKamaa from a functional standpoint, file "USAGE" could really need an example with steps for testing (eg: sale > click here > do this > ...) and a screenshot of result :)


Hey see sample I attached and also added a screenshot of the sample result on readme "USAGE". Also instructions on how to set it up.

https://github.com/OCA/web/assets/10249832/250d9270-571d-45f8-9786-a098df8fb817

KKamaa:
> +1 for a screenshot of the résult.
> 
> Is it possible to avoid extra configuration and add default value available for all the "add a line" Button in all odoo ?

- screenshot has been added

- I think its better we use configurations since if we do system wide, it might not be appreciated by all users. The idea is to make it flexible where required. e.g Those using odoo on mobile maybe in field activities. They may only need it on specific 0ne2many tree views. Also to allow users to add their own custom classes for buttons rather than force a particular one globally in case users have their own theme changes or wish to show different button color or designs.

francesco-ooops:
@KKamaa I'm talking from a non-technical perspective, the video you posted shows the end result but I don't know how to set it up for customer

> Install module and in the editable tree view in question add the
required options as show below

Can you refactor it in:

> go to... (eg: sale order)
click on...
add value...

PR #2729, opened 2024-01-23
[FIX] web_widget_numeric_step: list view column min width

yajo:
Default width is too narrow for these fields in list views, which makes it uncomfortable to use in desktops.

For example, a sale order after installing `sale_numeric_step`:

| Before | After |
| --- | --- |
| imagen | imagen |

@moduon MT-4472

### DIFF
--- a/web_widget_numeric_step/static/src/numeric_step.scss
+++ b/web_widget_numeric_step/static/src/numeric_step.scss
@@ -12,3 +12,11 @@
         visibility: visible;
     }
 }
+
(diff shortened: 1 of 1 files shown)


fcvalgar (approved):
Functional review.

LGTM, thanj you @yajo 

image

Gelojr (approved):
LGTM Thanks @yajo

Shide (approved):
Code lgtm

PR #2731, opened 2024-01-25
[16.0][MIG] web_widget_x2many_2d_matrix_example

norlinhenrik:
Migrated from 13.0


norlinhenrik:
I have added matrix example for **boolean** and **monetary**.

Monetary works in version 15.0 but not in version 16.0.

PR #2732, opened 2024-01-26
[14.0][ADD] web_field_empty_on_create

ilyasProgrammer:
Removes default values from selected fields when record created by user using GUI.


francesco-ooops (approved):
Functional ok!

Some required fields (eg: "Product Category" on product template) are automatically
populated on record creation. This module allows to force user to select a specific
record when creating a new record (in this case, a category when creating a new product)
in order to reduce margin of error.

PR #2734, opened 2024-01-31
[16.0][FIX] web_responsive: fix z-index on toolbar to prevent being on background in modals

olivierb2:
Prevent the tools bar which is appearing when selecting text to be behind the modal (e.g when sending quote by mail).


ivs-cetmix (approved):
Functional review LGTM

pedrobaeza:
@bvanjeelharia do you know why you put the `!important` there?

pedrobaeza:
As no answer, let's fix it and see if other problem arises:

/ocabot merge patch

pedrobaeza:
Please forward-port it to v17, where the problem seems to be also.

PR #2739, opened 2024-02-06
[14.0][IMP] web_field_required_invisible_manager: compatibility with tier v…

ilyasProgrammer:
…alidation

moved code from base model to ir.ui.view + minor refactoring.

### DIFF
--- a/web_field_required_invisible_manager/models/__init__.py
+++ b/web_field_required_invisible_manager/models/__init__.py
@@ -1,2 +1,4 @@
 from . import custom_field_restriction
-from . import models
+from . import base
+from . import ir_ui_view
+from . import ir_model
--- a/web_field_required_invisible_manager/models/base.py
+++ b/web_field_required_invisible_manager/models/base.py
@@ -1,0 +1,66 @@
+# Copyright 2023 ooops404
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+from odoo import models
+from odoo.tools.safe_eval import safe_eval
+
+
+class Base(models.AbstractModel):
+    _inherit = "base"
+
+    def default_get(self, fields_list):
+        res = super(Base, self).default_get(fields_list)
+        if self.env.user.has_group("base.group_user"):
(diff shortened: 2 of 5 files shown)


francesco-ooops (approved):
Functional ok!

web_field_required_invisible_manager/models/models.py
```diff
@@ -27,34 +27,32 @@ class IrModel(models.Model):
...
-            return arch
-        # TODO speed up somehow
+    def postprocess_and_fields(self, node, model=None, validate=False):
+        arch, new_fields = super().postprocess_and_fields(node, model, validate)
+        if self.type not in ["form", "tree"] and node.tag not in ["form", "tree"]:
+            return arch, new_fields
         restrictions = self.env["custom.field.restriction"].search(
             [
-                ("model_name", "=", self._name),
+                ("model_name", "=", self.model or model),
```
HekkiMelody: **issue**: from base Odoo, it looks like `model` should have priority over `self.model`

https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_ui_view.py#L857

web_field_required_invisible_manager/models/models.py
```diff
@@ -27,34 +27,32 @@ class IrModel(models.Model):
...
                 ("group_ids", "in", self.env.user.groups_id.ids),
             ]
         )
+        view_type = self.type or node.tag
```
HekkiMelody: **question**: I feel like here, too, `node.tag` should have priority over `self.type` (e.g.: a tree view embedded in a form view) but I'm not sure. What do you think?

HekkiMelody (changes requested):
Mostly ok, one comment and one question I'm not sure about.

Since you're already doing some cleanup and refactoring, could you please put each model in its own file, named after the model itself?

Thank you!

web_field_required_invisible_manager/models/models.py
ilyasProgrammer: i see no difference. But ok.

HekkiMelody (approved):

PR #2746, opened 2024-02-13
[MIG] web_edit_user_filter: Migration to 15.0

amh-mw:


amh-mw:
~~Unable to reproduce pre-commit failure locally on macOS Sonoma 14.5 with either pre-commit 2.20.0 or 3.7.1~~
I did a `pre-comment uninstall` then `pre-commit install` and that allowed me to reproduce.

PR #2750, opened 2024-02-16
[11.0][UPD] dotfiles

thomaspaulb:
Fix CI


thomaspaulb:
@OCA/web-maintainers Could you review?

requirements.txt
```diff
@@ -1,2 +1,2 @@
-# web_widget_bokeh_chart
-bokeh==0.12.7
+# generated from manifests external_dependencies
+bokeh
```
StefanRijnhart: Should the version be moved to https://github.com/OCA/web/blob/11.0/web_widget_bokeh_chart/__manifest__.py#L17?

thomaspaulb: Since this is 11.0, i think it needs to be an `external_dependencies_override`, I've pushed it, lets see

legalsylvain (approved):
Thanks for removing travis !

requirements.txt
StefanRijnhart: Great, thanks!

StefanRijnhart (approved):

PR #2744, opened 2024-02-09
[17.0][MIG] web_widget_x2many_2d_matrix

norlinhenrik:
The 16.0 fix for "editable boolean" is causing this in 17.0:
image
`Error: Patch "web_widget_x2many_2d_matrix": Second argument is not the patch name anymore, it should be the object containing the patched properties`

PR #2756, opened 2024-02-29
[17.0][MIG] web_widget_x2many_2d_matrix: Migration to 17.0

SodexisTeam:


web_widget_x2many_2d_matrix/static/src/views/fields/boolean/boolean_field.esm.js
Rad0van, line 1: This whole file is either unnecessary  or completely wrong. In this state it causes JS exception in Enterprise hr_appraisal which also does some JS magic with boolean_toggle The other thing is there is no isReadonly() method in 17.0 codebase so there is also no _super() to that which causes the exception...

Rad0van:
This does not work at all - it crashes upon loading. Have you tested it at all?

Rad0van:
Much better now. Actually we have attempted migration of this as well and came to very similar result. Few remarks on your version:

- The totals only get recomputed on first cell change. It does the same in our version. Don't know why.
- The center alignment of cells in matrix is wrong in my opinion. It's numbers there, they should have been aligned to right.
- The sums are always displayed with 2 decimal figures. This is wrong. In original implementation it used the format of the row/column being summed. So in case of integers, no decimal figures. In case of floats, as many decimal figures as the floats display.
- You have completely removed that boolean field stuff - even from the read-only override check. I was referring to the JS field part causing exceptions but not sure if the read-only oveeride is still needed.

norlinhenrik (approved):
Boolean works. Tested with #2782.

Rad0van:
> Boolean works. Tested with #2782.

Do all others I mentioned work for you? Because they still do not for me... Please see my comments above.

norlinhenrik:
Here is my testing:

- Totals get recomputed correctly.
- Cell alignment is to the right.
- Float sums: 2 decimals. Integer sums: 0 decimals.
- Boolean field stuff: not sure what to test.

@Rad0van Feel free to clone the PR and make a new commit with your desired changes. Then I guess @SodexisTeam can add it to this PR.

ioans73:
Which of the two PRs should we use? 🤔  This or https://github.com/OCA/web/pull/2744

@SodexisTeam @norlinhenrik

Rad0van:
> Here is my testing:
> 
> * Totals get recomputed correctly.
> * Cell alignment is to the right.
> * Float sums: 2 decimals. Integer sums: 0 decimals.
> * Boolean field stuff: not sure what to test.
> 
> @Rad0van Feel free to clone the PR and make a new commit with your desired changes. Then I guess @SodexisTeam can add it to this PR.

Totals get recomputed - nice! Alignment is almost OK - the totals are getting away slightly - see attached picture. Haven't tested integers and booleans yet.

image

norlinhenrik:
> Which of the two PRs should we use? 🤔 This or #2744
> 
> @SodexisTeam @norlinhenrik

We use this PR. I just closed #2744.

atchuthan (approved):

web_widget_x2many_2d_matrix/__manifest__.py
```diff
@@ -14,6 +14,7 @@
         "Camptocamp, "
         "CorporateHub, "
         "Onestein, "
+        "SodexisTeam, "
```
pedrobaeza: I don't think the amount of work for this migration qualifies for upgrading to co-authorship. Instead, you should put yourself in contributors.

pedrobaeza (changes requested):

web_widget_x2many_2d_matrix/__manifest__.py
SodexisTeam: @pedrobaeza Done the changes. Please review it.

pedrobaeza (approved):
/ocabot merge nobump

PR #2765, opened 2024-03-11
[17.0][MIG] web_favicon: Migration to 17.0

dc-oerp:
Odoo removed the favicon function and field in the company setting in V17.
Migrate from V12 to V17
This module is to add the favicon function back, and this can be helpful for multi-company/instance users.


pedrobaeza:
Thanks for the contribution. Indeed this one is needed since the removal of the feature in https://github.com/odoo/odoo/pull/116367

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

PR #2774, opened 2024-03-21
[16.0][IMP] web_timeline: Handle inverted date ranges

houzefa-abba:
Inverted dates make no functional sense, but when they do happen, this change allows displaying them as single points in the timeline (same as when begin=end).

We currently do have such demo data when installing project_timeline:
Screenshot at 2024-03-21 12-24-31

# Before: office design tasks are missing
Screenshot at 2024-03-21 11-53-24

# After: office design tasks are shown
Screenshot at 2024-03-21 11-50-51

### DIFF
--- a/web_timeline/static/src/js/timeline_renderer.js
+++ b/web_timeline/static/src/js/timeline_renderer.js
@@ -531,9 +531,9 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
                 evt: evt,
                 style: `background-color: ${this.color};`,
             };
-            // Check if the event is instantaneous,
-            // if so, display it with a point on the timeline (no 'end')
-            if (date_stop && !moment(date_start).isSame(date_stop)) {
+            // Only specify range end when there actually is one.
+            // ➔ Instantaneous events / those with inverted dates are displayed as points.
+            if (date_stop && moment(date_start).isBefore(date_stop)) {
                 r.end = date_stop;
             }
             this.color = null;


pedrobaeza:
I don't get the concept of "inverted date".

houzefa-abba:
> I don't get the concept of "inverted date".

Sorry I wasn't clear enough; this is when end date is **before** start date (in my example, March 21st ➔ March 19th).

Makes no sense, but at least with this change they won't be completely hidden.

pedrobaeza:
I think that should be forbidden by a constraint (SQL or Python).

pedrobaeza (approved):
But the constraint in `project_timeline` should be added IMO.

houzefa-abba:
> But the constraint in `project_timeline` should be added IMO.

Yes constraint and demo data, because demo data produces these "reversed dates"

I'll look into it soon

houzefa-abba:
> > But the constraint in `project_timeline` should be added IMO.
> 
> Yes constraint and demo data, because demo data produces these "reversed dates"
> 
> I'll look into it soon

Done with https://github.com/OCA/project/pull/1259 & https://github.com/OCA/project/pull/1260

carlos-lopez-tecnativa:
@houzefa-abba @pedrobaeza Is this PR still needed? The other PRs related to the `project` have already been merged.

TT50999

houzefa-abba:
> @houzefa-abba @pedrobaeza Is this PR still needed? The other PRs related to the `project` have already been merged.
> 
> TT50999

Hi @carlos-lopez-tecnativa this PR may still help the odd case where someone has data with start-date after end-date

We prevent it in the project part thanks to https://github.com/OCA/project/pull/1260 but web_timeline may be applied in other places and who knows, it may help someone? :)

carlos-lopez-tecnativa (approved):
I agree with @pedrobaeza and believe that it would be more effective to implement constraints that ensure data integrity at the source, preventing such scenarios from occurring in the first place. This is outside the scope of this module and must be controlled on each side. On the other hand, for fixed visual issues, it looks good to me. Thanks @houzefa-abba .

tarteo (commented):
Yes this makes sense, thanks. The constaint should be indeed done on model level, this way it shows events where end is before start date.

PR #2775, opened 2024-03-21
[16.0][IMP] web_timeline: Overflow text, add hover colors

houzefa-abba:
When an item label does not fit in its date-range box, overflow according to
https://visjs.github.io/vis-timeline/examples/timeline/items/rangeOverflowItem.html

Previous CSS code was already trying to do that, but was selecting `.vis-item.vis-item-content` instead of `.vis-item .vis-item-content`.

Displaying overflow text brings up layout issues solved by removing the forced-100% width instruction.

This change also adds highlight when hovering a box, which is useful on text that has overflown (as it has no borders).

In screenshots below I added an end date close to the start date on the first 2 tasks.
They almost have the same dates, to make sure the vis lib handles them correctly (they don't step on each other).

# Before: labels cut
Screenshot at 2024-03-21 11-47-11

# After: labels overflow
Screenshot at 2024-03-21 11-35-10

## With hover highlight
Screenshot at 2024-03-21 11-35-30

### DIFF
--- a/web_timeline/readme/ROADMAP.rst
+++ b/web_timeline/readme/ROADMAP.rst
@@ -1,5 +1,5 @@
 * Implement a more efficient way of refreshing timeline after a record update;
-* Make `attrs` attribute work;
+* Make ``attrs`` attribute work;
 * Make action attributes work (create, edit, delete) like in form and tree views.
 * When grouping by m2m and more than one record is set, the timeline item appears only
   on one group. Allow showing in both groups.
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

flotho (commented):
Is there any way to test this on runboat ?

pedrobaeza:
Well, if rebased after #2776, on the cron view

houzefa-abba:
Hi rebase is done

The demo cron view uses single points, so it allows testing hover highlight color but not text overflow:
Screenshot at 2024-04-04 09-50-57

I then edited the "ir.cron timeline" view on runboat to change:
```xml
date_start="nextcall"
```
to:
```xml
date_start="create_date" date_stop="nextcall"
```

Now items are displayed with range boxes instead of single points, and we can see text overflow works:
Screenshot at 2024-04-04 09-49-55

Before this PR, the "Mail: Fetchmail Service" text for example would be completely invisible as its range box is so small

flotho (approved):
functionnal test on ruboat
LGTM

PR #2777, opened 2024-03-21
17.0 mig web sheet full width

Wodran14:


CRogos:
Is this module still needed? I think this is also part of odoo standard now?

PR #2778, opened 2024-03-21
[16.0][FIX] web_timeline: Redraw issues with initial mode

houzefa-abba:
This commit fixes redraw issues when setting a `mode` attribute in the `timeline` view tag.

This mode specifies a default scale one would want to set; same as when clicking on Day/Week/Month buttons at the top of the view.

Initial rendering had issues here because data was loaded too soon, before the timeline component was rendered/ready. The fix is to load data into the component only after initial redraw event, called `changed` (see <https://visjs.github.io/vis-timeline/docs/timeline/#Events>).

There was old code attempting to call `on_scale_xxx_clicked` methods at load time to simulate clicks on these Day/Week/Month buttons, but these methods have been renamed so this code is no longer working.

This commit also removes the `current_window` instance variable, not needed and actually confusing as the timeline component already maintains its own start/end information (which we can query with `timeline.getWindow()`).

---

Screenshots below with project_timeline task view with `mode="month"`

# Before
Only 1 line at the bottom, rest is missing
Screenshot at 2024-03-21 17-36-09

# After
Screenshot at 2024-03-21 17-36-42

### DIFF
--- a/web_timeline/static/src/js/timeline_renderer.js
+++ b/web_timeline/static/src/js/timeline_renderer.js
@@ -41,6 +41,7 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
             this.fields = params.fields;
 
             this.timeline = false;
+            this.initial_data_loaded = false;
         },
 
         /**
@@ -48,11 +49,6 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
          */
         start: function () {
             const attrs = this.arch.attrs;
-            this.current_window = {
-                start: new moment(),
-                end: new moment().add(24, "hours"),
-            };
-
             this.$el.addClass(attrs.class);
             this.$timeline = this.$(".oe_timeline_widget");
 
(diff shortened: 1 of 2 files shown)


carlos-lopez-tecnativa:
@houzefa-abba, thanks for the PR. I tried to test it (on another runbot without these changes), but I can't reproduce the issue. I added `mode="month"` as you said, but the view is rendered correctly. Could you provide the steps to reproduce the issue?

http://oca-project-16-0-8ed45f433d53.runboat.odoo-community.org/

image


TT50999

carlos-lopez-tecnativa (approved):
Update: Locally, when I group by any field and set `mode="month"` in the view definition, the issue occurs. With this PR, the issue is resolved. However, on other runbots, it’s working fine. I think network latency has something to do with it, and it works fine.

Before
image
After
image

Anyway, the PR resolved the issue. Ping @pedrobaeza.

pedrobaeza (approved):
/ocabot merge patch

houzefa-abba:
> Update: Locally, when I group by any field and set `mode="month"` in the view definition, the issue occurs. With this PR, the issue is resolved. However, on other runbots, it’s working fine. I think network latency has something to do with it, and it works fine.
> 
> Before image After image
> 
> Anyway, the PR resolved the issue. Ping @pedrobaeza.

Hmm interesting, thanks for this thorough test

When I did this PR I always had this issue, both locally and in a deployed env

Yes network latency seems like a good cause for these divergences

PR #2788, opened 2024-04-01
[16.0][FIX] web_tree_duplicate: adapt readme to behavior

hbrunn:
that was a leftover from the version where it was the other way around


StefanRijnhart (approved):

flotho (approved):
make sense
LGTM

PR #2791, opened 2024-04-04
[16.0][IMP] web_timeline: Follow create/edit/delete attrs

houzefa-abba:
In addition to security rights (was already implemented), now follow `create="0"` / `edit="0"` / `delete="0"` attributes one can set onto the `timeline` tag, same as in other Odoo views.

I had to add a specific event for read-only double-clicks to distinguish with update ones (read-only double-clicks have item ID only, updates have more item info in a dict).

To test on runboat, open the ir.cron timeline view and play with:
* `create="0"` / 1
* `edit="0"` / 1
* `delete="0"` / 1
* `event_open_popup="0"` / 1

### DIFF
--- a/web_timeline/readme/ROADMAP.rst
+++ b/web_timeline/readme/ROADMAP.rst
@@ -1,6 +1,5 @@
 * Implement a more efficient way of refreshing timeline after a record update;
 * Make ``attrs`` attribute work;
-* Make action attributes work (create, edit, delete) like in form and tree views.
 * When grouping by m2m and more than one record is set, the timeline item appears only
   on one group. Allow showing in both groups.
 * When grouping by m2m and dragging for changing the time or the group, the changes on
--- a/web_timeline/static/src/js/timeline_controller.esm.js
+++ b/web_timeline/static/src/js/timeline_controller.esm.js
@@ -12,6 +12,7 @@ import {Component} from "@odoo/owl";
 export default AbstractController.extend({
     custom_events: _.extend({}, AbstractController.prototype.custom_events, {
         onGroupClick: "_onGroupClick",
+        onItemDoubleClick: "_onItemDoubleClick",
         onUpdate: "_onUpdate",
         onRemove: "_onRemove",
         onMove: "_onMove",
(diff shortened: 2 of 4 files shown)


OpenCode (approved):

carlos-lopez-tecnativa (commented):
I tested it functionally, and it works, but I have some doubts.

If I add edit="0", I can't edit, which is okay, but I also can't create (double-click). Is this intentional or an issue?

TT50999

houzefa-abba:
> I tested it functionally, and it works, but I have some doubts.
> 
> If I add edit="0", I can't edit, which is okay, but I also can't create (double-click). Is this intentional or an issue?
> 
> TT50999

Thanks for testing :)

That point was clearly intentional, but I can't recall why - maybe it wasn't working in my first iterations.

Functionally I compared with form & list views: it works as you say, the "New" button is available even without edit modes.

~~I just rebased + moved that `if this.can_create` outside of the `if this.can_update` block, and it's working fine.~~
edit: see my next message

houzefa-abba:
> > I tested it functionally, and it works, but I have some doubts.
> > If I add edit="0", I can't edit, which is okay, but I also can't create (double-click). Is this intentional or an issue?
> > TT50999
> 
> Thanks for testing :)
> 
> That point was clearly intentional, but I can't recall why - maybe it wasn't working in my first iterations.
> 
> Functionally I compared with form & list views: it works as you say, the "New" button is available even without edit modes.
> 
> ~I just rebased + moved that `if this.can_create` outside of the `if this.can_update` block, and it's working fine.~ edit: see my next message

I know what the issue with create=1 + edit=0 was: with `event_open_popup="0"` (default), it's working fine, but not with `event_open_popup="1"`. In that configuration, double-clicking an existing item opens a creation popup instead of readonly one.

I didn't push that case further as I doubt create=1 + edit=0 makes any sense functionally.
I suggest we leave it this way and work on it if someone really has a use case.

➔ I have moved the `if this.can_create` block back inside the `if this.can_update` one.

carlos-lopez-tecnativa (approved):
OK, LGTM, ping @pedrobaeza

PR #2792, opened 2024-04-05
[16.0][FIX] web_responsive: fix status widget visibility

chienandalu:
Before this change the status widget was hidden as it had no width. You can check it installing the project module and going to a task. Now you can see the status widget along the task name.

image

please review @pilarvargas-tecnativa @carlosdauden


pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2795, opened 2024-04-14
[17.0][FIX] web_responsive: disable default file preview

### LINKED ISSUE #2745: [17.0] web_responsive: Main document preview is activated by default
Hi @LazarusUA, it seems the auto-preview of the main attachment when entering a document like vendor bill is not deactivated, like it was in v16, and making other features like the possibility of previewing other attachments or seeing the chatter in the right are killed.

Can you please check to include that inhibition?

TesaLazarenko:
https://github.com/OCA/web/issues/2745

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -48,9 +48,7 @@
             "/web_responsive/static/src/components/chatter/*",
             "/web_responsive/static/src/components/control_panel/*",
             "/web_responsive/static/src/components/command_palette/*",
-            "/web_responsive/static/src/views/form/form_controller.scss",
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):
Working as expected, thanks!

/ocabot merge patch

cvinh:
Hello
We have a accountant customer who liked the attachment automatically opened in supplier invoices (for reviewing quickly)
Humm, it would be better to have a setup by user... what do you think ?
Best regards

pedrobaeza:
You can propose a pull request adding such option.

legalsylvain:
>  Humm, it would be better to have a setup by user... what do you think ?

A per user option looks great.

PR #2796, opened 2024-04-15
[15.0][FIX] web_pivot_computed_measure: Avoid Infinity errors by comparisons

CarlosRoca13:
As with the values that computed measurements can take, comparisons can also take these values which are formatted by the function getFormattedVariation. So we extend this method to take infinite values into account.

cc @Tecnativa TT48745

ping @pedrobaeza @chienandalu

### DIFF
--- a/web_pivot_computed_measure/static/src/pivot/pivot_renderer.esm.js
+++ b/web_pivot_computed_measure/static/src/pivot/pivot_renderer.esm.js
@@ -12,4 +12,10 @@ patch(PivotRenderer.prototype, "web_pivot_computed_measure.PivotRenderer", {
         }
         return this._super(...arguments);
     },
+    getFormattedVariation(cell) {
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Please always put steps to reproduce the problem, not only the explanation of the potential error it fixes. That should go later after that steps.

chienandalu (approved):

pedrobaeza:
@CarlosRoca13 can you indicate the steps to reproduce in the commit message?

CarlosRoca13:
I don't know exactly the precise steps to get to the error, I have only been able to reproduce it in our client instance and I know that the consequence is that at this point an Infinity value is arriving. But I could not reproduce it in a clean instance 🤔

BTW I test this code in the instance of the client by adding it with the console of the browser and it works well

PR #2803, opened 2024-04-22
[15.0][FIX] web_responsive: Use recordParams when accesing to searchView on mobile interface to avoid loosing domain of form view

CarlosRoca13:
You can reproduce the error by installing sale_order_secondary_unit and doing this steps:

1. Add secondary units of measure on different products
2. Create a sale order and add a product with a secondary uom
3. Display the window on a Mobile resolution
4. Refresh
5. Try to change the secondary uom of the line
6. You will see all the secondary uoms created

With this changes, we use the correct domain as it is used by _toggleAutoComplete normal way

cc @Tecnativa TT48837

ping @chienandalu @sergio-teruel @pedrobaeza

### DIFF
--- a/web_responsive/static/src/legacy/js/web_responsive.js
+++ b/web_responsive/static/src/legacy/js/web_responsive.js
@@ -146,6 +146,7 @@ odoo.define("web_responsive", function (require) {
             const options = this._super.apply(this, arguments);
             _.extend(options, {
                 on_clear: () => this.reinitialize(false),
+                domain: this.record.getDomain(this.recordParams),
             });
             return options;
         },


pedrobaeza (approved):
The context is passed already or should also be parsed explicitly?

web_responsive/static/src/legacy/js/web_responsive.js
```diff
@@ -146,6 +146,7 @@ odoo.define("web_responsive", function (require) {
             const options = this._super.apply(this, arguments);
             _.extend(options, {
                 on_clear: () => this.reinitialize(false),
+                domain: this.record.getDomain(this.recordParams),
```
chienandalu, line 149: Maybe this should be a fix to Odoo itself, but maybe it's hard to reproduce in core...

CarlosRoca13: The only way I found to reproduce it in core is by adding a many2one field in a line model like "sale.order.line" adding a domain in the form view of the line, and in mobile view pressing the button search more. I will open a PR to odoo to solve it

chienandalu: Well, that's precisely the case, isn't it? :sweat_smile:

PR #2804, opened 2024-04-23
[16.0][IMP] web_refresher: Charge t-inherit-mode extension views just after the main

CarlosRoca13:
By doing this, the extension view is charged just after the main, so if some other views are using this other, to make a primary one, like project is doing for project.task control_panel, the refresher will continue be shown.

Before the patch:
image

After the patch:
image

cc @Tecnativa TT48551

ping @chienandalu @pedrobaeza


chienandalu (approved):
Great! :+1:

web_refresher/__manifest__.py
```diff
@@ -11,7 +11,17 @@
         "web.assets_backend": [
             "web_refresher/static/src/**/*.scss",
             "web_refresher/static/src/**/*.esm.js",
-            "web_refresher/static/src/**/*.xml",
+            "web_refresher/static/src/xml/refresher.xml",
+            (
```
pedrobaeza, line 18: Always explain the whys

```suggestion
            # Load the modification of the master template just after it,
            # for having the modification in all the primary extensions.
            # Example: the project primary view.
            (
```

CarlosRoca13:
Added

pedrobaeza (approved):
/ocabot merge minor

PR #2806, opened 2024-04-23
[16.0][ADD] web_widget_pattern

hbrunn:
This module allows to use [patterns](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) in text fields in the backend.

Usage
=====

You can either define a pattern in your Python source code on a field as
in

```python
       my_field = fields.Char(pattern='[0-9]')
```

or on view level as in

```xml
       <field name="my_field" pattern="[0-9]" />
```

Demo data adds a pattern on the street2 field on partner forms to only
allow numbers.


OpenCode (approved):

Jaimermaccione (approved):
LGTM! Functional review on runboat

PR #2807, opened 2024-04-24
[17.0][IMP] web_refresher: Charge t-inherit-mode extension views just after the main

CarlosRoca13:
By doing this, the extension view is charged just after the main, so if some other views are using this other, to make a primary one, like project is doing for project.task control_panel, the refresher will continue be shown.

FW-PORT of: https://github.com/OCA/web/pull/2804

cc @Tecnativa TT48551

ping @pedrobaeza @chienandalu


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #2813, opened 2024-04-26
[17.0][MIG] web_company_color: Migration to 17.0 

peluko00:


NL66278:
I tested it, but setting company colors does not have any effect on the UI. I noted the same on Odoo 16.0 (even though filling all company colors to avoid the bug on 16.0). This did work before. Apparently something changed in core, both in 16.0 and 17.0, that makes the module ineffective. Unfortunately I am not well versed in either scss or bootstrap to find the cause myself, without a lot of time that I currently do not have.

EDIT: I got the 16.0 version finally working just by leaving debug mode. It kept working after reactivating debug mode.

CarlosRoca13:
I have added a fix to the problem mentioned in this PR, please note it for when merging to cherry-pick the commit.

https://github.com/OCA/web/pull/2902

micheledic:
Hello @peluko00 ,
the module is not working.
Instead of compiling the file, the file is printed on the HTML

LoisRForgeFlow (changes requested):
Functional test in local, it does not work:

image

PieterPaulussen:
I believe this PR can be closed because this has been merged: https://github.com/OCA/web/pull/3025

PR #2814, opened 2024-04-29
[14.0][IMP] web_company_color: UI improvements

ajaniszewska-dev:
1. add field to control input bg color : before this change "button background color" was changing not only button's bg color but also input fields bg color.

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -83,7 +83,7 @@ class ResCompany(models.Model):
           color: %(color_link_text)s !important;
         }
         .o_required_modifier.o_input, .o_required_modifier .o_input {
-          background-color: lighten(%(color_button_bg)s, 10%%) !important;
+          background-color: %(color_input_bg)s !important;
         }
     """
 
@@ -98,6 +98,7 @@ class ResCompany(models.Model):
     color_button_bg_hover = fields.Char(
         "Button Background Color Hover", sparse="company_colors"
     )
+    color_input_bg = fields.Char("Input Background Color", sparse="company_colors")
     color_link_text = fields.Char("Link Text Color", sparse="company_colors")
     color_link_text_hover = fields.Char(
         "Link Text Color Hover", sparse="company_colors"
@@ -129,6 +130,7 @@ class ResCompany(models.Model):
                 "color_button_text",
                 "color_link_text",
                 "color_link_text_hover",
+                "color_input_bg",
             )
             result = super().write(values)
             if any([field in values for field in fields_to_check]):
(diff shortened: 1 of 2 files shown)


JuMiSanAr (approved):
Looks good.
However, I'm assuming the previous color system was intended as is and someone might want to keep it, maybe this should go into its own module?

JuMiSanAr (approved):

ajaniszewska-dev:
@Tardo  could you review? :pray:

web_company_color/view/res_company.xml
```diff
@@ -28,6 +28,7 @@
                         <field name="color_button_bg" widget="color" />
                         <field name="color_button_bg_hover" widget="color" />
                         <field name="color_button_text" widget="color" />
+                        <field name="color_input_bg" widget="color" />
```
Tardo: Since it only affects required fields, a more specific name would be better.
```suggestion
                        <field name="color_required_input_bg" widget="color" />
```

web_company_color/models/res_company.py
```diff
@@ -196,6 +198,7 @@ def _scss_get_sanitized_values(self):
                 or 'theme-color("primary")',
                 "color_link_text_hover": values.get("color_link_text_hover")
                 or 'darken(theme-color("primary"), 10%)',
+                "color_input_bg": values.get("color_input_bg") or "#FFF",
```
Tardo: Use the default color for required fields. Unfortunately the color value is hardcoded :/
```suggestion
                "color_required_input_bg": values.get("color_required_input_bg") or "#D2D2FF",
```

web_company_color/models/res_company.py
```diff
@@ -98,6 +98,7 @@ class ResCompany(models.Model):
     color_button_bg_hover = fields.Char(
         "Button Background Color Hover", sparse="company_colors"
     )
+    color_input_bg = fields.Char("Input Background Color", sparse="company_colors")
```
Tardo:
```suggestion
    color_required_input_bg = fields.Char("Required Input Background Color", sparse="company_colors")
```

Tardo (changes requested):
Thanks for the work!

It would also be interesting to change the color of the border... perhaps simply darkening the selected color is enough...

Tardo:
@JuMiSanAr 
> Looks good. However, I'm assuming the previous color system was intended as is and someone might want to keep it, maybe this should go into its own module?

Yes, it may be a problem... but for me the previous color was not correct either. I don't know to what extent the previous color matched the real color.

Can always do a migration script and set the color of the new field to the value that the operation gave: ```lighten(%(color_button_bg)s, 10%)```

web_company_color/models/res_company.py
ajaniszewska-dev: thanks @Tardo , addressed in fixup commit.

PR #2816, opened 2024-05-06
[16.0] [IMP] web_responsive showing main menu when no application is selected

eLBati:
See

menu

### DIFF
--- a/web_responsive/static/src/components/apps_menu/apps_menu.esm.js
+++ b/web_responsive/static/src/components/apps_menu/apps_menu.esm.js
@@ -24,6 +24,14 @@ patch(WebClient.prototype, "web_responsive.DefaultAppsMenu", {
             document.body.classList.toggle("o_apps_menu_opened", state);
         });
     },
+    _loadDefaultApp() {
+        var menu_apps_dropdown = document.querySelector(
+            ".o_navbar_apps_menu .dropdown-toggle"
+        );
+        menu_apps_dropdown.click();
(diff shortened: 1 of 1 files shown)


Tardo:
Code OK. But idk how to "solve" the tests...

Tardo (approved):

Jaimermaccione (approved):
LGTM! Functional review on runboat

pedrobaeza (changes requested):
This must be something optional.

eLBati:
Hi @pedrobaeza , what problem do you see in setting [this](https://github.com/OCA/web/pull/2816#issue-2280800222) as default behavior o the module?
Thanks!

pedrobaeza:
For several reasons:

- It's not what users are used to.
- A lot of users have an initial action set. How does this combine with that?
- There are modules like https://github.com/OCA/server-ux/tree/16.0/announcement that shows popups on login that will conflict.

If you assure that there are no problems with points 2 and 3, then 1 can be ignored.

primes2h:
Thanks @eLBati. 
Whether it is optional or not, I consider this a must have feature.

primes2h:
> 
>     * A lot of users have an initial action set. How does this combine with that?


It works well with home action.

home_action



> 
>     * There are modules like https://github.com/OCA/server-ux/tree/16.0/announcement that shows popups on login that will conflict.

It works well with announcements too.

announcements

primes2h:
@eLBati 

I would hide the top left button after login.

web_responsive

PR #2817, opened 2024-05-07
[14.0][FIX][IMP] web_ir_actions_act_multi: uninstall issue

### LINKED ISSUE #2394: Uninstallation of web_ir_actions_act_multi, v14 ce, break database

sajadkk:
This commit fixes uninstallation issue of web_ir_actions_act_multi module. At the time of uninstallation, it removes several columns like name, binding_model_id, binding_type, binding_view_types... from ir_actions, ir_act_server, etc. This makes the database not usable anymore.

Fixes #2394

### DIFF
--- a/web_ir_actions_act_multi/models/ir_actions.py
+++ b/web_ir_actions_act_multi/models/ir_actions.py
@@ -7,7 +7,7 @@ class IrActionsActMulti(models.Model):
     _name = "ir.actions.act_multi"
     _description = "Action Mulit"
     _inherit = "ir.actions.actions"
-    _table = "ir_actions"
+    _table = "ir_act_multi"
 
     type = fields.Char(default="ir.actions.act_multi")
 
--- a/web_ir_actions_act_multi/README.rst
+++ b/web_ir_actions_act_multi/README.rst
@@ -82,6 +82,7 @@ Contributors
   * Alexey Pelykh <alexey.pelykh@corphub.eu>
 
 * Manuel Calero - Tecnativa
+* Sajad KK
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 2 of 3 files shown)


OpenCode (approved):

gaikaz (commented):
I have encountered this issue too and this seems like a way to fix it, but are we sure this doesn't break the intended functionality of the module?

I'm curious why such approach of using the original table was chosen in the first place.

sajadkk:
it doesn't break the intended functionality

gaikaz (approved):

baguenth (approved):

baguenth:
@matiasperalta1 Can you may also take a look? This was initially added by you. Maybe you can also remember why adding _table in the first place? In general I am confused why we need the .py-part at all.

web_ir_actions_act_multi/models/ir_actions.py
```diff
@@ -7,7 +7,7 @@ class IrActionsActMulti(models.Model):
     _name = "ir.actions.act_multi"
     _description = "Action Mulit"
     _inherit = "ir.actions.actions"
-    _table = "ir_actions"
+    _table = "ir_act_multi"
```
chienandalu, line 10: I think you should better do this with an uninstall hook to avoid having another relation. And maybe this could have migration issues when updating the module.

PR #2818, opened 2024-05-10
[16.0][FIX] web_calendar_slot_duration: fixing behavior on slot duration

imlopes:
Convert all to float :
if we use a context like {'calendar_slot_duration': '01:30:00'} we will have on the backend a duration of 10 hour and 30 minutes instead of 1 hour and 30 minutes

image

And using a slot_duration > 1 hour, the calendar view will be adapted to the slot duration but has not a good presentation.
So I'm adding a possibility to adapt or not the calendar view using `adapt_view_to_slot_duration` on the context too.

### DIFF
--- a/web_calendar_slot_duration/readme/CONFIGURE.rst
+++ b/web_calendar_slot_duration/readme/CONFIGURE.rst
@@ -7,10 +7,10 @@ action includes a context similar to this (example is the default value)::
 
 In addition, you can also configure the calendar view's default mode by adding::
 
-{"calendar_slot_duration": "00:30:00", "adapt_view_to_slot_duration": False}
+    {"calendar_slot_duration": "00:30:00", "keep_default_view_slot_duration": True}
 
-The ``adapt_view_to_slot_duration`` key is optional and defaults to ``True``.
-When set to ``False``, the calendar view will not adapt its view to the slot size.
+The ``keep_default_view_slot_duration`` key is optional and defaults to ``False``.
+When set to ``True``, the calendar view will not adapt its view to the slot size.
 
 For example, if you want to set the default slot duration to 1 hour and 30 minutes,
 by default the calendar view will adapt its view to show slots of 1 hour and 30 minutes.
--- a/web_calendar_slot_duration/readme/CONFIGURE.rst
+++ b/web_calendar_slot_duration/readme/CONFIGURE.rst
@@ -5,5 +5,16 @@ action includes a context similar to this (example is the default value)::
 
     {"calendar_slot_duration": "00:30:00"}
 
+In addition, you can also configure the calendar view's default mode by adding::
+
+{"calendar_slot_duration": "00:30:00", "adapt_view_to_slot_duration": False}
+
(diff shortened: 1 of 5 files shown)


web_calendar_slot_duration/static/src/js/calendar_common_renderer.esm.js
```diff
@@ -11,7 +11,10 @@ patch(
     {
         get options() {
             const options = this._super(...arguments);
-            if (this.env.searchModel.context.calendar_slot_duration) {
+            if (
+                this.env.searchModel.context.calendar_slot_duration &&
+                this.env.searchModel.context.adapt_view_to_slot_duration
```
yajo: I don't understand this fix. Could you post some screenshots explaining the problem you're fixing here please? Also, don't you think it would make sense to default to `true` for `adapt_view_to_slot_duration`?

yajo (changes requested):
The float fix is obvious, but the other one isn't.

web_calendar_slot_duration/static/src/js/calendar_common_renderer.esm.js
imlopes: Hello @yajo 
The `adapt_view_to_slot_duration` is to solve the way to display the calendar view.
If you check on my screen shot on description, If I put a `slot_duration` like 30m, on the calendar the view will be adapted to display slots 30min by 30min and IMO it's not very "beautiful". 

And yes, I'm agreed with you about the `default` to `true` to not change the standard behavior .

yajo: OK I see... Well, then it's OK, but could you add some docs in the readme please?

imlopes: @yajo Can you check ?

web_calendar_slot_duration/readme/CONFIGURE.rst
```diff
@@ -5,5 +5,16 @@ action includes a context similar to this (example is the default value)::
 
     {"calendar_slot_duration": "00:30:00"}
 
+In addition, you can also configure the calendar view's default mode by adding::
+
+{"calendar_slot_duration": "00:30:00", "adapt_view_to_slot_duration": False}
+
+The ``adapt_view_to_slot_duration`` key is optional and defaults to ``True``.
+When set to ``False``, the calendar view will not adapt its view to the slot size.
```
yajo:
```suggestion
    {"calendar_slot_duration": "00:30:00", "keep_default_view_slot_duration": True}

The ``keep_default_view_slot_duration`` key is optional and defaults to ``False``.
When set to ``True``, the calendar view will not adapt its view to the slot size.
```

yajo (changes requested):

imlopes:
@yajo can you check and validate?
Thanks a lot

yajo (approved):
/ocabot merge minor

pedrobaeza:
Please fw-port it to 17.0

PR #2819, opened 2024-05-10
[16.0][ADD] web_calendar_slot_duration_by_user

imlopes:


imlopes:
@yajo 

Can you check this PR too ?
Here my objective is to add a slot duration by user.

Context :

In some cases, an employee will plan by default meets each hour. But another employee (in another department) will plan his meets each 30min.

web_calendar_slot_duration_by_user/models/res_users.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    use_custom_calendar = fields.Boolean(
+        help="If checked, the user will use the custom calendar",
+    )
+    slot_duration = fields.Char(
+        help="The slot duration for the user",
+    )
```
yajo: I think you can just set a default slot duration for everybody and remove the boolean field.

Also the duration should be IMHO a `Float` using the `float_time` widget for better UI consistency.

web_calendar_slot_duration_by_user/views/res_users.xml
```diff
@@ -0,0 +1,22 @@
...
+    <record id="view_users_form" model="ir.ui.view">
+        <field name="model">res.users</field>
+        <field name="inherit_id" ref="base.view_users_form" />
+        <field name="arch" type="xml">
+            <page name="preferences" position="inside">
+                <group name="Calendar View Preference">
+                    <field name="use_custom_calendar" />
+                    <field
+                        name="slot_duration"
+                        help="Duration of the default duration for the calendar view in the format HH:MM:SS"
```
yajo: Add help directly in the model.

yajo (changes requested):

imlopes:
@yajo this one too, can you check ?
Thanks

web_calendar_slot_duration_by_user/models/res_users.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    slot_duration = fields.Float(
+        help="Duration of the default duration for the calendar view in the format HH:MM:SS",
+        default="1",
```
yajo, line 12: issue: if field is float, default value should be float/int:

```suggestion
        default=1.0,
```

yajo (changes requested):

PR #2823, opened 2024-05-15
[16.0][FIX] web_field_tooltip: access to ir.model and ir.model.fields…

glitchov:
If you are not admin you can't create any new tooltip as a tooltips manager because of the access to ir.model and ir.model.fields. Giving read access to this specific tooltip manager group seems to be acceptable.

### DIFF
--- a/web_field_tooltip/security/ir_model_access.xml
+++ b/web_field_tooltip/security/ir_model_access.xml
@@ -23,4 +23,24 @@
         <field name="perm_unlink" eval="1" />
     </record>
 
+    <record model="ir.model.access" id="ir_model_access_reader">
+        <field name="name">ir.model access reader</field>
+        <field name="model_id" ref="base.model_ir_model" />
+        <field name="group_id" ref="web_field_tooltip.group_tooltip_manager" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+    </record>
+
+    <record model="ir.model.access" id="ir_model_fields_access_reader">
+        <field name="name">ir.model.fields access reader</field>
+        <field name="model_id" ref="base.model_ir_model_fields" />
+        <field name="group_id" ref="web_field_tooltip.group_tooltip_manager" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+    </record>
+
 </odoo>
--- a/web_field_tooltip/views/ir_model_fields_tooltip.xml
+++ b/web_field_tooltip/views/ir_model_fields_tooltip.xml
@@ -18,10 +18,14 @@
                     <group name="first">
                         <group name="left">
                             <field name="active" invisible="1" />
-                            <field name="model_id" />
+                            <field
+                                name="model_id"
+                                options="{'create_edit': false, 'no_open': True}"
+                            />
                             <field
                                 name="field_id"
                                 domain="[('model_id', '=', model_id)]"
+                                options="{'create_edit': false, 'no_open': True}"
                             />
                         </group>
                         <group name="right">


web_field_tooltip/models/ir_model_fields_tooltip.py
```diff
@@ -75,3 +91,21 @@ def _compute_name(self):
...
                 tooltip.field_id.name, tooltip.model_id.name
             )
+
+    @api.depends("proxy_model_id")
+    def _compute_model_id(self):
+        for rec in self:
+            rec.model_id = (
+                rec.env["ir.model"].sudo().browse(rec.proxy_model_id.id)
+                if rec.proxy_model_id
+                else False
```
benwillig: I think we can avoid the browse as we already have the model ID.

web_field_tooltip/models/ir_model_fields_tooltip.py
```diff
@@ -75,3 +91,21 @@ def _compute_name(self):
...
+            )
+
+    @api.depends("proxy_field_id")
+    def _compute_field_id(self):
+        for rec in self:
+            rec.field_id = (
+                rec.env["ir.model.fields"].sudo().browse(rec.proxy_field_id.id)
+                if rec.proxy_field_id
+                else False
+            )
```
benwillig: same here with the field

web_field_tooltip/models/ir_model_fields_proxy.py
```diff
@@ -0,0 +1,41 @@
...
+
+    id = fields.Integer()
+    name = fields.Char()
+    description = fields.Char(translate=True)
+    ttype = fields.Char(string="Field Type")
+    index = fields.Boolean(string="Indexed")
+    store = fields.Boolean(string="Stored")
+    readonly = fields.Boolean()
+    relation = fields.Char(string="Related Model")
+    model_id = fields.Integer()
```
benwillig: can't we simply have the ID, name (and maybe description) ? same question for the ir.model.proxy model. I'm wondering if those models should not be declared in a specific odoo module, as it could be usefull somewhere else.

web_field_tooltip/models/ir_model_proxy.py
```diff
@@ -0,0 +1,39 @@
...
+class IrModelProxy(models.Model):
+
+    _name = "ir.model.proxy"
+    _description = "Kind of proxy for ir.model"
+    _auto = False
+    _rec_name = "description"
+
+    id = fields.Integer()
+    name = fields.Char(string="Model Name")
+    description = fields.Char(translate=True)
```
benwillig: the SQL view won't handle the translate properly. So maybe a compute would be better here. All fields should be readonly too.

web_field_tooltip/models/ir_model_fields_tooltip.py
```diff
@@ -13,17 +13,25 @@ class IrModelFieldsTooltip(models.Model):
...
+        compute="_compute_model_id",
+        store=True,
+    )
+    proxy_model_id = fields.Many2one(
+        comodel_name="ir.model.proxy",
         required=True,
         help="Model for the Field Tooltip.",
-        default=lambda self: self._get_default_model_id(),
+        default=lambda self: self._get_default_proxy_model_id(),
     )
```
benwillig: I think that proxy_model_id, proxy_field_id should not be stored (or required) as the related view might be dropped and recreated during an update. I think that the proxy fields should be used only inside the views, to prevent access error, and to populate automatically the model_id and field_id fields. The proxy fields are helpers, not the actual fields we want to target.

benwillig (changes requested):

web_field_tooltip/models/ir_model_fields_proxy.py
glitchov: The idea is to have a search more ... window more friendly.

web_field_tooltip/models/ir_model_proxy.py
glitchov: The SQL view extract the jsonb so declaring the field translate=True enable the Odoo mechanism to handle it properly.

web_field_tooltip/migrations/16.0.1.0.1/pre-migrate.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
+
+from openupgradelib import openupgrade
+
+_logger = logging.getLogger(__name__)
+
+
+@openupgrade.migrate(use_env=True)
```
benwillig: I don't think we should use openupgrade only for two simple queries.

ThomasBinsfeld (approved):

benwillig (approved):

PR #2835, opened 2024-05-24
[15.0][IMP] web_filter_header_button: show groupBy filters

chienandalu:
- [x] Show groupBy filters: Now we can configure group filters to show up along the other filters so users can easily access those as well. TT49232
- [x] Put buttons in their own row: When there are too many buttons, they won't fit in the window and they will overlap with other ui elemnets. We can prevent it putting them in their own row. TT49431

image

cc @Tecnativa 

check @sergio-teruel @pedrobaeza


web_filter_header_button/demo/ir_module_module_view.xml
```diff
@@ -14,6 +14,11 @@
                     name="context"
                 >{'shown_in_panel': {'icon': 'fa-toggle-off', 'hotkey': 'u'}}</attribute>
             </filter>
+            <filter name="state" position="attributes">
+                <attribute
+                    name="context"
+                >{'group_by':'state', 'shown_in_panel': {'icon': 'fa-th-list', 'hotkey': 's'}}</attribute>
```
pedrobaeza, line 20: Why do you need to put `group_by` here? What happen if a field is not present as `<filter>` in the search bar? Isn't possible to have it on the header but not on the search? Maybe `invisible="1"`?

sergio-teruel (approved):

web_filter_header_button/demo/ir_module_module_view.xml
pedrobaeza: What about this?

chienandalu: Yes, it's something I'd liked to have, but I didn't find a very obvious way to do it at first glance. If you set the invisble flag, the filter goes away from the webclient memory and when a query for the list I don't have it to be shown...

In the other hand working with the context key, has these limitations in groups, as if you inherit from an existing group filter you'll have to overwrite the context...

And you can't add a new independent attribute, as the view validator will avoid it...

pedrobaeza: OK, please explain all of this in ROADMAP.rst.

pedrobaeza (approved):
/ocabot merge minor

PR #2845, opened 2024-06-12
[15.0][ADD] web_widget_remote_measure_utilcell: New module

chienandalu:
Adds web_widget_remote_measure support for Utilcell scales

Depends on:

- [ ] #2808 

cc @Tecnativa TT42173


pedrobaeza:
I'm feeling that now this module and the base one should be hosted at https://github.com/OCA/stock-weighing/ (it was not created before merging it).

chienandalu:
We can do it when migrating maybe?

pedrobaeza:
Well, I think this repository is too crowded and there's still not too much spread about the module, so waiting for the change will worse this.

chienandalu:
Ok, but isn't moving `web_widget_remote_measure` to another repo going to cause issues to the projects that depend on it being in `web`?

pedrobaeza:
Not too much, as those that access through pypi won't notice it, and being this module very new, I don't think too many people have it installed.

pedrobaeza:
Moved to https://github.com/OCA/stock-weighing/pull/11

PR #2856, opened 2024-06-18
[IMP] web_environment_ribbon: hide the ribbon on hover, 

legalsylvain:
Hi all. 
First thanks for this great module. this trivial PR improve the UI. on hover of the ribbon, the ribbon should be hidden to let user see more easily, the elements that are behind the ribbon.

Thanks for your review ! 

Note: This PR is in draft state because I don't figure out how to achive the following things : 
- make ribbon invisible when the mouse is hover the ribbon
- make items under the ribbon clickable.

-> removing point-events:none make items not clickable.
-> Let point-events:none disable .test-ribbon:hover

### DIFF
--- a/web_environment_ribbon/static/src/css/ribbon.css
+++ b/web_environment_ribbon/static/src/css/ribbon.css
@@ -21,7 +21,10 @@
     position: fixed;
     box-shadow: 0 0 3px rgba(0, 0, 0, 0.3);
     background: rgba(255, 0, 0, 0.6);
-    pointer-events: none;
+    &:hover {
+        opacity: 0;
+        transition: 0.2s ease;
+    }
 }
 
 .test-ribbon b {


pedrobaeza:
Isn't already hidden when you hover on (which by the way I don't like at all)? Please put the before and after

legalsylvain:
thanks for your review !

> Isn't already hidden when you hover on (which by the way I don't like at all)

No, for the time being, the ribbon is hidden on nav and header hover. 

> Please put the before and after

? didn't understood your point. could you elaborate ?

pedrobaeza:
I ask for seeing visually the improvement you are doing.

legalsylvain:
I think it's better if you test it on runboat, to feel if it's OK for you. But well, here are the screenshots : 

**Before, when the cursor is hover the ribbon**

image

(http://oca-web-16-0-7711bc91876b.runboat.odoo-community.org)

Problem : the button "Start a meeting" (and all the other components are hidden)

**After, when the cursor is hover the ribbon :** 

image


(http://oca-web-16-0-pr2856-07e4bb634f4e.runboat.odoo-community.org)

-> better visibility.

PR #2857, opened 2024-06-18
[15.0][FIX] web_filter_header_button: compatibility issues

chienandalu:
In custom views, like the reconciliation one, that use the control panel we must prevent loading the FilterButton component.

cc @Tecnativa TT49664


pedrobaeza (approved):

CarlosRoca13 (approved):
Thanks!

PR #2861, opened 2024-06-25
[14.0] [ADD] Module web_responsive_company_color

anusriNPS:
Module web_responsive_company_color is introduced for customizing web elements that are part of web_responsive that needs customization

Limitations: If new web elements are included in module, one needs to reinstall web_responsive_company_color for customization to take effect


pedrobaeza:
Isn't this the same as https://github.com/OCA/web/tree/14.0/web_company_color ?

web_company_color/models/res_company.py
```diff
@@ -200,21 +201,26 @@ def _scss_get_sanitized_values(self):
...
 
-    def scss_create_or_update_attachment(self):
+    def scss_create_or_update_attachment(self, uninstall=False):
         IrAttachmentObj = self.env["ir.attachment"]
         for record in self:
-            datas = base64.b64encode(record._scss_generate_content().encode("utf-8"))
+            if not uninstall:
+              datas = base64.b64encode(record._scss_generate_content().encode("utf-8"))
+            else:
+              datas = base64.b64encode(record._scss_generate_content(uninstall).encode("utf-8"))
```
PicchiSeba: **suggestion**: in the method `_scss_generate_content` uninstall is already handled. We could just call the method without checking if uninstall is falsy

```suggestion
datas = base64.b64encode(record._scss_generate_content(uninstall).encode("utf-8"))```
```

web_responsive_company_color/models/res_company.py
```diff
@@ -0,0 +1,22 @@
...
+                    to bottom,
+                    %(color_navbar_bg)s,
+                    desaturate(lighten(%(color_navbar_bg)s, 20%%), 15)
+                  );
+             }
+            """
+        )
+      
+    def _get_scss_template_uninstall(self):
+        return super()._get_scss_template()
```
PicchiSeba: **question**: What does this method do?.

PicchiSeba (commented):
Code review

web_company_color/models/res_company.py
```diff
@@ -227,8 +230,8 @@ def scss_create_or_update_attachment(self):
                 "company_id": record.id,
             }
             if custom_attachment:
-                custom_attachment.sudo().write(values)
+               custom_attachment.sudo().write(values)
             else:
-                values.update({"type": "binary", "mimetype": "text/scss"})
-                IrAttachmentObj.sudo().create(values)
+               values.update({"type": "binary", "mimetype": "text/scss"})
+               IrAttachmentObj.sudo().create(values)
```
PicchiSeba: **issue**: we lost the indentation here

PicchiSeba (changes requested):

web_company_color/models/res_company.py
anusriNPS: Implemented suggested change

web_company_color/models/res_company.py
anusriNPS: Indentation is restored

web_responsive_company_color/models/res_company.py
anusriNPS: This method tries to remove web elements introduced by the module for customization during uninstall, so that when this module is uninstalled customization related to its web elements are removed.

During uninstall, web_company_color module tries to return self._get_scss_template() % self._scss_get_sanitized_values() using _scss_generate_content() method.

Whenever, self._get_scss_template() is called, it provides the entire scss_template including web element which has be removed for the module for which uninstalled is performed.

In order to avoid including web elements of module which are about to uninstall, calling a scss_template_uninstall method in the module which tries uninstall. As only parent of scss_template is called in this method, I am able to remove customized web elements of module for which uninstall is performed.

If not, though module is uninstalled, customization of web element introduced by the uninstalled module remains.

web_company_color/models/res_company.py
```diff
@@ -200,21 +201,27 @@ def _scss_get_sanitized_values(self):
...
-    def _scss_generate_content(self):
+    def _scss_generate_content(self, uninstall=False):
         self.ensure_one()
         # ir.attachment need files with content to work
         if not self.company_colors:
             return "// No Web Company Color SCSS Content\n"
-        return self.SCSS_TEMPLATE % self._scss_get_sanitized_values()
+        if uninstall:
+            return (
+                self._get_scss_template_uninstall() % self._scss_get_sanitized_values()
```
renda-dev: **issue:** `_get_scss_template_uninstall` does not exist in this module and it would raise an error on the module uninstall.

**suggestion:** instead of creating a new method you could add a parameter inside the context in the "uninstall_hook", check for that inside this method and call `super()` instead of `self`

web_responsive_company_color/models/res_company.py
renda-dev: **question:** shouldn't this be something like:
```py
return super()._get_scss_template() % super()._scss_get_sanitized_values
```

web_company_color/models/res_company.py
anusriNPS: I tried to uninstall web_company_color but did not see any error. However, you are right, it might raise error without _get_scss_template_uninstall function here
 
 I also tried to update context from uninstall_hook and use super() instead of self, but uninstall fails saying super has no attribute 
image
It is trying to call the parent of web_company_color instead of the parent which needs to be called for module which is uninstalled

web_responsive_company_color/models/res_company.py
```diff
@@ -0,0 +1,30 @@
...
+             .o_menu_apps .dropdown-menu {
+                background: url('/web_responsive/static/img/home-menu-bg-overlay.svg'),
+                  linear-gradient(
+                    to bottom,
+                    %(color_navbar_bg)s,
+                    desaturate(lighten(%(color_navbar_bg)s, 20%%), 15)
+                  );
+             }
+            """
+        )
```
renda-dev: suggestion: In my opinion you should:
- Delete `_get_scss_template_uninstall`, could only cause issues, not needed
- Add a context variable in the uninstall hook (something like: `uninstall_scss=True`) before executing `scss_create_or_update_attachment`
- Check inside `_get_scss_template` if there is the variable in the context:
    - If yes: just return `super`, you only need the basic scss template
    - If not: call `super` and add your customizations

anusriNPS: Sorry for misunderstanding the suggestion. Thank your for detailed response. Updated solution as suggested

web_responsive_company_color/hooks.py
```diff
@@ -0,0 +1,11 @@
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    env["res.company"].search([]).scss_create_or_update_attachment()
+
+
+def uninstall_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {"uninstall": True})
```
renda-dev: **nitpick:**  just `uninstall` could lead to conflicts / misunderstandings, could we make it more specific? Something like `uninstall_scss` would be enough

web_company_color/models/res_company.py
renda-dev: **suggestion:** now that `_get_scss_template_uninstall` doesn't actually exist, you could remove it (also from the method's parameters) and check for the context variable instead

web_company_color/models/res_company.py
```diff
@@ -200,21 +201,27 @@ def _scss_get_sanitized_values(self):
...
         self.ensure_one()
         return URL_SCSS_GEN_TEMPLATE % self.id
 
-    def scss_create_or_update_attachment(self):
+    def scss_create_or_update_attachment(self, uninstall=False):
         IrAttachmentObj = self.env["ir.attachment"]
         for record in self:
-            datas = base64.b64encode(record._scss_generate_content().encode("utf-8"))
+            datas = base64.b64encode(
+                record._scss_generate_content(uninstall).encode("utf-8")
```
renda-dev: **nitpick:** you can get `uninstall` directly from the context inside the `_scss_generate_content` method, these changes could be removed

renda-dev (changes requested):

web_responsive_company_color/hooks.py
anusriNPS: updated to uninstall_scss in the context

web_company_color/models/res_company.py
anusriNPS: I missed to push the changes for scss_generate as testing was handled in different repo.

Not checking context variable inside _scss_generate_content and it is checked only inside _get_scss_template

web_company_color/models/res_company.py
anusriNPS: Removed these changes

renda-dev (approved):
Code and functional review, LGTM

PR #2879, opened 2024-07-10
[16.0] Create module web_datetime_picker_default_time

grindtildeath:
This module customizes the datetime picker widget and allows to define a
default time to be applied in case the user selects only a Date.

For example, if a user wants to define a commitment date without having
to specify the time on that date, setting the default time value on the
field in the Form view allows to ensure the commitment date will be set
to this time instead of the time when the page was loaded by the
browser.


ivantodorovich (approved):

ivantodorovich (approved):
💯

santostelmo (approved):

gurneyalex (approved):

PR #2883, opened 2024-07-12
[15.0][IMP] web_widget_numeric_step: option to add classes

chienandalu:
New option so we can add new classes to the input or remove existing ones so we can style the input at our will.

cc @Tecnativa TT49390


sergio-teruel (approved):

CarlosRoca13 (approved):
Code review and tested on runboat :+1:

PR #2884, opened 2024-07-16
[ADD] new module web_hide_on_edit

sergiocorato:
This module add the ability to hide buttons with a custom class when in edit mode (see README).


hbrunn:
does this do anything different than the class [`oe_read_only`](https://github.com/OCA/OCB/blob/14.0/addons/web/static/src/scss/form_view.scss#L131)?

sergiocorato:
> does this do anything different than the class [`oe_read_only`](https://github.com/OCA/OCB/blob/14.0/addons/web/static/src/scss/form_view.scss#L131)?

Probably not, I missed it, many thanks!
I close the PR

PR #2848, opened 2024-06-13
[17.0][MIG] web_m2x_options Migration to 17.0

mohs8421:
Adding all files from 16.0
Modifications:
 - Updated version to refer to Odoo 17.0
 - Updated calls to patch(…) in js files to be compatible with Odoo 17.0

PR #2847, opened 2024-06-13
[MIG] web_m2x_options: Migration to 17.0

dyngnijs:

PR #2888, opened 2024-07-22
[17.0][MIG] web_m2x_options: Migration to 17

manuelregidor:
There has been a wide refactoring work in this migration to adapt the module to the v17 environment and get a more compact code.

It also avoids checking whether certain modules are installed, which happens in other migrations proposals.

Some changes have been applied too:

- The create option in the field's option dict is only available for m2o fields, as m2m fields already include this option in v17.
- The m2o_dialog options have been removed as they don't seem to be used in v17.
- The no_open_edit option has been removed as it makes no sense in v17, since edit mode is always active.
- The no_color_picker options for m2m fields has been removed as the option canEditColor is avaiable in v17 base.

The commit history has been reduced.

Performance has been tested and no issues have been found.

@HaraldPanten 
@ValentinVinagre 

T-5977


HaraldPanten:
@manuelregidor could you check @CRogos comments here --> https://github.com/OCA/web/pull/2847 about the commit history? I think it could be shorter. 

You can add this information (that the commit history has been shortened) as well in the first message of your PR.

THX

manuelregidor:
@HaraldPanten Commit history reduced

HaraldPanten:
@CRogos could we continue the reviews here?

THX!

web_m2x_options/static/src/components/form.esm.js
```diff
@@ -0,0 +1,332 @@
...
+    ...Many2ManyTagsFieldColorEditable.props,
+    searchLimit: {type: Number, optional: true},
+    fieldColor: {type: String, optional: true},
+    fieldColorOptions: {type: Object, optional: true},
+};
+
+patch(many2OneField, {
+    m2o_options_props(props, attrs, options) {
+        const ir_options = session.web_m2x_options;
+        // Create
```
mohs8421: since you already refactored a lot, maybe you want to replicate my refactoring attempts I did in this alternative approach: https://github.com/OCA/web/pull/2848/commits/8eb54f79f937d9ea1c8918394d05ca2c560063e6
There I essentially took the section comments and split the parts of m2o_options_props along these.

manuelregidor: Thank you for your feedback, but I don't think I'm following you. Could you reformulate your suggestion, please?

mohs8421: the m2o_options_props function is quite long and could be split into more reasonable sections. Each section is already introduced by a comment. So just follow this code smell rule: https://refactoring.guru/smells/comments
This would increase understandability of it as much as it makes it easier to adapt it when needed.

manuelregidor: @mohs8421 done

HaraldPanten (approved):
Functional review in runboat.

@pedrobaeza Could you review/merge?

As you suggested, we have considered/studied the performance and we didn't find any issue.

THX!

web_m2x_options/static/src/components/base.xml
```diff
@@ -15,32 +14,17 @@
...
-                <button class="btn btn-primary" t-on-click="onCreate">Create</button>
-                <button
-                    class="btn btn-primary"
-                    t-on-click="onCreateEdit"
-                >Create and Edit</button>
-                <button class="btn" t-on-click="() => props.close()">Discard</button>
-            </t>
-        </Dialog>
-    </t>
 </templates>
```
CRogos, line 30: Have you removed the confirm create dialog?

CRogos (approved):
LGTM beside the confirm create dialog.

We disabled create... globally so this is not an issue for me anymore, but in the past this was very helpful to prevent creating entries accidently.

web_m2x_options/static/src/components/base.xml
manuelregidor: @CRogos  I removed it as I don't think it was working in v15 and v16 (or at least I couldn't make this option work). Are we sure they work in previous versions? Thank you.

CRogos: You are right. Not working in v16 either.

CRogos:
@manuelregidor could you have a look into this https://github.com/OCA/web/pull/2901 ?

manuelregidor:
@CRogos As the feature was not working in v16, we decided that there was no point in keeping it in v17 as if it was properly working and that's why we removed it. At the moment, we are not planning to put the feature back I'm afraid, but we will keep you updated in case we do. Thank you.

cuongnmtm:
I found the `open` option in the many2many_tags widget very useful, but I noticed it hasn't been migrated to v17. How are you currently handling this functionality? Are you relying on smart buttons instead, or do you have an alternative approach?

manuelregidor:
@cuongnmtm  We don't have an alternative approach at the moment I'm afraid.

cuongnmtm:
@manuelregidor Thank you for your response.

Did you encounter any difficulties during the migration process for that feature, or are you considering adding it to the roadmap for future planning - to simplify the migration PR?

manuelregidor:
@cuongnmtm We are not planning to implement that feature at the momento, so feel free to implement it if you think it's interesting enough. Thank you.

PR #2891, opened 2024-07-24
[16.0][FIX] web_refresher: Define displayRefresher in setup instead of using a getter

CarlosRoca13:
Before this changes whe trying to access to a view with the refresher, in mobile view, an error was thrown.

As it is not necessary to define the value in a get function, defining the value in the setup solves the problem.

cc @Tecnativa TT50296

ping @pedrobaeza @victoralmau


pedrobaeza (approved):
Working perfectly. Let's fast-track it to fix the problem that is hitting all mobile users:

/ocabot merge patch

PR #2892, opened 2024-07-24
[17.0][MIG]web_field_tooltip: Migration to 17.0

manuelregidor:
Refactoring. The tooltips information is shown along with the help information set in fields definitions and help information defined in views.

@HaraldPanten 
@ValentinVinagre 

T-6240


manuelregidor:
@StefanRijnhart @legalsylvain @pedrobaeza @etobella
Could you review, please? Thank you

legalsylvain:
Hi @manuelregidor. I don't use this module. (i don't get the interest, in fact, because it is generating double "?" icon).

image

Please avoid to ping random users. better ping original authors.
thanks !

manuelregidor:
@legalsylvain Sorry about that. The module is not generation a double "?" icon. The second icon is the one that shows the field info in v17 when the user is in debug mode.

legalsylvain:
> The second icon is the one that shows the field info in v17 when the user is in debug mode.

I don't think so. step to reproduce : 
- go to runboat
- log with demo / demo.
- the "?" are available, even if we are not in debug mode.

image

Did I missed something ?

manuelregidor:
@legalsylvain You see that question mark because the date_order field defintion has the parameter help with the text "Creation date of draft/sent orders,\nConfirmation date of confirmed orders.". Therefore, it's completely normal to see that question mark in that specific field.

legalsylvain:
Yes. what is the difference between the help native field and the tooltip this module introduces ?  why not use only help  ?

(sorry if it's a little off topic)

manuelregidor:
@legalsylvain This module lets you easily create custom help messages for specific fields with no need of modifying any code.

legalsylvain:
Yes, I understand the need to create custom help via UI.
Maybe we could imagine a simpler implentation.
- keep ir.model.fields.tooltip model
- display in the UI the field help + tooltip concatened in the native way

-> very lighter implementation. 
-> no duplicated "?" icon

what do you think ?

manuelregidor:
@legalsylvain Ok, I'll do that

manuelregidor:
@HaraldPanten Could you review, please?

legalsylvain (approved):
Functional Review. LGTM !

HaraldPanten (approved):
Much better now, THX!

HaraldPanten:
@etobella Could you merge this one, please? 🙏

web_field_tooltip/models/ir_model_fields_tooltip.py
```diff
@@ -34,7 +34,7 @@ class IrModelFieldsTooltip(models.Model):
         help="Set active to false to hide the Tooltip without removing it.",
     )
     field_name = fields.Char(related="field_id.name")
-    tooltip_text = fields.Html(required=True)
+    tooltip_text = fields.Text(required=True, translate=True)
```
legalsylvain, line 37: maybe this change requires a migration script for users coming from V16.

not a blocking point, though. Can be done in another PR

etobella: I think now is the right time for this, you just need to use https://github.com/odoo/odoo/blob/d1f20d6b71203e6bb4b428ae1bcbb0b76e739f51/odoo/tools/mail.py#L350

manuelregidor: Migration script added

manuelregidor:
@Tisho99 Could yu review, please?

Tisho99 (approved):
LGTM, functional and code review

PR #2658, opened 2023-11-06
[16.0][FIX] web_responsive: overlapping issue

hafizhlf:
Hi there,

Based on the report by @gpothier here https://github.com/OCA/web/commit/5de47a859287bbd71cd0532a739a0f943d6384ad#commitcomment-131721969, it seems that my commit in PR #2654 is causing the User menu to disappear when the Apps menu is shown. I've made the necessary changes to both revert the problematic commit and introduce specific changes to address this issue.

Please let me know if there are any further adjustments.

Thank you for your review!

PR #2654, opened 2023-11-02
[16.0][FIX] web_responsive: overlapping issue

hafizhlf:
**Description:**

This pull request addresses the issue of overlapping UI elements in the application. The issue was causing elements to display inappropriately, leading to a less-than-optimal user experience. The changes made in this PR fix the issue by adjusting the z-index of apps_menu.scss.

**Screenshots:**

Odoo-Quotations

PR #2541, opened 2023-06-27
[15.0][FIX] web_responsive: attachment_viewer is displayed under clock of employees view

CarlosRoca13:
Before this change, when an attachment was displayed on the screen, the employee's clocks were being shown above the attachment viewer.
image

After making this change, everything is now displayed normally.
image

cc @Tecnativa TT44101

ping @pedrobaeza @pilarvargas-tecnativa

PR #2894, opened 2024-07-25
[16.0][IMP] web_responsive: Set AttachmentViewer z-index upper than chatter

CarlosRoca13:
Before this changes when oppening an attachment from the chatter the attachment was behind the chatter and if we try to hide it, the attachement disappears.

Example:
Error shown

With this changes the AttachmentViewer appears in front of the chatter and the problem does not exists anymore.

Example:
solución

cc @Tecnativa TT50272 

ping @pedrobaeza @pilarvargas-tecnativa

### DIFF
--- a/web_responsive/static/src/components/attachment_viewer/attachment_viewer.scss
+++ b/web_responsive/static/src/components/attachment_viewer/attachment_viewer.scss
@@ -14,7 +14,7 @@
         }
         .o_AttachmentViewer {
             // On-top of navbar
-            z-index: 10;
+            z-index: 1100;
             position: absolute;
             right: 0;
             top: 0;


pedrobaeza:
You have to put screenshot to illustrate the problem. Please also check all the previous PRs modifying z-indexes to check that none of them has reappeared with this change:

https://github.com/OCA/web/pull/2734
https://github.com/OCA/web/pull/2658
https://github.com/OCA/web/pull/2654
https://github.com/OCA/web/pull/2565
https://github.com/OCA/web/pull/2541
https://github.com/OCA/web/pull/2522

CarlosRoca13:
Checked and added 2 gifs to show the problem and the result on the PR description

pilarvargas-tecnativa (approved):
:+1:

pedrobaeza (approved):
/ocabot merge patch

Please fw-port it to 17 if applicable

CarlosRoca13:
On 17.0 it's not necessary 
pdf

PR #2899, opened 2024-07-30
[16.0][IMP] web_responsive: Change replace for less dangerous extension options

CarlosRoca13:
Replacing all the template is a bit dangerous if any other module tries to extend that tamplate.

So this changes make the same as it is done on before, but using the other options provided by xpath that are less dangerous.

cc @Tecnativa TT50275

ping @pedrobaeza @carolinafernandez-tecnativa

### DIFF
--- a/web_responsive/static/src/components/chatter_topbar/chatter_topbar.xml
+++ b/web_responsive/static/src/components/chatter_topbar/chatter_topbar.xml
@@ -11,11 +11,12 @@
         owl="1"
         t-inherit-mode="extension"
     >
-        <xpath expr="//div[contains(@class, 'o_ChatterTopbar')]" position="replace">
+        <xpath expr="//div[hasclass('o_ChatterTopbar')]" position="before">
             <t t-if="ui.isSmall">
                 <div
                     class="o_ChatterTopbar_rightSection d-flex border-bottom"
                     style="max-height:45%"
+                    id="mobileChatterTopbar"
                 >
                     <button
                         t-if="chatterTopbar.chatter.thread.allAttachments.length === 0"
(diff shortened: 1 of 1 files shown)


legalsylvain:
Nice simplification !

carolinafernandez-tecnativa (approved):
LGTM Thanks :)

pedrobaeza (changes requested):
The change is not correct I'm afraid:

image

image

image

CarlosRoca13:
I'm not having the same problem... @legalsylvain @carolinafernandez-tecnativa, can you confirm it?
scroll

pedrobaeza:
I'm working with Firefox. Have you tried there?

pedrobaeza:
And anyway, you can see in your sample that there's a bit of lateral scroll. Maybe you have selected a wider screen resolution than me.

CarlosRoca13:
Okay, I could reproduce it, but the problem is not introduced on this PR, you can see the same problem on 16.0 branch runboat

I will review it BTW

pedrobaeza:
OK, but please check on Firefox on my example, where the lateral scroll is wild.

CarlosRoca13:
@pedrobaeza the problem was related with modules [web_save_discard_button](https://github.com/OCA/web/tree/16.0/web_save_discard_button) and [web_theme_classic](https://github.com/OCA/web/tree/16.0/web_theme_classic)

Set both as rebel modules

legalsylvain:
> @pedrobaeza the problem was related with modules [web_save_discard_button](https://github.com/OCA/web/tree/16.0/web_save_discard_button) and [web_theme_classic](https://github.com/OCA/web/tree/16.0/web_theme_classic)

could you elaborate the problem. I can try to fix [web_theme_classic](https://github.com/OCA/web/tree/16.0/web_theme_classic) (or [web_save_discard_button](https://github.com/OCA/web/tree/16.0/web_save_discard_button)) when I have a time.

pedrobaeza:
@legalsylvain these problems are caused by that modules: https://github.com/OCA/web/pull/2899#pullrequestreview-2207211141

legalsylvain:
Well, I didn't develop and don't use web_save_discard_button, but if there a problem between web_theme_classic and web_responsive, I can try to take a look.

CarlosRoca13:
Removed commit to set as rebel, because it does not avoid the installation of the modules

CarlosRoca13:
I think that this PR can be merged, if you agree

pedrobaeza (approved):
Working without problems uninstalling both undesired modules:

/ocabot merge minor

PR #2902, opened 2024-07-31
[16.0][FIX] web_company_color: Set defaults hardcoded to avoid errors charging the assetbundle

CarlosRoca13:
When trying to charge the assetbundle the theme-color("primary") function is not found what causes problems when trying to set the default values.
Example_no_change

With this changes the default value is the value set on https://github.com/odoo/odoo/blob/ad9f8fae717aae4cad5b3715c1f73ee906b64f8d/addons/web/static/src/scss/primary_variables.scss#L55 this change avoid the problems and the module works fine.
Example_change

Addapted too the class for required fields.

cc @Tecnativa TT50098

ping @sergio-teruel @pilarvargas-tecnativa

### DIFF
--- a/web_company_color/__manifest__.py
+++ b/web_company_color/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Web Company Color",
     "category": "web",
-    "version": "16.0.1.2.0",
+    "version": "16.0.1.2.1",
     "author": "Alexandre Díaz, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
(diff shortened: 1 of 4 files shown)


CarlosRoca13:
Added too the migration script for the companies that the change has not beeing applied

pedrobaeza (approved):
To hardcode the colors is the only option here.

pilarvargas-tecnativa (approved):
Thanks! :+1:

chienandalu (approved):

PR #2909, opened 2024-08-08
[16.0][FIX] web_widget_numeric_step: Remove focus to avoid annoying flickering effect

CarlosRoca13:
The focus introduced on the 16.0 migration is causing a flickering effect that is a bit annoying and does not make complete sense, because if you click more than once the cursor is positioned at the end of the text, which makes it difficult to edit the input.

The counterpart of these changes is that to edit the input using the keyboard we will have to click on the field. Although this is how the module has been operating since its inception.

Before this changes:
parpadeo

After this changes:
no parpadeo

cc @Tecnativa TT50439

ping @chienandalu @pedrobaeza

### DIFF
--- a/web_widget_numeric_step/static/src/numeric_step.esm.js
+++ b/web_widget_numeric_step/static/src/numeric_step.esm.js
@@ -4,17 +4,12 @@ import {registry} from "@web/core/registry";
 import {standardFieldProps} from "@web/views/fields/standard_field_props";
 import {_lt} from "@web/core/l10n/translation";
 import {FloatField} from "@web/views/fields/float/float_field";
-import {hasTouch} from "@web/core/browser/feature_detection";
 
 export class NumericStep extends FloatField {
     setup() {
         super.setup();
     }
     _onStepClick(ev) {
-        const $el = $(ev.target).parent().parent().find("input");
-        if (!hasTouch()) {
-            $el.focus();
-        }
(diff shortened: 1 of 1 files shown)


chienandalu (commented):
Oh, I see now that the flickering is a little bit less annoying with the regular theme (I tested it with web_theme_classic installed in the runboat)... In the one hand, we lived without this focus for several versions without noticing it, in the other hand is a nice UX touch...

pedrobaeza:
It was changed by @yajo. Let's see the reasons for that change. At first sight, I would say the idea was to click on -/+ and have the focus for refining the amount. If the flickering is not very annoying, we can live with it.

chienandalu:
Not really, it was introduced in the migration for v16 and @yajo disabled it for touch devices

pedrobaeza:
Uhm, then it was @dsolanki-initos

CarlosRoca13:
I think the focus doesn't contribute much, since in the end, you're going to have to position the cursor where you want to modify the value. Also, I think the module's scope isn't meant to cover the need to manually modify the input... That's what the buttons and the scroll are for. Don't you think the same?

yajo (commented):
I agree it doesn't add value in desktop, and it works better now. Thanks!

yajo (approved):
Ah sorry, bad button. Approving with the correct button now 😆

PR #2684, opened 2023-11-24
[17.0][MIG] web_responsive: Migration to 17.0

TesaLazarenko:
- [x] change the background of the app menu
- [x] change search list items
- [x] fix sticky buttons panel
- [x] Readme / Description with screenshots
- [x] fix app search

PR #2910, opened 2024-08-08
[16.0][FIX] web_responsive: Remove overflow:visible to tables to avoid side effects

victoralmau:
Remove overflow:visible to tables to avoid side effects

**Without `web_responsive`**
sin-web-responsive

**With `web_responsive` - before**
antes-web_responsive

**With `web_responsive` - after**
despues-web_responsive

Please @CarlosRoca13 can you review it?

@Tecnativa TT50468

### DIFF
--- a/web_responsive/static/src/legacy/scss/web_responsive.scss
+++ b/web_responsive/static/src/legacy/scss/web_responsive.scss
@@ -250,7 +250,6 @@ html .o_web_client .o_action_manager .o_action {
 // Sticky Header & Footer in List View
 .o_list_view {
     .table-responsive {
-        overflow: visible;
         .o_list_table {
             // th & td are here for compatibility with chrome
             thead tr:nth-child(1) th {


pedrobaeza:
Please check why that was introduced, as maybe removing it has another "side effect". You can ask as well the original contributor.

victoralmau:
These changes were added at https://github.com/OCA/web/pull/2405 (I don't know exactly why) and removed at https://github.com/OCA/web/pull/2684.

Ping @bvanjeelharia Can you tell us why that code was added, is this change correct and will not cause side effects or do you propose a better one? Thanks

pedrobaeza:
The ping wasn't correct. Trying again @bvanjeelharia

pedrobaeza:
If removed in 17.0, it seems it's not very needed.

victoralmau:
Specifically, this piece of code was removed in v17 https://github.com/OCA/web/blob/16.0/web_responsive/static/src/legacy/scss/web_responsive.scss#L250-L288

pedrobaeza:
Well, it was removed because Odoo incorporates the sticky feature natively, so maybe the overflow was not removed specifically, but inside that block. Anyway, it looks more and more a side effect.

pedrobaeza (approved):

chienandalu (approved):

CarlosRoca13 (approved):
/ocabot merge patch

PR #2915, opened 2024-08-13
[17.0][MIG] web_widget_dropdown_dynamic: Migration to 17.0

CarlosRoca13:
cc @Tecnativa TT50100

ping @carlos-lopez-tecnativa @chienandalu


carlos-lopez-tecnativa:
Sorry, I couldn't test it.

image

web_widget_dropdown_dynamic/__manifest__.py
```diff
@@ -5,15 +5,17 @@
...
     "license": "AGPL-3",
     "author": "CorporateHub, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
     "depends": ["web"],
     "installable": True,
     "assets": {
         "web.assets_backend": [
-            "web_widget_dropdown_dynamic/**/*",
+            "web_widget_dropdown_dynamic/static/src/js/field_dynamic_dropdown.esm.js",
+            "web_widget_dropdown_dynamic/static/tests/web_widget_dropdown_dynamic_tests.esm.js",
```
chienandalu, line 19: Maybe this should be in `'web.qunit_suite_tests`?

CarlosRoca13:
Changes done

carlos-lopez-tecnativa (commented):
Thanks for the demo data; it helps with testing.

One question: is this module incompatible with the `web_m2x_options` module?
image

CarlosRoca13:
Thanks for the notice, Carlos. The problem comes from the lack of a security control in the `web_m2x_options` module. I have added the fix in a new commit within this same PR.

web_widget_dropdown_dynamic/models/ir_filters.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class IrFilters(models.Model):
+    _inherit = "ir.filters"
+
+    dynamic_dropdown_int = fields.Integer("Dropdown Integer")
```
chienandalu: Doesn't this create a presistent column which is useless outside its demo purposes?

web_widget_dropdown_dynamic/README.rst
```diff
@@ -71,6 +71,13 @@ Usage
        context="{'depending_on': other_field}"
    />
 
+**DEMO**
+
+On User defined filters added new field to show the feature, it is
+called **Dropdown Integer**. If any user selected just One option shoud
+appear, but if Mitchell Admin it should be possible to select option One
+and Two.
```
chienandalu, line 79: I don't know if its expected, but in a new record, the behaviour doesn't seem to work right:

image

CarlosRoca13: It's depending on the field "Available for User"

web_widget_dropdown_dynamic/models/ir_filters.py
CarlosRoca13: Hmm yes... any other option to do this?

chienandalu: You can load it as data. Something like this should work, I think

```xml
  <record id="demo_web_widget_dropdown_dynamic" model="ir.model.fields">
    <field name="model">ir.filters</field>
    <field name="model_id" ref="base.model_ir_filters"/>
    <field name="name">dynamic_dropdown_int</field>
    <field name="selectable" eval="True"/>
    <field name="selection">[('one', 'one'), ('two', 'two')]</field>
    <field name="ttype">selection</field>
  </record>
```

web_widget_dropdown_dynamic/README.rst
chienandalu: <strike>But it works when the record is already saved...</strike>

forget about it, you're right :smile:

CarlosRoca13:
@chienandalu changes done :smile:

chienandalu (approved):
Great! :+1:

carlos-lopez-tecnativa (approved):
LGTM, thanks!

PR #2920, opened 2024-08-20
[16.0][ADD] web_quick_start_screen: New module

chienandalu:
This module adds a configurable start screen so users can go right into
their more common tasks.

image

cc @Tecnativa TT50477

please check @pedrobaeza @CarlosRoca13


web_quick_start_screen/wizards/quick_start_screen.py
```diff
@@ -0,0 +1,29 @@
...
+            actions.append(
+                {
+                    "title": action.name,
+                    "description": (action.description),
+                    "image": action.image,
+                    "icon": (
+                        f"fa {action.icon_name or 'fa-flash'} "
+                        f"o_start_icon_color_{action.color}"
+                    ),
+                    "action_name": action.action_ref_id.id,
```
leemannd: Warning the field is non mandatory and will break with the following stacktrace
```
  File "/opt/odoo/odoo/http.py", line 1649, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1676, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1880, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 732, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/mnt/data/odoo-addons-dir/web_quick_start_screen/controllers/quick_start_screen.py", line 17, in quick_start_screen
    ]._get_start_screen_actions()
  File "/mnt/data/odoo-addons-dir/web_quick_start_screen/wizards/quick_start_screen.py", line 25, in _get_start_screen_actions
    "action_name": action.action_ref_id.id,
AttributeError: 'NoneType' object has no attribute 'id'
```

web_quick_start_screen/README.rst
```diff
@@ -0,0 +1,110 @@
...
+Some users use repeatedly the same actions on and on and sometimes those
+actions are so many clicks away which tends to be frustrating.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+-  Go to *Settings > Technical > User Interface > Screen actions* and
+   create o modify an existing one.
```
pedrobaeza, line 54:
```suggestion
   create or modify an existing one.
```

web_quick_start_screen/README.rst
```diff
@@ -0,0 +1,110 @@
...
+-  Choose the desired action, a name, a description and an icon.
+-  Once save, it will show up in the quick start screen.
+
+Usage
+=====
+
+Once you login, you'll go to the quick actions screen. Just click in and
```
pedrobaeza: This shouldn't be something out of the box for all the users. I think this module should provide an action that can be put as initial action in the users configuration to go to the screen. Other options:

- A security group to include users in.
- A check in the user.

web_quick_start_screen/models/quick_start_screen_action.py
```diff
@@ -0,0 +1,27 @@
...
+    active = fields.Boolean(default=True)
+    sequence = fields.Integer()
+    name = fields.Char()
+    image = fields.Image("Start screen icon", max_width=128, max_height=128, store=True)
+    icon_name = fields.Char(help="Just set the Font Awesome icon name. e.g.: fa-truck")
+    color = fields.Integer(help="Choose the icon color")
+    description = fields.Html()
+    action_ref_id = fields.Reference(
+        selection=[
+            ("ir.actions.report", "ir.actions.report"),
```
pedrobaeza, line 21: Should we put nicer labels?

web_quick_start_screen/wizards/quick_start_screen.py
pedrobaeza: Shouldn't this be an evalued expression?

web_quick_start_screen/models/quick_start_screen_action.py
chienandalu: In this case, I think it's better to reflect for sure what model we're taking the record from

web_quick_start_screen/wizards/quick_start_screen.py
chienandalu: > Warning the field is non mandatory

Sure it should be. Thanks :)

> Shouldn't this be an evalued expression?

Not really, but the key name is a bit missleading, changing it

pedrobaeza: OK, but we talked about some evalued thing (¿the context or the domain?)

chienandalu: Yeah, I was thinking that being able to just run a custom server action there's no need to add such logic as server actions already provide a full fledged code evaluator. There's a simple optianal context, that can be used for manual actions thou.

pedrobaeza: OK, please indicate these hints in the README for others to be able to discover that usefulness.

pedrobaeza:
I'm also thinking about having a parent model for quick start templates, to have several of them, so they can be assigned for different users in the initial action/user preference. What do you think?

chienandalu:
> I'm also thinking about having a parent model for quick start templates, to have several of them, so they can be assigned for different users in the initial action/user preference. What do you think?

I was thinking in that too, but simpler just making the model regular instead of transient. And then the actions be linked in many2many relation

chienandalu:
Refactor done so we can have multiple start screens. I found some limitations with the banner controller, but maybe they aren't important right now

pedrobaeza (approved):
/ocabot merge nobump

PR #2921, opened 2024-08-22
[17.0][FIX] web_m2x_options: Error when installing employee module

CRogos:
When web_m2x_options is installed, and you install the employee module, you cannot open the employee app and get this error message (development mode needs to be active): 

image

The components Many2OneAvatarUserField and KanbanMany2OneAvatarUserField causing this error.

~~I am not sure if this is the best solution, because it also changes the dependency from "web" to "mail".~~
Review feedback highly welcome.

### DIFF
--- a/web_m2x_options/static/src/components/form.esm.js
+++ b/web_m2x_options/static/src/components/form.esm.js
@@ -6,7 +6,11 @@ import {
 } from "@web/views/fields/many2many_tags/many2many_tags_field";
 import {Many2OneField, many2OneField} from "@web/views/fields/many2one/many2one_field";
 import {FormController} from "@web/views/form/form_controller";
-import {Many2OneAvatarField} from "@web/views/fields/many2one_avatar/many2one_avatar_field";
+import {
+    KanbanMany2OneAvatarField,
+    Many2OneAvatarField,
+} from "@web/views/fields/many2one_avatar/many2one_avatar_field";
+
 import {Many2XAutocomplete} from "@web/views/fields/relational_utils";
 import {evaluateBooleanExpr} from "@web/core/py_js/py";
 import {isX2Many} from "@web/views/utils";
@@ -21,6 +25,12 @@ Many2OneField.props = {
 };
 Many2XAutocomplete.props = {
     ...Many2XAutocomplete.props,
+    fieldColor: {type: String, optional: true},
+    fieldColorOptions: {type: Object, optional: true},
+};
+
+KanbanMany2OneAvatarField.props = {
+    ...KanbanMany2OneAvatarField.props,
     fieldColor: {type: String, optional: true},
     fieldColorOptions: {type: Object, optional: true},
 };


CarlosRoca13:
Are you sure the problem is with this module? I've installed it with `hr` locally and haven't had any issue (all without including your changes).

Do you have any extra module installed on your instance?

CRogos:
I could reproduced the issue in runbot, by installing the hr module.

But I just realized this issue only occurs in the development mode! (?debug=1)

http://oca-web-17-0-f4885276a6e4.runboat.odoo-community.org/web?debug=1#menu_id=106&action=160&model=res.partner&view_type=form

web_m2x_options/static/src/components/form.esm.js
```diff
@@ -25,6 +30,18 @@ Many2XAutocomplete.props = {
...
+    fieldColor: {type: String, optional: true},
+    fieldColorOptions: {type: Object, optional: true},
+};
+
+Many2OneAvatarUserField.props = {
+    ...Many2OneAvatarUserField.props,
+    fieldColor: {type: String, optional: true},
+    fieldColorOptions: {type: Object, optional: true},
+};
+
```
CarlosRoca13:
```suggestion
KanbanMany2OneAvatarField.props = {
    ...KanbanMany2OneAvatarField.props,
    fieldColor: {type: String, optional: true},
    fieldColorOptions: {type: Object, optional: true},
};
```

And change this line https://github.com/OCA/web/pull/2921/files#diff-ca954d78dad7b4338381958be4abd93f5fbea46f766a6af5e914a9bd74017737R9 by this 
```javascript
import {Many2OneAvatarField, KanbanMany2OneAvatarField} from "@web/views/fields/many2one_avatar/many2one_avatar_field";

File: web_m2x_options/__manifest__.py [hunk 2]
```diff
@@ -16,7 +16,7 @@
     "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
     "license": "AGPL-3",
-    "depends": ["web"],
+    "depends": ["mail"],
```
CarlosRoca13: Don't change the dependency

web_m2x_options/static/src/components/form.esm.js
```diff
@@ -13,6 +13,11 @@ import {isX2Many} from "@web/views/utils";
 import {patch} from "@web/core/utils/patch";
 import {session} from "@web/session";
 
+import {
+    KanbanMany2OneAvatarUserField,
+    Many2OneAvatarUserField,
+} from "@mail/views/web/fields/many2one_avatar_user_field/many2one_avatar_user_field";
+
```
CarlosRoca13:
```suggestion
```

CarlosRoca13 (changes requested):
Make this changes to avoid changing the dependency

pedrobaeza:
It seems that we will have the same problem with any other custom m2o widget, isn't it? Can't we do anything that don't force to declare that extra options on each widget?

CarlosRoca13:
I think it's not possible to make it more abstract... I have tried different options to avoid that extensions but nothing works... @chienandalu do you have any ideas to avoid them?

chienandalu (commented):
> I think it's not possible to make it more abstract... I have tried different options to avoid that extensions but nothing works... @chienandalu do you have any ideas to avoid them?

What catches my eye is that these components already take their props from the `Many2OneField.props` so it should work with no extension somehow: https://github.com/OCA/OCB/blob/5eb4c06a5d26476cd47cfff0df23d5e05f5318b6/addons/web/static/src/views/fields/many2one_avatar/many2one_avatar_field.js#L16-L18

Did you try tweaking the assets loading order?

CarlosRoca13:
Yes, I tried changing the order of asset loading, but I wasn't able to make the error go away.

chienandalu:
What about using `patch()`: https://github.com/odoo/documentation/blob/17.0/content/developer/reference/frontend/javascript_reference.rst?plain=1#L168-L169

chienandalu:
> What about using `patch()`: https://github.com/odoo/documentation/blob/17.0/content/developer/reference/frontend/javascript_reference.rst?plain=1#L168-L169

An example: https://github.com/OCA/OCB/blob/5eb4c06a5d26476cd47cfff0df23d5e05f5318b6/addons/sms/static/src/components/phone_field/phone_field.js#L8-L21

CarlosRoca13:
Ups :sweat_smile:

CarlosRoca13:
I will try it, I have tried with patch but patching the props directly

CRogos:
Any idea why this only occur in developer mode?

CarlosRoca13:
> Any idea why this only occur in developer

Sometimes non-blocking errors are hidden to users, so they are only visible to developers. However, this doesn't mean they shouldn't be fixed.

CarlosRoca13:
> What about using `patch()`: https://github.com/odoo/documentation/blob/17.0/content/developer/reference/frontend/javascript_reference.rst?plain=1#L168-L169

It does not work either

CRogos:
@CarlosRoca13 @chienandalu can we go with this solution as long we do not have any better way?

CarlosRoca13:
What do you think @chienandalu?

CRogos:
@CarlosRoca13 shall we continue without @chienandalu ?
Maybe this solution is not the best, but it fixes the js error in developer mode.

CarlosRoca13 (approved):
Yes I think it's the only way to achieve this

PR #2926, opened 2024-09-03
[MIG] web_edit_user_filter: Migration to 16.0

amh-mw:


flotho (commented):
Tried on runboat, impossible to see the widget.
image

Nothing special in the console logs

PR #2929, opened 2024-09-05
 [16.0][IMP] web_timeline - for relational group fields use default order in related model, fix for default_group_by attribute no relational

IJOL:


carlos-lopez-tecnativa:
@IJOL, could you please give me an example or use case to test this PR? Thanks.

TT50999

IJOL:
> @IJOL, could you please give me an example or use case to test this PR? Thanks.
> 
> TT50999

if your timeline default_group_by field is a many2* field and the related model has a sequence field it used to order the groups in the timeline left area, in the example in https://github.com/OCA/project/blob/16.0/project_timeline/views/project_project_view.xml if you change default_group_by to project_id which i think you can in that context,  with my change the sequence field in the project  model should be used as order for the groups in the timeline, thus giving the same order in the timeline that projects has

pedrobaeza:
The order should be defined by the model itself with their `_order` clause, being the sequence or whatever, but it shouldn't be hardcoded. The module should rely on the existing order.

IJOL:
this is exactly what the sequence field does in other contexts like list , trees or whatever, if the group is a model wiith no seuqence, or a plain field it does nothing

pedrobaeza:
That's only if the widget="handle" is present AFAIK, but it's not the same context here, as we can't drag and drop for changing such sequence. I insist that we must rely on the model order.

IJOL:
The problem is that Js part reworks the groups, because your are no reading from a projects query that should be ordered you are doingthe groups by the extension it self, and gives up any order i think in the context i use it, and in the example given you order the group as it is stored in the tasks, so no order in projects at all

IJOL:
Just confused the example i use it in project.task context model and i try to group by project_id..

pedrobaeza:
When grouping by project, you should query Odoo to return the ordered recordset including all the existing projects.

IJOL:
> When grouping by project, you should query Odoo to return the ordered recordset including all the existing projects.

Ok, get your point it should be easy to implement that way too, not actually able to remember why i ended in this solution, not so sure i have now the time to do it.. i will push something if i can

pedrobaeza:
Great, let us know. @carlos-lopez-tecnativa FYI

tarteo:
@IJOL are you saying when grouping records, the groups are not ordered by the corresponding model's order? 

> this is exactly what the sequence field does in other contexts like list , trees or whatever, if the group is a model wiith no seuqence, or a plain field it does nothing

Can you send me a link to where Odoo does this in the tree view etc. Does it there look at hard-coded sequence or _order? Thanks in advance

IJOL:
Actually the js code for extension, only gets the group id from data ( in
the examples i put in from tasks and the group_id is the project_id) and if
it a relational field gets the name from the corresponding model, but the
order is the one obtained from the data, the order in which the groups are
named in the data to be exact, mi first proposed change tried to not mess
too much with the Js part and simply try to assume that if the related
model in the group_id has a sequence field  this was the order the group id
model was supposed to do, and do nothing in any other case, just a cheap
way to do what actually the extension do badly, that is the groups had no
order at all, last friday i got convinced by Pedro Baeza to getting the
order from the model itself by doing a simple search of all related model
data and extract the order from there, as i was assuming naively that the
order of the related model comes from a field named sequence in the
model....


Saludos,
--------------------------------------------
Ignacio J.Ortega
skype:ignacioj.ortega

IJOL:
As suggested the group's order is obtained from the related model if the default_group_by is relational ( Many2one or Many2many ), i added a  fix for  other issues  i'm having  when the group field if is not relational, need reviewers though..

@carlos-lopez-tecnativa @tarteo @pedrobaeza

pedrobaeza:
ping @carlos-lopez-tecnativa

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -380,66 +380,87 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
...
+            const groups = [{id: -1, content: _t("<b>UNASSIGNED</b>"), order: -1}];
+            const grouped_field = _.first(group_bys);
+            const fields = await this._rpc({
+                model: this.modelName,
+                method: "fields_get",
+                args: [[grouped_field]],
+                context: this.getSession().user_context,
+            });
+            const fieldType = fields[grouped_field].type;
+            const relation = fields[grouped_field].relation;
```
carlos-lopez-tecnativa: I think this RPC call can be avoided if you use `this.fields`, which already contains metadata for fields (such as type, relation, etc.).
image


```suggestion
            const fieldType = this.fields[grouped_field].type;
            const relation = this.fields[grouped_field].relation;
```

carlos-lopez-tecnativa (changes requested):
Functionally tested, and it works. Just a minor technical change needed for performance reasons.

web_timeline/static/src/js/timeline_renderer.js
IJOL: Thank you, good catch!!

carlos-lopez-tecnativa (approved):
LGTM

carlos-lopez-tecnativa:
ping @pedrobaeza

web_timeline/static/src/js/timeline_controller.esm.js
```diff
@@ -93,15 +93,19 @@ export default AbstractController.extend({
...
+        if (field.relation) {
+            return this.do_action({
+                type: "ir.actions.act_window",
+                res_model: field.relation,
+                res_id: event.data.item.group,
+                target: "new",
+                views: [[false, "form"]],
+            });
+        }
+        console.warn(`Field ${groupField} is not a relational field.`);
```
pedrobaeza, line 106: Remove this

web_timeline/README.rst
```diff
@@ -2,7 +2,7 @@
 Web timeline
 ============
 
-.. 
+..
```
pedrobaeza, line 5: These changes seem to be done by the IDE stripping out leading spaces. They should be removed.

web_timeline/static/description/index.html
```diff
@@ -8,11 +8,10 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
+:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
```
pedrobaeza, line 11: This also shouldn't change.

IJOL: I dont know how that change was done, but it seems legit to me, maybe is the precommit thing..

pedrobaeza: Yeah, because you have a different libraries than the bot, but it will be undone again by the bot. I was trying to avoid noise.

enriquemartin (approved):

enriquemartin:
@IJOL can you rebase, please? Thanks!

PR #2943, opened 2024-10-01
[16.0][FIX] web_field_tooltip: set max width for the tooltip and fixed odoo dark mode

benwillig:
- set max width for the tooltip to prevent weird behaviors where a tooltip could be placed outside the window when its a content was a long text
- fixed popup colors when using odoo dark mode


flotho (approved):
small review , tested on runboat,
LGTM

ThomasBinsfeld (approved):

PR #2944, opened 2024-10-04
[FIX] web_company_color: Colors where not shown correctly

etobella:
Without this change, we have the following top:

Captura desde 2024-10-04 10-18-22

Now we have:

image


luisDIXMIT (approved):
Tested on runbot. LGTM!

pedrobaeza (approved):
/ocabot merge patch

PR #2950, opened 2024-10-09
[15.0] Update copier template

sergio-teruel:
cc @Tecnativa 

ping @chienandalu @pedrobaeza


pedrobaeza (approved):
I don't see any appreciable change (ES6 is for 16+), but it's OK to update it:

/ocabot merge patch

PR #2917, opened 2024-08-14
Widget map first commit

drkpkg:
New widget for OCA web!

I bring you the map field and widget. This widget will help you to add a map (using leaflet) inside the forms.

Xml form:
```xml
...
<field name="map_location" widget="location_map" readonly="0"/>
...
```

```python
from odoo.addons.widget_map.models.map import Map

...
map_location = Map(string="Location", help="Employee location on the map", store=True, default="0.0,0.0")

```

Feel free to suggest, this is my first contribution to OCA and I hope contribute more!

PR #2953, opened 2024-10-11
17.0: Widget map and new map field

drkpkg:
Previous PR: https://github.com/OCA/web/pull/2917

I decided do a refactor for this module and let the previous PR stay in closed and dropped.

This new widget is a combination of the `map` field and the `map` widget. It allows you
to display a map in the form and to select a location by dragging a marker on the map.

I did the pre commit and all was good this time.

The python instance of the field is `fields.Map`.

```python

from odoo import models, fields, api, _

_logger = logging.getLogger(__name__)

class MyModel(models.Model):
    _name = 'my.model'

    name = fields.Char(string='Name')
    location = fields.Map(string=_('Location'))

    @api.onchange('location')
    def _onchange_location(self):
        """
        This will be called when the location is changed.
        """
        if self.location:
            _logger.info('Location: %s', self.location)

```

The xml file should look like this:

```xml
<odoo>
    <data>
        <record id="view_my_model_form" model="ir.ui.view">
            <field name="name">my.model.form</field>
                <form string="My Model">
                    <sheet>
                        <group>
                            <field name="name" />
                            <field name="location" widget="map" />
                        </group>
                    </sheet>
                </form>
            </field>
        </record>
    </data>
</odoo>
```

Note: 

Sorry if i duplicate this PR but I think it is the correct way due the refactor.


legalsylvain (changes requested):
Hi @drkpkg. Thanks a lot for sharing. It looks very nice ! 

Some remarks : 
- blocking point : OCA module still contains leaflet lib. I think we should mutualize. [For exemp](https://github.com/OCA/geospatial/tree/16.0/web_view_leaflet_map/static/lib/leaflet). for exemple, create a specific module named "web_leafet_lib" and make both module depending on the "lib" module. What do you think ?

- Question : why use lib version 1.4.0 ? it looks quite old (the latest version is 1.9.4).

- I think that the name of the field is incorrect. (fields.Map) It doesn't contains a map but a GPS coordinate, AFAIU. So it should be something like fields.Coordinate, (or something similar)


I tried to test it locally. 

### first error

when trying to display a map widget
```
    float(value.split(",")[0])
AttributeError: 'NoneType' object has no attribute 'split'
```

### Second error

(when trying to display a map widget)

UncaughtPromiseError > OwlError

Uncaught Promise > The following error occurred in onWillStart: "The loading of /map_field/static/lib/leaflet/leaflet.js failed"

I set the PR as draft, feel free to put it as ready, when finished. Thanks !

drkpkg:
> Hi @drkpkg. Thanks a lot for sharing. It looks very nice !
> 
> Some remarks :
> 
> * blocking point : OCA module still contains leaflet lib. I think we should mutualize. [For exemp](https://github.com/OCA/geospatial/tree/16.0/web_view_leaflet_map/static/lib/leaflet). for exemple, create a specific module named "web_leafet_lib" and make both module depending on the "lib" module. What do you think ?
> * Question : why use lib version 1.4.0 ? it looks quite old (the latest version is 1.9.4).
> * I think that the name of the field is incorrect. (fields.Map) It doesn't contains a map but a GPS coordinate, AFAIU. So it should be something like fields.Coordinate, (or something similar)

I think create a new contribution should be better. After that we can document about the new module. This will be a good start to prevent break modules due this change.

> 
> I tried to test it locally.
> 
> ### first error
> when trying to display a map widget
> 
> ```
>     float(value.split(",")[0])
> AttributeError: 'NoneType' object has no attribute 'split'
> ```
> 
> ### Second error
> (when trying to display a map widget)
> 
> UncaughtPromiseError > OwlError
> 
> Uncaught Promise > The following error occurred in onWillStart: "The loading of /map_field/static/lib/leaflet/leaflet.js failed"
> 
> I set the PR as draft, feel free to put it as ready, when finished. Thanks !

Yes, I was working on it and I finished when I read this, sorry!

PS: Now leaflet 1.94 is used.

legalsylvain:
Hi @drkpkg. I splitted the a module that contains leaflet lib in two modules here : https://github.com/OCA/geospatial/pull/380
- could you review this PR ? 
- once merged, we can port the module ``web_leaflet_lib`` in V17.0, and make your module depends on the generic module.

Note : the module provide a parameter to define the tile system to be used. (https://github.com/OCA/geospatial/blob/16.0/web_view_leaflet_map/demo/ir_config_parameter.xml#L14)
So you can use it in your module, to avoid to : 
- hardcode the url here : https://github.com/OCA/web/pull/2953/files#diff-69e21aab3b05c2fd8aea9018645ebcc17a3bd36543044cf283b2a738cb3340e0R87
- force users to use a system that can not be legal in case of heavy use of OSM (https://operations.osmfoundation.org/policies/tiles/)
thanks !

drkpkg:
> Hi @drkpkg. I splitted the a module that contains leaflet lib in two modules here : [OCA/geospatial#380](https://github.com/OCA/geospatial/pull/380)
> 
> * could you review this PR ?
> * once merged, we can port the module `web_leaflet_lib` in V17.0, and make your module depends on the generic module.
> 
> Note : the module provide a parameter to define the tile system to be used. (https://github.com/OCA/geospatial/blob/16.0/web_view_leaflet_map/demo/ir_config_parameter.xml#L14) So you can use it in your module, to avoid to :
> 
> * hardcode the url here : https://github.com/OCA/web/pull/2953/files#diff-69e21aab3b05c2fd8aea9018645ebcc17a3bd36543044cf283b2a738cb3340e0R87
> * force users to use a system that can not be legal in case of heavy use of OSM (https://operations.osmfoundation.org/policies/tiles/)
>   thanks !

Dependency updated. Runboat failed, but now the module will have the dependency from web_view_leaflet_map.

legalsylvain:
> Dependency updated. Runboat failed, but now the module will have the dependency from web_view_leaflet_map.

That can not work immediately ! 
- you have to review and approve first the V16 PR (https://github.com/OCA/geospatial/pull/380). 
- wait to be merged
- then we can port it to V17. (trivial should not be take time).
- then your dependency will work.

note : the name is not web_view_leaflet_map but web_leaflet_lib.

thanks !

drkpkg:
> > Dependency updated. Runboat failed, but now the module will have the dependency from web_view_leaflet_map.
> 
> That can not work immediately !
> 
> * you have to review and approve first the V16 PR ([[16][REF] split web_view_leaflet_map into web_view_leaflet_map and web_leaflet_lib geospatial#380](https://github.com/OCA/geospatial/pull/380)).
> * wait to be merged
> * then we can port it to V17. (trivial should not be take time).
> * then your dependency will work.
> 
> note : the name is not web_view_leaflet_map but web_leaflet_lib.
> 
> thanks !

I did a little PR for leaflet https://github.com/OCA/geospatial/pull/381
I think we can start with this, what do you think?

legalsylvain:
Hi @drkpkg. You can review  https://github.com/OCA/geospatial/pull/383 (and then, base your work on this PR ).

thanks !

web_widget_map/static/src/components/widget_map.esm.js
```diff
@@ -0,0 +1,137 @@
...
+        });
+    }
+
+    initializeMap() {
+        const mapOptions = {
+            center: {lat: this.state.location.lat, lng: this.state.location.lng},
+            zoom: 13,
+        };
+        // eslint-disable-next-line no-undef
+        this.map = L.tileLayer("https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png", {
```
legalsylvain, line 87: Hi could you reuse settings introduced by web_leaflet_lib module ? 

See : 

- https://github.com/OCA/geospatial/blob/16.0/web_view_leaflet_map/static/src/js/view/map/map_renderer.js#L15-L16

- https://github.com/OCA/geospatial/blob/16.0/web_view_leaflet_map/static/src/js/view/map/map_renderer.js#L169-L172


values are set here : https://github.com/OCA/geospatial/pull/383/files#diff-5be4d3d480aee272fac8a7719d7470286c43ba57485ca6fc98da0cc374ee8910R11

Thanks !

Issue #2934, opened 2024-09-23
[17.0] web_m2x_options error when opening hr.skill.type


Paulius11:
Module: hr (Human Resources)
Describe the bug

Developer mode needs to be activated.

An error occurs when trying to access the "hr.skill.type" view in Odoo. The error message indicates that the code is trying to access a property ("kanban") of an undefined variable.
To Reproduce

Affected versions: 17.0

    Navigate to the Human Resources module in Odoo.
    Try to access the "Skill Types" view (usually under "Configuration").
    The error should appear.

Expected behavior
The "Skill Types" view should load and display a list of skill types.

```
UncaughtPromiseError > TypeError

Uncaught Promise > Cannot read properties of undefined (reading 'kanban')

TypeError: Cannot read properties of undefined (reading 'kanban')

    at FormController._setSubViewLimit (https://odoo.com/web/assets/04b5dff/web.assets_web.min.js:7605:39)

    at FormController.setup (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:7600:92)

    at FormController.setup (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:12859:446)

    at new ComponentNode (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:1002:202)

    at https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:1540:6

    at View.slot1 (eval at compile (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:1500:374), <anonymous>:13:36)

    at callSlot (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:1109:25)

    at WithSearch.template (eval at compile (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:1500:374), <anonymous>:8:12)

    at Fiber._render (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:940:96)

    at Fiber.render (https://odoo..com/web/assets/04b5dff/web.assets_web.min.js:939:6)
 ```

Paulius11:
Also error is triggered only when developer mode is activated and I go to: Time Off -> Management- Time Off

`UncaughtPromiseError > OwlError
Uncaught Promise > Invalid props for component 'ListMany2ManyTagsAvatarField': unknown key 'fieldColor', unknown key 'fieldColorOptions'
[...]

PR #2957, opened 2024-10-14
[17.0][FIX] web_m2x_options: add fieldColor to ListMany2ManyTagsAvatarField and Many2ManyTagsAvatarField

JordiMForgeFlow:
Fix for issue https://github.com/OCA/web/issues/2934

Based on similar fixes in https://github.com/OCA/web/pull/2921

### DIFF
--- a/web_m2x_options/static/src/components/form.esm.js
+++ b/web_m2x_options/static/src/components/form.esm.js
@@ -10,6 +10,10 @@ import {
     KanbanMany2OneAvatarField,
     Many2OneAvatarField,
 } from "@web/views/fields/many2one_avatar/many2one_avatar_field";
+import {
+    ListMany2ManyTagsAvatarField,
+    Many2ManyTagsAvatarField,
+} from "@web/views/fields/many2many_tags_avatar/many2many_tags_avatar_field";
 
 import {Many2XAutocomplete} from "@web/views/fields/relational_utils";
 import {evaluateBooleanExpr} from "@web/core/py_js/py";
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@CarlosRoca13 I think we have faced this before, isn't it?

pedrobaeza:
I mean this one: https://github.com/OCA/web/pull/2921 and we avoid to add the `mail` dependency. cc @chienandalu @CRogos

JordiMForgeFlow:
@pedrobaeza yes, I found the correction you did and applied the same mechanism without having to depend on mail.

This is now fixing https://github.com/OCA/web/issues/2934

pedrobaeza (approved):

chienandalu (approved):
Thanks :)

web_m2x_options/static/src/components/form.esm.js
```diff
@@ -56,6 +60,18 @@ Many2ManyTagsFieldColorEditable.props = {
     fieldColorOptions: {type: Object, optional: true},
 };
 
+ListMany2ManyTagsAvatarField.props = {
+    ...ListMany2ManyTagsAvatarField.props,
+    fieldColor: {type: String, optional: true},
+    fieldColorOptions: {type: Object, optional: true},
+};
```
CRogos, line 67: Question: is ListMany2ManyTagsAvatarField necessary, because it derives from Many2ManyTagsAvatarField:
export class ListMany2ManyTagsAvatarField extends Many2ManyTagsAvatarField

CRogos (commented):
#2921 was only addressing Many2OneAvatarField while this patch addresses ListMany2ManyTagsAvatarField and Many2ManyTagsAvatarField with the same solution.

web_m2x_options/static/src/components/form.esm.js
JordiMForgeFlow: That is a good point, actually I fixed first the ListMany2ManyTagsAvatarField and then noticed that I was still getting the issue with Many2ManyTagsAvatarField so I applied directly the same fix.

I will check by just fixing the latter to make sure we do not add unnecessary code and open a new PR. Thanks for noticing!

PR #2959, opened 2024-10-15
[17.0][FIX] web_m2x_options: add fieldColor to KanbanMany2ManyTagsAvatarField

JordiMForgeFlow:
Fixes new issued identified in https://github.com/OCA/web/issues/2934

Based on same fixes applied on https://github.com/OCA/web/pull/2921

CC @CRogos @chienandalu @pedrobaeza

### DIFF
--- a/web_m2x_options/static/src/components/form.esm.js
+++ b/web_m2x_options/static/src/components/form.esm.js
@@ -21,55 +21,52 @@ import {isX2Many} from "@web/views/utils";
 import {patch} from "@web/core/utils/patch";
 import {session} from "@web/session";
 
-Many2OneField.props = {
-    ...Many2OneField.props,
-    noSearchMore: {type: Boolean, optional: true},
-    fieldColor: {type: String, optional: true},
-    fieldColorOptions: {type: Object, optional: true},
-};
-Many2XAutocomplete.props = {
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
I wonder if there's a generic way to do that, as I already stated before.

chienandalu (commented):
Maybe we could avoid some boilerplate code using a common object. Somethin like:

```js
const fieldColorProps = {
    fieldColor: {type: String, optional: true},
}

//...

WhatEverField.props = {
    ...WhatEverField.props,
    ...fieldColorProps,
};

JordiMForgeFlow:
> Maybe we could avoid some boilerplate code using a common object. Somethin like:
> 
> ```js
> const fieldColorProps = {
>     fieldColor: {type: String, optional: true},
>     fieldColorOptions: {type: Object, optional: true},
> }
> 
> //...
> 
> WhatEverField.props = {
>     ...WhatEverField.props,
>     ...fieldColorProps,
> };
> ```

Makes sense, changes applied! :)

chienandalu (approved):
:+1:

pedrobaeza:
This is still requiring to inherit each widget. I meant a way to do it on the "base", but anyway, let's merge meanwhile:

/ocabot merge patch

PR #2964, opened 2024-10-16
[16.0][FIX] web_advanced_search: Fixed changes from code editor not propagated

chienandalu:
Propagate changes in the debug mode domain editor to the domain itself.

cc @Tecnativa TT51315

please review @CarlosRoca13 @carlosdauden

### DIFF
--- a/web_advanced_search/static/src/js/DomainSelector.esm.js
+++ b/web_advanced_search/static/src/js/DomainSelector.esm.js
@@ -1,0 +1,57 @@
+/** @odoo-module **/
+
+import {patch} from "@web/core/utils/patch";
+import {_t} from "web.core";
+import Domain from "web.Domain";
+import DomainSelector from "web.DomainSelector";
+import basic_fields from "web.basic_fields";
+/**
+ * The redraw in the Debug Field does not trigger correctly
+ * so we overwrite it with the v14 Version
+ *
+ */
+patch(DomainSelector.prototype, "web.DomainSelector", {
+    /**
+     * @override
+     */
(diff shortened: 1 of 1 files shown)


CarlosRoca13 (approved):
Code review :+1:

web_advanced_search/static/src/js/DomainSelector.esm.js
```diff
@@ -0,0 +1,57 @@
...
+            return this._super(...arguments);
+        }
+        const rawDomain = e.currentTarget.value;
+        try {
+            Domain.prototype.stringToArray(rawDomain);
+        } catch (err) {
+            // If there is a syntax error, just ignore the change
+            this.displayNotification({
+                title: _t("Syntax error"),
+                message: _t("Domain not properly formed"),
```
pedrobaeza, line 28: For this to be exported for translation, we need to override http.py, isn't it?

chienandalu: That's for frontend

pedrobaeza: OK then

pedrobaeza (approved):
/ocabot merge patch

carlosdauden (commented):
Ok to this PR.
Only comment that the "add new value" option does not work correctly, I understand that this problem is different from the one solved by this PR and that functionally it is not important because the only way to use the "in" operator is using the debug mode.

But as I have detected it in my functional review, I leave it commented.

image

PR #2970, opened 2024-10-22
18.0 [ADD] web_hierarchy_list

IT-Ideas:
This PR is about adding a **readonly** hierarchy  view in odoo based on the used of the existing list view and by adding an embedded breadcrumb representing the parents path during the exploration.

Remaining Improvements to be done in a later roadmap:
- [ ] Prevent fetching all child_ids by adding the field name option in the parent field and performing a read_group for the ids that are actually rendered.
- [ ] Implement CRUD in one2many widget.

image

image


pedrobaeza:
Any illustrating screenshot?

IT-Ideas:
> Any illustrating screenshot?

Here you go!

IT-Ideas:
Hey @aab-odoo, @ged-odoo 

If you would have a minute for me and have a look at the concept I would be very happy to get your feedback.
Proceeding this way (with an embedded breadcrumb) removes the issues we had foreseen regarding the Filters and Search. Indeed this way you always get the result corresponding to your criteria and then you can start your drill downs (keeping you path at sight)

IT-Ideas:
@lmignon, @sbidoul,

Here is already the very very draft PR 🙂

web_hierarchy_list/static/src/hierarchy_list_breadcrumb.esm.js
```diff
@@ -0,0 +1,15 @@
+import {Component} from "@odoo/owl";
+import {HierarchyListBreadcrumbItem} from "./hierarchy_list_breadcrumb_item.esm";
+
+export class HierarchyListBreadcrumb extends Component {
+    static components = {
+        HierarchyListBreadcrumbItem,
+    };
+    static props = {
+        parentRecords: Array[Object],
```
ged-odoo: i am pretty sure `Array[Object]` is just undefined. You probably mean `{ type: Array, element: Object}` (too bad, Array[Object] would be cool)

web_hierarchy_list/static/src/hierarchy_list_renderer.esm.js
```diff
@@ -0,0 +1,108 @@
...
+                    breadcrumbState: this._getBreadcrumbState(),
+                };
+            },
+        });
+        // As the breadcrumb state is not provided when the renderer is mounted into a x2Many
+        // field, we need to have a fallback value.
+        this.state = useState({
+            breadcrumbState: this.env.breadcrumbState || [],
+        });
+        onWillStart(this.willStart);
```
ged-odoo, line 40: haha, i was going to comment that this does not work, because this.willStart is unbound, but it actually does, because owl is calling it with the component as `this`. I had forgotten about that

web_hierarchy_list/static/src/hierarchy_list_x2many_field.esm.js
```diff
@@ -0,0 +1,78 @@
...
+            this.archInfo,
+            this.props.record.fields[this.props.name].relation,
+            this.archInfo.fields
+        );
+        this.state = useState({
+            parent: false,
+        });
+        const services = {};
+        for (const key of RelationalModel.services) {
+            services[key] = useService(key);
```
ged-odoo, line 35: i wish that we could just get the services directly from the model itself. Technically, it could work, because the new RelationalModel is done in the setup, so we can actually call useService in the model constructor, but it feels wrong, which is why we kind of have to do this.  still feels awkward

ged-odoo:
just had a quick look. x2manys are hard, so it's quite impressive you managed to do that, good job. From a technical point of view, looks very good. Nothing really stands out as a problem, code seems well organized.

Functionally, it makes sense to me, but i am not an expert. (also, note that odoo will likely work on a tree view in the future, not sure how much it will overlap this, but i guess it's good to know)

IT-Ideas:
> just had a quick look. x2manys are hard, so it's quite impressive you managed to do that, good job. From a technical point of view, looks very good. Nothing really stands out as a problem, code seems well organized.
> 
> Functionally, it makes sense to me, but i am not an expert. (also, note that odoo will likely work on a tree view in the future, not sure how much it will overlap this, but i guess it's good to know)

Hey @ged-odoo thanks so much for having taken the time to have a look 🙂. I owe you some beers 😉 I'll pass by the farm soon to hang out with the Vidange / Services team. I did not get too much time to share with you guys at the OXP.

sbidoul:
:medal_sports: @IT-Ideas  :)

flotho:
Hi @IT-Ideas , 
any chance to rebase ?

IT-Ideas:
> Hi @IT-Ideas , 
> 
> any chance to rebase ?

Hello 🙂!
Wow that is a while ago! I'll try yes.

PR #2971, opened 2024-10-23
[16.0][IMP] web_notify: close everywhere

BT-cjimeno:
If a user gets a notification with sticky: True and has multiple tabs open, it needs to be closed manually in each of the tabs.
With these changes, closing it from any tab should also close it in the rest of the tabs.


flotho (approved):
Hi,
Thanks, I was expecting for this one !

xaviedoanhduy (approved):

PR #2378, opened 2023-01-02
[MIG] web_timeline: Migration to 16.0

bvanjeelharia:

PR #2972, opened 2024-10-23
[17.0][IMP] web_timeline: Update vis-timeline-graph2d lib: 7.7.0 -> 7.7.3

carlos-lopez-tecnativa:
no breakchanges, see https://github.com/visjs/vis-timeline/releases/tag/v7.7.3
TT50999
@Tecnativa @pedrobaeza @CarlosRoca13 @chienandalu  could you please review this


CarlosRoca13 (approved):

chienandalu (commented):
isn't the diff much larger than needed? I don't see that many changes in the related release notes :thinking: for example, looks like css stuff is just moved from one place to another in the css file...

web_timeline/static/lib/vis-timeline/vis-timeline-graph2d.css
```diff
@@ -889,6 +889,21 @@
   border-style: solid;
   border-radius: 4px;
 }
+.vis .overlay {
```
pedrobaeza, line 892: Why increasing the diff moving the styles ?

pedrobaeza:
Yes, and the library JS seems also very different. Maybe one version is minified and the other not?

carlos-lopez-tecnativa:
I only copied the files generated by executing the `npm run build` command. The result is similar to previous updates, as seen in the commit.

https://github.com/OCA/web/pull/2378/commits/94323daa0cb913e5862c7f7cbb92a63c24db095d

pedrobaeza:
Well, it's the double in fact (48k additions instead of 24k). Not sure if there can be any command line for changing somehow the result.

carlos-lopez-tecnativa:
> Well, it's the double in fact. Not sure if there can be any command line for changing somehow the result.

I'm not sure, but if @chienandalu or @CarlosRoca13 can help me with the command used in the previous case, that would be great. The difference is mainly in the indentation.

image

chienandalu:
> The difference is mainly in the indentation.

That's not hard to deal with. Just select the whole block and indent it to the right :)

carlos-lopez-tecnativa:
> > The difference is mainly in the indentation.
> 
> That's not hard to deal with. Just select the whole block and indent it to the right :)

Yes, but I meant that the current spacing is 2
image

and now it's 4. I think the new one is better, although it shows more differences. In the next updates, the differences should be smaller.
image

chienandalu:
> Yes, but I meant that the current spacing is 2

Ah, ok. In that case I think is ok :) I just saw your first capture where all the code was pulled to the left

pedrobaeza:
Can't you configure a different indentation or open it and change it for minimizing the diff?

carlos-lopez-tecnativa:
> Can't you configure a different indentation or open it and change it for minimizing the diff?

I took the file directly from https://unpkg.com/vis-timeline@7.7.3/standalone/umd/vis-timeline-graph2d.js
If I manually adapt the code, the result will be very different again in the next update.

pedrobaeza:
Are we sure the next time it will be indented the same way?

pedrobaeza (approved):
As the diff seems to not be avoidable, let's merge it:

/ocabot merge patch

PR #2978, opened 2024-11-04
[17.0][FIX] web_responsive: Fix test related to ir.actions.act_window

victoralmau:
Related to https://github.com/odoo/odoo/commit/cac20c5f2f0da7be6a75bd1281d528df7f34bfe6

Now it is necessary to have an `ir.actions.act_window` record

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa

### DIFF
--- a/web_responsive/tests/test_res_users.py
+++ b/web_responsive/tests/test_res_users.py
@@ -1,32 +1,23 @@
 # Copyright 2023 Taras Shabaranskyi
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+from odoo.tests import new_test_user
 
-from odoo.tests.common import TransactionCase
+from odoo.addons.base.tests.common import BaseCommon
 
 
-class TestResUsers(TransactionCase):
+class TestResUsers(BaseCommon):
     def test_compute_redirect_home(self):
-        record = self.env["res.users"].create(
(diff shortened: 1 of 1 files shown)


web_responsive/tests/test_res_users.py
```diff
@@ -1,32 +1,27 @@
...
-            {"name": "Test Action", "type": "ir.actions.act_window"}
+        action = self.env["ir.actions.act_window"].create(
+            {
+                "name": "Test Action",
+                "type": "ir.actions.act_window",
+                "res_model": record._name,
+            }
         )
-        record._compute_redirect_home()
+        record.action_id = self.env["ir.actions.actions"].browse(action.id)
```
pedrobaeza: Is there any reason to rebrowse?

```suggestion
        record.action_id = action.id
```

chienandalu: According to Odoo's reference PR...

image

pedrobaeza: OK, then the same kind of "comment" should be needed here for future reference.

victoralmau: Changed, thank you. Yes, it is possible to do it without rebrowse (although it may be confusing).

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2979, opened 2024-11-05
[18.0][MIG] web_ir_actions_close_wizard_refresh_view: Migration to 18.0

IT-Ideas:


IT-Ideas:
Does not seem to work

StefanRijnhart:
Seems like a soft refresh is the default action after calling a wizard model method in the UI without a return value. If you need this behaviour in other cases, you can try `{'type': 'ir.actions.client', 'tag': 'soft_reload'}` (see https://github.com/odoo/odoo/pull/133643)

PR #2980, opened 2024-11-07
[16.0] web_timeline - implement attribute group_edit to control group updates

IJOL:
Added attribute group_edit to control group changing for events,  defaults true


IJOL:
@carlos-lopez-tecnativa can you take a look please

carlos-lopez-tecnativa (changes requested):
Please ensure the new attribute is documented in the [CONFIGURE.rst](https://github.com/OCA/web/blob/16.0/web_timeline/readme/CONFIGURE.rst) file. This will help users understand its purpose and usage more clearly.

PR #2981, opened 2024-11-07
[16.0][IMP] web_timeline: implementing readonly_field attribute: a boolean field to mark the event as not editable

IJOL:
Marks events as non editable following a field in the model

### DIFF
--- a/web_timeline/static/src/js/timeline_renderer.js
+++ b/web_timeline/static/src/js/timeline_renderer.js
@@ -37,6 +37,8 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
             this.date_stop = params.date_stop;
             this.date_delay = params.date_delay;
             this.colors = params.colors;
+            this.readonly_field = params.readonly_field;
+
             this.fieldNames = params.fieldNames;
             this.default_group_by = params.default_group_by;
             this.dependency_arrow = params.dependency_arrow;
@@ -537,6 +539,9 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
             // ➔ Instantaneous events / those with inverted dates are displayed as points.
             if (date_stop && moment(date_start).isBefore(date_stop)) {
                 r.end = date_stop;
+            }
+            if (this.readonly_field && evt[this.readonly_field]) {
+                r.editable = false;
             }
             this.color = null;
             return r;
--- a/web_timeline/static/src/js/timeline_view.js
+++ b/web_timeline/static/src/js/timeline_view.js
@@ -75,6 +75,10 @@ odoo.define("web_timeline.TimelineView", function (require) {
                 (item) => item.attrs.name
             );
             fieldNames = _.union(fieldNames, archFieldNames);
+            if (attrs.readonly_field) {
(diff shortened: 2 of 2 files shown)


IJOL:
@carlos-lopez-tecnativa and this one too, thanks in advance

carlos-lopez-tecnativa (commented):
I'm not sure I fully understand the intended use case. Could you provide an example?

I tested it, but I'm not certain it's working as expected. I added readonly_field="active" to ir.cron and refreshed the view. However, when I double-click on a record, an empty popup opens without loading the record’s information.

Would it be better to use edit="0" or edit="expression" instead of adding a new option? What do you think? Otherwise, please add the new option to the README.
image

image

IJOL:
You are right i see the same, open a blank form, looking into this RN, the use you propose of the new attribute is perfectly valid, and that is the intended use case be able to put readonly bars in the timeline, i will see if can solve, at least not present a blank form, can you think of a way to show a readonly view of the form? if not the only solution now is to not show the popup

IJOL:
Btw nice use of the timeline... i will add it to my rather long todo list

carlos-lopez-tecnativa:
@IJOL I believe these changes are covered in this PR. You can add `edit="0"`, or expand it to support expressions if needed.
https://github.com/OCA/web/pull/2791

IJOL:
The issue is that you can control only group change ( lane change ) without forbid editing, i need to control group but not the data itself, and thats is because i use synthetic non relational groups in my use case where is appropiate to edit the event linked but not to change the group, makes only the group readonly essentially

IJOL:
And my use case is related to listing rentals where the group is the name of the product, you can change the dates but not the product in the line, and for those use cases I need too the readonly_field attribute, because i need lines in confirmed sale orders to be readonly..

IJOL:
The blank form is correct, is an add operation, that can be controlled with the create attribute, so my change is not the culprit is simply working correctly :)

pedrobaeza:
Shouldn't that be done with a special attribute like in kanban for not allowing drag and drop between groups? And for the form part, define a special form view to view used when opening the record, with the readonly fields you need. On one2many widget, that can be done passing through context `tree_view_ref` or `form_view_ref`.

IJOL:
> Shouldn't that be done with a special attribute like in kanban for not allowing drag and drop between groups? And for 

this is exactly what i'm proposing a new attribute "edit_group" to control exactly this, i'm open to any name change that fits better

IJOL:
this last comment was wrong, the pr where i propose the edit_group is another, sorry bout that

pedrobaeza:
Check how it's called in kanban, and you can use the same for homogeneity.

IJOL:
We are talking here of https://github.com/OCA/web/pull/2980, I had selected the opposite naming :100:  for me, i will change it to fit, good catch..

carlos-lopez-tecnativa (changes requested):
Please ensure the new attribute is documented in the [CONFIGURE.rst](https://github.com/OCA/web/blob/16.0/web_timeline/readme/CONFIGURE.rst) file. This will help users understand its purpose and usage more clearly.

PR #2983, opened 2024-11-08
[16.0][FIX] web_advanced_search: display label name in partners search

yajo:
Without this patch, when searching in the `res.partner` view for 'Tags is equal to "Something"', the search bar would display "Tags is equal to "123"' (the tag ID instead of its display name).

Before:
image

After:
image

@moduon MT-7870


chienandalu (approved):
Nice :+1:

edlopen (approved):
LGTM. Also I've tested it in runboat. Such a nice feature @yajo !!

PR #2989, opened 2024-11-13
[16.0][ADD] web_widget_progressbar_gradient

rousseldenis:
Rendering example:

image


legalsylvain (approved):
Very Nice !
Thanks !

web_widget_progressbar_color/static/src/xml/progressbar.xml
```diff
@@ -0,0 +1,20 @@
...
+    <t
+        t-name="web_widget_progressbar_color.ProgressBarFieldColor"
+        t-inherit="web.ProgressBarField"
+        t-inherit-mode="primary"
+        owl="1"
+    >
+        <xpath expr="//div[@t-attf-class]" position="attributes">
+            <attribute
+                name="t-attf-class"
+                add="{{ (state.currentValue > 40 and state.currentValue &lt;= 70) ? 'o_progress_warning' : state.currentValue > 70 ?  'o_progress_danger' : ''}}"
```
jbaudoux:
```suggestion
                add="{{ (state.currentValue > 40 * state.maxValue/100 and state.currentValue &lt;= 70 * state.maxValue/100) ? 'o_progress_warning' : state.currentValue > 70 * state.maxValue/100 ?  'o_progress_danger' : ''}}"
```

jbaudoux (approved):
Nice.

jbaudoux (commented):
Why not making a gradient progressbar instead ? Then you don't have to configure the thresholds

rousseldenis:
> Why not making a gradient progressbar instead ? Then you don't have to configure the thresholds

Could you develop ?

jbaudoux:
> > Why not making a gradient progressbar instead ? Then you don't have to configure the thresholds
> 
> Could you develop ?

Something like applying always this class:
```
.o_progressbar .o_progress  o_progressbar_gradient {
  background: linear-gradient(to right, green, yellow, red)
}
```

rousseldenis:
> > > Why not making a gradient progressbar instead ? Then you don't have to configure the thresholds
> > 
> > 
> > Could you develop ?
> 
> Something like applying always this class:
> 
> ```
> .o_progressbar .o_progress  o_progressbar_gradient {
>   background: linear-gradient(to right, green, yellow, red)
> }
> ```

The result:

image

This is not the expecting result

rousseldenis:
@legalsylvain @pedrobaeza @jbaudoux 

I will change the imlpementaiton to this (need a little bit tweaking) as this is our needs:

image

rousseldenis:
I think also changing the module name to `web_widget_progressbar_gradient` to reflect more the final implementation.

legalsylvain:
Hi @rousseldenis : looks nice ! 

A "matter of taste" remarks : 
- Generally, red is "bad / problematic", yellow is not "very good", and green is "good / allowed". for traffic lights, for example.
- Generally also, 0% means an unfulfilled objective, and 100% means a successful target
So, my proposal is to have red -> Yellow -> green. (instead of green -> yellow -> green).

However, just a suggestion. If you think it's better like that, let's move on.

rousseldenis:
> * So, my proposal is to have red -> Yellow -> green. (instead of green -> yellow -> green).

The current implementation is to reflect the "occupation" which means that the less is better, the more is worse.

I'll try to add an option to the widget to allow the `inverse`mode (in fact, mine :sweat_smile: )

rousseldenis:
@legalsylvain 

Now:

image


So both options are supported. The normal rendering is red to green. Inverse mode, green to red.

legalsylvain:
Thanks ! 

It looks like there is a little offset when 100 is displayed.

rousseldenis:
> It looks like there is a little offset when 100 is displayed.

That's also the case for the standard progressbar widget :sweat_smile:

legalsylvain:
> That's also the case for the standard progressbar widget 😅

;-)

I think we can merge it. Agree ? Improvment can be done in other PRs.

rousseldenis:
> Agree ?

Yes. Thanks !

web_widget_progressbar_gradient/static/src/scss/progressbar.scss
```diff
@@ -0,0 +1,30 @@
...
+    mask: linear-gradient(#fff 0 0);
+}
+
+.o_progressbar .o_progress .o_progressbar_gradient.o_inverse::before {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
```
jbaudoux: I think you can drop this. Maybe putting this section after the other would improve readability

rousseldenis: What do you mean by `after the other` ?

apineux (approved):

web_widget_progressbar_gradient/static/src/scss/progressbar.scss
jbaudoux: the other css block just below that repeats the same

web_widget_progressbar_gradient/static/src/scss/progressbar.scss
```diff
@@ -0,0 +1,30 @@
...
+
+.o_progressbar .o_progress .o_progressbar_gradient::before {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    background-image: linear-gradient(to right, #dc3545, #ffc107, #198754);
+}
```
jbaudoux: Like this:
```suggestion
.o_progressbar .o_progress .o_progressbar_gradient::before {
    content: "";
    position: absolute;
    top: 0;
    left: 0;
    right: 0;
    bottom: 0;
    background-image: linear-gradient(to right, #dc3545, #ffc107, #198754);
}

.o_progressbar .o_progress .o_progressbar_gradient.o_inverse::before {
    background-image: linear-gradient(to right, #198754, #ffc107, #dc3545);
}
```

rousseldenis: I will try but I had problems applying the inverse options in that case

FrancoMaxime (approved):
LGTM: code review

web_widget_progressbar_gradient/static/src/scss/progressbar.scss
rousseldenis: @jbaudoux Thanks. Working well.

PR #2996, opened 2024-11-21
[FIX] add many2ManyTagsAvatarUserField and kanbanMany2ManyTagsAvatarU…

elhamdaoui:
Add color props to many2ManyTagsAvatarUserField and kanbanMany2ManyTagsAvatarUserField Components from the 'mail' module. This resolves the error occurring after installing the project module: 'OwlError: Invalid props for component 'KanbanMany2ManyTagsAvatarUserField': unknown key 'fieldColor' and unknown key 'fieldColorOptions'.


AlvaroRM11:
Thanks for your PR. I think this is more an [IMP] than a [FIX], so you should consider to change the title. Also, you're missing the pre-commit stuff rightnow.

AlvaroRM11:
Thanks for your PR, you're missing the pre-commit stuff right now

PR #3005, opened 2024-11-27
[18.0][MIG] web_pivot_computed_measure: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT51903

ping @chienandalu @carlos-lopez-tecnativa


test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-test-helper
```
chienandalu, line 1: Maybe this should go in a commit appart from the migration one

chienandalu (approved):
Good work! Tested and looking good. Just a tiny doubt

test-requirements.txt
CarlosRoca13: @pedrobaeza Can you show us the way?

pedrobaeza: There's no problem in going on the same one, as in future migrations, the filter directory git command strips it out, and in fact it's somehow part of the migration. The pity is that this is not automatizable.

carlos-lopez-tecnativa (changes requested):
I tested the module, but I believe something isn’t working correctly. When grouping by a Many2one field (e.g., by Customer or Salesperson) in the columns:

Without a computed measure, the values are displayed correctly:
image

However, with a computed measure (e.g., Total - Taxes), the values appear empty:
image

Could you please check this?

CarlosRoca13:
@carlos-lopez-tecnativa thanks 4 the review :) It should be solved right now

carlos-lopez-tecnativa (approved):
LGTM

PR #3006, opened 2024-12-02
[16.0][IMP] web_timeline - support multiple group-by levels and allow to group by m2m field

cuongnmtm:
Allowing to add multiple group-by levels

image

Test cases:

- [x] group by one field (group by m2o, selection, integer, m2m, date, ...)
- [x] group by multiple fields
- [x] group by 0 field: back to default group by
- [x] create a new record
- [x] drag and drop record to update
- [x] click on the group by to view the group by record detail
- [x] when group by m2m field, record is displayed multiple times if linked to multiple m2m records
- [x] when drag and drop, the value of the m2m field should not be updated as there is no best way to handle this
- [x] when creating a new record, set the default value for all grouped fields
- [x] handle group by non-relation fields: click on the group of selection or char field do not popup record form view to avoid error.
- [x] dependency_arrow between multiple collapse/expand groups

### DIFF
--- a/web_timeline/static/src/js/timeline_renderer.js
+++ b/web_timeline/static/src/js/timeline_renderer.js
@@ -342,7 +342,6 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
                         event
                     )
                 );
-                this.fieldsGet = await this.get_fields_get(group_bys);
                 return this.on_data_loaded_2(nevents, group_bys, adjust_window);
             });
         },
@@ -369,25 +368,36 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
                 }
             }
             this.split_groups(events, group_bys).then((groups) => {
+                // This.groups contains all unique group objects
                 this.groups = groups;
+                // Ensure relevant groups are visible
                 for (const d of data) {
-                    // Check if the group should be visible
-                    // If d.group is 'partner_id-35/category_id-4/city_id-false'
-                    // it means there are 3 groups to check
-                    // partner_id-35, partner_id-35/category_id-4, partner_id-35/category_id-4/city_id-false
-                    const groupParts = d.group.split("/");
-                    let groupPath = "";
-                    const groupsToCheck = groupParts.map((part, index) => {
-                        groupPath = index === 0 ? part : `${groupPath}/${part}`;
-                        return groupPath;
-                    });
-                    for (const gtc of groupsToCheck) {
-                        if (gtc.endsWith("-false")) {
-                            const group = groups.find((g) => g.id === gtc);
-                            if (group) {
+                    // D is a vis.js item, d.group is a single JSON string path
+                    const itemGroupPathJson = d.group;
+                    try {
+                        // Parse the item's full group path
+                        const itemPathSegments = JSON.parse(itemGroupPathJson);
+
+                        // Iterate through all prefixes of this item's group path (e.g., [seg1], [seg1, seg2], ...)
+                        for (let i = 0; i < itemPathSegments.length; i++) {
+                            const subPathJson = JSON.stringify(
+                                itemPathSegments.slice(0, i + 1)
+                            );
+                            // Find the actual group object corresponding to this sub-path
+                            const group = groups.find((g) => g.id === subPathJson);
+                            if (group && !group.visible) {
+                                // If a group in the item's hierarchy was initially not visible (e.g., an "UNASSIGNED" group),
+                                // make it visible because an item actually belongs to this path.
                                 group.visible = true;
                             }
                         }
+                    } catch (e) {
+                        console.error(
+                            "Error processing group visibility for item. Group JSON string was:",
+                            itemGroupPathJson,
+                            "Error:",
+                            e
+                        );
                     }
                 }
                 this.timeline.setGroups(groups);
@@ -410,7 +420,13 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
          */
         split_groups: async function (events, group_bys) {
             if (group_bys.length === 0) {
-                return events;
+                // No grouping, but vis.js expects groups if items have a 'group' property.
+                // Create a single default group if items might have group properties.
+                // However, event_data_transform might assign 'undefined-false' if no groups.
+                // For safety, ensure at least one group if group_bys is empty but items might specify groups.
+                // This case should ideally be handled by ensuring items don't have group props if no group_bys.
+                // For now, returning an empty array, assuming items won't have group property.
+                return [];
             }
             const groups = [];
             let seq = 1;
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Can you please put an screenshot of the feature?

cuongnmtm:
@pedrobaeza I have included a screenshot of the feature in the PR description.

pedrobaeza:
Great improvement!

tarteo (commented):
Awesome! 
I get an error when I quickly add and remove a group (by double clicking). However if I click normally I don't get an error.  There's probably a race condition somewhere

Error (I have assets debugging enabled):
`
UncaughtPromiseError > TypeError
Uncaught Promise > Cannot read properties of undefined (reading 'relation')
TypeError: Cannot read properties of undefined (reading 'relation')
    at Class.split_groups (http://localhost:8069/web/assets/debug/web.assets_backend.js:190546:71) (/web_timeline/static/src/js/timeline_renderer.js:509)
`

cuongnmtm:
@tarteo I cannot reproduce your issue, but I found another related defect and fixed it (error when clicking on the group when group by non-relation field). Could you please help to check again? Thanks.

And other things had been done (inthe  description).

cuongnmtm:
For non-technical who wish to test this feature, you can use the XML view arch below to create a timeline view for the project model and modify the project window action to include the timeline view mode for testing.

```xml
<timeline date_start="date_start" date_stop="date" string="Projects" default_group_by="user_id,partner_id,privacy_visibility" mode="week" event_open_popup="true" stack="true">
    <field name="user_id"/>
    <field name="privacy_visibility"/>
    <field name="tag_ids"/>
    <field name="name"/>
    <templates>
        <t t-name="timeline-item">
            <div t-att-title="record.name">
                <small>
                    <div>
                      <span>
                          <t t-esc="record.name"/>
                      </span>
                    </div>
                    <div>
                      PM:
                      <span>
                          <t t-esc="record.user_id[1]"/>
                      </span>
                    </div>
                  
                </small>
            </div>
        </t>
    </templates>
</timeline>
```

pedrobaeza:
We added on the scheduled actions the timeline view as demo data for demonstration purposes without requiring extra modules. Maybe you can modify that demo view to showcase this feature.

cuongnmtm:
> We added on the scheduled actions the timeline view as demo data for demonstration purposes without requiring extra modules. Maybe you can modify that demo view to showcase this feature.

I missed that. Thanks.

antoniodavid:
Awesome @cuongnmtm. It would be great to update the version of the addon.

pedrobaeza:
> Awesome @cuongnmtm. It would be great to update the version of the addon.

No, that's done on merge by the bot, when we issue the command.

minhthie (commented):
Functional test: LGTM

hitrosol:
This is nice new feature. 
One input from me:
Every time I click the arrow on the left or any field there, it will automatically show the popup form related to the field. If it's related to partner, then it will open the pop up of the partner, if it's related to the stage, then it will open the pop up form in of the Stage. Is it possible to disable the behaviour, because it's bit annoying to close the popup every time user click the arrow. Thanks.

cuongnmtm:
Hi @hitrosol ,

I will check if it is easy to do (if the arrow and the label are in the same element, it will be annoying to implement the change as we need to split them).

cuongnmtm:
BTW, I think a toggle for the stack option is nice to have as well. Right now, the timeline needs to be either stack or no-stack.

cuongnmtm:
> This is nice new feature. One input from me: Every time I click the arrow on the left or any field there, it will automatically show the popup form related to the field. If it's related to partner, then it will open the pop up of the partner, if it's related to the stage, then it will open the pop up form in of the Stage. Is it possible to disable the behaviour, because it's bit annoying to close the popup every time user click the arrow. Thanks.

@hitrosol I addressed this issue by changing the event trigger from a single "click" to a "double-click", thereby enhancing the user experience slightly.

web_timeline/static/src/js/timeline_view.js
```diff
@@ -70,6 +70,20 @@ odoo.define("web_timeline.TimelineView", function (require) {
...
+                if (field.includes(",")) {
+                    // Multiple group by fields
+                    fieldNames.pop(field);
+                    fieldNames.push(...field.split(","));
+                }
+                if (field.includes(":")) {
+                    // Group by date field may have a group operator, e.g. date:month
+                    fieldNames.pop(field);
+                    fieldNames.push(field.split(":")[0]);
+                }
```
carlos-lopez-tecnativa, line 84: When I group by a `Date` or `Datetime` field and drag and drop any element, I get this error. I attached a video:

https://github.com/user-attachments/assets/0690e0dc-58c8-4acf-89c5-16c822998f2c


Additionally, with the same field types, I don't see the correct format. In the example, I grouped by Year and Month, but the data is always displayed as YYYY-mm-dd HH:MM:SS. I expected it to appear as shown in my image, respecting the date format configured in the language settings
image

carlos-lopez-tecnativa (changes requested):
Great feature, thanks.
I remember testing a similar feature in this PR: https://github.com/OCA/web/pull/2929, but this one looks better with support for multi-level grouping.
I'm leaving the results of my functional testing below. Please let me know if you need any additional information.

hitrosol:
@cuongnmtm, would you please make the runboat available so I can try it again ?

cuongnmtm:
Hi @hitrosol , the new runboat is sailing. I also took the opportunity to improve the behavior when grouping by date and datetime fields, as noted by carlos-lopez-tecnativa above.

web_timeline/static/src/js/timeline_renderer.js
```diff
@@ -455,6 +733,16 @@ odoo.define("web_timeline.TimelineRenderer", function (require) {
...
         },
 
+        get_fields_get: async function (group_bys) {
+            const groupByFields = group_bys.map((spec) => spec.split(":")[0]);
+            return await this._rpc({
+                model: this.modelName,
+                method: "fields_get",
+                args: [groupByFields],
+                context: this.getSession().user_context,
+            });
```
carlos-lopez-tecnativa: Could you review the comment here: https://github.com/OCA/web/pull/2929#discussion_r1830977095? I think it applies and avoid making an RPC call.

carlos-lopez-tecnativa (changes requested):
I tested it locally and it works, but I found this error when collapsing a group in project.task. In other views, however, the error is not present. I'm not sure if this view needs some adaptation or if the issue comes from these changes. Could you please review it?
image

web_timeline/static/src/js/timeline_renderer.js
cuongnmtm: Thanks @carlos-lopez-tecnativa, I replaced the redundant fieldsGet by the existing fields.

Regarding the collapsing issue, I cannot reproduce it on my side, even for the proteck.task timeline. Could you please share the timeline view code and the group-by fields you are using?

carlos-lopez-tecnativa: > Regarding the collapsing issue, I cannot reproduce it on my side, even for the proteck.task timeline. Could you please share the timeline view code and the group-by fields you are using?

When grouping by two or more levels and collapsing the first level.
https://github.com/OCA/project/blob/16.0/project_timeline/views/project_task_view.xml

https://github.com/user-attachments/assets/1fe7103c-9eed-46cc-8998-60ae5d32e36d

cuongnmtm: I fixed the issue. It was caused by dependency arrows between records belonging to multiple groups, which triggered errors when those groups were collapsed.

tarteo (approved):
Work perfectly now thanks

web_timeline/static/src/js/timeline_view.js
cuongnmtm: Solved.

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza:
/ocabot merge major

Can you forward port it to upper versions?

cuongnmtm:
Thank you all. I will work on forward-porting this to the newer versions.

PR #3007, opened 2024-12-02
[16.0][MIG]: web_edit_user_filter: Migration to 16.0

anusriNPS:
Migrating web_edit_user_filter to 16.0


anusriNPS:
Verified Scenarios:
==================
1. User able to make a filter as favorite and when tries to click on popover of favorite, it unpacks the filter which is stored as favorite
2. Similarly, user able to mark groupBy as favorite and able to unpack when click on the popover of the favorite.
3. When both filter and groupby is marked as favorite, both groupby and filter values are unpacked.
4. Filter/GroupBy facet on click shows popover and user able to remove filter/groupBy values from popover.
5. When a new favorite is added, user need not reload the page in order to unpack filter/groupBy from favorites. Warning which is available in 14.0 is no longer needed, so removed it.

Unverified Scenarios:
=================
1. In unpack_filter function, there is a branch deals with tentative filter/groupBy which could not be verified as not able to understand the feature from README file or in general.
So, I would like to know how this scenario could be verified? Any suggestions would be helpful to verify it.

web_edit_user_filter/models/ir_filters.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @api.model
+    def get_filters(self, model, action_id=None):
+        res = super().get_filters(model, action_id)
+        ids = map(lambda f: f["id"], res)
+        # Browse filters that are in res
+        filters = self.browse(ids)
+        for i, res_filter in enumerate(res):
+            # Add the field 'facet' to the result
+            res[i]["facet"] = filters.filtered(lambda f: f.id == res_filter["id"]).facet
```
HekkiMelody: **suggestion**:

```suggestion
        for res_filter in res:
            # Add the field 'facet' to the result
            res_filter["facet"] = filters.filtered(lambda f: f.id == res_filter["id"]).facet
```

web_edit_user_filter/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "development_status": "Production/Stable",
+    "license": "AGPL-3",
+    "depends": ["web"],
+    "installable": True,
+    "assets": {
+        "web.assets_backend": [
+            "web_edit_user_filter/static/src/js/*.js",
+            "web_edit_user_filter/static/src/scss/*.scss",
+            "web_edit_user_filter/static/src/xml/backend.xml",
+            "web_edit_user_filter/static/src/xml/search_extended.xml",
```
HekkiMelody: **chore**: for consistence,

```suggestion
            "web_edit_user_filter/static/src/xml/*.xml",
```

web_edit_user_filter/static/src/js/customFavorite.esm.js
```diff
@@ -0,0 +1,54 @@
...
+            );
+            ev.stopPropagation();
+            return this.descriptionRef.el.focus();
+        }
+        var searchItems = this.env.searchModel.getSearchItems(
+            (s) => s.type === "favorite"
+        );
+        const favorites = searchItems.filter(
+            (f) => f.description === this.state.description
+        );
```
HekkiMelody: **suggestion**:

```suggestion
        var searchItems = this.env.searchModel.getSearchItems(
            (s) => s.type === "favorite" && s.description === this.state.description
        );
```

(there are a few similar simplifications that could be made around the code)

HekkiMelody:
> # Unverified Scenarios:
> 
> 1. In unpack_filter function, there is a branch deals with tentative filter/groupBy which could not be verified as not able to understand the feature from README file or in general.
>    So, I would like to know how this scenario could be verified? Any suggestions would be helpful to verify it.

From testing on v14, `tentativeFilter` ([here](https://github.com/OCA/web/blob/14.0/web_edit_user_filter/static/src/js/backend.js#L250) and [here](https://github.com/OCA/web/blob/14.0/web_edit_user_filter/static/src/js/backend.js#L301)) is used when customized filters are used via the "Add custom filter" and "Add custom group" buttons. That's when the pre-defined filter cannot be found. In that case, the module tries to compensate for it and find a similar filter, but it doesn't seem to me like it works very well at all (in v14).

In fact, in v14, contacts view:
1) Group by Country
2) Add Custom Group -> City (not in the pre-defined group-by options)
3) Save as favourite, and reload the page
4) Apply the filter, and then click on the pencil to edit it
5) ISSUE: The filter is taken apart but only the Country groupby remains, instead of having the two nested filters like before.

I'd even go as far as suggesting removing that code path entirely, and adding to the Roadmap that the module is not currently compatible with custom filters. I don't think it's within the scope of the migration to fix that problem, necessarily.

Though of course the last word is up to a maintainer.

web_edit_user_filter/models/ir_filters.py
anusriNPS: Done

web_edit_user_filter/__manifest__.py
anusriNPS: Done

web_edit_user_filter/static/src/js/customFavorite.esm.js
anusriNPS: Modified as suggested. Identified in few other places and modified it. However, could not modify in places like [here](https://github.com/OCA/web/pull/3007/files#diff-6d0be0c9045ebd4c2ce6f7e9cac9d59acdcf7491d10856a32f5b6dd1f546f03bR26) and popover is not visible when modifying it.

anusriNPS:
Tried to use custom groups and filters as mentioned and it works in v16 and it is able to unpack the filters/groupBy which are selected from Add custom filter/group

For eg: (Both "Application" and "Created By" custom groups under App module)
1. Click on "Add Custom Group -> "Application"
2. Click on "Add Custom Group -> "Created By"
3. Save as favorite.
4. Click on favorite facet and then clicking on pencil icon, it unpacks selected custom groups.

However, for this case as well it did not go inside tentative filter related branch code, so removed for now from the PR.

In order to include as part of ROADMAP, I am not sure what to be mentioned as above custom related filters/groupBy works in v16(though it does not work in v14).

As part of latest patchset, corrected issue with displaying facet values and unpacking facets  that related to type "field".

HekkiMelody:
> Tried to use custom groups and filters as mentioned and it works in v16 and it is able to unpack the filters/groupBy which are selected from Add custom filter/group
> 
> For eg: (Both "Application" and "Created By" custom groups under App module)
> 
>     1. Click on "Add Custom Group -> "Application"
> 
>     2. Click on "Add Custom Group -> "Created By"
> 
>     3. Save as favorite.
> 
>     4. Click on favorite facet and then clicking on pencil icon, it unpacks selected custom groups.
> 
> 
> However, for this case as well it did not go inside tentative filter related branch code, so removed for now from the PR.
> 
> In order to include as part of ROADMAP, I am not sure what to be mentioned as above custom related filters/groupBy works in v16(though it does not work in v14).
> 
> As part of latest patchset, corrected issue with displaying facet values and unpacking facets that related to type "field".

I had not tested on v16, if it works correctly then great, no need to mention anything on the roadmap. Thank you

web_edit_user_filter/static/src/js/searchBar.esm.js
```diff
@@ -0,0 +1,159 @@
...
+        );
+
+        if (!selectedFacet.length) {
+            return;
+        }
+        if (facet_type === "favorite") {
+            var FavFacets = [];
+            var currentFacet = this.env.searchModel.getSearchItems(
+                (f) => f.type === facet_type && f.groupId == facetId
+            );
```
HekkiMelody: **comment (non-blocking)**: I think in this case using `searchItems.filter` was better, because you're filtering a subset of the `searchItems` you had already filtered beforehand, instead of searching from every filter in the model.

On the other hand, if `searchItems` is already filtered by `facet_type` and you use `.filter` here, you can drop the `f.type === facet_type` condition because it was already checked earlier.

HekkiMelody (approved):
Code review, LGTM, just some minor non-blocking code cleanup possible

web_edit_user_filter/static/src/js/searchBar.esm.js
anusriNPS: Updated as suggested

anusriNPS:
Identified an item for ROADMAP and included it on the ROADMAP file.

HekkiMelody (approved):

AnizR (approved):
Code LGTM and it works as described.

Thanks!

SAnnabelle (commented):
There is an error on the filters that already existed before the module was installed. 

`UncaughtClientError > SyntaxError
Erreur Javascript non interceptée > "undefined" is not valid JSON
SyntaxError: "undefined" is not valid JSON
    at JSON.parse (<anonymous>)
    at SearchBar._unpackFilter (https://odoo/web/assets/1941417-e5f98c9/web.assets_backend.min.js:18312:577)
    at HTMLAnchorElement.<anonymous> (https://odoo/web/assets/1941417-e5f98c9/web.assets_backend.min.js:18311:822)
    at HTMLAnchorElement.dispatch (https://odoo/web/assets/1941416-2c43b8a/web.assets_common.min.js:2469:447)
    at elemData.handle (https://odoo/web/assets/1941416-2c43b8a/web.assets_common.min.js:2455:166)`

web_edit_user_filter/static/src/js/searchBar.esm.js
```diff
@@ -0,0 +1,151 @@
...
+        });
+    },
+
+    _process_filters($el, facet_type, facetId) {
+        var searchItems = this.env.searchModel.getSearchItems(
+            (facet) => facet.type === facet_type
+        );
+
+        var selectedFacet = searchItems.filter(
+            (f) => f.groupId == facetId && f.isActive === true
```
AnizR:
```suggestion
            (f) => f.groupId == facetId && f.isActive === true && (f.type !== 'favorite' || f.facet !== undefined)
```

I would recommend adding this.
Otherwise, as @SAnnabelle pointed out, we will have an error when editing a favorite filter that was created before installing this module.

AnizR (changes requested):

web_edit_user_filter/static/src/js/searchBar.esm.js
anusriNPS: I agree to implement the suggestion, but when I tried to reproduce the issue locally, it could not reproduce it.

I tried editing favorites thinking latest code base would introduced such errors, but while testing on latest code base I could not reproduce it.

@SAnnabelle , could you please let me know how can I reproduce it? So, I can push the required changes.

AnizR: To reproduce the error, I did this:
 1. Create a favorite filter
 2. Install the module 'web_edit_user_filter'
 3. Try to edit the favorite filter that I created
 4. Error is thrown
 
 The problem is that `facet` wasn't set when creating my filter since the module wasn't yet installed.

AnizR (approved):

web_edit_user_filter/static/src/js/searchBar.esm.js
anusriNPS: Thank you for detailed response. I am able to reproduce it. With provided suggestion, user won't be able to see edit button for favorites whose facet is not available.

I have commented reason for the condition added. Should we notify the user that it can't be unpacked?

AnizR: Well, not displaying the edit button is enough IMO

anusriNPS:
Updated commit to handle error seen below when clicked on "Add to My Dashboard" under "Favorites"

UncaughtPromiseError > TypeError
Uncaught Promise > right-hand side of 'in' should be an object, got undefined
TypeError: right-hand side of 'in' should be an object, got undefined
    _getIrFilterDescription@http://localhost:16069/web/assets/56995-84bd7fb/web.assets_backend.min.js:16716:372
    _getIrFilterDescription (patch web_edit_user_filter.SearchModel)@http://localhost:16069/web/assets/56995-84bd7fb/web.assets_backend.min.js:1691:22
    getPreFavoriteValues@http://localhost:16069/web/assets/56995-84bd7fb/web.assets_backend.min.js:2103:54
    addToBoard@http://localhost:16069/web/assets/56995-84bd7fb/web.assets_backend.min.js:15430:122
    mainEventHandler@http://localhost:16069/web/assets/56988-fc17d32/web.assets_common.min.js:2139:77
    listener@http://localhost:16069/web/assets/56988-fc17d32/web.assets_common.min.js:1294:15

ivs-cetmix:
Hey @anusriNPS thank you for your contribution! @SAnnabelle could you please check again if everything works as expected?

SirPyTech:
@OCA/web-maintainers this PR has been reviewed, fixed, approved and set as "ready to merge" by the bot about 6 months ago, can someone please have a look?
@amh-mw you have authored a few commits in this module's history, would you like to have a look too?
Thanks!

The only pending review https://github.com/OCA/web/pull/3007#pullrequestreview-2845692213 is mentioning an error that happens before the PR so it should not interfere with merge.

tarteo:
Sorry I'm super busy so I can't look at it right now.

ivs-cetmix:
Hey @SirPyTech , could you please squash the bot commits is you are ok with this?

SirPyTech:
> Hey @SirPyTech , could you please squash the bot commits is you are ok with this?

Sure, fixed up bots' commits based on https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate.

ivs-cetmix:
Perfect! @SirPyTech thank you for your contribution!
/ocabot merge nobump

PR #3008, opened 2024-12-02
[16.0][MIG] web_widget_char_size: Migration to 16.0

anusriNPS:
Migrating web_widget_char_size to 16.0


HekkiMelody (approved):
Functional review, LGTM

web_widget_char_size/static/src/js/char_widget.esm.js
```diff
@@ -0,0 +1,10 @@
+/** @odoo-module **/
+
+import {CharField} from "@web/views/fields/char/char_field";
+
+const _extractProps = CharField.extractProps;
+CharField.extractProps = ({attrs, field}) => {
+    return Object.assign(_extractProps({attrs, field}), {
+        maxLength: field.size || attrs.options.size,
```
renda-dev, line 8: nitpick(non-blocking): According to https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength it should be without the capital 'L'

Still works anyways tho

renda-dev (approved):
Code review, LGTM

web_widget_char_size/static/src/js/char_widget.esm.js
anusriNPS: Required to update actual definition in order to be aligned with HTML attributes name:
https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/views/fields/char/char_field.js#L80

renda-dev: Everything is clear now, it's good as it is, thanks!

renda-dev:
Greetings @OCA/web-maintainers, would anyone mind giving this one a look? 

I personally think this is pretty useful :)

CarlosRoca13 (changes requested):
Thank you for your effort! Please check that when fields have a widget assigned, it is not working. For example, by adding the options to the phone field in the res.partner form. Try loading the asset right after the char_field is loaded. Here’s how to do it: https://www.odoo.com/documentation/17.0/es/developer/reference/frontend/assets.html#after 

You also have examples in the code :smile:

HekkiMelody:
> Thank you for your effort! Please check that when fields have a widget assigned, it is not working. For example, by adding the options to the phone field in the res.partner form. Try loading the asset right after the char_field is loaded. Here’s how to do it: https://www.odoo.com/documentation/17.0/es/developer/reference/frontend/assets.html#after
> 
> You also have examples in the code 😄

I didn't even know this was a thing, I thought it was a given that it would only work with the default char widget.

Thank you for pointing it out!

anusriNPS:
> Thank you for your effort! Please check that when fields have a widget assigned, it is not working. For example, by adding the options to the phone field in the res.partner form. Try loading the asset right after the char_field is loaded. Here’s how to do it: https://www.odoo.com/documentation/17.0/es/developer/reference/frontend/assets.html#after
> 
> You also have examples in the code 😄

Hi,

  I tried to implement as suggested by keeping the web_widget_char_size definition after target file of actual char_field.js file. But, it did not help to use option="{'size': 10}" from the view definition.
  I also tried to check option="{'size': 10}" usage in phone field in the contact module in version 14. But, i did not see it's usage in version 14 as well. 
form_view_widget_phonev14
phone_uses_FieldPhone


I also think that in order to use "option={size: 10}" in the char field which uses "widget=phone", updating PhoneField, FormPhoneField definition would help to achieve the same. 

As version 14 does not support this feature from web_widget_char_size module, is this addition required here? 

Please let me know your views on this observation and also correct me if I am wrong.

CarlosRoca13 (approved):
Okay, I see that PhoneField is not extending CharField. It’s a bit inconvenient, but I think we can do without it. If someone needs it in the future, we could create a web_widget_phone_size module.

/ocabot migration web_widget_char_size
/ocabot merge nobump

PR #3018, opened 2024-12-13
[14.0] [IMP] web_m2x_options: use the correct context in search function

renda-dev:
To make the search function more inheritable we are using the same context used in the same base search function (ref: https://github.com/odoo/odoo/blob/14.0/addons/web/static/src/js/fields/relational_fields.js#L568-L571)

### DIFF
--- a/web_m2x_options/static/src/js/form.js
+++ b/web_m2x_options/static/src/js/form.js
@@ -127,7 +127,10 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
                 self.field_color = self.nodeOptions.field_color;
                 self.colors = self.nodeOptions.colors;
 
-                var context = self.record.getContext(self.recordParams);
+                const context = Object.assign(
+                    self.record.getContext(self.recordParams),
+                    self.additionalContext
+                );
                 var domain = self.record.getDomain(self.recordParams);
 
                 var blacklisted_ids = self._getSearchBlacklist();


francesco-ooops (approved):
Functional ok

web_m2x_options/static/src/js/form.js
```diff
@@ -127,7 +127,10 @@ odoo.define("web_m2x_options.web_m2x_options", function (require) {
                 self.field_color = self.nodeOptions.field_color;
                 self.colors = self.nodeOptions.colors;
 
-                var context = self.record.getContext(self.recordParams);
+                const context = Object.assign(
+                    this.record.getContext(this.recordParams),
+                    this.additionalContext
+                );
```
HekkiMelody: **chore** (non-blocking): since all the surrounding code is using `self`, let's also use `self` here for consistency.

HekkiMelody (approved):
Code review, LGTM

web_m2x_options/static/src/js/form.js
renda-dev: Thanks!

HekkiMelody (approved):

francesco-ooops:
@pedrobaeza any opinion on this?

pedrobaeza (approved):
It seems OK according the reference. Please check if this applies the same to v15:

/ocabot merge patch

PR #3024, opened 2024-12-18
[18.0][MIG] web_datetime_picker_default_time: Migration to 18.0

natuan9:


ivantodorovich (approved):
Seems good.

It's a shame we don't have unit tests, though.
How do you feel about adding some?

natuan9:
Hi @ivantodorovich , I added some test cases for this module.

eslint.config.cjs
```diff
@@ -191,7 +191,7 @@ const config = [{
     },
 
 }, {
-    files: ["**/*.esm.js"],
```
ivantodorovich, line 194: These files are synchronized from time to time from the https://github.com/oca/oca-addons-repo-template
This change will then be lost..

I recommend to simply use the `.test.esm.js` suffix on your filename, then it'd be selected by this existing pattern

ivantodorovich (changes requested):
Thanks a lot!

As far as I remember, and as far as I can see on the CI logs, we need to explicitly execute these tests from a python test.

Here's an example on a 16.0 module: https://github.com/OCA/web/blob/16.0/web_domain_field/tests/test_qunit.py

eslint.config.cjs
natuan9: I tried using the suffix `.test.esm.js`, but Hoot doesn't work with this file extension. The guy in this video mentioned that test files must end with `.test.js`.
[Test your JS Code with Hoot](https://www.youtube.com/watch?v=QNHwykNNg2E&t=235s)
image

grindtildeath (changes requested):
Looks good but setting of default time while entering manually the date with keyboard does not work. (it used to in v16.0)

natuan9:
> Looks good but setting of default time while entering manually the date with keyboard does not work

Hi @grindtildeath , is your default time field a JSON field? Do you mean editing the JSON field directly?

natuan9:
Hi @grindtildeath , I resolved issue

web_datetime_picker_default_time/static/src/js/datetimepicker_service.esm.js
```diff
@@ -0,0 +1,295 @@
...
+                const saveAndClose = () => {
+                    if (popoverService.isOpen) {
+                        // Apply will be done in the "onClose" callback
+                        popoverService.close();
+                    } else {
+                        apply();
+                    }
+                };
+
+                const onInputChange = (ev) => {
```
ivantodorovich: Can we avoid re-defining methods that we don't want to override/change?

If the method is exactly the same as the original, then you can probably reference the one in `originalService`

web_datetime_picker_default_time/static/src/js/datetimepicker_service.esm.js
```diff
@@ -0,0 +1,295 @@
...
+const FOCUS_CLASSNAME = "text-primary";
+
+const formatters = {
+    date: formatDate,
+    datetime: formatDateTime,
+};
+
+const listenedElements = new WeakSet();
+
+// OVERRIDE:
```
ivantodorovich: See my other comment about trying to keep here only the methods that are actually overriden

ivantodorovich (commented):
The code has gotten a lot more complicated than before. But I think it's mostly due to the amount of full overwrites we have in here, which should be avoided as much as possible

If you could try to reduce them, that'd be great

natuan9:
Hi @ivantodorovich , I updated, please help to review

ivantodorovich (approved):
Thanks @natuan9 ! LG technically now 👍🏻

sebalix (approved):
Tested locally, working well :+1:

chienandalu (approved):
Tested :+1:

simahawk:
Rebased and replaced here to speed it up https://github.com/OCA/web/pull/3311
Thanks!

PR #3028, opened 2024-12-20
[18.0][MIG] web_assets_warmup: Migration to 18.0

xaviedoanhduy:
### Note
- Since Odoo v17.0, assets are no longer generated in the same transaction that generates a report. They are now generated when retrieved, through the `/web/assets` route. See https://github.com/odoo/odoo/pull/140124
- As a consequence, the problematic case that led to the creation of `web_assets_warmup` can no longer occur: https://github.com/OCA/web/pull/2614#issue-1898108560
- That said, module still has some value in "warming" up common assets for reports
- Since 16.0, odoo introduced a more generic `env['ir.qweb']._pregenerate_assets_bundles()`, but used for tests only, see https://github.com/odoo/odoo/pull/99176
- `commit_assetsbundle` was removed when rendering the report in commit: https://github.com/odoo/odoo/commit/ad4deba
- "declare assets in manifest, not in xml" so this commit changed xmlid parameter to bundles, see in: https://github.com/odoo/odoo/pull/60632/commits/8cc06617
- `web.assets_common` was removed in commit https://github.com/odoo/odoo/pull/132190/commits/02cba88d3 

I also conducted some speed tests when printing the report for the first time (checked when calling the `_render_qweb_pdf` method)([ref](https://github.com/odoo/odoo/blob/730247f96611813cf65cd30a237d29d58b568a96/odoo/addons/base/models/ir_actions_report.py#L1001C9-L1001C25))

- when installing the `sale_management` module, and not installing `web_assets_warmup`: 3.861 (s)
image

- when installing the `sale_management` and `web_assets_warmup` modules but not manually executing the schedule action task (Generate report assets): 3.949 (s)
image

- when installing the `sale_management` and `web_assets_warmup` modules then manually executing the schedule action task (Generate report assets): 2.860 (s)
image


web_assets_warmup/models/ir_actions_report.py
```diff
@@ -0,0 +1,28 @@
...
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    @api.model
+    def cron_generate_assets(self):
+        """Ensure that the assets are well-generated in the database."""
+        logger.info("Ensure that assets are generated and stored in the database...")
+        bundles = [
+            "web.report_assets_common",
+            "web.assets_common",
```
nilshamerlinck, line 20: removed: see `https://github.com/odoo/odoo/pull/132190`

henrybackman (approved):

flotho:
Hi, 
Thanks for the PR,
Any chance to rebase ?

xaviedoanhduy:
> Hi, Thanks for the PR, Any chance to rebase ?

hello, rebase is done

PR #3033, opened 2024-12-24
[16.0][IMP] web_widget_numeric_step: throttle update calls when scrolling on field

CarlosRoca13:
Before this changes, when scrolling on a field, the price update was collapsed because lots of calls to the function was made in a few seconds.

With this changes, the field update, when using the scroll is limited to 1 each 0,1 seconds and the info is updated correctly.

ping @carlos-lopez-tecnativa @chienandalu


carlos-lopez-tecnativa (approved):

yajo (approved):
/ocabot merge patch

PR #3041, opened 2024-12-26
[16.0][ADD] web_widget_many2many_binary_preview

hbrunn:
This module allows developers to enable a preview on fields displayed
with widget ``many2many_binary``.

Easiest way to test this is to edit the view of mail templates and add the attribute ``show_preview="True"`` which I just did on [runboat](http://oca-web-16-0-pr3041-51849d264f23.runboat.odoo-community.org/web?debug=1#id=2&cids=1&menu_id=4&action=125&model=mail.template&view_type=form)


web_widget_many2many_binary_preview/readme/USAGE.md
```diff
@@ -0,0 +1 @@
+To use this module, you need to set the attribute `show_preview` to ``True`` on a field with widget `many2many_binary`.
```
thomaspaulb: I like `preview` better than `show_preview`, but ah well.

web_widget_many2many_binary_preview/static/src/web_widget_many2many_binary_preview.xml
```diff
@@ -0,0 +1,19 @@
...
+        <xpath expr="//div[hasclass('o_image_box')]/a/span" position="attributes">
+            <attribute name="t-if">!show_preview(file)</attribute>
+        </xpath>
+        <xpath expr="//div[hasclass('o_image_box')]/a/span" position="after">
+            <img
+                t-if="show_preview(file)"
+                t-att-src="getUrl(file.id)"
+                aria-label="Download"
+                download=""
+            />
```
thomaspaulb, line 16: As you already mentioned, the image previews can be rather sizy as compared to the more thumbnaily ones shown in the chatter attachment widget. I've noticed that this this is because of the URL; while here, getUrl returns `web/image/XXX?download=true`, in the other widget, it returns `/web/image/XXX/1920x160` which returns the thumbnail. Maybe you can inject another getUrl and do the same here as well. I don't think multiple sizes need to be supported at the moment, but it could be a roadmap item to provide parameters for that.

web_widget_many2many_binary_preview/static/src/web_widget_many2many_binary_preview.xml
```diff
@@ -0,0 +1,19 @@
+<templates>
+    <t
+        t-name="Many2ManyBinaryField.attachment_preview"
+        t-inherit="web.Many2ManyBinaryField.attachment_preview"
```
thomaspaulb, line 4: Weird that the parent `attachment_preview` template does not in any way interface with the set of more advanced [attachment preview components](https://github.com/OCA/OCB/blob/16.0/addons/mail/static/src/components/attachment_image/attachment_image.xml#L3) that are in the `mail` module. Maybe that's a pending refactor that didn't make the 16.0 release.

thomaspaulb: I'd almost think that it's better to make a new widget based on those components, since they have all kinds of [size logic](https://github.com/OCA/OCB/blob/16.0/addons/mail/static/src/models/attachment_image.js#L92) already. But that will probably open other cans of worms, so maybe just copy some of the logic to a new `getPreviewUrl` function and be done with it.

hbrunn: > Maybe that's a pending refactor that didn't make the 16.0 release.

I don't [think](https://github.com/OCA/OCB/blob/18.0/addons/web/static/src/views/fields/many2many_binary/many2many_binary_field.xml#L28) so

> But that will probably open other cans of worms

None that I see, it will just be *much* more work to migrate this back and forth as the mail components have been refactored/rewritten a lot in the past releases, while the many2many_binary widget has barely changed.

I suggest for version 1 we just use the `/web/image` controller and expose its parameters (crop, width, height), that will already be very useful. Whenever inline preview should be implemented, we can look into using the mail components for that, but then we're already pretty much on [attachment_preview](https://odoo-community.org/shop/preview-attachments-531) territory and could just make that support images on many2many_binary fields.

thomaspaulb: Agreed!

hbrunn: I've implemented all your comments, and there's a v14 version of this at https://github.com/OCA/web/pull/3047

thomaspaulb:
I've tested and noticed that even though you limit width and height to very low numbers (I tried 50x50), and I make the container width very high (by embedding the field in a <group colspan="2"/>, or artificially by adding width: 100% to the field div element, I can't get the thumbnails to `float: left` with eachother nicely. The block with the image and its nearby caption just stretches, and the next one always comes below:

image

hbrunn:
that's what the binary attachment widget does, not sure if we should change this here. Would adding `display: flex` to the `.o_attachments` div solve the issue for your use case?

PR #3042, opened 2024-12-27
[18.0][MIG] web_dialog_size: Migration to 18.0

jeromeguerriat-msf:
re-ordered import to avoid "expected multiple syntax before single syntax" error for eslint
removed useless odoo-module in js file
changed the version
removed undefined "stop" parameter in if()


pedrobaeza:
/ocabot migration web_dialog_size

cc @CarlosRoca13

web_dialog_size/web_dialog_size/__manifest__.py
chienandalu, line 1: You have duplicated the whole module inside its own folder :thinking:

chienandalu (changes requested):

web_dialog_size/web_dialog_size/__manifest__.py
jeromeguerriat-msf: oopsy, i fixed that!

CarlosRoca13 (approved):
:+1: Thanks for your work!

youring:
FYI, if you install this module with website_sale module, go to product form, Add Media in the Sales tab, upload and select a image for the product, save the product (manually). Open the uploaded image, there exist two expand icon.
web_dialog_size
The odoo form expand icon is introduced by this commit https://github.com/odoo/odoo/commit/31834b333bc12fa79f5fc468aa69471b76bb666a#diff-21d8a9654933e8e45d6bf0f45e66c0320252bbecc91282381f94e00104c9ca1dR45 . Click it will expand the form while closing the dialog.

So do we have to accept the two almost identical buttons in that corner?

youring:
Personal suggestion, since changing core is hard.
fa-expand > fa-window-maximize
fa-compress > fa-window-restore
max
restore

bealdav:
@chienandalu I see that requested changes are done. Could you approve ?
@youring can your suggestion be done in another PR to avoid a merge delay?

Thanks for all

youring:
@bealdav Sure. non-blocking since it's only cosmetic.

chienandalu (changes requested):
Good catch @youring !

IMO this is far from a corner case, as it's used in x2ManyX2ManyFieldDialog and FormViewDialog with the limitations described in the commit (an with a different behavior from the one in this module).

It's not just cosmetic as the result is a missleading UI where the user can't know which button does what...

For me the changes should be made (and they're really trivial)

bealdav:
Ok I hope @jeromeguerriat-msf will answer

jeromeguerriat-msf:
@bealdav i just pushed the suggestion ;)

bealdav:
@jeromeguerriat-msf thanks a lot cc @chienandalu

chienandalu (approved):
Working fine. Thanks!

CarlosRoca13:
I think it is ready to be merged :smile: Thanks

/ocabot merge nobump

PR #3048, opened 2025-01-07
[18.0][MIG] web_widget_x2many_2d_matrix: Migration to 18.0

yankinmax:


TDu (approved):
LG, but noway to test on runboat

ajaniszewska-dev (approved):
seems good

web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_field/x2many_2d_matrix_field.esm.js
```diff
@@ -0,0 +1,57 @@
+/** @odoo-module **/
+
+import {Component} from "@odoo/owl";
+import {X2Many2DMatrixRenderer} from "@web_widget_x2many_2d_matrix/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm";
+import {archParseBoolean} from "@web/views/utils";
```
BT-mlopez: As of https://github.com/odoo/odoo/commit/0bc69d757f7c3429c6112aa5aaa1ceb999428189 archParseBoolean is not available anymore, exprToBoolean should be used.
` import { exprToBoolean } from "@web/core/utils/strings";`

web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.xml
```diff
@@ -0,0 +1,58 @@
...
+                </tr>
+            </thead>
+            <tbody>
+                <tr t-foreach="rows" t-as="row" t-key="row.value">
+                    <td>
+                        <t t-esc="row.text" />
+                    </td>
+                    <td t-foreach="columns" t-as="column" t-key="column.value">
+                        <t
+                            t-component="ValueFieldComponent"
```
BT-mlopez: When there is no record for any combination of row/column an error is raised from owl, it can be avoided adding:
`t-if="getValueFieldProps(column.value, row.value).record"`

BT-mlopez (changes requested):
Doing a functional review I found a couple of issues that prevent its success.

BT-mlopez (approved):
Thanks for the changes.

PR #3053, opened 2025-01-08
[16.0][IMP] web_widget_dropdown_dynamic: Change license

ivs-cetmix:
Change license to LGPL-3 in order to be able to use this module with other modules with non-AGPL-3 licenses.

@alexey-pelykh  @NL66278 @BankyKmitl

### DIFF
--- a/web_widget_dropdown_dynamic/__init__.py
+++ b/web_widget_dropdown_dynamic/__init__.py
@@ -1,1 +1,1 @@
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
--- a/web_widget_dropdown_dynamic/__manifest__.py
+++ b/web_widget_dropdown_dynamic/__manifest__.py
@@ -1,12 +1,12 @@
 # Copyright 2019-2020 Brainbean Apps (https://brainbeanapps.com)
 # Copyright 2020 CorporateHub (https://corporatehub.eu)
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 {
     "name": "Dynamic Dropdown Widget",
     "summary": "This module adds support for dynamic dropdown widget",
     "category": "Web",
     "version": "16.0.1.0.0",
-    "license": "AGPL-3",
+    "license": "LGPL-3",
     "author": "CorporateHub, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
     "depends": ["web"],
--- a/web_widget_dropdown_dynamic/README.rst
+++ b/web_widget_dropdown_dynamic/README.rst
@@ -13,9 +13,9 @@ Dynamic Dropdown Widget
 .. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
     :target: https://odoo-community.org/page/development-status
     :alt: Beta
-.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
-    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
-    :alt: License: AGPL-3
+.. |badge2| image:: https://img.shields.io/badge/licence-LGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+    :alt: License: LGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fweb-lightgray.png?logo=github
     :target: https://github.com/OCA/web/tree/16.0/web_widget_dropdown_dynamic
     :alt: OCA/web
--- a/web_widget_dropdown_dynamic/static/description/index.html
+++ b/web_widget_dropdown_dynamic/static/description/index.html
@@ -1,20 +1,20 @@
-<?xml version="1.0" encoding="utf-8" ?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils: https://docutils.sourceforge.io/" />
 <title>Dynamic Dropdown Widget</title>
 <style type="text/css">
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 7952 2016-07-26 18:15:59Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
 :Copyright: This stylesheet has been placed in the public domain.
 
 Default cascading style sheet for the HTML output of Docutils.
+Despite the name, some widely supported CSS2 features are used.
 
-See http://docutils.sf.net/docs/howto/html-stylesheets.html for how to
+See https://docutils.sourceforge.io/docs/howto/html-stylesheets.html for how to
 customize this style sheet.
 */
 
@@ -275,7 +275,7 @@ pre.literal-block, pre.doctest-block, pre.math, pre.code {
   margin-left: 2em ;
   margin-right: 2em }
 
-pre.code .ln { color: grey; } /* line numbers */
+pre.code .ln { color: gray; } /* line numbers */
 pre.code, code { background-color: #eeeeee }
 pre.code .comment, code .comment { color: #5C6576 }
 pre.code .keyword, code .keyword { color: #3B0D06; font-weight: bold }
@@ -301,7 +301,7 @@ span.option {
 span.pre {
   white-space: pre }
 
-span.problematic {
+span.problematic, pre.problematic {
   color: red }
 
 span.section-subtitle {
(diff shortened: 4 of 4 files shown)


alexey-pelykh (approved):
I'm fine with this change

Bearnard21 (approved):
We just contributed a similar module and also changed the license to LGPL base on the OCA member request: https://github.com/OCA/server-ux/pull/942

pedrobaeza:
I don't like to open the door to private extensions on top of this one, but it's not my choice in this case as I'm not a contributor of this module.

You have to change though all the copyright texts in the file headers if the switch is done.

ivs-cetmix:
> I don't like to open the door to private extensions on top of this one, but it's not my choice in this case as I'm not a contributor of this module.
> 
> You have to change though all the copyright texts in the file headers if the switch is done.

May I ask you why exactly? I prefer the AGPL license as well, however in some cases it creates havoc while giving no benefit to community. 
I would personally prefer to have things like widgets and tools being LGPL, while core industry modules stay AGPL. This is because with LGPL we can increase potential auditory and benefit from the widespread of the OCA solutions making them a crucial part of other tools. Exactly the way it happened with queue_job

pedrobaeza:
This is a backend tool, so I don't want my developments that I'm giving for free are used for building other things that are not shared. I don't mind the audience. I mind that everything stays open, so if you don't want to follow that rules, be on your own. At the end, I have seen that I have to maintain the same modules being AGPL / LGPL, so I don't benefit on any hypothetical audience increase.

I just make the exception on website ones, where being accessible to the public, and with customized layers like themes and so, I don't want that someone may require the custom/theme code to my customer.

ivs-cetmix:
> This is a backend tool, so I don't want my developments that I'm giving for free are used for building other things that are not shared. I don't mind the audience. I mind that everything stays open, so if you don't want to follow that rules, be on your own. At the end, I have seen that I have to maintain the same modules being AGPL / LGPL, so I don't benefit on any hypothetical audience increase.
> 
> I just make the exception on website ones, where being accessible to the public, and with customized layers like themes and so, I don't want that someone may require the custom/theme code to my customer.

I understand and share your concerns. However I would mention once again: this is just a widget. This module has zero business logic in it.
Keeping it AGPL will force to develop a slightly modified copy of it under LGPL license and will end up with two similar widgets. 
Regarding the benefit of hypothetical audience increase, this is something that should be address at the Organisation level and is more related to marketing rather than to technical part. Recent changes raise some hopes that this can be the case.

P.S. as @Bearnard21 mentioned above, we have changed license in our module because @rvalyi asked us about it. 
P.S.S. there were already [license changes in modules](https://github.com/OCA/web/pulls?q=is%3Apr+in%3Atitle+LGPL+is%3Aclosed+) in this repo.

ivs-cetmix:
P.P.P.S updated license in code too.

ivs-cetmix:
Hi @rvalyi would appreciate your review here too.

ivs-cetmix:
Dear @OCA/web-maintainers would appreciate your merging and reviewing it as it is "ready to merge". Thank you in advance!

rvalyi:
@ivs-cetmix I share the same view on licensing. However you should get the agreement from CroporateHub the author here and from the contributors eventually.

Theoretically, the OCA board would have the power to pass over, but that would create a precendent that would undermine the trust of module authors. So you should go after the module authors and get them agree here...

ivs-cetmix:
> @ivs-cetmix I share the same view on licensing. However you should get the agreement from CroporateHub the author here and from the contributors eventually.
> 
> Theoretically, the OCA board would have the power to pass over, but that would create a precendent that would undermine the trust of module authors. So you should go after the module authors and get them agree here...

@rvalyi CroporateHub has already approved this PR as @alexey-pelykh is the CEO and the owner of it. So I don't see any legal obstacles to move forward.

legalsylvain:
Hi all. Two remarks here.

@Pedro said : 
> but it's not my choice in this case as I'm not a contributor of this module.

@rvalyi said : 
>  However you should get the agreement from CroporateHub the author here and from the contributors eventually.

I don't think so. As far as I understand, when we contribute modules to the OCA, we do it under a CLA. As soon as the code is accepted, the entity that can change the licence of the OCA module is not CroporateHub, but OCA.
Or did I missed something regarding the CLA ?
Ref : https://raw.githubusercontent.com/OCA/odoo-community.org/master/the-association/ICLA.pdf
Thanks for your insight here, because maybe I misunderstood some legal points here.

@ivs-cetmix said : 
>  Keeping it AGPL will force to develop a slightly modified copy of it under LGPL license and will end up with two similar widgets.

Well, you have another option : you can put an AGPL licence on the module you develop, that depends on web_widget_dropdown_dynamic. ;-)
In fact, that is exactly the objective of this licence.

That giving said, I abstain on that PR.

alexey-pelykh:
> Croporate

🙄 :) Corporate

ivs-cetmix:
Ok, we have a board meeting next week. Will raise this question. Thank you for your feedback!

web_widget_dropdown_dynamic/__init__.py
```diff
@@ -1 +1 @@
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/LGPL.html).
```
simahawk: you have to use lowercase, this link does not work.

web_widget_dropdown_dynamic/__manifest__.py
```diff
@@ -1,12 +1,12 @@
 # Copyright 2019-2020 Brainbean Apps (https://brainbeanapps.com)
 # Copyright 2020 CorporateHub (https://corporatehub.eu)
-# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/LGPL.html).
 {
     "name": "Dynamic Dropdown Widget",
     "summary": "This module adds support for dynamic dropdown widget",
     "category": "Web",
     "version": "16.0.1.0.0",
```
simahawk, line 8:
```suggestion
    "version": "16.0.2.0.0",
```

ivs-cetmix: Thank you! OCABot will bump the version when merging with the "minor" param.

web_widget_dropdown_dynamic/__init__.py
ivs-cetmix: Thank you! Will update.

ivs-cetmix:
@simahawk done, could you please check and merge?

web_widget_dropdown_dynamic/__manifest__.py
simahawk: Any reason to go for minor? I'd go for major when changing license (maybe something to add to future guidelines :wink: )

ivs-cetmix: Up to you, I don't have any experience with this, so I believe you know it better.

rvalyi:
@legalsylvain as far as I know, the module author retain the copyright but also delegates it to the OCA board. That is technically both can decide.

But if the OCA starts taking decisions that goes against modules authors we have to understand the deep implications that would have: that would de facto totally undermine the trust contributors have in the OCA entity and contributors will simply assume the OCA will be the new OpenERP, ready to screw everybody as it is more convenient for itself.

But if the module author simply disappeared, of course the situation is different than an author firmly opposed to a license change...

alexey-pelykh:
@rvalyi the author (me) approves and ~couldn't care less~ does not have a preference between the original and suggested license for this module :)

rvalyi:
> @rvalyi the author (me) approves and ~couldn't care less~ does not have a preference between the original and suggested license for this module :)

great! thank you for showing up.

rvalyi (approved):

rvalyi:
actually @alexey-pelykh it seems you coded the module while employed by CorporateHub which is the declared module author and by checking your Github it seems you are not part of this organization or not anymore.

Let's just remember that technically in general when employees do some work their employer or eventually customer retain the copyright.

So @alexey-pelykh does your agreement here for a license change legally represents CorporateHub? How do we know? That would fully protect the OCA from a potential shitstorm...

Now, my personal opinion here is as technically the OCA board also holds the copyright and as at least the person who coded the module agrees explicitly on the change, then that would already aliviate a lot the "moral" responsability of the OCA in doing the change. That's why I approved the PR.

alexey-pelykh:
@rvalyi

https://github.com/orgs/corporatehub/people - switched to public
https://euipo.europa.eu/eSearch/#details/trademarks/018787771

alexey-pelykh:
> great! thank you for showing up.

@rvalyi also, just in case, I was the very first one who approved this PR same day it was created.

P.S. OCA's speed of reaction is as always unmatched 🤣

rvalyi:
> > great! thank you for showing up.
> 
> @rvalyi also, just in case, I was the very first one who approved this PR same day it was created.
> 
> P.S. OCA's speed of reaction is as always unmatched 🤣

Ok thank you. Well, please simply understand that your early approval "I'm fine with this change" had nothing to help us to relate you to CorporateHub which is the module author and copyright holder in the files. Some people may know it's the company you are working for, but I'm afraid you cannot take it for granted...

But okay anyway this is solved now I think. cc @ivs-cetmix @pedrobaeza @legalsylvain

alexey-pelykh:
> Now, my personal opinion here is as technically the OCA board also holds the copyright and as at least the person who coded the module agrees explicitly on the change, then that would already aliviate a lot the "moral" responsability of the OCA in doing the change. That's why I approved the PR.

As for this, depends on presence of explicit clause in CLA about pass of copyright. And that would mean that only OCA should be indicated as copyright holder. And I'm not sure that's the case. So I guess you need approval from everyone who [owns copyright on the] contributed code to the module and thus - holds copyright to portions of work.

rvalyi:
Well here I think we fall into the "moral responsibility" of the OCA not to go against contributors. But IMHO minor contributors who didn't even got a copyright mention are not legit to oppose such a change. Again, legally speaking the OCA can do the change. I'm just here giving my opinion about the "moral responsibility" when doing so.

alexey-pelykh:
👍 for asking the legitimacy of approval - that question must be asked :)

PR #3055, opened 2025-01-13
[18.0][MIG] web_time_range_menu_custom: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT51904

ping @chienandalu @sergio-teruel


giarve (approved):

pedrobaeza (approved):
The utility of this module is now reduced, as we can filter the same in the custom search:

Image

But it's still useful to have the option directly in the date auto-filters.

/ocabot merge nobump

PR #3058, opened 2025-01-15
[16.0][ADD] web_responsive_attachment_preview

hbrunn:
This module restores the attachment preview on the right side of forms for example on invoices by standard Odoo when using ``web_responsive``. The chatter in that case is always below the form. Note Odoo's cutoff is 1534px width, smaller screens never show the preview.

Use case is a controller going through invoices to confirm them, in that case it's pretty annoying to have to open the attachments and then the preview of the invoice's pdf manually a hundred times a day.

Technically, it removes this [patch](https://github.com/OCA/web/blob/16.0/web_responsive/static/src/views/form/form_controller.esm.js#L9-L14). (I still think web_responsive does too much at once and should be split up into several ones)


flotho (approved):
Thanks,
Tested on runboat, make sense
LGTM

flotho:
any conflict with https://github.com/OCA/web/pull/2695 ?

hbrunn:
they do basically the same thing, so one or the other should be merged. People on the other PR didn't seem enthusiastic to have it in the main module, that's why I added it as extra one

pedrobaeza:
I think that a exclusive view for reviewing invoices with the preview should exist, but most of the work with invoices shouldn't show the preview, as you don't need it, and you have to go to bottom to check the chatter, which is what you use most.

Stephanvandenberg:
This PR is highly appreciated by our customers so It hink it should be part of the standard package.

Stephanvandenberg (approved):
Works like a charm - can we merge this?

PR #3059, opened 2025-01-16
[18.0][MIG] web_widget_pattern: Migration to 18.0

sensiblecs:


hbrunn (approved):

hbrunn (changes requested):
Further testing reveals several problems:

- the pattern doesn't work when setting the pattern parameter on a field in python code
  You need to patch Field.parseFieldNode for this:
  ```patch
  +import {Field} from  "@web/views/fields/field";
  ...
  +patch(Field, {
  +    parseFieldNode(node, models, modelName, viewType, jsClass) {
  +        const fieldInfo = super.parseFieldNode(...arguments);
  +        fieldInfo.pattern = models[modelName].fields[fieldInfo.name].pattern;
  +        return fieldInfo;
  +    }
  +});
  ```
  and change `extractProps` accordingly
- function `base#fields_get` should check for `attributes` being falsy, not `None`
- function `base#_get_view_field_attributes` doesn't exist any more, so this override should be removed
- the autocomplete component now inherits from Char, so this override should be removed too
- please fix the tests

hbrunn:
superseded by https://github.com/OCA/web/pull/3511

PR #3066, opened 2025-01-21
[16.0][FIX] web_dialog_size: Fixed the issue that form view is changing its…

baguenth:
The size of a form is inconsistent when using web_dialog_size. It changes when changing any field in the form (which leads to rerendering the form).

Affected versions: 16.0

Scenario 1:

1) Set the system parameter "web_dialog_size.default_maximize" to true, so the form will open in full screen by default
2) Open a form view --> It will open in full screen
3) Change any field, so an onchange event is fired and the form gets re-rendered --> The form is no longer showing in full screen

Scenario 2:

1) Open a form view
2) Manually maximise it so it shows in full screen
3) Change any field, so an onchange event is fired and the form gets re-rendered --> The form is no longer showing in full screen

Expected behavior

The form should keep it's current size when changing any field values

@pedrobaeza @chienandalu Kindly asking for a review :)

### DIFF
--- a/web_dialog_size/static/src/js/web_dialog_size.esm.js
+++ b/web_dialog_size/static/src/js/web_dialog_size.esm.js
@@ -3,35 +3,49 @@
 import {ActionDialog} from "@web/webclient/actions/action_dialog";
 import {patch} from "@web/core/utils/patch";
 import rpc from "web.rpc";
-import {Component, onMounted} from "@odoo/owl";
+import {Component, onWillRender} from "@odoo/owl";
 import {Dialog} from "@web/core/dialog/dialog";
 import {SelectCreateDialog} from "@web/views/view_dialogs/select_create_dialog";
 
 export class ExpandButton extends Component {
     setup() {
-        this.last_size = this.props.getsize();
+        this.lastSize = this.currentSize = this.props.getsize();
+        this.sizeRestored = false;
         this.config = rpc.query({
             model: "ir.config_parameter",
             method: "get_web_dialog_size_config",
         });
 
-        onMounted(() => {
+        onWillRender(() => {
             var self = this;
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not able to reproduce the scenario you are mentioning in the issue. Can you please put exact steps to reproduce it?

And please don't open an issue and then the PR. You can put the issue text here in the PR and close the issue.

baguenth:
@pedrobaeza 

First of all: Sorry, I thought creating an issue before raising a pull request is the official workflow.

I have recorded the issue which I can easily reproduce in the web-runboat. 

Precondition for the video: I have modified the sale order line to open a form instead of inline edit.

https://github.com/user-attachments/assets/a3f45214-5cfb-4d8f-826a-797ccd2d8a8e

You can try yourself here: http://oca-web-16-0-913230bed5f6.runboat.odoo-community.org/web?#cids=1&menu_id=186&action=305&model=sale.order&view_type=form

Steps: 

1) Create a new sale order line --> Form will open
2) Expand the form to full screen
2) Fill the Description field
3) Click somewhere outside of the description field --> Form will change it's size

chienandalu (commented):
I tested in your runbot and looks like v17 has an issue in that conditions as well. You didn't mention that you modified that sales order view to raise the form dialog for the order lines.

@CarlosRoca13 can you take a look on the fix?

CarlosRoca13 (approved):
The changes made make sense and I’ve tested it as well, and it resolves the issue. From my side, it’s good to go :smile:

chienandalu (approved):
:+1:

CarlosRoca13:
/ocabot merge patch 

It will not be necessary on 17.0 so don't need to port the changes. Thanks 4 the patch :smile:

PR #3069, opened 2025-01-27
[17.0][FIX] web_responsive: Make showAppsMenuItem work again

carlos-lopez-tecnativa:
In Odoo 17, the Class `TourStepUtils` was removed, and `stepUtils` is now an `object` instead of a `class`. When this module is installed, the tours stop working and throw the error:` Uncaught Error: Dependencies should be defined by an array:`

This commit adapts the code to align with these changes, ensuring that tours work correctly.

V16
https://github.com/odoo/odoo/blob/b97f0cfbd6b39d176f9b7e465f6da0722222b010/addons/web_tour/static/src/js/tour_step_utils.js#L7
V17
https://github.com/odoo/odoo/blob/77243bc85db87fad2196052036390f65bd451eeb/addons/web_tour/static/src/tour_service/tour_utils.js#L426

image

@Tecnativa @CarlosRoca13 @pedrobaeza @yajo could you please review this, Thanks!

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -16,7 +16,7 @@
     "Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "installable": True,
-    "depends": ["web", "mail"],
+    "depends": ["web_tour", "mail"],
     "development_status": "Production/Stable",
     "maintainers": ["Tardo", "SplashS"],
     "excludes": ["web_enterprise"],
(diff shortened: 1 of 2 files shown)


CarlosRoca13 (approved):
LGTM :+1:

web_responsive/__manifest__.py
```diff
@@ -16,7 +16,7 @@
     "Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "installable": True,
-    "depends": ["web", "mail"],
+    "depends": ["web", "mail", "web_tour"],
```
sergio-teruel: Remove web module

sergio-teruel (changes requested):

web_responsive/__manifest__.py
carlos-lopez-tecnativa: Done

pedrobaeza (approved):
The change of dependencies is reasonable, as `web_tour` is required by most of the top-level applications (sale, account, purchase...).

/ocabot merge patch

PR #2776, opened 2024-03-21
[16.0][IMP] web_timeline: Add demo view on cron tasks

houzefa-abba:
This way we can showcase this module without project_timeline.

Enable debug mode, go to Technical > Scheduled Actions, switch to timeline view:
Screenshot at 2024-03-21 13-34-48

PR #3073, opened 2025-01-30
[16.0][IMP] web_timeline: Safer demo view action override

houzefa-abba:
For the demo view added in https://github.com/OCA/web/pull/2776

This is ahead of another module I am working on where I will also be adding a demo view on cron tasks; that new module & web_timeline both override this same action to add their own demo view, so only 1 or the other would be testable on runboat.

I am not aware of a slicker way to do this (similar to [base_view_inheritance_extension](https://github.com/OCA/server-tools/tree/16.0/base_view_inheritance_extension)), I hope with the comment I added this remains readable.

Runboat check: Enable debug mode, go to Technical > Scheduled Actions, ensure timeline view got added same as before:
Screenshot at 2025-01-30 10-31-34

### DIFF
--- a/web_timeline/demo/ir_cron_view.xml
+++ b/web_timeline/demo/ir_cron_view.xml
@@ -8,7 +8,12 @@
             <timeline date_start="nextcall" default_group_by="model_id" />
         </field>
     </record>
-    <record id="base.ir_cron_act" model="ir.actions.act_window">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
I think it's ok, and more being demo data.

tarteo (approved):

PR #3079, opened 2025-02-04
[15.0][ADD] web_field_text_limit: New module to add limits to texts

CarlosRoca13:
This module allows adding limits in the same way as for float decimals. However, for now, it only applies to fields in QWeb views. The intention is to eventually apply it to form, tree, and other views as well.

cc @Tecnativa TT54838


sergio-teruel:
@CarlosRoca13 Some test??

carlosdauden (approved):

PR #3086, opened 2025-02-10
[FIX][16.0] web_refresher: error when dom destroyed during timeout

miikanissi:
If you navigate out of the current view while the refresh animation is running, you will receive the following error:

```
UncaughtClientError > TypeError
Uncaught Javascript Error > Cannot read properties of null (reading 'remove')
TypeError: Cannot read properties of null (reading 'remove')
    at http://127.0.0.1:8089/web/assets/5713-67df1f9/web.assets_backend.min.js:1860:125
```

This is due to the fact that the dom element does not exist anymore. This patch adds an extra check to ensure the content exists after the animation timeout finishes.

### DIFF
--- a/web_refresher/static/src/js/refresher.esm.js
+++ b/web_refresher/static/src/js/refresher.esm.js
@@ -12,14 +12,6 @@ export function useRefreshAnimation(timeout) {
     const refreshClass = "o_content__refresh";
     let timeoutId = null;
 
-    /**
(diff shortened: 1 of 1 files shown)


CarlosRoca13 (approved):
LGTM :+1:

pedrobaeza (approved):
Please check if needed in upper versions:

/ocabot merge patch

PR #3093, opened 2025-02-17
[17.0][IMP] web_widget_x2many_2d_matrix: Fix domain issue

JasminSForgeFlow:

PR #3116, opened 2025-03-08
[17.0][IMP] web_widget_x2many_2d_matrix: fix to prevent errors and update readme usage

cuongnmtm:
**1. Fix the issue reported here https://github.com/OCA/web/pull/3093#discussion_r1986026469 (1st commit).**

Steps to reproduce:

- Test with the field of field_value has no domain specified, or domain = [] (domain = "[]" still works)

**2. Fix the issue where an error is raised when no record exists for a cell (2nd commit).**

W/o the fix (2 errors):

image

image

With the fix:

image

**3. Update the readme usage cause Odoo renamed some fields.**


JasminSForgeFlow:
@cuongnmtm domain issue hasn't occurred to me while testing a field without a domain.

PR #3118, opened 2025-03-11
[16.0][OU-FIX] web_chatter_position: Ensure correct values for 'chatter_position'

baguenth:
Field 'chatter_position' was defined in module 'web_responsive' before it was moved to 'web_chatter_position'.
In 'web_responsive' the allowed values have been ['normal', 'sided'] while the allowed values in 'web_chatter_position' are ['auto', 'bottom', 'sided']

In case users installed 'web_chatter_position' along with 'web_responsive', field 'chatter_position' potentially was already in the db with old values, which are not supported.

This migration covers this scenario by correcting unsupported values.

The issue was also fixed for future migrations in https://github.com/OCA/OpenUpgrade/pull/4807

@pedrobaeza Kindly asking for a review

### DIFF
--- a/web_chatter_position/__manifest__.py
+++ b/web_chatter_position/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Chatter Position",
     "summary": "Add an option to change the chatter position",
-    "version": "16.0.1.0.2",
+    "version": "16.0.1.0.3",
     "author": "Hynsys Technologies, Camptocamp, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
     "license": "LGPL-3",
--- a/web_chatter_position/migrations/16.0.1.0.3/post-migration.py
+++ b/web_chatter_position/migrations/16.0.1.0.3/post-migration.py
@@ -1,0 +1,21 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    # Field 'chatter_position' was defined in module 'web_responsive' before it was
+    # moved to 'web_chatter_position'. In 'web_responsive' the allowed values have
+    # been ['normal', 'sided'].
+    # In case users installed 'web_chatter_position' along with 'web_responsive' field
+    # 'chatter_position' potentially was already in the db with old values, which are
+    # not supported.
+    # This migration covers this scenario by correcting unsupported values.
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE res_users
+        SET chatter_position = 'auto'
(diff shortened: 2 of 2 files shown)


pedrobaeza:
I don't use nor recommend the usage of this module, as for me the intended usage of having the chatter at the right part is the one to enforce. Optimizations should go at other level, like reducing the number of columns in the list or design the screens in a better way, as if you need more width, it only highlights a worst use experience.

baguenth:
@pedrobaeza In can understand you don't like the idea of the module in general. We use it in production and therefore the choice was already made.

Maybe some of the previous contributors can take a look: @dreispt @ndd-odoo @BT-jmontero

web_chatter_position/migrations/16.0.1.0.3/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+    # In case users installed 'web_chatter_position' along with 'web_responsive' field
+    # 'chatter_position' potentially was already in the db with old values, which are
+    # not supported.
+    # This migration covers this scenario by correcting unsupported values.
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE res_users
+        SET chatter_position = 'auto'
+        WHERE chatter_position not in ('auto', 'bottom', 'sided')
```
MiquelRForgeFlow:
```suggestion
        WHERE chatter_position = 'normal'
```
Better this way, as the 'web_responsive' non-allowed option was 'normal', and because other custom modules may have other options.

web_chatter_position/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     "name": "Chatter Position",
     "summary": "Add an option to change the chatter position",
-    "version": "16.0.1.0.2",
+    "version": "16.0.1.0.3",
```
MiquelRForgeFlow, line 7: I am not sure, but I think this is not needed, as the migration folder should be `16.0.1.0.0`.

But I understand that maybe you have already migrated, so you may want to execute this migration script in a v16 database. So ok with the change.

web_chatter_position/migrations/16.0.1.0.3/post-migration.py
baguenth: Thank you for your review. Updated the pr.

MiquelRForgeFlow (approved):

web_chatter_position/__manifest__.py
baguenth: Yes exactly. Module is already rolled out and therefore I need a version higher than the current version.

baguenth:
@dreispt @pedrobaeza Can we may integrate this?

pedrobaeza (approved):
Supporting you although I don't use it and being the patch OK.

PR #3119, opened 2025-03-11
[14.0][FIX] web_ir_actions_act_multi: Uninstallation Issue

carlos-lopez-tecnativa:
During uninstallation, Odoo attempts to delete columns and tables related to the `ir.actions.act_multi` model. However, this model shares the same table (`ir_actions`) with other core modules, meaning the table and its columns should not be removed.

Before this commit, this issue could render the database unusable. After this commit, the database remains intact, and the module can be uninstalled without problems.

Fix https://github.com/OCA/web/issues/2394

TT51961
@Tecnativa @pedrobaeza @chienandalu @augusto-weiss @gaikaz could you please review this?

I believe these changes are less risky than https://github.com/OCA/web/pull/2817
Note: Forward-port to v16 and included in the migration to v17: https://github.com/OCA/web/pull/2704

### DIFF
--- a/web_ir_actions_act_multi/models/__init__.py
+++ b/web_ir_actions_act_multi/models/__init__.py
@@ -1,1 +1,2 @@
 from . import ir_actions
+from . import ir_model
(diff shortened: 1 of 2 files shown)


chienandalu (approved):
Looks good :)

pedrobaeza (approved):
/ocabot merge patch

PR #3120, opened 2025-03-11
[15.0][FIX] web_ir_actions_act_multi: Uninstallation Issue

carlos-lopez-tecnativa:
During uninstallation, Odoo attempts to delete columns and tables related to the ir.actions.act_multi model. However, this model shares the same table (ir_actions) with other core modules, meaning the table and its columns should not be removed.

Before this commit, this issue could render the database unusable. After this commit, the database remains intact, and the module can be uninstalled without problems.

@Tecnativa @pedrobaeza @chienandalu @augusto-weiss @gaikaz could you please review this?
fw-port of https://github.com/OCA/web/pull/3119

### DIFF
--- a/web_ir_actions_act_multi/models/__init__.py
+++ b/web_ir_actions_act_multi/models/__init__.py
@@ -1,1 +1,2 @@
 from . import ir_actions
+from . import ir_model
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Fast-tracking it for solving the issue:

/ocabot merge patch

PR #3128, opened 2025-03-18
[18.0][IMP] web_widget_x2many_2d_matrix: remove bg-colors to work with Odoo EE dark mode.

DavidJForgeFlow:
The background color was needed in previous versions, however now the color is inherited by parent elements. Also removing shadow from totals.

Forward port of #3126


LoisRForgeFlow (approved):
Thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #3133, opened 2025-03-26
[18.0][MIG] web_filter_header_button: Migration to 18.0 

chienandalu:
Tasks done:

- [x] Refactor to the new web client framework
- [x] Adapt to new layout

Peek 27-03-2025 11-20

cc @Tecnativa TT54366


CarlosRoca13 (approved):
LGTM :+1: please @sergio-teruel review

CarlosRoca13:
I've found a bug and I'm looking for a way to fix it. I'll let you know as soon as I have it.

CarlosRoca13:
The error was caused by field-type filters, but it doesn't make sense for them to be in the header buttons because they require a value to be evaluated. So, the easiest solution is to exclude them in the initial filter.

CarlosRoca13:
Please @sergio-teruel review this PR :smile:

sergio-teruel (approved):
Functional review Ok!!

PR #3134, opened 2025-03-27
[18.0][MIG] web_ir_actions_act_multi: Migration to 18.0

chienandalu:
Standard migration.

Pending:

- [ ] Test in real use case

cc @Tecnativa @carlos-lopez-tecnativa @CarlosRoca13 TT54367


web_ir_actions_act_multi/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+crud_ir_actions_act_multi,access_ir_actions_act_multi,model_ir_actions_act_multi,,1,1,1,1
```
carlos-lopez-tecnativa: Please add a group to prevent the error in CI.

carlos-lopez-tecnativa (changes requested):

web_ir_actions_act_multi/security/ir.model.access.csv
pedrobaeza: Never add a wildcard ACL. At least it should be restricted to internal users.

pedrobaeza: Anyway, follow the standard way of working, that I think it's using it with `sudo`.

chienandalu: This is how Odoo handles the same case: https://github.com/odoo/odoo/blame/ec92d92ec8dff80a6a8f0c9d9291e55123f429c5/odoo/addons/base/security/ir.model.access.csv#L102

carlos-lopez-tecnativa (approved):
LGTM

CarlosRoca13 (approved):
Working fine :smile: 

/ocabot merge nobump

PR #3135, opened 2025-03-27
[18.0][MIG] web_listview_range_select

JasminSForgeFlow:
Standard Migration

@ForgeFlow


pedrobaeza:
Isn't this already included in core?

LoisRForgeFlow:
Yes, this is standard in both CE and EE since 17.0 at least. I think the PR can be closed.

PR #3139, opened 2025-03-28
[18.0][IMP] web_widget_numeric_step: be able to add classes

chienandalu:
- Adapted forward of https://github.com/OCA/web/pull/2883

If we set a class in the field it will get to the widget template.

cc @Tecnativa @sergio-teruel @CarlosRoca13 TT54358


CarlosRoca13 (approved):
LGTM :+1: Thanks :smile:

sergio-teruel (approved):

PR #3140, opened 2025-03-28
[18.0][MIG] web_widget_mpld3_chart

ThiagoMForgeFlow:
Migration to 18.0

@ForgeFlow


web_widget_mpld3_chart/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "summary": "This widget allows to display charts using MPLD3 library.",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "version": "18.0.1.0.0",
+    "website": "https://github.com/OCA/web",
+    "depends": ["web"],
+    "data": [],
+    "external_dependencies": {"python": ["mpld3==0.5.10", "beautifulsoup4"]},
+    "auto_install": False,
+    "development_status": "Beta",
+    "maintainers": ["JordiBForgeFlow", "ChrisOForgeFlow"],
```
LoisRForgeFlow: @ThiagoMForgeFlow Can you replace Chris as maintainer? (he's not active anymore)

ThiagoMForgeFlow: Changed!

LoisRForgeFlow (approved):
Functional test, working fine :+1:

web_widget_mpld3_chart/static/src/js/web_widget_mpld3_chart.esm.js
```diff
@@ -15,14 +15,12 @@ export default class Mpld3ChartJsonWidget extends Component {
...
         });
-        onWillStart(() =>
-            loadBundle({
-                jsLibs: [
-                    "/web_widget_mpld3_chart/static/src/lib/d3/d3.v5.js",
-                    "/web_widget_mpld3_chart/static/src/lib/mpld3/mpld3.v0.5.10.js",
-                ],
-            })
-        );
+        onWillStart(async () => {
```
etobella, line 18: A small suggestion. Why don't you create an specific asset and load directly as loadBundle? This way you may add extra libraries if needed.

WDYT?

ThiagoMForgeFlow: I changed loadBundle to loadJS funtion because the old method now only accepts an asset as a parameter. For simplicity, I have added the two lines of loadJS. But you are right, if we create an asset for these two files also works. Do you prefer to change it?

etobella: I think I was not clear enough.

The idea is to create a new asset called mpld3 and add the libraries. This is done by adding the list on the assets key in the manifest. This way we can extend in the future with child libraries of mpld or adding specific css or whatever we want

ThiagoMForgeFlow: The loadBundle in version 18 loads the asset in parallel. The before and after directives can be used to enforce a specific order. However, with the current code, the following error is displayed:

`Uncaught Promise > The following error occurred in onWillStart: "JSON.parse: unexpected character at line 1 column 1 of the JSON data"`

ThiagoMForgeFlow: I've been trying to fix the error, but it seems that Odoo doesn't load the bundle correctly. I'm going back to the first version using `loadJS`.

etobella: Yes, it loads them without order, giving errors. Sorry for the confusion :cry:

etobella (approved):
/ocabot merge nobump

Issue #3010, opened 2024-12-03
Dark Mode calendar picker too dark


skalika:
## Module
Dark Mode v16

## Describe the bug
Dark mode on v16 calendar picker is too dark and choose date is dificult.

## To Reproduce
16.0.1.3.2

1. Change to dark mode
2. Choose any module and create record that require date to be choosed
3.
![Screenshot_2024-12-03-17-24-07-974_com android chrome-edit](https://github.com/user-attachments/assets/ac1d68d0-e3cf-47ed-8a56-a38047f2879c)

**Expected behavior**

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

PR #3141, opened 2025-03-29
[16.0] web_dark_mode: adapt date picker coloring

BitStab:
Issue #3010 

Adapted slightly the date picker and date range picker to more contrast
image


pedrobaeza (approved):
Tested on runboat uninstalling classic theme.

PR #3143, opened 2025-03-31
[IMP] web_environment_ribbon: prepare running_env for ribbon name vals

kafai-lam:
Allow developer to construct ribbon name by `running_env` (from [server_environment](https://github.com/OCA/server-env/tree/18.0/server_environment) module)

Reference: https://github.com/OCA/server-env/blob/894c98587156042a02b8d02646b7513f75642720/server_environment/server_env.py#L47-L61

### DIFF
--- a/web_environment_ribbon/models/web_environment_ribbon_backend.py
+++ b/web_environment_ribbon/models/web_environment_ribbon_backend.py
@@ -1,7 +1,10 @@
 # Copyright 2017 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import os
+
 from odoo import api, models
+from odoo.tools.config import config
 
 
 class WebEnvironmentRibbonBackend(models.AbstractModel):
@@ -10,7 +13,16 @@ class WebEnvironmentRibbonBackend(models.AbstractModel):
 
     @api.model
     def _prepare_ribbon_format_vals(self):
-        return {"db_name": self.env.cr.dbname}
+        running_env = (
+            config.get("running_env")
+            or os.environ.get("RUNNING_ENV")
+            or os.environ.get("ODOO_STAGE")
+            or "test"
+        )
+        return {
+            "db_name": self.env.cr.dbname,
+            "running_env": running_env,
+        }
 
     @api.model
     def _prepare_ribbon_name(self):


web_environment_ribbon/models/web_environment_ribbon_backend.py
```diff
@@ -10,7 +13,16 @@ class WebEnvironmentRibbonBackend(models.AbstractModel):
 
     @api.model
     def _prepare_ribbon_format_vals(self):
-        return {"db_name": self.env.cr.dbname}
+        running_env = (
+            config.get("running_env")
+            or os.environ.get("RUNNING_ENV")
```
ap-wtioit, line 18: Mostly because I'm curious is this list of environment variables supporting a specific odoo docker image or some kind of setup? Could we add [DOODBA_ENVIRONMENT](https://github.com/Tecnativa/doodba-copier-template/blob/main/prod.yaml.jinja#L25C7-L25C25) to this list? 

And it seems that [odoo](https://github.com/odoo/odoo/blob/18.0/odoo/tools/config.py) does not support `config.get("running_env")` are you using a special Odoo Version to support this or am i missunderstanding the code?

kafai-lam: `running_env` is **not** a standard odoo config, it was added from the oca module [server_environment](https://github.com/OCA/server-env/tree/18.0/server_environment). You can find the usage from this [reference](https://github.com/OCA/server-env/blob/894c98587156042a02b8d02646b7513f75642720/server_environment/server_env.py#L47-L61)

hparfr: Does https://github.com/OCA/server-env/tree/18.0/server_environment_ir_config_parameter#configuration do what you need ?

kafai-lam: I have also downloaded this module to configure the ribbon display like following
```
[ir.config_parameter]
ribbon.name={db_name}-staging
```
But I have to manually change the ribbon for every environment by myself
What I want to do is using single ribbon config for all environment(test/develop/staging/preproduction), like following
```
[ir.config_parameter]
ribbon.name={db_name}-{running_env}
```

PR #3144, opened 2025-03-31
[18.0][FIX]web_environment_ribbon: hide ribbon by default

cem-adhoc:
Since the showRibbon method doesn’t work in KnowledgePortalWebClient and the ribbon always appears empty.
image


ap-wtioit (approved):
LGTM, tested on runboat and `display: none` is the same initial value as used in the JS now it's just used from the beginning.

pedrobaeza (approved):
Don't observe any side problem with this change.

/ocabot merge patch

PR #3148, opened 2025-04-04
[16.0][FIX] web_tree_many2one_clickable: Charge assets after list renderer component

CarlosRoca13:
Before these changes, if a module with many2one fields was loaded before this one, an error occurred saying that the TreeMany2oneClickableButton component could not be found.

To reproduce the error: Install dms and web_tree_many2one_clickable, go to Documents > Files, switch to tree view, and display the Directory field.

By forcing the assets to load right after the web module, we prevent the other module from loading first, thus avoiding this error.

cc @Tecnativa TT55813

ping @pedrobaeza @victoralmau

### DIFF
--- a/web_tree_many2one_clickable/__manifest__.py
+++ b/web_tree_many2one_clickable/__manifest__.py
@@ -21,12 +21,24 @@
     "data": [],
     "assets": {
         "web.assets_backend": [
-            "web_tree_many2one_clickable/static/src/components/"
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza (approved):
Yes, at the end, it seems all the basic stuff that is being altered in the web client requires to be "pre-loaded". Let's keep it in mind when doing new changes.

/ocabot merge patch

PR #3152, opened 2025-04-11
[FIXED] TypeError: env._t is not a function

zedcentury:


StefanRijnhart (changes requested):
Thanks for the fix! Please fix the formatting to make pre-commit happy.

PR #3153, opened 2025-04-14
[17.0][FIX] web_responsive: Don't override z-index of toolbar

CarlosRoca13:
Before this changes when trying to edit a message using web_editor, the toolbar is showed behind the wizard.

After this changes, it is showed correctly.

cc @Tecnativa TT55967

ping @pilarvargas-tecnativa @pedrobaeza


pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #3154, opened 2025-04-14
[17.0][MIG] web_widget_pattern: Migration to 17.0

kafai-lam:


marcelsavegnago (approved):

tarteo (changes requested):
I'm testing on runboat, I'm trying to enter a number ([0-9]+) in the street2 field (from the demo data), but it is never correct:

image

PR #2961, opened 2024-10-16
[18.0][MIG] web_m2x_options

bud-e:

PR #3164, opened 2025-04-29
[18.0][MIG] web_m2x_options: Migration to 18.0

denaux:
New PR for OCA module `web_m2x_options` for v18, based on [PR #2961](https://github.com/OCA/web/pull/2961). Said PR has been open for 6+ months with several requested changes still pending. With this PR, I hope to get things moving:

- Applied changes in accordance with the [migration guidelines](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#tasks-to-do-in-the-migration).
- Incorporated new commits coming from older versions.
- Implementing [my suggestion](https://github.com/OCA/web/pull/2961#discussion_r1802632737) to add unregistered methods to O.W.L.'s validation flow to avoid errors in dev mode.
- Allowing falsy/truthy system parameters in a separate (backportable) commit, as requested [here](https://github.com/OCA/web/pull/2961#pullrequestreview-2401100405).
- Fixing all linting warning as requested [here](https://github.com/OCA/web/pull/2961#discussion_r1942460356).
- Restoring the default (system parameter) behavior, as requested [here](https://github.com/OCA/web/pull/2961#pullrequestreview-2703360342) and some other comments.
- Implementing [my suggestion](https://github.com/OCA/web/pull/2961#pullrequestreview-2749405113) to remove the many manual `props` updates on relational objects to avoid errors in dev mode.
- Proposing a working migrated version of this module for v18 @bosd (based on your comment [here](https://github.com/OCA/web/pull/2961#issuecomment-2813457087)).

Feel free to share suggestions or missing functionality, I commit myself to respond quickly and get this module migrated.


bosd (approved):
Functional test LGTM :+1
Many thanks for the work on this!

bizzappdev (approved):
Code and Functional Test LGTM

web_m2x_options/models/ir_config_parameter.py
```diff
@@ -0,0 +1,18 @@
...
+class IrConfigParameter(models.Model):
+    _inherit = "ir.config_parameter"
+
+    @api.model
+    def get_web_m2x_options(self):
+        opts = [
+            "web_m2x_options.create",
+            "web_m2x_options.create_edit",
+            "web_m2x_options.limit",
+            "web_m2x_options.search_more",
```
carlos-lopez-tecnativa, line 13: This parameter is not working on many2many_avatar_user, but it works fine on other relational widgets.

image

carlos-lopez-tecnativa (changes requested):
Tested and working. There's just one remaining widget that isn't working with search_more. Could you review this, please?
Additionally (non-blocking comment), I think some administrative commits can be squashed.

web_m2x_options/models/ir_config_parameter.py
denaux: This should be fixed now. Can you check? I've put it in a separate commit to backport if necessary.

I've squashed some administrative commits as well.

bizzappdev (commented):
When we do create=True in field with widget many2many_avatar_user 

```xml
<field name="user_ids" class="o_task_user_field" options="{'no_open': True, 'no_quick_create': True, 'limit': 10, 'create': true, 'create_edit': false, 'search_more': true, 'field_color':'type', 'colors':{'contact':'green', 'invoice': 'red', 'delivery': 'blue'}}" widget="many2many_avatar_user"/>
```

It is raising error on opening form view.

image (17)

denaux:
> When we do create=True in field with widget many2many_avatar_user
> 
> ```
> <field name="user_ids" class="o_task_user_field" options="{'no_open': True, 'no_quick_create': True, 'limit': 10, 'create': true, 'create_edit': false, 'search_more': true, 'field_color':'type', 'colors':{'contact':'green', 'invoice': 'red', 'delivery': 'blue'}}" widget="many2many_avatar_user"/>
> ```
> It is raising error on opening form view.

@bizzappdev This is not an issue in the scope of this module but rather a coding error? The field widget you're trying to use accepts a `create` option, assuming the corresponding value is a domain. The error you receive pretty much conveys the same message.
image

bizzappdev:
> @bizzappdev This is not an issue in the scope of this module but rather a coding error? The field widget you're trying to use accepts a `create` option, assuming the corresponding value is a domain. The error you receive pretty much conveys 

Indeed, this does not belong to our module, but it is the default behaviour.

bizzappdev (approved):

carlos-lopez-tecnativa (commented):
I'm not sure if this comment has already been discussed in another conversation thread. If it has, please share the link with me.

I noticed that Odoo has supported the options `no_quick_create` and `no_create_edit` via the options dictionary for some versions now. Why does this module add new options for the same purpose? I think perhaps the module was created back when older Odoo versions didn’t support this, but since the framework now does, is it still necessary to support these custom options?

**Odoo native:**
`no_quick_create: True`
**This module:**
`create: False`

**Odoo native:**
`no_create_edit: True`
**This module:**
`create_edit: False`

My concern is that **with this module, the native options are being ignored**. For example, in the `project.task` form view, the field `user_ids` uses `no_quick_create`, and `tag_ids` uses `no_create_edit`.
image

Without this module, the native options work correctly, tested on Runboat. The field `tag_ids` has `no_create_edit`, and the option is correctly hidden.
image

But with this OCA module, the option is still shown to the user, which breaks the expected native behavior.
image

Conclusion: I think the `create` and `create_edit` options can be removed from this module since they are already supported natively by the framework, or at the very least, **the module should maintain compatibility with the native options**.

Note: I only referred to the `options via field`, but the options via `system parameters` should still be kept.

What do you think?
cc @pedrobaeza

pedrobaeza:
Yes, they can be removed, or at least, that they don't break the core features.

denaux:
@carlos-lopez-tecnativa 

That is a valid concern.

Let me first clear up some things to make sure we're on the same level.

> Why does this module add new options for the same purpose?

This module does not add any additional options for a developer to use in the `options` attribute of a particular field tag. As far as I'm aware, only certain system parameters are taken into account and acted upon accordingly. *There is nothing 'to remove', so it is necessary for the module to indeed be compatible with existing options.*

The undesired behavior you described *does indeed* result from the fact that the native options `no_quick_create` and `no_create_edit` are considered non-existent (due to -- as you have guessed -- the module being based on an older Odoo version that did not have these options). The cause of the undesired behavior is as follows:

1. The system parameter `web_m2x_options.create_edit` is non-existent, making the module consider it as `true`.
2. The option `no_create_edit` is ignored, which makes the module force the field to set `create_edit` to `true`.

I have proposed a fix for this just now. Let me know if you're happy with the results.

web_m2x_options/readme/USAGE.md
```diff
@@ -0,0 +1,83 @@
+## in the field's options dict
+
+`create` *boolean* (Default: depends if user have create rights)
+
+> Whether to display the "Create..." entry in dropdown panel. Only for m2o fields.
+
+`create_edit` *boolean* (Default: depends if user have create rights)
+
+> Whether to display "Create and Edit..." entry in dropdown panel
```
carlos-lopez-tecnativa: > 
> This module does not add any additional options for a developer to use in the `options` attribute of a particular field tag. As far as I'm aware, only certain system parameters are taken into account and acted upon accordingly. _There is nothing 'to remove', so it is necessary for the module to indeed be compatible with existing options._

If this module does not add new options, then the README is outdated.
When I read this file, I understand that these options can be set in the field's options.

web_m2x_options/readme/USAGE.md
```diff
@@ -0,0 +1,83 @@
...
+- web_m2x_options.search_more: True
+- web_m2x_options.field_limit_entries: 5
+
+## Example
+
+Your XML form view definition could contain:
+
+``` xml
+...
+<field name="partner_id" options="{'limit': 10, 'create': false, 'create_edit': false, 'search_more': true, 'field_color':'type', 'colors':{'contact':'green', 'invoice': 'red', 'delivery': 'blue'}}"/>
```
carlos-lopez-tecnativa: Please update the example too
```suggestion
<field name="partner_id" options="{'limit': 10, 'search_more': true, 'field_color':'type', 'colors':{'contact':'green', 'invoice': 'red', 'delivery': 'blue'}}"/>
```

carlos-lopez-tecnativa (changes requested):
Thanks for the changes, this works fine. Just update the README, and it's OK for me.

denaux:
@carlos-lopez-tecnativa 

Seems I was mistaken, apologies. Analyzing the code again, the module seems to assume the existence of said options. Therefore, the USAGE.md file doesn't seem out of date. Do contradict me if I'm wrong, though.

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa
> 
> Seems I was mistaken, apologies. Analyzing the code again, the module seems to assume the existence of said options. Therefore, the USAGE.md file doesn't seem out of date. Do contradict me if I'm wrong, though.

Ok, but this contradicts your previous comment.

> 
> > Why does this module add new options for the same purpose?
> 
> This module does not add any additional options for a developer to use in the `options` attribute of a particular field tag. As far as I'm aware, only certain system parameters are taken into account and acted upon accordingly. _There is nothing 'to remove', so it is necessary for the module to indeed be compatible with existing options._


In conclusion:
This module adds new options to fields, in addition to those already supported by the Odoo core (which seems redundant to me and could be unified).
The README is correct regarding the currently supported options, but if we decide to remove these custom options and only keep the core ones, then it will be necessary to update the README as well.

My final question: do you plan to unify the options by removing the custom ones and keeping only the core-supported options, or not?

denaux:
@carlos-lopez-tecnativa 

> Ok, but this contradicts your previous comment.

That is kind of the nature of a mistake being made and admitting to it? No need to unnecessarily point this out.

> My final question: do you plan to unify the options by removing the custom ones and keeping only the core-supported options, or not?

Thinking about it, it seems safer to keep them in for easier migrations on existing projects that use said options. This is just one opinion, though. If you think otherwise, I'm happy to incorporate your suggestion, if you could make the necessary code changes.

pedrobaeza:
The migration is the moment for doing "breaking changes". If there's a feature that already exists in core, keeping it here only complicates the maintenance and future migrations of the module, so I vote for removing them. People migrating dependent modules may be aware of this removal for switching to the core option.

carlos-lopez-tecnativa:
> > Ok, but this contradicts your previous comment.
> 
> That is kind of the nature of a mistake being made and admitting to it? No need to unnecessarily point this out.

My apologies, my comment wasn't intended to cause any discomfort. I just wanted to clarify the points.


> > My final question: do you plan to unify the options by removing the custom ones and keeping only the core-supported options, or not?
> 
> Thinking about it, it seems safer to keep them in for easier migrations on existing projects that use said options. This is just one opinion, though. If you think otherwise, I'm happy to incorporate your suggestion, if you could make the necessary code changes.

My main concern was that the core options weren’t working, and now that’s resolved. But I agree with @pedrobaeza that this is a good time to make this kind of change.

denaux:
@pedrobaeza @carlos-lopez-tecnativa 

Thank you both for your input. I'm outvoted but also convinced of your arguments, so I've removed the `create` and `create_edit` options due to Odoo having the native options `no_quick_create` and `no_create_edit`. I have also done the same with the `open` option, as Odoo has the native option `no_open`.

I've also done some code cleanup.

Feel free to test things through!

carlos-lopez-tecnativa (approved):
LGTM, thanks for the work

CRogos (commented):
Could you squash some of the commits?
image

denaux:
@CRogos 
> Could you squash some of the commits?

Done!

web_m2x_options/tests/test_ir_config_parameter.py
```diff
@@ -1,10 +1,10 @@
 # Copyright 2020 initOS GmbH.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo.tests import common
+from odoo.tests import TransactionCase
```
CRogos, line 4: Use BaseCommon as base test class to reduce overhead from tracking and boost the test suite.

CRogos (approved):
Tested and LGTM, 

minor finding on the tests

PR #3168, opened 2025-05-07
[17.0][FIX] Opining Task Menu's

Murtaza-OSI:


legalsylvain:
Hi. Thanks for your contribution.

Could you elaborate what is the bug ?
Could you fix precommit ?

Thanks !

web_m2x_options/static/src/components/form.esm.js
```diff
@@ -14,7 +14,7 @@ import {
     KanbanMany2ManyTagsAvatarField,
     Many2ManyTagsAvatarField,
 } from "@web/views/fields/many2many_tags_avatar/many2many_tags_avatar_field";
-
+import { KanbanMany2ManyTagsAvatarUserField ,ListMany2ManyTagsAvatarUserField } from "@mail/views/web/fields/many2many_avatar_user_field/many2many_avatar_user_field";
```
chienandalu, line 17: This is coming from the `mail` module, but this one depends just on `web`.

The issue is a regular one with this module and it would require glue module for this case. If I'm not wrong, in the v18 it was addressed with an alternative approach...

PR #2525, opened 2023-06-02
[16.0][FIX] web_timeline: css and styles not charged when rendering t…

Yadier-Tecnativa:
On library version update a new feature was added at Timeline creation level to allow modifying the xss. By keeping it active the css classes are not showed

As the view is on backend, has no sense to keep it active. So we deactivate the feature to allow show the timeline correctly.

TT43238

PR #3169, opened 2025-05-09
[17.0][FIX] web_timeline: Enable XSS filtering with whitelist for safer HTML rendering

carlos-lopez-tecnativa:
Before this commit, XSS filtering was disabled, which could introduce potential security risks. This commit adopts a safer approach, similar to PR https://github.com/OCA/web/pull/2525. After this commit, XSS filtering is re-enabled, and a whitelist of allowed HTML elements is provided.

This change can be tested with the `project_timeline_hr_timesheet` module, which adds custom classes. The elements are rendered correctly, and no more console warnings appear.

image

TT56064
@Tecnativa @pedrobaeza @CarlosRoca13 @JasminSForgeFlow could you please review this?

### DIFF
--- a/web_timeline/static/src/views/timeline/timeline_renderer.esm.js
+++ b/web_timeline/static/src/views/timeline/timeline_renderer.esm.js
@@ -195,7 +195,17 @@ export class TimelineRenderer extends Component {
             // Delete an item by tapping the delete button top right
             this.options.editable.remove = true;
         }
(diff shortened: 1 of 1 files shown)


CarlosRoca13 (approved):
Okay, seems to be ok :smile:

pedrobaeza (approved):
/ocabot merge patch

PR #3091, opened 2025-02-14
[18.0][MIG] web_widget_open_tab: Migration to 18.0

widadelbekkouri-odoo:

PR #3029, opened 2024-12-20
[MIG] web_widget_open_tab: Migration to 18.0

amh-mw:

PR #3173, opened 2025-05-15
[MIG] web_widget_open_tab: Migration to 18.0

CRogos:
Followup of:
https://github.com/OCA/web/pull/3091
https://github.com/OCA/web/pull/3029


CRogos:
@amh-mw could you review?

amh-mw (approved):
Code review only. LGTM.

@yostashiro @widadelbekkouri-odoo @jtl-mw Additional reviews?

MohamedOsman7 (approved):
Code & functional review LGTM

web_widget_open_tab/static/src/js/open_tab_widget.esm.js
```diff
@@ -10,21 +8,7 @@ export class OpenTabWidget extends Component {
...
-            searchParams.get("model") !== this.props.record.resModel
-        ) {
-            searchParams.set("model", this.props.record.resModel);
-            searchParams.delete("action");
-        }
-        return url.split("#")[0] + "#" + searchParams.toString();
-    }
-    loadAttrs(ev) {
-        $(ev.target).tooltip();
+        return window.location.pathname + "/" + this.props.record.data.id;
```
etobella: This is not working properly. Let me explain the case:

1- Access Apps menu
2- Refresh (this way we set the window.location.pathname
3- Access the User menu (do not press F5)
4- PRess the Open Tab widget

etobella (changes requested):

web_widget_open_tab/static/src/js/open_tab_widget.esm.js
etobella: Use this:

```
return `/odoo/${this.props.record.resModel}/${this.props.record.data.id}`;
```

etobella (changes requested):

web_widget_open_tab/static/src/js/open_tab_widget.esm.js
CRogos: I've changed it this way, because I also could not find a better solution, but the bread-crumbs and context get lost in the new tab.

etobella: Might be, but there is no better solution, as we need to use the model if the widget is aplied in a list of a on2many field

etobella (approved):

PR #2838, opened 2024-05-28
[17.0][MIG]web_sheet_full_width Migration to 17.0

Dati-lom:

PR #3180, opened 2025-05-28
[18.0][MIG] web_sheet_full_width: Migration to 18.0

MohamedOsman7:
PR v17: https://github.com/OCA/web/pull/2838


MohamedOsman7:
@anmarmo1, @zamberjo could you please make a review

anmarmo1:
I'm not sure this module is necessary yet; I don't see any noticeable changes, at least in the Community version.

web_sheet_full_width/static/src/scss/web_sheet_full_width.scss
```diff
@@ -0,0 +1,9 @@
+.o_form_view_container {
+    .o_form_sheet_bg .o_form_sheet {
+        max-width: none !important;
+        width: auto !important;
+    }
+    .o_FormRenderer_chatterContainer {
```
StefanRijnhart, line 6: Class no longer seems used after Odoo 16.

StefanRijnhart (changes requested):
Agreed with @anmarmo1. Under what circumstances does this module make a visible difference? Also, this module contains an older recommendation to use web_responsive instead, which reads like a deprecation warning. That makes a strong case to discontinue the module.

StefanRijnhart:
Gotta correct myself here: this module is effective when web_chatter_position is installed. Installing web_chatter_position and forcing the chatter downwards frees up space on the side of the form, but it is this module that allows the main form view use that space. If that is your use case as well, maybe we could consider moving the CSS into web_chatter_position instead.

MohamedOsman7:
Moved to https://github.com/OCA/web/pull/3183

anajuaristi:
@StefanRijnhart I agree with you
We are always using web_chatter_position. With this module installed, grey emply spaces are shown in any form. 

We will test the PR and let you know.
Thank you for your work!

StefanRijnhart:
@anajuaristi be sure to test the code from https://github.com/c4a8-odoo/module-oca-web/pull/17!

PR #3182, opened 2025-06-01
[UPD][14.0] web_widget_x2many_2d_matrix: license to LGPL

thomaspaulb:
Funny, I've been having discussions about this in the board before, and now it happens to me for the first time, that I'm aware of.

Issue I have:

- Customer is using [mrp_production_serial_matrix](https://github.com/OCA/manufacture/tree/14.0/mrp_production_serial_matrix) which depends on this module `web_widget_x2many_2d_matrix`. Because the matrix module was AGPL, all dependencies are, and rightfully should be.
- However, customer is also using Odoo `quality_mrp` module.
- `quality_mrp` does not play well with the matrix, so I need to make a glue module, which I think I'm not allowed to do, because it would depend both on an AGPL module and an Enterprise module, which have incompatible licenses.

Do I have any other choice than this?

### DIFF
--- a/web_widget_x2many_2d_matrix/__manifest__.py
+++ b/web_widget_x2many_2d_matrix/__manifest__.py
@@ -14,7 +14,7 @@
         "Odoo Community Association (OCA)"
     ),
     "website": "https://github.com/OCA/web",
-    "license": "AGPL-3",
+    "license": "LGPL-3",
     "category": "Hidden/Dependency",
     "summary": "Show list fields as a matrix",
     "depends": ["web"],


thomaspaulb:
@hbrunn @simahawk @alexey-pelykh @pedrobaeza

NL66278:
@thomaspaulb There is also a module called quality_mrp_oca. Did not study this in detail, but could it not be a plugin compatible module for quality_mrp?

thomaspaulb:
That will have to involve them switching fully from Enterprise quality suite to OCA quality suite. I'm not ready to bear the burden of that currently

pedrobaeza (approved):
I personally won't relicense any of my backend modules to LGPL. Anyway, I would say that the license incompatibility is the other direction, isn't it? @dreispt can say about it.

rvalyi:
I think you should also adapt the headers in the files.

@pedrobaeza you approved the PR but your comment kind of mean you would disagree. Did you approved by mistake? Did I miss something?

pedrobaeza (changes requested):
I personally won't relicense any of my backend modules to LGPL. Anyway, I would say that the license incompatibility is the other direction, isn't it? @dreispt can say about it.

rafaelbn (approved):
To increase adoption and attract more contributors, we propose changing the license of this module to LGPL, with consent from all authors.

LGPL reduces barriers for companies and SaaS providers, without compromising the spirit of open collaboration.

It’s a simple step to keep this module alive, useful, and growing.

dreispt:
Although the OCA can use the CLA to change a license (more precisely, can sublicense to an OSI-approved license), we don't want to do that against the will of the original authors.

Note that later contributions, migrations, refactors, etc. do not grant authorship.

Looking at the [history of this module](https://github.com/OCA/web/commit/e8d68ba07f5d093255172466fb0b61fbb3ca04c6#diff-53dd23ab1c797f81c4d2e3e9052fecc8d83cfd5577427368f4014d1682cd6258), it seems that the original author is @hbrunn under Therp BV.
If this is inaccurate, we can hear from other people that may claim co-authorship.

At this point, we would need this PR to be approved by a Therp representative.

dreispt:
@thomaspaulb 
> quality_mrp does not play well with the matrix, so I need to make a glue module, which I think I'm not allowed to do, because it would depend both on an AGPL module and an Enterprise module, which have incompatible licenses.

This is correct.
Is it an option that the source of the incompatibility can be resolved directly in `web_widget_x2many_2d_matrix`?

thomaspaulb:
The issue is actually with `mrp_serial_number_matrix`, which depends on this module, so both modules would need a license change.

I did find a "goat's path" though: we can make a module that depends only on `quality_mrp`, which will make it play well.

thomaspaulb:
Me being the Therp representative, I could decide now then to merge this. I'm personally leaning towards the license change because this matrix module is meant as a tool to be widely used in all kinds of dependent modules. Though it could also mean that it's going to be used as part of third party closed source modules, on the app store. Tough choice. @hbrunn what's your opinion on the matter?

dreispt:
Yes, indeed Tom. 
If Holger wrote this under contract with Therp, then Therp would hold the authorship
Let's wait a bit to hear from Holger or other people.

ivs-cetmix (changes requested):
Please update license in files as well. Eg [here](https://github.com/Therp/web/blob/e4225d97b414cb3f98186ea74e84ae105ef8dc6c/web_widget_x2many_2d_matrix/static/src/js/2d_matrix_renderer.js#L3)

hbrunn:
@thomaspaulb I typically neither approve nor block this kind of license change, so do what you feel you need to be doing

PR #3183, opened 2025-06-04
[18.0][IMP] web_chatter_position: improve form and chatter layout styles

MohamedOsman7:
As web_sheet_full_width will be deprecated (see discussion(https://github.com/OCA/web/pull/3180#issuecomment-2930126999)), i moved the layout improvements it provided to web_chatter_position as recommended by @StefanRijnhart.

### DIFF
--- a/web_chatter_position/static/src/scss/chatter_form_adjustments.scss
+++ b/web_chatter_position/static/src/scss/chatter_form_adjustments.scss
@@ -1,0 +1,9 @@
+.o_form_view_container {
+    .o_form_sheet_bg .o_form_sheet {
+        max-width: none !important;
+        width: auto !important;
+    }
+    .o-mail-Form-chatter {
+        max-width: none;
+    }
+}
--- a/web_chatter_position/README.rst
+++ b/web_chatter_position/README.rst
@@ -32,6 +32,9 @@ Configurable chatter position from the user preferences.
 
 Supports Both Community & Enterprise Edition.
 
+Extends the functionality of the web client to get full width in the
+form view sheet.
+
 **Table of contents**
 
 .. contents::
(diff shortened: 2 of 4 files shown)


anmarmo1 (approved):
Functional review is ok

Before
image


After
image

web_chatter_position/static/src/scss/chatter_form_adjustments.scss
```diff
@@ -0,0 +1,9 @@
+.o_form_view_container {
+    .o_form_sheet_bg .o_form_sheet {
+        max-width: none !important;
+        width: auto !important;
+    }
+    .o-mail-Form-chatter {
+        max-width: none;
```
StefanRijnhart, line 7: While this class name is actually in use, it does not make a difference when I test locally (on runbot, the chatter is full width, but that's probably because of web_responsive).
image

StefanRijnhart (commented):
Thanks! The form sheet is now using the margins that become available when the chatter is forced downward. ~~However, on the invoice form this causes the totals to be unaligned with the invoice lines. Is there anything you can do about that?~~
image

-- edit -- To answer my own question, that is this colspan of size 4 that fits both widgets side by side, so it's not an issue with the CSS. https://github.com/odoo/odoo/blob/5842c9f4821772fe44cfe7244162edec1ffff88e/addons/account/views/account_move_views.xml#L1261

web_chatter_position/static/src/scss/chatter_form_adjustments.scss
StefanRijnhart: I'm removing this definition in https://github.com/c4a8-odoo/module-oca-web/pull/17, along with some other suggestions.

StefanRijnhart: Any opinion on the suggestions I put in my PR for your branch?

anajuaristi (approved):

anajuaristi:
@StefanRijnhart Perfect. I've tested de new aproach, and It's quite good, since you are avoiding to install one module but you obtain almost similar functionality installing only this one.

I like the aproach, so 👍 LGTM (functionally tested)

StefanRijnhart:
Thanks for checking @anajuaristi. Assuming you tested the code from https://github.com/c4a8-odoo/module-oca-web/pull/17 as I suggested, I'm superseding this PR with https://github.com/OCA/web/pull/3272

PR #2923, opened 2024-08-28
[16.0][IMP] web_notify:t3894 Allow notify with sound

GabbasovDinar:
These changes allow you to send notifications with an audible alarm.

PR #3184, opened 2025-06-04
[16.0][IMP] web_notify: license change

Bearnard21:
We are using this module extensively in our projects. Moreover, we have recently contributed to it by adding another [useful feature](https://github.com/OCA/web/pull/2923)

However, recently we encountered an issue in a new project, where we were unfortunately not allowed to use AGPL-3 licensed modules.

This is a bit disgusting because this will force us to create a dedicated module just for this project instead of putting those efforts towards the OCA contribution.

As a possible solution, we would like to suggest changing the license of this module to LGPL-3.

Why? This is a helper module which doesn't implement any business logic. At the same time, it is really a good module for a so-called "OCA addiction" which, having a less restrictive license, could attract both users and contributors to the OCA

I'd like to kindly ask the contributors to approve the PR. Thank you.

@lmignon
@aitorbouzas


ivs-cetmix (approved):
Speaking not as the owner of Cetmix, but as the OCA member:
I think this is a right way to strengthen the OCA position by centralising and consolidation the entire Odoo (non the SA) ecosystem around the OCA. 
Would appreciate your opinion on this @OCA/board @simahawk  @sbidoul @pedrobaeza @rvalyi @dreispt

dreispt:
See [this similar discussion](https://github.com/OCA/web/pull/3182#issuecomment-2940075984).

sbidoul (approved):

dreispt:
@mike-cetmix In this case the [original author](https://github.com/OCA/web/commit/7b6096266fb5a677f2fd378ae728d1e8a33f7197#diff-c8828560a49458be3c9c2da95e9af252e705d6fe8661c0b4d6a37e08e99bc8cd) seems to be ACSONE, I believe @sbidoul can represent it, and he just approved the change.

ivs-cetmix:
> @mike-cetmix In this case the [original author](https://github.com/OCA/web/commit/7b6096266fb5a677f2fd378ae728d1e8a33f7197#diff-c8828560a49458be3c9c2da95e9af252e705d6fe8661c0b4d6a37e08e99bc8cd) seems to be ACSONE, I believe @sbidoul can represent it, and he just approved the change.

@dreispt thank you! Should we wait for the others to approve it? Or you can simply merge it yourself?

dreispt:
Let's give it the regular "dust settling" period for PRs, to give opportunity for other people to comment.

yajo (approved):

Shide (approved):

lmignon:
:+1:

lmignon (approved):

sbidoul:
@mike-cetmix would you like to forward port?

ivs-cetmix:
> @mike-cetmix would you like to forward port?

@sbidoul thank you, yes we will fp to all the other versions today

PR #3185, opened 2025-06-05
[16.0][MIG] web_responsive_company_color: Migration to 16.0

anusriNPS:
Migrating PR: https://github.com/OCA/web/pull/2861

Module web_responsive_company_color is introduced for customizing web elements that are part of web_responsive that needs customization


anusriNPS:
navbar_app_menu background is customized as part of web_responsive_company_color which uses color from company_style customized for the company:

image

Uninstalling web_responsive_company_color removes customized web elements handled from it. navbar_app_menu background customization is removed

image

web_company_color/models/res_company.py
SirPyTech, line 1: **chore**: Please move this refactoring to a dedicated `[REF]` commit before the other commits of this PR.
This is because commits should affect only one module as much as possible, see:
> Avoid commits which simultaneously impact lots of modules.
Try to split into different commits where impacted modules are different.
This is helpful if we need to revert changes on a module separately.

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message)

web_responsive_company_color/models/ir_module.py
```diff
@@ -0,0 +1,17 @@
+from odoo import models
+
+
+class Module(models.Model):
+    _inherit = "ir.module.module"
+
+    def button_immediate_install(self):
+        res = super().button_immediate_install()
+        self.env["res.company"].search([]).scss_create_or_update_attachment()
```
SirPyTech: **question**: Why do we have to `scss_create_or_update_attachment` every time _any_ module is installed? Based on the README, maybe this has to be done only when `web_responsive` is updated.

Same concern also for the `uninstall` method below.

web_responsive_company_color/README.rst
```diff
@@ -0,0 +1,87 @@
...
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+Simply install the app. See web_company_color for instructions on how to customize colors.
+
+If new elements are included for customization under scss_template, one needs to reinstall the app to view customized colors.
```
SirPyTech: **question**: Do you mean when someone modifies the string returned by method `res.company._get_scss_template`?

Please give more details about this issue because it's very strange that the module has to be updated when something changes: if the failing use-case is clear maybe we can fix it before the PR is merged.

web_responsive_company_color/models/ir_module.py
```diff
@@ -0,0 +1,17 @@
...
+        return res
+
+    def button_immediate_uninstall(self):
+        res = super().button_immediate_uninstall()
+        self.env["res.company"].search([]).with_context(
+            uninstall_scss=True
```
SirPyTech: **thought**: This context value is not used in the module, so I don't think it should be in this module, but it does no harm and could be useful for downstream methods to know that we are uninstalling.

It would be better if downstream modules that need it could find another way to recognize this situation, like:
```py
    def button_immediate_uninstall(self):
        return super(self.with_context(...), ClassName).button_immediate_uninstall()
```

web_responsive_company_color/__manifest__.py
```diff
@@ -1,16 +1,17 @@
 {
     "name": "Web Responsive Company Color",
     "summary": "Styling hooks for web_responsive elements",
-    "version": "14.0.1.0.0",
+    "version": "16.0.1.0.0",
     "category": "Web",
     "website": "https://github.com/OCA/web",
-    "author": "PyTech SRL, Alessandro Uffreduzzi, Odoo Community Association (OCA)",
+    "author": "Anusri Veerappan Prakasam, Odoo Community Association (OCA)",
```
SirPyTech: **chore**: Usually you only add yourself as a contributor when you work on a module, but I think that you worked on this from the beginning so this is fine, just don't remove the other authors.

web_responsive_company_color/models/res_company.py
```diff
@@ -0,0 +1,20 @@
...
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    def _get_scss_template(self):
+        return (
+            super()._get_scss_template()
+            + """
+        .o_navbar_apps_menu .dropdown-menu-custom {
+            background: url('/web_responsive/static/img/home-menu-bg-overlay.svg'),
```
SirPyTech: **question**: I can't find this image in https://github.com/OCA/web/tree/9bfd8c348ad94233606aa04954bdb7d21b640a8e/web_responsive/static/img, how is it being loaded?

SirPyTech (changes requested):
Thanks for the PR!

web_company_color/models/res_company.py
anusriNPS: Seperate commit created for web_company_color module.

web_responsive_company_color/models/ir_module.py
anusriNPS: Agree the approach leads to calling of "scss_create_or_update_attachment()" method every time each module is installed, which was missed by me to address in 14.0 when implemented.

**Objective**: When a module adds new elements to the SCSS template, the added modules is customized only when the method scss_create_or_update_attachment() is invoked. The same needs to be taken care when module is uninstalled (to remove elements introduced by module)

This is achieved using hooks(post install and uninstall) in web_company_color.

Approaches:
  1. Using same hooks in every module(web_responsive_company_color, other modules which are introduced based on this bridge module) becomes duplication of code and in order to remove redundant code, used ir_module button_install() in 14.0 and adpated for 16.0 here. - This definitely have the flaw
  2. Tried to use JS Component to call the scss_create_or_update_attachment() whenever module  gets installed/uninstalled based on rpc query - Since component is not used anywhere in templates as such, component is not getting loaded.
  3. Add an eventListener to a bus in this bridge module and send an event from backend using "bus.bus" module whenever install/uninstall of module happens so that we can call "scss_create_or_update_attachment()" for customization to take effect from bridge module and its dependencies. - I have not tried implementation of this approach as I am not sure it's a right way.
  
  Your inputs would be helpful. Please let me know your views.

web_responsive_company_color/README.rst
anusriNPS: Given details of the issue faced in the above comment.

web_responsive_company_color/models/ir_module.py
anusriNPS: Once approach is concluded, we try to address it if required

web_company_color/models/res_company.py
SirPyTech: :+1:

web_responsive_company_color/models/ir_module.py
SirPyTech: Thanks for clarifying the issue and proposing a few approaches.
In my opinion you cannot ask the user to remember to reinstall the module every time a string is changed, and calling this method every time _any_ module is installed is overkill because most modules won't need it.
So I'd go with the installation hooks (that is `web_company_color`'s approach).

I appreciate very much your concern about code duplication but that is why functions/methods exist!
We are not duplicating the logic: all the logic has already been carefully extracted in the method `res.company.scss_create_or_update_attachment`, then this method is called wherever it is needed (in hooks of modules that edit the SCSS).

anusriNPS: I have updated using hooks as discussed earlier and removed using "ir.module" logic based on "button_install and button_uninstall".

web_responsive_company_color/README.rst
anusriNPS: As method `_get_scss_template()` is invoked only during `post_init_hook` and `uninstall_hook`, hence when an new element is inserted inside method `get_scss_template()` to customize styling,  one need to take care of migrartion in which `post_init_hook()` needs to be invoked, which makes customization to take effect. So, removed comment mentioned earlier in the usage. Also, I would like to know whether this information should be updated as part of documentation.

Please let me know your thoughts.

web_responsive_company_color/models/ir_module.py
anusriNPS: Not using `ir.module` with this implementation. Hence, omitted this comment.

web_responsive_company_color/models/res_company.py
anusriNPS: Updated with correct url path: /web_responsive/static/src/img/home-menu-bg-overlay.svg

web_responsive_company_color/__manifest__.py
SirPyTech: 👍

web_responsive_company_color/__manifest__.py
```diff
@@ -1,17 +1,18 @@
...
     "summary": "Styling hooks for web_responsive elements",
-    "license": "AGPL-3",
-    "author": "Anusri Veerappan Prakasam, Odoo Community Association (OCA)",
-    "website": "https://github.com/OCA/web",
+    "version": "16.0.1.0.0",
     "category": "Web",
-    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/web",
+    "author": "Anusri Veerappan Prakasam, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
```
SirPyTech: **polish**: Please keep the lines in the order they were so it's clear from the diff that only the version has changed.

web_responsive_company_color/models/ir_module.py
SirPyTech: Thanks! Now this is much more convenient for the user :clap:

web_responsive_company_color/README.rst
SirPyTech: So if a developer adds a new element in the string returned by `_get_scss_template`, they will have to create a migration to call `scss_create_or_update_attachment` (or `_get_scss_template`) as done in the `post_init` hook.

Yes please add it into a comment in the `_get_scss_template` method: it is something that is only interesting for developers so it can stay in the code in my opinion.

web_responsive_company_color/models/ir_module.py
SirPyTech: :+1:

web_responsive_company_color/models/res_company.py
SirPyTech: Now it links to https://github.com/OCA/web/blob/6d3312e9b47a23bf695b48169793775452467f24/web_responsive/static/src/img/home-menu-bg-overlay.svg.
:+1:

SirPyTech (changes requested):
I tried the module and reviewed the code, it looks good to me 👍

**chore**: Please edit the commit history, now the migration commit si duplicated: 

Image

Looking at their content, I think it's better to merge them together into only one commit, but let me know what you think!
If they have to stay on their own please give a more appropriate title to the 2nd one.

web_responsive_company_color/README.rst
anusriNPS: Updated comment above the method "_get_scss_template()"

anusriNPS:
> I tried the module and reviewed the code, it looks good to me 👍
> 
> **chore**: Please edit the commit history, now the migration commit si duplicated:
> Image
> 
> Looking at their content, I think it's better to merge them together into only one commit, but let me know what you think! If they have to stay on their own please give a more appropriate title to the 2nd one.

Apologies.It was not intentional to add two commits, merged into one

web_responsive_company_color/__manifest__.py
anusriNPS: Only modified required changes in migration commit of 16.0 and all others are kept untouched.

SirPyTech: :+1:

web_responsive_company_color/README.rst
SirPyTech: :+1:

SirPyTech (approved):

quirino95 (approved):
Code and functional review: LGTM!

PR #3190, opened 2025-06-08
[16.0][FIX] web_responsive: Do not duplicate buttons

etobella:
Duplicating buttons will give problems on other modules that want to add buttons on the chatter

With this change:

image
(you can scroll to the right)

Before:

image

Before looked better. However, it came with the handicap of removing extensibility of this specific component.

### DIFF
--- a/web_responsive/__manifest__.py
+++ b/web_responsive/__manifest__.py
@@ -39,6 +39,7 @@
             "/web_responsive/static/src/components/hotkey/hotkey.xml",
             "/web_responsive/static/src/components/chatter_topbar/chatter_topbar.esm.js",
             "/web_responsive/static/src/components/chatter_topbar/chatter_topbar.xml",
+            "/web_responsive/static/src/components/chatter_topbar/chatter_topbar.scss",
             "/web_responsive/static/src/components/attachment_viewer/attachment_viewer.scss",
             "/web_responsive/static/src/components/attachment_viewer/attachment_viewer.esm.js",
             "/web_responsive/static/src/components/attachment_viewer/attachment_viewer.xml",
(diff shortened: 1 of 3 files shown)


ValentinVinagre:
Error reference: https://github.com/OCA/ai/pull/2#discussion_r2134757238

pedrobaeza:
Uhm, but it was done this way to have the buttons without doing side scroll. That's the key...

etobella:
yes, but it collides with other modules (not only ai, also a module in mangament-system for example)

We need to avoid duplication. In any case, now it is a problem.

ValentinVinagre:
But then the inheritance breaks down. IMO, another way of doing it should be considered if necessary.

ValentinVinagre (approved):
The change may affect other modules.

etobella:
@pedrobaeza An easy option would be to add a small SCSS that will allow this:

image

It is not as nice as it was before, but that should work.

Does it work for you?

pedrobaeza:
What about changing the priority of the template for applying the change at the end to try to not interfere?

etobella:
Because it will affect mobile view then.... We need to find a solution without duplicating, and odoo didn't helped with the position of the items

pedrobaeza:
Can you use `position="move"` then?

etobella:
With some engineering I can do something like this:

image

I can try to make it centered....

Cons: a line before and under activities is lost in desktop

Before:

image

After:

image

pedrobaeza:
Looking at enterprise runbot, they are putting everything in one line:

imagen


Let's do the following:

- Let it everything in one line
- Show only activity icon for reducing horizontal space.
- Maybe also put "Send message" and "Log note" as icons to check how it's seen.

etobella:
image

WDYT?

pedrobaeza (approved):

pedrobaeza:
Let's apply the same on upper versions

pedrobaeza:
Actually it seems we already took the same path in 17+

etobella:
Yes, there is no need to forward port the change. When doing the migration to 17, we (as a community) removed the duplication :tada:

PR #3191, opened 2025-06-09
[18.0][FIX] web_m2x_options: prevent crash on Many2OneReferenceField

sbejaoui:
`Many2OneReferenceField` inherits the props definition from `Many2OneField` but does not inherit the prototype methods added via `patch` on `many2OneField`, including `m2o_options_props`

without this patch, calling `extractProps` on `Many2OneReferenceField` leads to `"this.m2o_options_props is not a function"` errors because this method is undefined

this fix adds a no-op `m2o_options_props` implementation on `Many2OneReferenceField` to ensure compatibility and safely bypass the `m2o_options_props` logic since it's not relevant for reference fields

to reproduce this bug, open any view with many2one reference field, example attachment form view

image

### DIFF
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -105,11 +105,11 @@ create rights)
 To add these parameters go to Configuration -> Technical -> Parameters
 -> System Parameters and add new parameters like:
 
-- web_m2x_options.create: False
-- web_m2x_options.create_edit: False
-- web_m2x_options.limit: 10
-- web_m2x_options.search_more: True
-- web_m2x_options.field_limit_entries: 5
+-  web_m2x_options.create: False
+-  web_m2x_options.create_edit: False
+-  web_m2x_options.limit: 10
+-  web_m2x_options.search_more: True
+-  web_m2x_options.field_limit_entries: 5
 
 Example
 -------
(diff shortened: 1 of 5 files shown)


sbejaoui:
cc/ @glitchov ,

carlos-lopez-tecnativa (changes requested):
These changes prevent the error, but the Many2OneReferenceField is not working with the options.
Please check and add support to make it work correctly.

image

sbejaoui:
> These changes prevent the error, but the Many2OneReferenceField is not working with the options. Please check and add support to make it work correctly.
> 

I can't work on this right now. if someone else is available to take it over, I'd appreciate it
if not, I propose we merge it as-is with a FIXME comment in the code and document it as a known issue — otherwise, the view doesn't open at all in its current state

sbiosca-s73 (approved):
LGTM

Botakely:
LGTM

CasVissers-360ERP (approved):

StefanRijnhart:
@sbejaoui Please add the roadmap entry so this can be merged quickly, given the impact of the bug.

sbejaoui:
> @sbejaoui Please add the roadmap entry so this can be merged quickly, given the impact of the bug.

done

StefanRijnhart (approved):
@carlos-lopez-tecnativa can you live with this for now?

carlos-lopez-tecnativa (approved):

PR #3192, opened 2025-06-09
[18.0][IMP] web_responsive: Quick Record Search

mohammedshahil:
## Usage

1. Open the command palette (usually with `Ctrl+K`).
2. Type the model name and search term, e.g.:
   - `/res.partner John`
   - `/sale.order S00001`
   - `/product.product Chair`
3. Matching records will be listed. Click a record to open it.

https://github.com/user-attachments/assets/ee09006b-b03d-47ca-9207-8bd3763a93c0


pedrobaeza:
Thanks for this new feature. It seems interesting, but I wouldn't say this is something for the `web_responsive` module. In fact, I think this will work even without installing this module, as the command palette is a native feature, so you should extract it in a module `web_command_palette_model_search` or similar name. Check also the pre-commit.

web_responsive/static/description/index.html
```diff
@@ -497,6 +500,7 @@ <h2><a class="toc-backref" href="#toc-entry-6">Contributors</a></h2>
 <li>Taras Shabaranskyi &lt;<a class="reference external" href="mailto:shabaranskij&#64;gmail.com">shabaranskij&#64;gmail.com</a>&gt;</li>
 <li>Angel Patel &lt;<a class="reference external" href="mailto:patelangel1414&#64;gmail.com">patelangel1414&#64;gmail.com</a>&gt;</li>
 <li>Mohamed Alkobrosli &lt;<a class="reference external" href="mailto:malkobrosly&#64;kencove.com">malkobrosly&#64;kencove.com</a>&gt;</li>
+<li>Mohammed Shahil &lt;<a class="reference external" href="mailto:mohammedshahilmp&#64;gmail.com">mohammedshahilmp&#64;gmail.com</a>&gt;</li>
```
ivs-cetmix, line 503: This file is auto-generated. Please modify the [CONTRIBUTORS](https://github.com/mohammedshahil/web/blob/18.0-web_responsive-quick_record_search/web_responsive/readme/CONTRIBUTORS.md) file instead. Check the example [here](https://github.com/OCA/maintainer-tools/tree/master/template/module/readme).

web_responsive/static/img/quick_record_search.gif
ivs-cetmix, line 1: That's really good that you have added a screenshot! Please update the [USAGE](https://github.com/mohammedshahil/web/blob/18.0-web_responsive-quick_record_search/web_responsive/readme/USAGE.md) file and add the description of how this feature works in it. You can also embed this gif in it.

ivs-cetmix (changes requested):
Hi @mohammedshahil, thank you for your contribution!
Please check the comments below.

web_responsive/static/img/quick_record_search.gif
pedrobaeza: @ivs-cetmix check my comment about the convenience of this feature being here: https://github.com/OCA/web/pull/3192#issuecomment-2956070103

ivs-cetmix: > @ivs-cetmix check my comment about the convenience of this feature being here: [#3192 (comment)](https://github.com/OCA/web/pull/3192#issuecomment-2956070103)

@pedrobaeza yup, that makes sense. I see two options:
- Make search in the same search string that is used now to search for apps (main screen). Then we need to update the web_responsive
- Leave it "as is". In this case yes, this should be a separate module.

@mohammedshahil what do you think?

mohammedshahil: > > @ivs-cetmix check my comment about the convenience of this feature being here: [#3192 (comment)](https://github.com/OCA/web/pull/3192#issuecomment-2956070103)
> 
> @pedrobaeza yup, that makes sense. I see two options:
> 
>     * Make search in the same search string that is used now to search for apps (main screen). Then we need to update the web_responsive
> 
>     * Leave it "as is". In this case yes, this should be a separate module.
> 
> 
> @mohammedshahil what do you think?

yes separate module is better

PR #3196, opened 2025-06-12
[14.0][FIX] web_company_color: colors not changed

bjkumal:
It only worked when all the colors were set. Even if just one was missing, it wouldn’t work.


eantones (commented):
@bjkumal I found a bug when installing in production, please chek it out 7eff8d28c01944131cdf59d869903ef1250bb027 the default values of the text were wrong, I solved it. Next time do more tests to be sure everything works correctly. Thanks

PR #3199, opened 2025-06-17
[16.0][ADD] web_sort_menu: Module web_sort_menu

anusriNPS:
Module web_sort_menu is introduced to sort apps in alphabetical order while displayed on NavBar


anusriNPS:
image

web_sort_menu/static/src/xml/sort_app_menu.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates id="template" xml:space="preserve">
+   <t
+        t-name="web_sort_menu.sortMenu"
+        t-inherit="web.NavBar"
+        t-inherit-mode="extension"
+        owl="1"
+    >
+         <xpath expr="//header[@class='o_navbar']//t/t" position="replace">
```
SirPyTech: **suggestion**: Please use a more focused `xpath`: if you want to find the element that has `t-set="apps"`, then you can use `t[@t-set='apps']` in the `xpath`.
Searching for `//t/t` is difficult to maintain because it is not clear which node you are looking for.

SirPyTech: **suggestion**: `replace` is risky because other modules might expect that node to be present, so it should be avoided as much as possible.
In this case I suggest to go `after` the node that sets the `apps` variable and substitute its value with the sorted list.

web_sort_menu/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Web Sort Menu",
+    "summary": "Sorts Apps in NavBar Menu alphabetically",
+    "version": "16.0.1.0.0",
+    "category": "Web",
+    "author": "Anusri Veerappan Prakasam, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    "license": "AGPL-3",
+    "depends": ["web", "web_responsive"],
```
SirPyTech: **suggestion**
```suggestion
    "depends": ["web",],
```
I can't find where the dependency from `web_responsive` is needed, could you please check if that's actually necessary?
Without `web_responsive`, I expect that the dropdown menu is sorted.

web_sort_menu/static/src/js/sort_app_menu.esm.js
```diff
@@ -0,0 +1,17 @@
...
+import {patch} from "@web/core/utils/patch";
+
+patch(NavBar.prototype, "web_sort_menu.sortmenu", {
+    setup() {
+        this._super();
+        this.sortedApps = [];
+
+        onWillStart(() => {
+            const apps = this.menuService.getApps();
+            this.sortedApps = apps.sort((a, b) => a.name.localeCompare(b.name));
```
SirPyTech: **note**: It is worth noting that apps are sorted based on the translated name, so users with different languages will see them in a different order

web_sort_menu/static/src/js/sort_app_menu.esm.js
```diff
@@ -0,0 +1,17 @@
...
+import {onWillStart} from "@odoo/owl";
+import {patch} from "@web/core/utils/patch";
+
+patch(NavBar.prototype, "web_sort_menu.sortmenu", {
+    setup() {
+        this._super();
+        this.sortedApps = [];
+
+        onWillStart(() => {
+            const apps = this.menuService.getApps();
```
SirPyTech: **thought**: I don't think this search is _that_ expensive, but we could avoid it by sorting directly in the template, or exposing a method and calling it in the template.
This would also preserve any edit that other modules have done to the `apps` variable in the template.

web_sort_menu/static/src/xml/sort_app_menu.xml
SirPyTech: **issue**: The usage of `@class` is discouraged (see https://github.com/odoo/odoo/commit/cbffd99341b5c9229e7ca75eb9ee328d26b4f084), please use `hasclass` instead.

SirPyTech (changes requested):
Thanks for the PR!
It is a very clean approach, please see the comments below.

web_sort_menu/static/src/xml/sort_app_menu.xml
anusriNPS: Thank you for the input. Updated as suggested.

anusriNPS: Updated using "attributes"

web_sort_menu/__manifest__.py
anusriNPS: I missed to think about dropdown menu getting sorted, I thought only NavBar would be sorted hence added web_responsive in the dependency.

I see dropdown menu getting sorted, hence removed web_responsive module as dependency.

web_sort_menu/static/src/js/sort_app_menu.esm.js
anusriNPS: Yes, based on different languages sorting happens according to it. 

Updated Description of module with this information.

web_sort_menu/static/src/xml/sort_app_menu.xml
anusriNPS: Updated using hasclass

web_sort_menu/static/src/js/sort_app_menu.esm.js
anusriNPS: I agree with you that we preseve any other module editing the value of apps,.

I tried to implement this search explicitly in the sort menu. However, I see we would need to use menuService in order to get information about available apps to sort them. If so, we could directly use the function exposed in menuService to know apps information.

Please let me know your views on it.

web_sort_menu/static/src/js/sort_app_menu.esm.js
SirPyTech: **praise**: Thanks for updating the documentation, this is something that developers rarely do in response to this kind of comments.
:+1:

web_sort_menu/static/src/js/sort_app_menu.esm.js
SirPyTech: I meant that you could create a new method in this patched class like
```js
sortApps((apps) => {
    return apps.sort((a, b) => a.name.localeCompare(b.name));
}),
```
and then in the template just add a call to this method:
```xml
<xpath expr="<select the t-set="apps" node>" position="after">
    <t t-set="apps" t-value="sortApps(apps)"/>
</xpath>
```
since the method is really simple, everything could be in the template too, but I think it's better to keep the code where code should be (JS in this case) and let the templates handle only the rendering as much as possible.

web_sort_menu/static/src/xml/sort_app_menu.xml
SirPyTech: :+1:

SirPyTech: :+1:

web_sort_menu/__manifest__.py
SirPyTech: :+1:

web_sort_menu/static/src/xml/sort_app_menu.xml
SirPyTech: :+1:

SirPyTech (approved):
Code and functional review, looks good to me.

I have just added an example for the improvement I suggested in https://github.com/OCA/web/pull/3199#discussion_r2153920638, but it's already good for merge in my opinion.

anusriNPS:
Suggestion was super good and precise, so implemented the same.

SirPyTech (approved):

HekkiMelody (approved):

andreampiovesana (approved):
LGTM, merge?

web_sort_menu/static/description/icon.png
ivs-cetmix, line 1: Please remove this icon, it will be added automatically on merge. Or use the actually one from

ivs-cetmix (changes requested):
Hey @anusriNPS thank you for your contribution! Some small change and LGTM 😉

web_sort_menu/static/description/icon.png
anusriNPS: Removed the icon as suggested.

ivs-cetmix:
@anusriNPS thank you for prompt action 😉 
/ocabot merge nobump

PR #3174, opened 2025-05-16
[17.0][FIX]: web_widget_x2many_2d_matrix: limit options values to x2many fields.

DavidJForgeFlow:
Created from issue: https://github.com/OCA/web/issues/3167

@SodexisTeam @BT-mlopez

PR #3200, opened 2025-06-18
[18.0][FIX]: web_widget_x2many_2d_matrix: limit options values to x2m…

jcadhoc:
…any fields.

Improve proposed by @DavidJForgeFlow.

Already merged in 17.0

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm.js
+++ b/web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm.js
@@ -1,7 +1,7 @@
 import {Component, onWillUpdateProps} from "@odoo/owl";
-import {registry} from "@web/core/registry";
 import {Domain} from "@web/core/domain";
 import {evaluateExpr} from "@web/core/py_js/py";
+import {registry} from "@web/core/registry";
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why this limit?

DavidJForgeFlow:
Because we only tested modules with the changes in the options into many2one fields. And in [the original PR](https://github.com/OCA/web/pull/3174) I was fixing an error caused by a many2many field getting attributes it won't support. Limiting the options to many2one ensures that no other field type would be affected and, if in a future someone needs this to work on any other type, can simply test and include the type in the method.

pedrobaeza:
@hbrunn should say if this limitation was something in mind when he created the module.

DavidJForgeFlow:
The options feature in v17 and ahead was included by myself and it was only taken into account to many2one fields. I agree that maybe original intention by @hbrunn was getting all the field attributes included by default (as it was happening after the migration to OWL). If his intention is to have all field types being able to get options, first someone would need to adapt how the matrix fields get all the attributes as right now only takes the ones we tested to be able to get. So maybe we could create an issue to don't loose any track of this and as this same features are already in v17.0 we could go ahead with this, what do you think @pedrobaeza ?

pedrobaeza:
OK on my side.

DavidJForgeFlow (approved):

zamberjo (approved):

PR #3203, opened 2025-06-20
[18.0][FIX] web_widget_x2many_2d_matrix: restore pre-v16 functionality

hbrunn:
https://github.com/OCA/web/pull/3124 shouldn't have been merged. Besides breaking most use cases, it only fixed some symptom of the underlying issue that the migration dropped the ability to configure the value field like you'd do it everywhere else. And on top of that didn't describe the non-standard functionality in the readme.

So this PR restores the ability to say

```xml
   <field widget="x2many_2d_matrix" ... field_value="value_field">
       <list>
           ...
           <field name="value_field" options="{}" domain="[]" etc/>
       </list>
   </field>
```

While being on it, I also restored the functionality for clickable x/y axes when they are many2one fields, and added myself as maintainer.

All of this with tests to be sure subsequent changes don't mess with existing functionality.

PR #3204, opened 2025-06-24
[18.0][FIX]: web_widget_x2many_2d_matrix: limit options values to x2many fields.

jcadhoc:
Only for testing for now, if it is okey we change the state of PR


hbrunn:
@jcadhoc I've implemented the right way to do this in https://github.com/OCA/web/pull/3203, I strongly recommend to review this instead of continuing this PR

PR #3205, opened 2025-06-25
[17.0][FIX] web_m2x_options: prevent crash on Many2OneReferenceField

Maarten-Van-Briel:
without this patch, calling extractProps on `Many2OneField` with type "many2one_reference" to "this.m2o_options_props is not a function" errors because this method is undefined

this fix adds a no-op m2o_options_props implementation on `Many2OneField` with type "many2one_reference" to ensure compatibility and safely bypass the m2o_options_props logic since it's not relevant for reference fields

to reproduce this bug, open any view with many2one reference field, example automation rule form view with a "**Vallues Updated**"-trigger
image

### DIFF
--- a/web_m2x_options/static/src/components/form.esm.js
+++ b/web_m2x_options/static/src/components/form.esm.js
@@ -176,6 +176,12 @@ patch(many2OneField, {
             {attrs, context, decorations, options, string},
             dynamicInfo
         );
+        // FIXME: field of type === "many2one_reference" does not support m2o_options_props.
+        // This no-op prevents crashes, but proper option support is still missing.
+        // See roadmap note in PR #3205
+        if (!this.m2m_options_props) {
+            return props;
+        }
         const new_props = this.m2o_options_props(props, attrs, options);
         return new_props;
     },
--- a/web_m2x_options/README.rst
+++ b/web_m2x_options/README.rst
@@ -147,6 +147,8 @@ verify your installation.
 -  Instead of making the tags rectangle clickable, I think it's better
    to put the text as a clickable link, so we will get a consistent
    behaviour/aspect with other clickable elements (many2one...).
+-  Properly support web_m2x_options on field with type ===
+   "many2one_reference".
 
 Bug Tracker
 ===========
--- a/web_m2x_options/readme/ROADMAP.md
+++ b/web_m2x_options/readme/ROADMAP.md
@@ -7,3 +7,4 @@ your installation.
 - Instead of making the tags rectangle clickable, I think it's better to
   put the text as a clickable link, so we will get a consistent
   behaviour/aspect with other clickable elements (many2one...).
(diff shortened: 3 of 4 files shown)


web_m2x_options/static/src/components/form.esm.js
```diff
@@ -171,11 +171,14 @@ patch(many2OneField, {
...
     },
-    extractProps({attrs, context, decorations, options, string}, dynamicInfo) {
+    extractProps({attrs, context, decorations, options, string, type}, dynamicInfo) {
         const props = super.extractProps(
             {attrs, context, decorations, options, string},
             dynamicInfo
         );
+        if (type === "many2one_reference") {
+            return props;
+        }
```
CRogos: How about solving this more generic?
```suggestion
        if (!this.m2m_options_props){
            return props;
        }
```

CRogos (commented):
I agree that this solves the issue, but I don't understand how this can happen? Is the extractProps mothod patched on an object where it should not be patched?

CRogos (commented):
LGTM, but could you squash both commits?

CRogos (approved):

Maarten-Van-Briel:
@CRogos when will this PR be merged?

CRogos:
@carlos-lopez-tecnativa, @bosd a second reviewer is needed.

carlos-lopez-tecnativa:
Why not backport this PR https://github.com/OCA/web/pull/3191?

Maarten-Van-Briel:
@carlos-lopez-tecnativa because in 17.0 there is no Many2OneReferenceField class like in 18.0 to inherit

**17.0:** https://github.com/odoo/odoo/blob/17.0/addons/web/static/src/views/fields/many2one_reference/many2one_reference_field.js
**18.0:**:  https://github.com/odoo/odoo/blob/18.0/addons/web/static/src/views/fields/many2one_reference/many2one_reference_field.js

carlos-lopez-tecnativa (changes requested):
> @carlos-lopez-tecnativa because in 17.0 there is no Many2OneReferenceField class like in 18.0 to inherit
> 
> **17.0:** https://github.com/odoo/odoo/blob/17.0/addons/web/static/src/views/fields/many2one_reference/many2one_reference_field.js **18.0:**: https://github.com/odoo/odoo/blob/18.0/addons/web/static/src/views/fields/many2one_reference/many2one_reference_field.js

Ok, I tested it and the error seems to be there, but this PR only hides the error and does not properly add the support, as I mentioned in the comment: https://github.com/OCA/web/pull/3191#pullrequestreview-2924501960
So, to avoid blocking the merge, please add it to the roadmap as was done in V18.

carlos-lopez-tecnativa (approved):

CRogos:
@pedrobaeza could you merge?

PR #3209, opened 2025-06-27
[FIX] web_widget_x2many_2d_matrix: don't break non-many2one value fields

david-s73:
fixes OCA#3167

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm.js
+++ b/web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm.js
@@ -144,11 +144,6 @@ export class X2Many2DMatrixRenderer extends Component {
         );
     }
 
-    // More options can be included in the future, here are only added the ones tested initially.
-    _canOptions() {
-        return ["many2one"].includes(this.list.fields[this.matrixFields.value].type);
-    }
-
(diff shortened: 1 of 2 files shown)


david-s73:
@pedrobaeza When you have a moment, can you check the pull request?

pedrobaeza:
I can't say about this one. It should be @hbrunn or others (ForgeFlow I think) involved more in this module.

hbrunn:
I've proposed https://github.com/OCA/web/pull/3203 weeks ago that fixes the problem on a fundamental level, which you should backport instead of heaping more ad-hoc code on the preexsisting ad-hoc code.

zamberjo:
I agree with @hbrunn, could you see about making a backport of his PR?

david-s73:
@hbrunn @zamberjo I have already added the changes you have asked for

hbrunn:
but did you also test if what you did works? Hint: It doesn't, so I find reviewing this as it is pointless currently.

Also note that the tests from v18 won't work in v17, so this part will need to be rewritten. I would accept dropping them altogether for now, as long as the v17 version comes into a usable state soon.

david-s73:
@hbrunn I prefer to leave only the first commit he made since it solved the error that popped up.

hbrunn:
well I think it's a tragedy that nowadays so many people aren't interested in making an actually good contribution, but just want to scratch their very specific itch. So I did it myself in https://github.com/OCA/web/pull/3219 and suggest to close this PR

PR #3211, opened 2025-07-01
[18.0][FIX] web_session_auto_close: Increase default timeout

pedrobaeza:
The previous time was ridiculously small, affecting also any test on runboat where it's installed by default, but also not being a real case for anyone having to start session again after 15 seconds of inactivity.

Let's adjust it to a more reasonable 10 minutes delay.

@Tecnativa

### DIFF
--- a/web_session_auto_close/controllers/main.py
+++ b/web_session_auto_close/controllers/main.py
@@ -11,6 +11,6 @@ class WebSessionAutoCloseController(http.Controller):
         timeout_sec = (
             request.env["ir.config_parameter"]
             .sudo()
-            .get_param("web_session_auto_close.timeout", 15)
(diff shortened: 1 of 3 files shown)


rvalyi (approved):

carlos-lopez-tecnativa (changes requested):
Please change the default value when the system parameter does not exist
https://github.com/Tecnativa/web/blob/9250d4e700baff6293a565c2075cf07af3aca8fa/web_session_auto_close/controllers/main.py#L14

pedrobaeza:
@carlos-lopez-tecnativa done

carlos-lopez-tecnativa (approved):

victoralmau (approved):

PR #3212, opened 2025-07-03
[18.0][FIX] web_m2x_options: Handle undefined m2m_options_props in many2one_reference

CRogos:
inspired by: https://github.com/OCA/web/pull/3205#pullrequestreview-2983192860

Messages cannot be opened in Odoo 18. Same error like PR above.
image


CRogos:
In think this already solved the issue: https://github.com/OCA/web/pull/3191

PR #3214, opened 2025-07-09
[15.0][FIX] web_responsive: Remove test patching that is breaking tests

CarlosRoca13:
This patch is breaking tours that are using the function showAppsMenuItem, so just remove this code to be able to run this tests normally.

cc @Tecnativa 

ping @pedrobaeza @carlosdauden


pedrobaeza (approved):
Click everywhere test continues working.

carlos-lopez-tecnativa (approved):

PR #3215, opened 2025-07-09
[17.0][IMP] web_widget_x2many_2d_matrix: Support selection type

zamberjo:
Enhance column value handling in X2Many2DMatrixRenderer to support selection type


zamberjo:
Hi @hbrunn, could you please take a look at this when you have a moment?. Thanks in advance! 🙏

hbrunn (approved):
I'd like to see a refactoring of this eventually such that column and row headers are rendered by the field type's widget, but that might be a bit too much to ask for this specific case.

Would be good to port to v18 too

hbrunn:
@zamberjo this should be superseded with #3246

zamberjo:
I'll check it out.

zamberjo:
No doubt your solution is much more complete, so I’m canceling this. Thanks a lot for your contribution!

PR #3219, opened 2025-07-10
[17.0][FIX] web_widget_x2many_2d_matrix: restore pre-v16 functionality

hbrunn:
see https://github.com/OCA/web/pull/3203

I didn't backport the tests because that would be a lot of work for just trying to make a better offer instead of butchering my baby even more in recent PRs. I did however add a tab to the group form in demo data where you can easily check on runbot that the different kinds of widgets work now in the matrix, and can be manipulated via attributes in the view definition.

### DIFF
--- a/web_widget_x2many_2d_matrix/README.rst
+++ b/web_widget_x2many_2d_matrix/README.rst
@@ -95,6 +95,12 @@ want to calculate row totals. True by default
 show_column_totals If field_value is a numeric field, it indicates if
 you want to calculate column totals. True by default
 
+x_axis_clickable If the x axis field is a many2one field, render the
+values as links to the record in question
+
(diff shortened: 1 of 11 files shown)


pedrobaeza (approved):
Let's merge it for normalizing the situation.

/ocabot merge major

MiquelRForgeFlow:
@hbrunn As you are maintainer now, could you please solve a simple issue I have with `hr_timesheet_sheet` that uses this module?

In v16:
Selection_4864

In v17 (and v18):
Selection_4865

The format has changed. Could you restore previous format?

hbrunn:
it's not simple, but see https://github.com/OCA/web/pull/3246, https://github.com/OCA/web/pull/3245

PR #3231, opened 2025-07-27
[18.0][ADD] web_json_widget

len-foss:
Add a Json editor widget.
Supports edits, dark mode (should work on OCA's version, since it relies on setting the `color_theme` cookie to `dark`). Can even be put in a tree view, although you probably don't want to do that.

This module uses the library [jsoneditor](https://github.com/josdejong/jsoneditor).
It is released under the [Apache-2.0](https://github.com/josdejong/jsoneditor/blob/master/LICENSE) license.

Screenshot_20250727_140139
Screenshot_20250727_140217


ivs-cetmix:
Hello @len-foss and thank you for your contribution!
Could you please create a simple `web_json_widget_demo` module like [this one](https://github.com/OCA/rest-framework/tree/17.0/fastapi_auth_jwt_demo) that would serve as an example and also will allow to do functional review of your PR? Just add a field to the `res.partner` module same way as on your screenshot.

len-foss:
@ivs-cetmix should be ready to test :+1:

ivs-cetmix (changes requested):
Perfect, thank you @len-foss!

Here are some glitches I encountered during testing on MacOs 15.6 (24G84) in Safari Version 18.6 (20621.3.11.11.3)

## Issue 1

**Steps to reproduce (on runboat)**

- In the second (text) key start typing "qwerty"

**Expected behaviour**

- The entire string can be typed in without loosing the focus

**Current behaviour**

- Need to click in the widget after each symbol is typed in. Eg "q" -> click, "w" -> click , ....

## Issue2

Field properties drop down is rendered below the chatter buttons.


[Video](https://drive.google.com/file/d/18Qynmfr-guEFuZdef2b69Jwnq20KnJYY/view?usp=share_link) of both issues.

len-foss:
@ivs-cetmix 
First, thank you for the testing. This is valuable feedback.

Easy point 2; I've added a CSS rule to increase the z-index. Good catch.

So, point 1. I actually broke editing trying to do a "quick patch" of a very specific problem: if the user calls the record switcher while the recordset is dirty, the component does not go through `onWillUpdateProps`, like it does if changes were already saved. As a result, you'd get outdated data in these cases. I'm not sure what is the proper solution, but I followed the approach of the chatter (which is to keep track of the displayed record, so that we can update if it changed). In any case, everything should be working now.

ivs-cetmix (approved):
Functional review LGTM
@len-foss works now, thank you!
P.S. If you have LinkedIn we can promote this module to get more attention and have it merged faster 😄

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,131 @@
...
+export class JsonEditorComponent extends Component {
+    static template = "web_json_widget.JsonEditor";
+    static components = {};
+    static props = {
+        ...standardFieldProps,
+        // Placeholder: { type: String, optional: true },
+    };
+    static defaultProps = {};
+
+    setup() {
```
ShehbajTechultra, line 26: Your createEditor() already has a great docstring. Adding the same for setup() and detect_dark_mode() helps future developers:
/**
 * Setup lifecycle hooks and state initialization.
 * - Tracks record resId and resModel
 * - Binds editor creation on mount
 * - Cleans up editor on unmount
 */
setup() { ... }

/**
 * Detect if dark mode is active.
 * Currently relies on cookie `color_scheme`.
 */
detect_dark_mode() { ... }

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,131 @@
...
+        const options = {
+            mode: "tree",
+            modes: ["tree", "text", "view"],
+            onChange: () => {
+                try {
+                    const new_value = this.editor.get();
+                    this.state.value = new_value;
+                    this.props.record.update({[this.props.name]: new_value});
+                } catch (error) {
+                    console.error("Error updating JSON value:", error);
```
ShehbajTechultra, line 84: Replace console.error with Odoo’s Notification service
this.env.services.notification.add(_t("Error updating JSON: ") + error.message, { type: "danger" });

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,131 @@
...
+            this.editorRef.el.classList.add("json-editor-theme-dark");
+        }
+
+        const options = {
+            mode: "tree",
+            modes: ["tree", "text", "view"],
+            onChange: () => {
+                try {
+                    const new_value = this.editor.get();
+                    this.state.value = new_value;
```
ShehbajTechultra: also update state with setter
this.state.value = { ...newValue };

antoniodavid (approved):
LGTM

ivs-cetmix:
Hi @len-foss , could you please check the code review comment and reply or apply them? Would be really great to have your excellent module merged!

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: Currently, the code is pretty short, at barely a 100 lines of logic. Adding such fluff would only make it less readable, and would only serve a purpose if it was used by some automatic documentation generation tool, which aren't used within the project.

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: Done :+1:

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: Done :+1:

len-foss:
Hi @ivs-cetmix done :+1: 

Thank you @ShehbajTechultra for the suggestions.

ivs-cetmix:
@ShehbajTechultra could you please do an after-update review? Than you in advance!

ShehbajTechultra (approved):
Approved

ivs-cetmix:
> Approved

@ShehbajTechultra , perfect, thank you!
@len-foss could you please prepare the PR for merge?

len-foss:
@ivs-cetmix do you mean you prefer to have the commits squashed?

ivs-cetmix:
@len-foss , yes, we need to squash commits so we have two - one per module.

ivs-cetmix:
Hey @OCA/web-maintainers looks like we have a new cool widget ready to join the OCA family. Would appreciate your review)

web_json_widget/__manifest__.py
```diff
@@ -0,0 +1,41 @@
...
+    "data": [],
+    "assets": {
+        "web.assets_backend": [
+            "/web_json_widget/static/src/components/json_editor/json_editor.esm.js",
+            "/web_json_widget/static/src/components/json_editor/json_editor.xml",
+            "/web_json_widget/static/src/components/json_editor/json_editor.css",
+            # content from https://app.unpkg.com/jsoneditor@10.2.0/files/dist
+            # source: https://github.com/josdejong/jsoneditor?tab=readme-ov-file
+            # licenced under Apache-2.0 license
+            "/web_json_widget/static/src/lib/dark-theme.css",
```
hbrunn: those usually go into /static/lib

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export const JsonEditor = {
+    component: JsonEditorComponent,
+    displayName: _t("Json Editor"),
+    supportedTypes: ["json"],
+    // Improvements: add supportedOptions, ExtractProps...
+};
+
+registry.category("fields").add("json_editor", JsonEditor);
```
hbrunn: I'd suggest

```suggestion
registry.category("fields").add("json", JsonEditor);
```

because then Odoo should pick the widget automatically, given there's not standard json widget

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,133 @@
...
+
+        onPatched(() => {
+            if (
+                this.resId !== this.props.record.resId ||
+                this.resModel !== this.props.record.resModel
+            ) {
+                this.resId = this.props.record.resId;
+                this.resModel = this.props.record.resModel;
+                if (this.editor) {
+                    this.editor.set(this.props.record.data[this.props.name] || {});
```
hbrunn, line 47:
```suggestion
                    this.editor.set(this.state.value);
```

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,133 @@
...
+            return;
+        }
+        if (typeof JSONEditor === "undefined") {
+            console.error("JSONEditor not available");
+            return;
+        }
+
+        const isDarkMode = this.detect_dark_mode();
+        if (isDarkMode) {
+            this.editorRef.el.classList.add("json-editor-theme-dark");
```
hbrunn, line 72: I wonder why you don't do this in the template?

web_json_widget/static/src/components/json_editor/json_editor.esm.js
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        const options = {
+            mode: "tree",
+            modes: ["tree", "text", "view"],
+            onChange: () => {
+                try {
+                    const new_value = this.editor.get();
+                    this.state.value = {...new_value};
```
hbrunn: isn't it the point of useState that you don't need to update this manually? Actually, what use is the state proxy in this widget in the first place?

hbrunn (dismissed):
please check out https://github.com/OCA/web/blob/18.0/web_widget_x2many_2d_matrix/static/tests/web_widget_x2many_2d_matrix.test.js for an example of how easy you can have your code tested nowadays.

This is called by https://github.com/OCA/web/blob/18.0/web_widget_x2many_2d_matrix/tests/test_web_widget_x2many_2d_matrix.py

web_json_widget/__manifest__.py
len-foss: :+1:

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: :+1:

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: You're right, it's not needed at all. I started by copying a basic widget, it's a 'PoC' remnant.

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: Not sure exactly but the doc about the Json editor was a bit confusing to me, and I found confusing answers as to how to instantiate a dark version. I think the 'Odoo way' is to put that in a CSS without the dark mangling which is put into the dark assets bundle. As of now I find it simpler to keep it this way because I copy-pasted this from recommended setup, but it wouldn't be hard to change afterwards.

web_json_widget/static/src/components/json_editor/json_editor.esm.js
len-foss: So, this one is pretty bad. Because of the very stupid OCA rule I squashed my commits, but basically if your recordset is dirty, using the record switcher makes it so that none of the widget lifecycle hooks are called. So, `this.state.value` is not updated, so you switch and the widget is not updated. I couldn't find anything that seemed to make sense and the odoo dev I asked was a clueless. This is inspired by how it's done by the mail JS code.

To clarify, onParched is a lifecycle hook but that is called continuously, whereas you'd want a 1-time one like [onWillUpdateProps](https://github.com/odoo/owl/blob/master/doc/reference/component.md#willupdateprops).

So, if there's a good explanation and solution, I'd like to know it.

len-foss:
@hbrunn 
Regarding the test, in my opinion, it would be much more cost much more that what it's worth, so I won't implement it on my limited free time.
You're free to disagree of course, but I'm seeing `test_web_systray_button_init_action` make the tests red for everything, so consider the time lost for all (potential) contributors.

flotho:
Hi @len-foss thnaks for your contribution.
Any chance to rebase ?

web_json_widget/readme/USAGE.md
```diff
@@ -0,0 +1,11 @@
+To insert a json editor in a form view, you can use the widget:
+```xml
+    <field name="json_field" widget="json_editor"/>
```
simahawk, line 3: is this needed since it's now registered for all `json` fields?

web_json_widget_demo/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2025 360ERP (<https://www.360erp.com>)
```
simahawk: headers to update?

simahawk (commented):
LG overall. Couldn't test as Runboat is hanging.

web_json_widget_demo/models/res_partner.py
len-foss: Thanks :-)

web_json_widget/readme/USAGE.md
len-foss: I'm not sure this was a good idea, since in 19.0 there is a default json widget -- that is just a json string dump on the content (which I don't understand why you didn't have that since the introduction of the field).

So if you use json_editor, you can migrate without any view change, whereas json means it won't, you'll get the default one.

len-foss:
Rebased :+1:

simahawk:
@hbrunn ping

simahawk:
@len-foss pls squash and I would set the maturity level to alpha if you don't mind (cheaper than adding tests :wink: )

etobella:
@simahawk it is true that alpha doesn't require tests, but we should encourage people to add them, isn't it? :wink:

simahawk:
> @simahawk it is true that alpha doesn't require tests, but we should encourage people to add them, isn't it? 😉

We should and we did (Holger did) and @len-foss didn't ignore that, he provided an answer.
Now, we have 2 choices: 
1. keep this PR stuck forever
2. let it go and have someone using it for real

In this particular case, I'm for opt 2, but this is a democracy and you can still block the PR :wink:

etobella:
@simahawk I was not going to block the PR don't worry for that. I already know how hard it is to add tests (even more JS). I can understand that it is not an easy step and it was not blocking me.

My only concern was that your comment could be interpreted as "add everything as alpha and then we don't need to care about tests" :wink: (I know what you were thinking and I agree, however I wanted to clarify to avoid confusion).

One last thing, @len-foss  can you squash all commits in a single one?

simahawk:
> My only concern was that your comment could be interpreted as "add everything as alpha and then we don't need to care about tests" 😉 (I know what you were thinking and I agree, however I wanted to clarify to avoid confusion).

Totally fair :relaxed: Thank you!

len-foss:
There's just the fact that in that version the widget is registered as `json`, which in 19.0 will migrate to the default Odoo `json`, which isn't editable; so I'd change that to `json_editor` by default, so that it's clean across migrations. Rebase/squashed commits.

PR #3237, opened 2025-07-30
[18.0][FIX] web_responsive: Tours that uses showAppsMenuItem are failing and click everywhere not working 

CarlosRoca13:
cc @Tecnativa

The patch of stepUtils is breaking tests that uses this function when this module is installed. So remove it to avoid them fail.

The selectors used in clickbot.js from the Community Edition do not match any elements in the view when web_responsive is installed.

Since the functions are not exported, we can't intercept them directly.
To solve this, we override the querySelector functions specifically within the clickbot assets, allowing us to adjust the selectors when they are incompatible with web_responsive.

ping @pedrobaeza @carlos-lopez-tecnativa


pedrobaeza (approved):
Thanks for fixing this.

carlos-lopez-tecnativa (approved):
Code review done and tested the `ClickEverywhere`. Regarding the patch in the test, I can't test it, but this code always causes problems. I think it's OK to remove it if you tested it in a real instance with tests.

PR #3241, opened 2025-08-05
[17.0][FIX] web_responsive: prevent error in tests when accessing an undefined variable

carlos-lopez-tecnativa:
For an unknown reason, when running the tests, the `session` does not have all variables filled, so `apps_menu` is undefined. This causes an error in tours of other modules: `Cannot read properties of undefined (reading 'theme')`.
image
image

The curious thing is that if you run the tests of this module alone, they pass.
But when running tests together with another module, or starting the test manually from the browser, tests from the other module fail.
For example, when running tests with the `voip_oca` module.

After this commit, the tests pass correctly:

image

@Tecnativa @pedrobaeza @CarlosRoca13 could you please review this?

### DIFF
--- a/web_responsive/static/src/components/apps_menu/apps_menu.esm.js
+++ b/web_responsive/static/src/components/apps_menu/apps_menu.esm.js
@@ -49,7 +49,7 @@ export class AppsMenu extends Component {
     setup() {
         super.setup();
         this.state = useState({open: false});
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
It seems safe.

CarlosRoca13 (approved):

PR #3247, opened 2025-08-08
[16.0][IMP] product_label_section_and_note_field: force a width on the column

carlos-lopez-tecnativa:
Odoo does not facilitate inheritance in the function https://github.com/odoo/odoo/blob/3bc173a6b22f2e360ec363277800ee5efa12088f/addons/web/static/src/views/list/list_renderer.js#L406 so it is necessary to force a max width at this low level.

In v18, Odoo provides an attribute at the field level
https://github.com/odoo/odoo/blob/1702cb44ab7692d70bfabfc25db9592dac85e7a2/addons/web/static/src/views/list/column_width_hook.js#L299  and in this widget Odoo uses a default value, which we also use here, see https://github.com/odoo/odoo/blob/1702cb44ab7692d70bfabfc25db9592dac85e7a2/addons/account/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.js#L249

TT57583

@Tecnativa @pedrobaeza

### DIFF
--- a/web_widget_product_label_section_and_note/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.esm.js
+++ b/web_widget_product_label_section_and_note/static/src/components/product_label_section_and_note_field/product_label_section_and_note_field.esm.js
@@ -285,3 +285,21 @@ ProductLabelSectionAndNoteField.template =
 registry
     .category("fields")
     .add("product_label_section_and_note_field", ProductLabelSectionAndNoteField);
+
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
It's not the most elegant one, but it works

CarlosRoca13 (approved):

pedrobaeza:
Is this needed in 17?

/ocabot merge patch

PR #3254, opened 2025-08-19
[ADD] Adds web forms

samirGuesmi:
This module can be used to collect data from external users (or partners without a portal account) via a simple web form, then create records in any Odoo model using those submitted values. Each form is tied to a partner-specific secure URL using a one-time rotatable token.


samirGuesmi:
@sbejaoui

yvaucher (approved):

simahawk:
FYI For extendable and reproducible forms you might want to have a look at https://github.com/OCA/website-cms/tree/16.0. Note that it depends only on portal and not on website anymore (repo should be renamed probably).

sbejaoui (approved):

PR #3264, opened 2025-08-28
[18.0][FIX] web_ir_actions_act_window_message: env.services.rpc is not a function

sergioM-S73:
1 - When creating an "ir.actions.act_window.message" and clicking on the buttons, this error appears:

image

I have refactored the component to use "useService" for ORM, removing the use of env.services.rpc

2 - After, testing with an ir_actions_act_window_message in sale.order, which contains a button that triggers the action_confirm action, I encountered this error.

image

This happens because if the page is refreshed within the form view of the sale.order, the context of the list view is lost, so when the switchView function is called, it executes this code:

image

In this case, the `controllerStack` array contains a list view, but it doesn't contain the `view` field.

image

If we access the sale.order form view directly from the list view without refreshing the page, the `controllerStack` array correctly contains the list view information.

image

The _refreshWidget function has been modified to prevent this error when the user refreshes the page in the form view, and it now triggers a `window.location.reload()` to maintain functionality.


sergioM-S73:
@yvaucher @miguel-S73 When using this module I get this errors, can you review it??

Reyes4711-S73 (approved):
LGTM

miguel-S73 (approved):

adrip-s73 (approved):
LGTM

Reyes4711-S73:
@OCA/website-maintainers Please, can you merge this PR?

web_ir_actions_act_window_message/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Client side message boxes",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
```
hbrunn: don't do this, we increase version numbers during merge

web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_msg_component.esm.js
```diff
@@ -16,21 +20,25 @@ export class ActWindowMessageDialog extends Component {
...
     generateButtons() {
         var self = this;
         const action = self.props.action;
-        const env = self.props.env;
         if (action.close_button_title !== false) {
             self.state.buttons.push({
                 name: action.close_button_title || _t("Close"),
                 click: () => {
-                    // Refresh the view before closing the dialog
-                    self._refreshWidget(env);
```
hbrunn, line 33: we need to keep this, as one pretty relevant use case for this addon is: Button is called, server side something happens, returns a message box action talking about this. And without the reload whatever the button did won't reflect in the currently shown view

hbrunn (changes requested):
thanks for this, I wonder if this ever worked or if the v18 migration just was a horrible hack job.

Given `env` from props isn't used any more, please remove all references to this.

And please review https://github.com/OCA/web/pull/3353 to avoid this kind of unpleasantness in the future (actual unittests pending)

web_ir_actions_act_window_message/static/src/js/web_ir_actions_act_window_msg_component.esm.js
```diff
@@ -1,9 +1,13 @@
-import {Component, onWillStart, useState} from "@odoo/owl";
+import {Component, onWillStart, useEnv, useState} from "@odoo/owl";
 import {Dialog} from "@web/core/dialog/dialog";
 import {_t} from "@web/core/l10n/translation";
+import {useService} from "@web/core/utils/hooks";
+import {router} from "@web/core/browser/router";
 
 export class ActWindowMessageDialog extends Component {
     setup() {
+        this.env = useEnv();
```
hbrunn: why not `useService("action");`?

sergioM-S73:
@hbrunn Changes made, thanks for the feedback!

hbrunn:
please remove env in `web_ir_actions_act_window_message.esm.js` too

sergioM-S73:
@hbrunn done

hbrunn (approved):
thanks, if you put this into your v19 migration, that should be fine too.

You might want to review the PR I pointed to for having this merged in v18

/ocabot merge patch

PR #3267, opened 2025-08-30
[MIG] web_dark_mode: Migration to 18.0

len-foss:
Note: this is a migration of the code. It seems the CSS would need some polish, even though it seems mostly alright (contrast of many elements is quite low for my tastes).


mathben:
I tested and it works for me. Thanks

I notice visual issue, example with required field combine with module [web_theme_classic](https://github.com/OCA/web/tree/18.0/web_theme_classic)
Screen Shot 2025-09-15 at 01 00 32

PR #3272, opened 2025-09-03
[18][IMP] web_chatter_position: improve form and chatter layout styles

StefanRijnhart:
Move functional elements of web_sheet_full_width into web_chatter_position, as it is otherwise deprecated.

Supersedes https://github.com/OCA/web/pull/3183 (after first having proposed the changes into that PR's branch in https://github.com/c4a8-odoo/module-oca-web/pull/17).

### DIFF
--- a/web_chatter_position/README.rst
+++ b/web_chatter_position/README.rst
@@ -32,6 +32,9 @@ Configurable chatter position from the user preferences.
 
 Supports Both Community & Enterprise Edition.
 
+Extends the functionality of the web client to get full width in the
+form view sheet.
+
 **Table of contents**
 
 .. contents::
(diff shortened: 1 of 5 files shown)


bealdav:
18.0 broken by web_systray_button_init_action module

StefanRijnhart:
@bealdav green in https://github.com/OCA/web/pull/3275

MohamedOsman7 (approved):
Code & functional review LGTM

bealdav (approved):

anajuaristi:
Hello @StefanRijnhart 
Just to let you know there is a bug in web_chatter_position that is being solved in 
https://github.com/OCA/web/pull/3282

I don't know if this is also happening in V18, but if yes, IMHO the patch should be also included here or in another PR.
My2cents.

StefanRijnhart:
@anajuaristi I saw that. That patch seems to break attachment rendering in our 16.0 setups for users that force the chatter below, possibly in combination with web_enterprise. We're now moving away from using this module. I might investigate during the OCA days, but you are right that this change is not compatible with that PR and we should probably not be merging this now.

trisdoan:
Hello @StefanRijnhart, fyi, that patch tries to fix a feature, which was added in https://github.com/OCA/web/pull/3040 for 16.0. And it's not forwarded to 18.0 yet

StefanRijnhart:
@trisdoan Perfect, thanks. I now see that #3282 is indeed a proposed fix for the issue we encountered in our setups, caused by #3040. In any case, this PR needs to be refactored to fit onto any port of #3040 which is likely to happen.

DavidJForgeFlow:
Hi, can someone merge this?

PR #3273, opened 2025-09-04
[18.0][FIX] web_remember_tree_column_width: fetch element from start event

StefanRijnhart:
Analogous to the inherited code, the event listener would work on the resize stop event, but this event would regularly not have the same target as the resize start event in which case the field name and size could not be determined.


cuongnmtm (approved):

AaronHForgeFlow (commented):
I could reproduce the issue (resize embedded lists with buttons on it) and this PR seem to fix it.

MiquelRForgeFlow (approved):

PR #3274, opened 2025-09-04
[18.0][FIX] web_company_color: show company colors from active company

ArnauCForgeFlow:
When you have multiple active companies, changing the active company doesn’t update the company colors.


StefanRijnhart (approved):

pedrobaeza (approved):
Checked on the developer tools that the way to store this in the cookie is with `-`.

/ocabot merge patch

pedrobaeza:
OK, we have the known problem...

LoisRForgeFlow (approved):
Tested, it works fine now :+1:

PR #3280, opened 2025-09-09
[ADD] web_chatter_camera

etobella:
This new module allows to use the camera inside odoo.

The functionality existed in Previous Odoo versions, but it was lost at some point.

image

The camera button will try to launch the camera directly, but that might depend on your mobile OS.

If you want to test it, use a mobile, with a computer, nothing will change


luisDIXMIT (approved):
Tested on runboat and LGTM!!

ValentinVinagre:
@etobella I seem to remember there was a similar module for the photo widget, and it was actually pretty good. This one, on the other hand, takes you to attachments. Why make it just for phones?

web_chatter_camera/static/src/components/chatter_topbar/chatter_topbar.esm.js
```diff
@@ -0,0 +1,33 @@
...
+import {ChatterTopbar} from "@mail/components/chatter_topbar/chatter_topbar";
+import {isMobileOS} from "@web/core/browser/feature_detection";
+import {patch} from "@web/core/utils/patch";
+import {useRef} from "@odoo/owl";
+
+const components = {ChatterTopbar};
+
+patch(
+    components.ChatterTopbar.prototype,
+    "mail_activity_board/static/src/components/chatter_topbar/chatter_topbar.esm.js",
```
ValentinVinagre: Wouldn't it be necessary to add this module to dependencies? --> [mail_activity_board](https://github.com/OCA/social/tree/17.0/mail_activity_board)

etobella: No, it is a typo error, sorry

etobella:
@ValentinVinagre It is done this way because it is done directly by your browser that decides if you should open the camera or not.

If we want to do as you say, it gets more complicated.

ValentinVinagre (commented):
Reviewed on laptop:
- **Widget functionality:**
Over http doesn't work, cannot be tested with runboat.
image

 https://stackoverflow.com/questions/62575901/webcam-js-error-no-supported-webcam-interface-found
https://github.com/jhuckaby/webcamjs/issues/307#issuecomment-708180340

- **Functionality to take a photo in chatter:**
image


- **Pending testing in an environment with https**

etobella:
@ValentinVinagre You are right. Tested in all the cases and now it should work properly.

etobella:
@ValentinVinagre can you retest?

ValentinVinagre (approved):
LGTM 😸

PR #3286, opened 2025-09-15
[12.0][ADD] web_form_banner

yostashiro:
The module adds configurable banners for backend form views. Define rules per model

on form load/save/reload.

@qrtl QT5611


AungKoKoLin1997 (approved):
Code and functional review: LGTM

yostashiro:
@AungKoKoLin1997 Just added a commit since the message translation wasn't working. Also re-allocated some eval context vars from _get_banner_eval_context() to _banner_base_eval_ctx_static().

Please squash once you confirm this, and apply the same in the migration PR.

yostashiro:
@AungKoKoLin1997 It was mistaken when I said the update didn't work. The updated code actually works fine.

PR #3290, opened 2025-09-23
[18.0][IMP] web_chatter_position: pre-commit fixes

DavidJForgeFlow:
When using side mode, the chatter is by default responsive by addind the isChatterAside and o-aside class. This can cause some 'inconveniencies' if the chatter contains a long text, making the form view almost unreadable while helping reading the chatter. So this PR simply removes the responsive when used in side mode.

Chatter Responsive previous behavior:
image

Chatter Responsive disabled:
image

### DIFF
--- a/web_chatter_position/models/res_users.py
+++ b/web_chatter_position/models/res_users.py
@@ -9,7 +9,7 @@ class ResUsers(models.Model):
 
     chatter_position = fields.Selection(
         [
-            ("auto", "Responsive"),
+            ("auto", "Automatic"),
             ("bottom", "Bottom"),
             ("sided", "Sided"),
         ],
@@ -22,4 +22,6 @@ class ResUsers(models.Model):
 
     @property
     def SELF_WRITEABLE_FIELDS(self):
-        return super().SELF_WRITEABLE_FIELDS + ["chatter_position"]
+        return super().SELF_WRITEABLE_FIELDS + [
+            "chatter_position",
+        ]
--- a/web_chatter_position/static/src/js/web_chatter_position.esm.js
+++ b/web_chatter_position/static/src/js/web_chatter_position.esm.js
@@ -3,10 +3,10 @@
     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 */
 
-import {FormCompiler} from "@web/views/form/form_compiler";
-import {patch} from "@web/core/utils/patch";
 import {append, setAttributes} from "@web/core/utils/xml";
+import {FormCompiler} from "@web/views/form/form_compiler";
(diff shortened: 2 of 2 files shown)


web_chatter_position/models/res_users.py
```diff
@@ -9,17 +9,25 @@ class ResUsers(models.Model):
...
             ("bottom", "Bottom"),
             ("sided", "Sided"),
         ],
         default="auto",
     )
+    chatter_responsive = fields.Boolean(
+        default=True,
+        help="Defines if the chatter should be responsive (extended)"
+        "or not based on the content length.",
+    )
```
LoisRForgeFlow: In any case, why adding an extra boolean on top of a selection field? adding a new option to the selection (auto_no_responsive or similar) makes more sense IMO

LoisRForgeFlow (commented):
I'm not sure if it is a good idea to keep that behavior, who would want the chatter to potentially consume the 90% of the screen? and not be consistent when navigating records? I would think about setting at least a max width.

Let's see what do others think. In any case in the current approach I would adjust as per my comment below

pedrobaeza:
My suggestion is that you get rid off this module, hehe.

web_chatter_position/models/res_users.py
DavidJForgeFlow: I thought it as two different choices: Where do you want the chatter? Do you want it to adapt to the content?
Getting more options on the position IMO should not be the way.

Also, getting the width limited makes sense in UX, however, taking into account hard-coding would make me cry, we would need to add also a % value to define how much you want the chatter to fill. And this setting is maybe to much.

DavidJForgeFlow: @LoisRForgeFlow you were right, I removed the responsive feature to simply get things work as before.

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
> My suggestion is that you get rid off this module, hehe.

Not my favourity module TBH haha but as it is already migrated, the minimum for the OCA is that it works

ThiagoMForgeFlow (approved):
LGTM

DavidJForgeFlow:
@ThiagoMForgeFlow @pedrobaeza 
At the end, the issue here was fixed by another PR #3272. However I changed the commit name to actually contain only pre-commit fixes!

etobella (approved):
Merging for getting rid of the errors.

However, there are Pre-commit warnings in other modules, but shouldn't be the scope of this PR.

/ocabot merge nobump

etobella:
I forced the stop of the merge to do it with patch, not nobump

/ocabot merge patch

PR #3292, opened 2025-09-25
[16.0][FIX] web_dialog_size: fix dialog shrink issue

BhaveshHeliconia:
image

Whenever we set this system parameter to set size for the dialog default to True, some of the wizards can not be restore to it's original size by clicking on the below button.

image

One of the example for this issue is product selection wizard in any order line form view.

image
image

When we click on "Search More..." it will open a wizard which can not be restored to it's original size.

It only happens when we set the system parameter "web_dialog_size.default_maximize" = "True"

https://github.com/user-attachments/assets/36095d06-eee8-4db4-93cf-5547d562aebb

### DIFF
--- a/web_dialog_size/static/src/xml/ExpandButton.xml
+++ b/web_dialog_size/static/src/xml/ExpandButton.xml
@@ -5,7 +5,7 @@
             t-if="props.getsize() == 'dialog_full_screen'"
             type="button"
             class="btn btn-secondary dialog_button_extend"
-            t-on-click="dialog_button_restore"
+            t-on-click="toggleSize"
         >
             <i class="fa fa-compress" />
         </button>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you describe the issue?

BhaveshHeliconia:
> Can you describe the issue?

yes

web_dialog_size/static/src/xml/ExpandButton.xml
pedrobaeza, line 4: Why putting 2 buttons now?

pedrobaeza: Forget it. I see that the icon is different.

pedrobaeza (approved):
Tested on runboat and working.

DavidJForgeFlow (approved):

pedrobaeza:
Please fw-port it to upper versions if applicable.

BhaveshHeliconia:
> Please fw-port it to upper versions if applicable.

No forward port needed; the issue doesn’t exist in upper versions.

PR #3294, opened 2025-09-25
[18.0][FIX] Add !default to all color variable definitions

ljmnoonan:
This will allow web_dark_mode or any other theme to override colors defined here.
This is especially necessary for the `$input-background-color-required` variable because it looks hideous in a dark theme as mentioned by [mathben](https://github.com/OCA/web/pull/3267#issuecomment-3290540385) in #3267 

image

@legalsylvain


DavidJForgeFlow (approved):

ljmnoonan:
After further thought, I have decided to build a full dark theme for this module rather than trying to override things from web_dark_mode. I will open a pull request for this soon

PR #3287, opened 2025-09-16
[15.0][MIG] web_form_banner: Migration to 15.0

AungKoKoLin1997:
Based on https://github.com/OCA/web/pull/3286

The module adds configurable banners for backend form views. Define rules per model
(and optionally per view) to show context-aware alerts with a chosen severity (info/warning/danger).

Messages can be plain text with ${placeholders} or fully custom HTML; visibility,
severity, and values are computed server-side via a safe Python expression.

Banners are injected just before or after a target node (default: //sheet) and refresh
on form load/save/reload.

@qrtl QT5747

PR #3298, opened 2025-10-01
[16.0][MIG] web_form_banner: Migration to 16.0

AungKoKoLin1997:
Based on https://github.com/OCA/web/pull/3287

The module adds configurable banners for backend **form** views. Define rules per model

on form load/save/reload.

@qrtl QT5747


bealdav:
Hi @AungKoKoLin1997 and @Yoshiro Thanks a lot for your nice module.
I'm asking myself if there is no way to have no js code and just python code.
I suppose with python, migration over versions could be more easy.
Yes it require to save page to evaluate python code but I suppose that's not an hard inconvenient

yostashiro:
Thanks @bealdav !

We'd also prefer to minimize JS, but dynamic banner updates are a hard requirement from a client, and I'm sure some people will find it useful as well. I also assume that JS changes required in future migrations will be relatively limited after this version.

We still need to tidy up the commits, but the code is ready for review! :)

yostashiro (approved):
Code review and functional test.

bealdav:
Thanks a lot @yostashiro for your answer. Do you plan to port this module to upper version like 18.0 ?

bealdav (approved):
Functionnaly tested, thanks a lot.

AungKoKoLin1997:
> Do you plan to port this module to upper version like 18.0 ?

@bealdav Yes, we plan to migrate to v17 and v18 in the next couple of days.

PR #3297, opened 2025-09-30
[19.0] [MIG] web_environment_ribbon: Migration to 19.0

rven:

PR #3299, opened 2025-10-01
[17.0][FIX] web_environment_ribbon: fix configuration image in description

ap-wtioit:
While reviewing for 19.0 #3297 i discovered that:

The description inside Odoo's Module Info featured a broken image:
image

With this change it is fixed:
image

If this is ok, i will forward port to 18.0 and 19.0 (17.0 seems to be the first version with the broken description)

Info @wt-io-it

### DIFF
--- a/web_environment_ribbon/README.rst
+++ b/web_environment_ribbon/README.rst
@@ -39,19 +39,21 @@ every page
 Configuration
 =============
 
--  You can change the ribbon's name ("TEST") by editing the default
-   system parameter "ribbon.name" (in the menu Settings > Technical >
-   Parameters > System Parameters) To hide the ribbon, set this
-   parameter to "False" or delete it and refresh the page.
+- You can change the ribbon's name ("TEST") by editing the default
+  system parameter "ribbon.name" (in the menu Settings > Technical >
+  Parameters > System Parameters) To hide the ribbon, set this parameter
+  to "False" or delete it and refresh the page.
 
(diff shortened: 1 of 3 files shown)


web_environment_ribbon/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     "name": "Web Environment Ribbon",
-    "version": "17.0.1.0.2",
+    "version": "17.0.1.0.3",
```
pedrobaeza: Please don't change module version manually.

ap-wtioit: Well, if it happens automatic, which place would it increase? I fear that the automatic one would increase it to 17.0.1.1.0 while this is only a fix without functional changes so it feels like the last place needs to be incremented. Do you know were i can lookup the automatic OCA version increment code?

pedrobaeza: It depends on how we (PSCs/core maintainers/maintainers) launch the command. We have `ocabot merge patch/minor/major`.

pedrobaeza (approved):

web_environment_ribbon/__manifest__.py
ap-wtioit: Thanks for the explanation, i removed the change of the version number from my commit.

hoangtrann (approved):
Looks good :+1:

PR #3305, opened 2025-10-07
[17.0][MIG] web_form_banner: Migration to 17.0

AungKoKoLin1997:
Based on https://github.com/OCA/web/pull/3298

The module adds configurable banners for backend form views. Define rules per model

on form load/save/reload.

@qrtl QT5747


web_form_banner/models/web_form_banner_rule.py
```diff
@@ -39,7 +38,7 @@ def _extract_m2o_id(v):
     """Normalize many2one values to an integer id or False."""
     if isinstance(v, int):
         return v
-    if isinstance(v, (list, tuple)) and v and isinstance(v[0], int):
+    if isinstance(v, list | tuple) and v and isinstance(v[0], int):
```
yostashiro, line 41: I don't quite understand this change. Why do we need to do this?

web_form_banner/models/web_form_banner_rule.py
```diff
@@ -172,7 +171,6 @@ def _build_form_url(self, rec):
             _logger.exception("Failed building form URL for %s", rec)
             return ""
 
-    @lru_cache(maxsize=1)
     def _base_eval_ctx_static(self):
         # Only static, import-heavy items
```
yostashiro: Please add @api.model. We may perhaps remove the the comment as it doesn't say much more than the method name.

I suppose the `lru_cache` removal should also be applied to the source versions?

web_form_banner/static/src/js/web_form_banner.esm.js
```diff
@@ -26,16 +27,20 @@ const safe = async (fn, fb) => {
 };
 
 function normalizeValue(v) {
```
yostashiro, line 29: Please avoid removing comments unless they are just unnecessary noise. Also avoid immaterial style changes.

yostashiro:
Dynamic banner update seems a bit slower than before. Is there room for improvement?

web_form_banner/models/web_form_banner_rule.py
AungKoKoLin1997: This is the new style, introduced in PEP 604 (Python 3.10+) and style guide (Ruff) forces us to use that style in the pre-commit.

AungKoKoLin1997:
> Dynamic banner update seems a bit slower than before. Is there room for improvement?

@yostashiro I compared between v16 and v17 in runboat with the latest changes and it is like the same.

yostashiro (approved):
Code review and functional test. LGTM.

yostashiro (changes requested):
There is an issue with banner reload with previous/next button. See https://github.com/OCA/web/pull/3308#pullrequestreview-3329902995.

web_form_banner/views/web_form_banner_rule_views.xml
```diff
@@ -0,0 +1,202 @@
...
+                                    >dict(env.context)</code>).</li>
+                                <li><code
+                                    >record</code>: Current record (the form's record).</li>
+                                <li><code
+                                    >draft</code>: The persisted field values of the ORM record (before applying the current
+                                    form's unsaved changes) + the current unsaved changes on trigger fields.
+                                    Should be used instead of <code
+                                    >record</code> when your rule is triggered dynamically by an
+                                    update to a trigger field. It doesn't include any values from complex fields
+                                    (x2many/reference, etc).
```
yostashiro: Minor adjustment.
```suggestion
                                    (one2many/reference, etc).
```

yostashiro (approved):
Code review and functional test. LGTM.

nobuQuartile (approved):
LGTM
Functional test

PR #3326, opened 2025-10-17
[19.0][MIG] web_company_color: Migration to 19.0

BhaveshHeliconia:


codeagencybe:
What's the status on this ? I see it's been open for 3 months already. What's the blocking?

BhaveshHeliconia:
@codeagencybe This PR is currently waiting for a reviewer. That’s the main blocker at the moment.

odoowired (approved):

Linekio:
Hello,
Seems it's not compatible with dark mode ? I just clicked on "Compute colors from logo"".
image
image

BhaveshHeliconia:
@Linekio Thanks for the feedback. Could you please elaborate on what’s not working in dark mode? That will help me better understand and investigate the issue.

Linekio:
Hello,
Most of the menu buttons are invisible by default
image

BhaveshHeliconia:
@Linekio I’ve tested it in light mode and it’s working as expected. I haven’t tested it in dark mode, as Odoo doesn’t provide native dark mode support, and I’m not aware of any official OCA module for it.

BhaveshHeliconia:
> @Linekio I’ve tested it in light mode and it’s working as expected. I haven’t tested it in dark mode, as Odoo doesn’t provide native dark mode support, and I’m not aware of any official OCA module for it.

@pedrobaeza do we need to consider compatibility with third-party dark mode modules in this case?

pedrobaeza:
I think he's talking about https://github.com/OCA/web/tree/19.0/web_dark_mode

BhaveshHeliconia:
> I think he's talking about https://github.com/OCA/web/tree/19.0/web_dark_mode

Thanks, I’ll work on that.

BhaveshHeliconia:
@Linekio I’ve reviewed the functionality of the module. It extracts colors from the uploaded company logo and applies them to the corresponding fields.

This logic is independent of light or dark mode, as the module itself is not aware of any theme context.

To support dark mode, we would need to introduce a dependency on a module like web_dark_mode and adapt the color extraction logic based on whether dark mode is enabled.

zigatrcek:
Hi, what's the status of this?

In my opinion, supporting dark mode would warrant a separate connector module in order to avoid any unwanted dependency chains. I think the right decision would be to merge this without dark mode support.

gbrito (approved):
Tested and looks good

web_company_color/models/res_company.py
```diff
@@ -115,10 +115,7 @@ class ResCompany(models.Model):
             background-color: %(color_navbar_bg_hover)s !important;
           }
         }
-        .o_menu_systray button,
-        .o_navbar_breadcrumbs,
```
pedrobaeza, line 119: Why removing this?

pedrobaeza (changes requested):
Please include #3448 #3568 #3426 #3427 #3484 

/ocabot migration web_company_color

web_company_color/models/res_company.py
BhaveshHeliconia: In Odoo 19 the breadcrumbs element is now `<div class="o_navbar_breadcrumbs d-contents">`. Because of d-contents the div renders no box of its own, so applying color/background-color hover rules to `.o_navbar_breadcrumbs` directly is a no-op the breadcrumb text/links are already covered by the navbar text and link color rules. So I dropped it as dead CSS rather than keep a selector that no longer does anything.

`.o_menu_systray button` wasn't actually removed — I split it into its own dedicated rule lower in the template `(.o_menu_systray button { color: ...; &:hover/&:focus { background-color: ... } })`, keeping the exact same color and hover behavior. The diff just shows it leaving the old grouped selector.

PR #3327, opened 2025-10-17
[19.0][MIG] web_copy_confirm: Migration to 19.0

BhaveshHeliconia:


BhaveshHeliconia:
@StefanRijnhart @ArnauCForgeFlow Could you please review this? Thanks!

ArnauCForgeFlow (approved):
LGTM

ArnauCForgeFlow (changes requested):
After reviewing more in detail i've realized that the message is not appearing when duplicating on tree view:

Image

BhaveshHeliconia:
@ArnauCForgeFlow You're right — the message was not being displayed when duplicating records from the tree view. I've fixed this issue, and the message now appears correctly when duplicating from the tree view as well.

ArnauCForgeFlow:
Hello @BhaveshHeliconia now the dialog appears correctly, but when i select multiple records and try to duplicate them, there are 2 dialogs appearing:

This is the first dialog that appears when duplicating multiple records:
image

This is the second dialog after accepting the first one:
image

BhaveshHeliconia:
Hello @ArnauCForgeFlow, thanks for the review. Yes, you're right — when duplicating multiple records, two dialogs were being displayed: the standard Odoo confirmation dialog followed by our custom dialog.

I've adjusted the logic to prevent this duplicate confirmation, so only the appropriate dialog is now displayed.

ArnauCForgeFlow (approved):
Thanks! LGTM

legalsylvain:
@andres-jpv : could you review this one ?

PR #3328, opened 2025-10-17
[19.0][MIG] web_editor_class_selector: Migration to 19.0

BhaveshHeliconia:


web_editor_class_selector/static/src/js/css_selector/css_selector.esm.js
```diff
@@ -1,15 +1,16 @@
...
 export class CssSelector extends Component {
     static template = "web_editor_class_selector.CssSelector";
     static props = {
         getItems: Function,
         getDisplay: Function,
         onSelected: Function,
-        ...toolbarButtonProps,
+        title: {type: [String, Function], optional: true},
+        getSelection: {type: Function, optional: true},
+        isDisabled: {type: [Function, Boolean], optional: true},
```
carlos-lopez-tecnativa: What is the reason for this change?
I see that this code already exists in v19:
https://github.com/odoo/odoo/blob/4986ffe1047e5e2b02f039314cd2dad2388db94d/addons/html_editor/static/src/main/toolbar/toolbar.js#L66-L70

carlos-lopez-tecnativa (commented):
LGTM, working fine, just a technical detail.

web_editor_class_selector/static/src/js/css_selector/css_selector.esm.js
BhaveshHeliconia: You're right, I hadn’t noticed that toolbarButtonProps was already exported from the core. I’ve updated the component to reuse the existing one instead of redefining the props manually. I also adjusted the template accordingly, following the same approach used in FontSelector. Thanks for pointing that out!

carlos-lopez-tecnativa: Sorry, but with these changes I now get this error. I understand that the title is empty, but in toolbarButtonProps the title is required. I now understand that this may be the reason why you changed this, but it could be useful to add a comment explaining the reason.

Image

BhaveshHeliconia: I've now marked those props as optional in the component and added a comment explaining why (since the Toolbar might pass undefined if the description is missing). I also updated the plugin to use description instead of title to match what the core Toolbar expects. That should fix the validation error you saw. Thanks!

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge nobump

PR #3333, opened 2025-10-17
[19.0] web_remember_tree_column_width: Migration to 19.0

BhaveshHeliconia:


AungKoKoLin1997 (approved):
Functional Review: LGTM

RyanTranAVB:
Test on ruboat and it doesn't seem to work. Could you check?

https://github.com/user-attachments/assets/0b9e1424-7bfd-4389-aed5-217f8138ba41

BhaveshHeliconia:
@RyanTranAVB : Yes, I checked it. In the video you shared, everything seems to be working correctly. I’m also sharing my own video — as you can see, when I increase the width of the number field column and reduce the customer column width, then refresh, the widths remain exactly as I set them. That’s the expected behavior of the module, and it is working as intended.


https://github.com/user-attachments/assets/80faebf1-b437-4441-8a4d-e61966b7cf05

RyanTranAVB:
Oh I see, so it remembers the column width but only the resized ones and won't remember the overall layout

RyanTranAVB (approved):

web_remember_tree_column_width/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+        "luisg123v",
+        "cuongnmtm",
+    ],
+    "depends": [
+        "web",
+    ],
+    "data": [],
+    "assets": {
+        "web.assets_backend": [
+            "web_remember_tree_column_width/static/src/**/*",
```
thomaspaulb: If you change this to `web_remember_tree_column_width/static/src/js/list_renderer.esm.js`, the module will become compatible with Odoo Online.

thomaspaulb (changes requested):

web_remember_tree_column_width/__manifest__.py
BhaveshHeliconia: @thomaspaulb : Done. Thank you!

thomaspaulb (approved):

PR #3343, opened 2025-10-23
Claude/odoo navbar responsive 011 cu qg qx gdb9zqa m6i cmf4x

agusmoncada:


ivs-cetmix:
@agusmoncada using AI presumes that the one who uses fully understands the result. With this said, please restrain from using the OCA repos as your vibecoding sandbox. Thank you!

PR #3344, opened 2025-10-24
[18.0][ADD] web_widget_product_label_section_and_note_full_label

LauraCForgeFlow:


pedrobaeza:
I think you already have it covered with https://github.com/OCA/web/tree/18.0/web_widget_product_label_section_and_note_name_visibility

LauraCForgeFlow:
Hi @pedrobaeza, thanks for taking a look. We actually developed this module since the [web_widget_product_label_section_and_note_name_visibility](https://github.com/OCA/web/tree/18.0/web_widget_product_label_section_and_note_name_visibility) module did not cover some necessities we had. 

The main limitation we found is that it only allows toggling the visibility of the product name in the description while the line is editable. Once the line becomes read-only, the eye icon disappears and it’s no longer possible to see or guess the full content of the description directly on screen (so users have to open the print preview or send the document to check it).

That behavior works fine when you occasionally want to hide product names, but in business cases where line descriptions are frequently customized, we think it’s more practical and safer to always display the full information. This helps avoid mistakes sending information, and ensures what users see on screen matches what will appear on the final reports.

pedrobaeza:
OK, as you prefer. I think the cons you mention is easily avoidable showing the icon in readonly.

LoisRForgeFlow (approved):
For me this is simpler for the business need. 

I actually tested web_widget_product_label_section_and_note_name_visibility and it didn't even work as expected because with the "eye enabled" I still cannot see the full description

image

 
image

(runboat, only web_widget_product_label_section_and_note_name_visibility installed)

DavidJForgeFlow:
@pedrobaeza did you see last Lois message? Do you agree with merging this?

pedrobaeza:
Yes, I'm not blocking this. Go ahead with the second review and merge.

DavidJForgeFlow (approved):
Functional review :+1: Could you merge @pedrobaeza ?

pedrobaeza:
Hehe, OK.

/ocabot merge nobump

PR #3122, opened 2025-03-12
[FIX]web_responsive: fixed the issue of file viewer not display in form view.

parvezqureshi:
The attachments of customer invoices appear automatically (see screenshot). When we install oca/web/web_responsive, the attachments do not appear automatically (see screenshot). This is on a computer screen of course.

PR #3347, opened 2025-10-30
[IMP]web_responsive: fixed the issue of file viewer not display in form view

len-foss:
Forward port of https://github.com/OCA/web/pull/3122/

The removal of the feature by the module has been buggy, causing the wrong file to be displayed on the wrong record. Not sure why this feature is bundled with web_responsive to begin with, but it caused a lot of trouble -- accountant thinking the file was missing, etc because of it. Having to click 3 times to be able to see the pdf of the bill is also a very poor user experience.


cvinh (approved):
Works as expected on our production instance

PR #3363, opened 2025-11-12
[16.0][ADD]web_refresh_from_backend: Refresh views from backend

DemchukM:
# Backend UI Reload Module

This is a **technical module** that allows triggering a **UI reload** from the backend.
It enables triggering the reload action for selected users and record IDs.

**NB:** we depend on the `mail` module because it's where a [partner is added](https://github.com/odoo/odoo/blob/5fab688f9bb91542f40622a9b55489304b1551b9/addons/mail/models/ir_websocket.py#L30) to the notification channel.

---

## 🔧 Helper Function: `reload_views`

A special helper function `reload_views` is added to the `res.users` model.

### **Arguments**

| Argument | Type | Description |
|-----------|------|-------------|
| **model** | `Char` | Model name, e.g. `'res.partner'` |
| **view_types** | `List of Char` *(optional)* | View types to reload, e.g. `["form", "kanban"]`. Leave blank to reload all views. |
| **rec_ids** | `List of Integer` *(optional)* | The view will be reloaded only if a record with an ID from this list is present in the view. |

---


web_refresh_from_backend/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+
+{
+    "name": "Web Refresh From Backend",
+    "summary": "Refresh views from backend",
+    "version": "16.0.1.0.0",
+    "category": "Web",
+    "license": "LGPL-3",
+    "author": "Cetmix, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/web",
+    "depends": ["web", "bus", "base"],
```
ivs-cetmix: Can we  leave just "web" as it already depends on both "bus" and "base"?

web_refresh_from_backend/static/src/views/form/form_controller_patch.esm.js
```diff
@@ -0,0 +1,98 @@
...
+
+        // Reload data from server
+        await record.load();
+
+        // Update the view (only if component is still mounted)
+        if (this.model && this.model.root) {
+            this.render(true);
+        }
+    },
+});
```
GabbasovDinar, line 181: I think soft_reload action is better to use to reload records:

https://github.com/odoo/odoo/blob/16.0/addons/web/static/src/webclient/actions/client_actions.js#L56

We can avoid being tied to the view type and call soft reload, which updates the data without reloading the browser.

GabbasovDinar (changes requested):

web_refresh_from_backend/static/src/views/form/form_controller_patch.esm.js
GabbasovDinar: You can, by analogy with the [web_notify](https://github.com/OCA/web/tree/16.0/web_notify) module, receive a message from the backend and call doAction(“soft_reload”);

JordiBForgeFlow:
What if the user is making edits to that view that will be refreshed? She would use the info?

ivs-cetmix:
> What if the user is making edits to that view that will be refreshed? She would use the info?

@JordiBForgeFlow yes, this may happen. This is why there is a [special warning](https://github.com/cetmix/oca-web/blob/16.0-t5066-web_refresh_from_backend-refresh_views_from_backend/web_refresh_from_backend/readme/DESCRIPTION.md#%EF%B8%8F-important-notes) in the module description.

web_refresh_from_backend/static/src/views/form/form_controller_patch.esm.js
DemchukM: Yes, you're right, that would be better, I've made the changes.

ivs-cetmix (changes requested):
After the last commit that implements the standard 'reload' mechanism wizard windows remain opened when a refresh is called from the backend. 
At the same time the form in the background is being refreshed.

Image

Before the last commit it was working fine.

[Code](https://github.com/cetmix/cetmix-tower/blob/b304cf53e3b1f7e7c4c7dd64388ab27ac7138a2a/cetmix_tower_server/wizards/cx_tower_jet_state_wizard.py#L74)
NB: adding 
```python
return {
            "type": "ir.actions.act_window_close",
        }
```
doesn't solve the issue.

ivs-cetmix (approved):
Works now. Functional review LGTM

GabbasovDinar (approved):

ivs-cetmix:
Hey @OCA/web-maintainers would appreciate your review! We would like to forward port it to newer Odoo versions asap.

mathben:
This module needs python code to action this refresh, from backend. Do you have existing module that use this module with this usage? I am curious to know a real use case.
@ivs-cetmix @DemchukM 
https://github.com/cetmix/oca-web/blob/85ab885c213badf0254d23c3968203574a1c8a43/web_refresh_from_backend/readme/USAGE.md

jasalt:
As one possible use case, I am interested to use this with Project module as it could allow updating the task kanban view (`https://myodoo.test/odoo/action-198/1/tasks`) "real-time" for all viewers when there's updates in task states.

Currently each viewer needs to do a full refresh for the kanban view every time when there's updates which is clumsy and disrupts calls/video conferencing within the same browser tab.

zamberjo:
> As one possible use case, I am interested to use this with Project module as it could allow updating the task kanban view (`https://myodoo.test/odoo/action-198/1/tasks`) "real-time" for all viewers when there's updates in task states.
> 
> Currently each viewer needs to do a full refresh for the kanban view every time when there's updates which is clumsy and disrupts calls/video conferencing within the same browser tab.

Hi @jasalt , I have actually implemented this specific feature in PR #3396. I believe it covers the use case you are looking for and provides a clean solution.
Since you have a clear view of the requirements for this, your feedback on that PR would be incredibly valuable. Could you please take a look or give it a test run? It would really help us move forward and get it merged.
Thanks in advance!

jasalt:
@zamberjo Thank's, I'll check it too.

jasalt:
I read the code of this module yesterday evening (the later commit in the https://github.com/cetmix/cetmix-tower/pull/469 PR and tried to get it running in Odoo 19. `_sendmany` method seems to have been removed from `self.env["bus.bus"]`) but it ran without error with `res_users.py` patch:

```
-        notifications = [
-            [user.partner_id, "web.refresh_view", bus_message] for user in self
-        ]
-        self.env["bus.bus"]._sendmany(notifications)
+        bus_service = self.env["bus.bus"]
+        for user in self:
+            bus_service._sendone(
+                user.partner_id,
+                "web.refresh_view",
+                bus_message.copy(),
+            )
```
Trying to trigger the task kanban view change however failed from `odoo shell`:
```
users = env["res.users"].search([])
users.reload_views(
model="project.task",
view_types=["kanban"], # only kanban boards
rec_ids=[] # leave empty to refresh every kanban list
)
```
Command didn't give feedback and client wasn't receiving anything for what I could see. Couldn't get the JavaScript patches loading to the client and not sure if there's been some change in asset registering or something else.

ivs-cetmix:
> This module needs python code to action this refresh, from backend. Do you have existing module that use this module with this usage? I am curious to know a real use case. @ivs-cetmix @DemchukM https://github.com/cetmix/oca-web/blob/85ab885c213badf0254d23c3968203574a1c8a43/web_refresh_from_backend/readme/USAGE.md

@mathben 
Yes, this module was developed for the Cetmix Tower new feature - Jets -  and it's already being used in prod. Here is the PR: https://github.com/cetmix/cetmix-tower/pull/465
Check [this video](https://drive.google.com/file/d/1k-kAdzI_hkTB5iwH4rdQuUWf7nd9UGnv/view?usp=sharing) and [this video](https://drive.google.com/file/d/14-6wlPTCr701aR9xXc_iPQpkXmItWNMK/view?usp=sharing) to see how it works.

However because this PR got stuck in the OCA we had to create a copy of the module in our own repo: https://github.com/cetmix/cetmix-tower/tree/16.0/cx_web_refresh_from_backend

NB: we changed the module name so in case OCA will merge the PR we can remove our copy.

pedrobaeza:
Isn't this the same as https://odoo-community.org/shop/web-actions-view-reload-4380#attr=21694 ?

ivs-cetmix:
> Isn't this the same as https://odoo-community.org/shop/web-actions-view-reload-4380#attr=21694 ?

@pedrobaeza Similar but not the same.
First of all it uses direct refresh instead of sending an action. Same way as it's done in the web_refresher: https://github.com/OCA/web/tree/18.0/web_refresher
Second it's more precise because it allows to select specific users and views to refresh.
Important note: we tried to use the "send action" approach, however it has some issues. Eg if you have a confirmation in the wizard (`confirm="Are you sure?"`) then this confirmation gets stuck.

IMPORTANT: there are some updates and minor fixes that we implemented in our copy of the module that are not ported to this PR yet. We need to apply them before merge.

ivs-cetmix:
> I read the code of this module yesterday evening (the later commit in the [cetmix/cetmix-tower#469](https://github.com/cetmix/cetmix-tower/pull/469) PR and tried to get it running in Odoo 19. `_sendmany` method seems to have been removed from `self.env["bus.bus"]`) but it ran without error with `res_users.py` patch:
> 
> ```
> -        notifications = [
> -            [user.partner_id, "web.refresh_view", bus_message] for user in self
> -        ]
> -        self.env["bus.bus"]._sendmany(notifications)
> +        bus_service = self.env["bus.bus"]
> +        for user in self:
> +            bus_service._sendone(
> +                user.partner_id,
> +                "web.refresh_view",
> +                bus_message.copy(),
> +            )
> ```
> 
> Trying to trigger the task kanban view change however failed from `odoo shell`:
> 
> ```
> users = env["res.users"].search([])
> users.reload_views(
> model="project.task",
> view_types=["kanban"], # only kanban boards
> rec_ids=[] # leave empty to refresh every kanban list
> )
> ```
> 
> Command didn't give feedback and client wasn't receiving anything for what I could see. Couldn't get the JavaScript patches loading to the client and not sure if there's been some change in asset registering or something else.

@jasalt  well I wouldn't expect a frontend module developed for 16.0 running on 19.0 without any modifications. 
So I suggest to have it merged and if it will be then migrate it. We are planning to migrate it to 18.0 asap, you can pick it up from there when ready.

pedrobaeza:
OK, it was just for not duplicating features. You can highlight the differences in the README for others with the same doubts.

ivs-cetmix:
> OK, it was just for not duplicating features. You can highlight the differences in the README for others with the same doubts.

Good point! 👍  Will do.

ivs-cetmix:
Hi @pedrobaeza I've updated the module description as per your request. Also added some improvements that we did in [our own copy](https://github.com/cetmix/cetmix-tower/tree/16.0/cx_web_refresh_from_backend) of this module that we are already using in prod. Would appreciate a review)

etobella (approved):

pedrobaeza:
Please fix the README fragments

ivs-cetmix:
@pedrobaeza updated the description file to align it with the original one that was successfully merged in our local repo using the OCA bot before.

PR #3375, opened 2025-11-29
[18.0][MIG] web_widget_remaining_days_exact_date: Migration to 18.0

CILC98:


web_widget_remaining_days_exact_date/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 Tecnativa - Carlos Roca Trescloud - César León
```
celm1990:
```suggestion
# Copyright 2025 Tecnativa - Carlos Roca 
# Copyright 2025 Trescloud - César León
```

web_widget_remaining_days_exact_date/models/disable_remaining_days_rule.py
```diff
@@ -0,0 +1,210 @@
...
+                ("company_id", "=", record.company_id.id),
+            ]
+            if self.search_count(domain) > 0:
+                raise models.ValidationError(
+                    _(
+                        "There is already a rule for the model '%s'."
+                        " You cannot create two rules for the same model."
+                    )
+                    % record.res_model_id.name
+                )
```
celm1990, line 85:
```suggestion
                raise models.ValidationError(
                    self.env._(
                        "There is already a rule for the model '%s'."
                        " You cannot create two rules for the same model.",
                        record.res_model_id.name
                   )
                )
```

web_widget_remaining_days_exact_date/models/disable_remaining_days_rule.py
```diff
@@ -0,0 +1,210 @@
...
+                            0,
+                            0,
+                            {
+                                "name": name,
+                                "technical_name": technical_name,
+                                "active": active,
+                            },
+                        )
+                        for name, technical_name, active in view_types
+                    ],
```
celm1990:
```suggestion
                    "line_ids": [
                        Command.create(
                            {
                                "name": name,
                                "technical_name": technical_name,
                                "active": active,
                            },
                        )
                        for name, technical_name, active in view_types
                    ],
```

web_widget_remaining_days_exact_date/models/disable_remaining_days_rule.py
celm1990, line 1: Please follow the OCA guidelines. The order of fields, CRUD methods, etc. is described here:
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#37models

web_widget_remaining_days_exact_date/tests/test_main.py
```diff
@@ -0,0 +1,110 @@
+from odoo import exceptions
+from odoo.tests import common, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestWidgetRemainingDaysExactDate(common.TransactionCase):
```
celm1990: Please inherit from [BaseCommon](https://github.com/odoo/odoo/blob/9fc9559ef06908eb788c491132d3aa1a1de91b8f/odoo/addons/base/tests/common.py#L20)
```suggestion
class TestWidgetRemainingDaysExactDate(BaseCommon):
```

web_widget_remaining_days_exact_date/tests/test_main.py
```diff
@@ -0,0 +1,110 @@
...
+    @classmethod
+    def setUpClass(cls):
+        return super().setUpClass()
```
celm1990: If this function does nothing, please remove it. Otherwise, implement the logic here to prepare the data for the test.
```suggestion
```

web_widget_remaining_days_exact_date/views/disable_remaining_days_rule_views.xml
```diff
@@ -0,0 +1,49 @@
...
+<odoo>
+    <record id="disable_remaining_days_rule_view_tree" model="ir.ui.view">
+        <field name="name">disable.remaining.days.rule.view.list</field>
+        <field name="model">disable.remaining.days.rule</field>
+        <field name="arch" type="xml">
+            <list
+                string="Disable Remaining Days Rules"
+                create="true"
+                edit="true"
+                delete="true"
```
celm1990: These options are the default values, so it is not necessary to set them.
```suggestion
```

web_widget_remaining_days_exact_date/static/src/widgets/remaining_days_fields.esm.js
```diff
@@ -0,0 +1,49 @@
+/** @odoo-module **/
```
celm1990:
```suggestion
```

web_widget_remaining_days_exact_date/wizards/disable_remaining_days_rule_line_wizard.py
```diff
@@ -0,0 +1,32 @@
...
+    technical_name = fields.Char(
+        string="Technical Name of view type",
+        required=True,
+        readonly=True,
+        help="Technical name of the view type where the remaining days widget"
+        " will be disabled.",
+    )
+    active = fields.Boolean(
+        string="Active record",
+    )
```
celm1990: I don’t see the point of having an active field in a TransientModel. Why do you need this field? I think it can be removed.
```suggestion
```

celm1990 (changes requested):
Code review: I think it is necessary to split the commits. Please keep a single commit with only the changes related to the migration. If you want to improve this module, I think it is better to create a new PR with the new features, or at least put them in a separate commit and not in the migration commit.
What do you think, @CarlosRoca13 ?

CILC98:
@celm1990 Hello and Happy New Year. I've already made two commits as you recommended. I have also corrected the suggestions.

CarlosRoca13:
Hello! Thank you for your contribution. Could you please add an explanation of what the changes you’ve added consist of, in order to make them easier to understand?

giarve:
I don't think a wizard or any model is needed for this module. I would rather close and do the straightforward migration for it. Alternatively create a new module with a new name that is clear

EDIT: see https://github.com/OCA/web/pull/3276

PR #3376, opened 2025-11-29
[19.0][FIX] web_responsive: support create/edit attributes in form/list buttons

yanchirino:
- Refactor form and list view button templates to respect create="false" and edit="false" attributes passed from views.
- Allows dynamic enabling/disabling of create/edit actions based on view configuration.
- Improves UX and compatibility with Odoo standard view options.

### DIFF
--- a/web_responsive/static/src/legacy/xml/form_buttons.xml
+++ b/web_responsive/static/src/legacy/xml/form_buttons.xml
@@ -37,6 +37,7 @@
                 class="btn btn-primary o_form_button_save"
                 data-hotkey="s"
                 t-on-click.stop="() => this.saveButtonClicked({closable: true})"
+                t-if="model.root.isInEdition"
             >
                 <t t-call="web_responsive.icon_button_save" />
             </button>
@@ -47,6 +48,7 @@
                 class="btn btn-secondary o_form_button_cancel"
                 data-hotkey="j"
                 t-on-click.stop="discard"
+                t-if="model.root.isInEdition"
             >
                 <t t-call="web_responsive.icon_button_discard" />
             </button>
@@ -57,6 +59,7 @@
                 class="btn btn-secondary o_form_button_create"
                 data-hotkey="c"
                 t-on-click.stop="create"
+                t-if="!model.root.isInEdition and canCreate"
             >
                 <t t-call="web_responsive.icon_button_create" />
             </button>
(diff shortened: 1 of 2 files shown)


yanchirino:
@OCA/web-maintainers 
Hi team, I’ve submitted a small contribution aimed at improving the flexibility and consistency of the interface. Please let me know if any adjustments are needed. Thanks!

pedrobaeza:
Can you specify what "Improves UX and compatibility with Odoo standard view options." means? Maybe with screenshots?

yanchirino:
Before the adjustment, when the `create="false"` and `edit="false"` attributes were included, they were ignored by the theme. As a result, buttons like the `create` button in different views were not hidden. After the adjustment, buttons with this attribute inserted in the view are hidden, just as they would be in the native operation.

image
image
image
image

@pedrobaeza

web_responsive/static/src/legacy/xml/form_buttons.xml
```diff
@@ -77,19 +80,7 @@
                 class="btn btn-outline-primary o_form_button_create"
                 data-hotkey="c"
                 t-on-click.stop="create"
-            >
```
pedrobaeza, line 80: But I mean the extra stuff and specifically this change.

yanchirino: image

Screenshot at Dec 03 13-57-17

I made that change because the replacement was duplicated. Both XPath expressions pointed to the same button, so the template was rewriting it twice.

In the first image, the original Odoo code only contains one button.
But in the second image, the template was replacing that same button twice, so I removed the duplicate block.

pedrobaeza: Not really, as the first xpath replaces the first found element with one thing, and then, the second xpath will replace the second one found.

yanchirino: You're right, I adjusted the commit just to keep the change related to the conditional about whether the button should be shown or not.

pedrobaeza (approved):

yanchirino:
Hi @OCA/web-maintainers could someone else please take a look at this PR so we can move forward with the merge process?

etobella (approved):
Tested and code review.

Thanks for the effort

/ocabot merge patch

PR #3381, opened 2025-12-04
[DRAFT][18.0][MIG] web_widget_text_markdown

mathieudelva:
Migration oca-port from 14.0


PaulGoubert (approved):
Functional review:
Tested on Runboat (no demo data)
Notes field on Groups displays raw markdown (no toolbar), which is expected
as the widget is only applied in demo views.
Module behaves as a reusable widget and does not alter standard views.
No JS error observed.

Aldeigja:
@mathieudelva could you please trigger a rebuild since the runboat is not available for testing?

ypapouin:
@mathieudelva, please rebase to trigger a rebuild

PaulGoubert (approved):

ypapouin:
FYI, I "made" (AI-assisted) an alternative module [1] for this functionality, built as a wrapper upon draftly [2] that is more in line with Odoo's philosophy of having a single view for viewing and editing.

[1] https://github.com/decgroupe/odoo-addons-dec/tree/18.0/web_widget_markdown
[2] https://draftly.nexul.in/playground

image

web_widget_text_markdown/static/src/js/web_widget_text_markdown.js
```diff
@@ -0,0 +1,148 @@
...
+/* Copyright 2014 Sudokeys <http://www.sudokeys.com>
+ * Copyright 2017 Komit - <http:///komit-consulting.com>
+ * Copyright 2019 Alexandre Díaz - <dev@redneboa.es>
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+odoo.define("web_widget_text_markdown.FieldTextMarkDown", function (require) {
+    "use strict";
+
+    var basic_fields = require("web.basic_fields");
+    var field_registry = require("web.field_registry");
+    var core = require("web.core");
```
Kev-Roche, line 11: I have  errors on my browser : 
- `The following modules are needed by other modules but have not been defined, they may not be present in the correct asset bundle:  Array(3) [ "web.basic_fields", "web.field_registry", "web.core" ]`
- `The following modules could not be loaded because they have unmet dependencies, this is a secondary error which is likely caused by one of the above problems:  Array [ "@web_widget_text_markdown/js/web_widget_text_markdown" ]`

Kev-Roche (changes requested):
It appears that the JS is not 18 compatible, it needs to be correctly migrated.

PR #3382, opened 2025-12-04
[19.0][MIG]: web_m2x_options

Alexgars73:
Migration to v19


Alexgars73:
@carlos-lopez-tecnativa @CRogos could you review it please?

CRogos (changes requested):
@Alexgars73 I've tested on the res_partner form and the system parameter 'web_m2x_options.create' and it worked for the country and tags field, but not for the res_partner_many2one widget
image

The "web_m2x_options.limit" setting and 'limit' option is working on the tags field, but did not work for country and parent_id (res_partner_many2one widget
).

Alexgars73:
@CRogos do we have the same problem in v18 and v17? Thanks for the review

CRogos:
@Alexgars73 In v18 web_m2x_options.limit and web_m2x_options.create is working on these fields.

Alexgars73:
@CRogos i made some changes to fix the problems, I've also removed the "search more" feature, since Odoo seems to have removed it. Could you confirm if this makes sense?

CRogos (changes requested):
I can confirm the limit, create, create_edit system paramters are now working.

The limit option on the parent_id (widget: res_partner_many2one) is not working. When I change the value to `options="{'limit':2}"` the global limit is taken.  Same on country_id, but on the category_id widget="many2many_tags") the limit option is working.

I also realized that the create + create and edit on country is visible, which is not the case when this module is not installed (this means that the odoo native options are not taken into account: `options="{&quot;no_open&quot;: True, &quot;no_create&quot;: True}"`):

image

The search more option controls this field in Odoo 17. If there are more results (in this screenshot limit=3), the show more option is displayed (or not if option is set to False)
image

Thank you for your work so fare.

CRogos:
The limit attribute in the xml is now working for category_id and parent_id, but not for country_id and county_id.

Alexgars73:
@CRogos I made some changes to correct that.
What I mean by "search more" is that Odoo versions prior to 19 allowed you to select the option to search more in the autocomplete feature, but this has now been removed, so in my opinion, this feature shouldn't be included in the migration.

V18:
image

V19:
image


This is the commit where is deleted: https://github.com/odoo/odoo/blob/0ecaa6e4c359681daf90c1757bcf71ba5e4d305c/addons/web/static/src/views/fields/relational_utils.js

CRogos:
I think this is still possible to select, as long there is one result... but not sure if this is needed to be configurable.
image

The other two findings above on the xml options are still not fixed.

Alexgars73:
@CRogos Yes, it is possible to search more that is right,  but what I’m referring to is that Odoo wants this to be not configurable. On the other hand, I tested and the limit option in the settings seems good. Could you tell me which field you tested? Many thanks.

CRogos:
@Alexgars73 I've also retested the limit on country_id and county_id and you are right, this is working correct. Probably I've forgot a reload last time.


> I also realized that the create + create and edit on country is visible, which is not the case when this module is not installed (this means that the odoo native options are not taken into account: `options="{&quot;no_open&quot;: True, &quot;no_create&quot;: True}"`):
> 
> image
> 

The on the country_id field on contact, there is a `no_create: True` set in the options. Without web_m2x_option, there is no create shown to the user on this field. So this module has the effect that the native options do not work anymore.

image

Alexgars73:
Hello @CRogos i made some changes to correct that could you review now? Thanks.

Alexgars73:
@pedrobaeza could you check please?

CRogos:
@Alexgars73 when the no_create option is set, this should also prevent create and edit...

image

rf-ife:
> @Alexgars73 when the no_create option is set, this should also prevent create and edit...
Even in v18.0, when you set web_m2x_options.create=False, it only disables the 'Create' option. 'Create and Edit' is still there, so I don't think it's a bug @CRogos

pedrobaeza:
Yes, indeed that's on purpose. In fact, I always disable quick-create "Create" option in my instances, but not "Create and edit...", that let's you create them, but having an extra step for avoiding accidental creations.

rf-ife:
@Alexgars73 , maybe search_more should be removed from the module description. If it's no longer supported

CRogos:
I think we are talking about different settings. 

In Odoo without web_m2x_options the UI looks like this:

image

When installing web_m2x_options, but not setting any option it looks like this:
image

Therefore by just installing this module, the behavior not to show any create options is changed. (And I don't think this is your intension?)

I am talking about this no_create option:
https://github.com/odoo/odoo/blob/8090d78c057766fffd5ef75a94b210fde643f3fe/odoo/addons/base/views/res_partner_views.xml#L130-L131

Alexis-SDK:
Anys news here ?

Alexis-SDK (approved):

CRogos (changes requested):
Still not working correctly.... see comments above.

Alexis-SDK:
@CRogos The module works the same as in v18

It’s stated in the usage documentation.

> web_m2x_options.create: boolean (Default: depends on whether the user has create rights)
> web_m2x_options.create_edit: boolean (Default: depends on whether the user has create rights)

Let’s merge it 🚀

CRogos (approved):
I was not aware that the bug is already in v18. After the installation of web_m2x_options, the options are not working like intended. If "no_create" is true, it should also disable "create and edit ..."

`<field name="country_id" placeholder="Country" class="o_address_country" options="{&quot;no_open&quot;: True, &quot;no_create&quot;: True}"/>
`
Image

Image

njeudy:
@CRogos in fact it's a feature, if you want tu force record create with all field and compelete form .. so it's ok for me, no bug at all !

njeudy (approved):
Ok for me !

njeudy:
@legalsylvain @pedrobaeza @denaux something wrong for merging this PR ?

CRogos:
> @CRogos in fact it's a feature, if you want tu force record create with all field and compelete form .. so it's ok for me, no bug at all !

@njeudy I disagree. If you install web_m2x_options and you do not set any options, the default behavior in odoo should not change. 
I agree that this behaviour should be configurable by setting `options="{'no_open': True, 'no_create': True, 'no_create_edit': False}"` on the field xml.

But the global setting should never overwrite field specific settings on any view. 

There is a reason, why countries relation not creatable by default, because I don't see any reason why it should be created manually. And changing this default is not a feature it is a bug to me.

njeudy:
@CRogos I just understood your need and your request — indeed, I did a review to indicate the modification to make for it to work. In Odoo, no_create hides both lines.

web_m2x_options/static/src/components/form.esm.js
```diff
@@ -0,0 +1,333 @@
...
+        props.canQuickCreate = attrs.can_create
+            ? evaluateBooleanExpr(attrs.can_create)
+            : true;
+    }
+    return props;
+}
+
+export function m2o_options_props_create_edit(props, attrs, options) {
+    const canCreateEdit = evaluateSystemParameterDefaultTrue("create_edit");
+    if (options.no_create_edit) {
```
njeudy: add if (options.no_create || here to keep odoo no_create options correct (odoo no_create option disable the two lines !)

njeudy (changes requested):

alexey-pelykh (commented):
Thanks for the migration work.

I've reviewed the JS approach for 19.0 and have a few observations:

**`no_create` should also disable `create_edit` (confirms @njeudy's comment)**

In `m2o_options_props_create` (line ~36), `options.no_create` is now correctly handled to disable quick create. However, the same `no_create` check is missing from `m2o_options_props_create_edit`. In stock Odoo, `no_create` disables both "Create" and "Create and edit..." -- this module should respect that:

```js
export function m2o_options_props_create_edit(props, attrs, options) {
    const canCreateEdit = evaluateSystemParameterDefaultTrue("create_edit");
    if (options.no_create || options.no_create_edit) {  // <-- add options.no_create here
        props.canCreateEdit = false;
    } else if ...
```

**`AvatarMany2XAutocomplete` props not extended**

The 18.0 version patches both `AvatarMany2XAutocomplete.props` and `Many2XAutocomplete.props` with `fieldColorProps`. The 19.0 version only patches `Many2XAutocomplete`. If `AvatarMany2XAutocomplete` still exists as a separate component in 19.0, the `field_color` option will break on avatar many2one fields.

**`search_more` support dropped silently**

The 18.0 version includes `m2o_options_props_search_more` and `m2m_options_props_search_more` plus the corresponding validation schema entries. These are all removed in 19.0. If this is intentional (e.g., Odoo 19 handles search_more natively now), it would be good to mention it in the PR description or ROADMAP. The `web_m2x_options.search_more` system parameter is still listed in `get_web_m2x_options()` but no JS code reads it, so it's dead config.

**Minor: `evaluateSystemParameterNumber` helper**

A new helper was added but only used for `limit` in the `Many2One.prototype` patch. The `m2o_options_props_limit` function still uses inline `Number()` calls. Consider using the helper consistently or removing it if not needed.

The CI "test with OCB" failure is a Docker login issue on ghcr.io, unrelated to the code.

Alexgars73:
@CRogos @Alexis-SDK @njeudy @alexey-pelykh  I did the change to reply odoo base, AvatarMany2XAutocomplete does not exists in odoo19, I deleted search more code.

dannyadair (approved):
LGTM

CRogos (approved):
Tested "create and edit" issue on country and now LGTM. thx.

gdgellatly:
I encountered an issue in debug mode on crm pipeline screen. The following patch to this PR resolved.

```bash
diff --git a/web_m2x_options/static/src/components/form.esm.js b/web_m2x_options/static/src/components/form.esm.js
index 0ea6685..6428fb6 100644
--- a/web_m2x_options/static/src/components/form.esm.js
+++ b/web_m2x_options/static/src/components/form.esm.js
@@ -1,7 +1,7 @@
 import * as many2OneField from "@web/views/fields/many2one/many2one_field";
 import * as many2one from "@web/views/fields/many2one/many2one";
 import {FormController} from "@web/views/form/form_controller";
-import {Many2One} from "@web/views/fields/many2one/many2one";
+import {Many2One, KanbanMany2One} from "@web/views/fields/many2one/many2one";
 import {Many2OneReferenceField} from "@web/views/fields/many2one_reference/many2one_reference_field";
 import {Many2XAutocomplete} from "@web/views/fields/relational_utils";
 import {evaluateBooleanExpr} from "@web/core/py_js/py";
@@ -128,6 +128,15 @@ patch(Many2OneReferenceField, {
     },
 });
 
+// FIXME: KanbanMany2One does not support m2o_options_props (e.g. searchLimit).
+// Accepting these props prevents "unknown key" validation errors when the
+// field description is built with m2o_options_props (e.g. kanban many2one avatar).
+KanbanMany2One.props = {
+    ...KanbanMany2One.props,
+    ...fieldColorProps,
+    searchLimit: {type: Number, optional: true},
+};
+
 patch(many2OneField.Many2OneField.prototype, {
     get m2oProps() {
         const props = super.m2oProps;
```

yankinmax:
Hello @Alexgars73 Can you pls address the review and update PR?

Alexgars73:
Hello @gdgellatly I applied the patch could you check if it is ok now, thanks for the review.

CRogos:
@njeudy could you update your review?

manuelcalerosolis:
Tested "create and edit" LGTM.

pedrobaeza:
@njeudy can you please update your review?

CRogos:
@Alexgars73 There are some commits which could be squashed:

image

@pedrobaeza I think @njeudy isn't active anymore and the change was done. Can this be overruled?

pedrobaeza:
Yes, as he doesn't answer, and his comments seem to be solved, we can overpass his requested changes. Waiting for the little squashing for launching the merge command.

Alexgars73:
@pedrobaeza @CRogos I'm not entirely clear on which commit to squash for the marked commits; should each one be squashed against the one before it?

pedrobaeza:
Yes

CRogos:
@Alexgars73 can you do the squash so that we can merge?

Alexgars73:
@pedrobaeza @CRogos done, if you can check if it's correct.

yankinmax:
Hello @OCA/web-maintainers can you pls take a look and possibly trigger merge?

Yusuke1998:
Ping

CRogos:
@pedrobaeza can you merge? I don't think @OCA/web-maintainers is working.

AungKoKoLin1997:
@Alexgars73 I found there is a issue when I test this module. 
Can you please look at this issue?
https://github.com/OCA/web/issues/3550

PR #3385, opened 2025-12-09
[18.0][IMP] Make is_redirect_home field always visible in user form

andreampiovesana:
set menu as initial view always activable
'because it's better for UX

### DIFF
--- a/web_responsive/views/res_users_views.xml
+++ b/web_responsive/views/res_users_views.xml
@@ -69,7 +69,7 @@
         <field name="arch" type="xml">
             <data>
                 <field name="action_id" position="after">
-                    <field name="is_redirect_home" invisible="action_id" />
+                    <field name="is_redirect_home" />
                 </field>
             </data>
         </field>


ivs-cetmix:
@andreampiovesana thank you for your contribution! 
Could you please update the PR name and put the correct tag (`[IMP]`) in it?
Also please update the commit message in accordance with the OCA guidelines. Eg

```git
[IMP] web_seponsive user settings

Put here the reason behind the changes. 
Eg 'because it's better for UX
```

andreampiovesana:
please merge

ivs-cetmix (approved):
Code review LGTM

pedrobaeza:
Please follow commit message format: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

andreampiovesana:
is an ultra simple change... we can close thi PR

pedrobaeza:
As you prefer. Anyway, I think the invisible condition is because if you have set an initial action, there's no redirection to home, so I think it's correct as it is now, or am I wrong?

PR #3371, opened 2025-11-21
[17.0] [MIG] web_font_size_report_layout

syci:
Odoo’s default report layout uses a fixed font size for all PDF documents.
In many real use cases, especially when invoice or sales order lines include long product descriptions or detailed content, this results in documents that are difficult to read or excessively long.

This module introduces a configurable font size in the Document Layout settings.
Administrators can choose between 9pt, 10pt, 12pt, and 14pt to optimize readability and space usage, adapting reports to company needs (compactness vs. visibility).

PR #3389, opened 2025-12-11
[18.0][ADD] web_font_size_report_layout: web_font_size_report_layout module added.

adasatorres:


rrebollo:
This Odoo module adds the capability to customize font size in all PDF reports throughout the system. Through a simple configuration, administrators can select from a range of predefined font sizes (measured in points) to optimize readability and appearance of invoices, quotations, sales orders, and other printed documents.

The module offers the following font size options (in points):

| Option | Size | Description |
|--------|------|-------------|
| Extra Small | 9pt | For maximum compactness |
| Standard | 10pt | Odoo's default size |
| Small | 11pt | Just a little larger |
| Large | 12pt | Better readability |
| Extra Large | 13pt | Really Large |
| 2Extra Large | 14pt | For high visibility needs |

web_font_size_report_layout/models/base_document_layout.py
```diff
@@ -0,0 +1,29 @@
...
+    @api.onchange("report_font_size")
+    def _onchange_report_font_size(self):
+        func = getattr(self, "_compute_preview", None)
+        if not callable(func):
+            return
+        try:
+            func()
+        except Exception as exc:  # pylint: disable=broad-except
+            _logger.debug("Failed computing document layout preview: %s", exc)
+            self.preview = False
```
rrebollo: Guessing the approach with:  
- Extending `_compute_preview`  
- Adding `report_font_size` to the `@api.depends`  
- Returning `super()._compute_preview()`  

…did not produce the expected result.

web_font_size_report_layout/models/res_company.py
```diff
@@ -0,0 +1,49 @@
+# Copyright 2025 Binhex
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models
+
```
rrebollo, line 4:
```suggestion

STYLE_FIELDS = {
    "external_report_layout_id",
    "font",
    "report_font_size",
    "primary_color",
    "secondary_color",
}
```

web_font_size_report_layout/models/res_company.py
```diff
@@ -0,0 +1,49 @@
...
+            ("12", "12 pt"),
+            ("13", "13 pt"),
+            ("14", "14 pt"),
+        ],
+        string="Report font size",
+        default="11",
+        help="Base font size for PDF content (in points), applied on the external "
+        "report layout.",
+    )
+
```
rrebollo, line 31:
```suggestion
    @api.model
    def _should_update_styles(vals_list):
        """Check if any values contain style-related fields."""
        return any(not STYLE_FIELDS.isdisjoint(values) for values in vals_list)
```

web_font_size_report_layout/models/res_company.py
```diff
@@ -0,0 +1,49 @@
...
+        companies = super().create(vals_list)
+        style_fields = {
+            "external_report_layout_id",
+            "font",
+            "report_font_size",
+            "primary_color",
+            "secondary_color",
+        }
+        if any(not style_fields.isdisjoint(values) for values in vals_list):
+            self._update_asset_style()
```
rrebollo:
```suggestion
        if self._should_update_styles(vals_list):
            self._update_asset_style()
```

web_font_size_report_layout/models/res_company.py
```diff
@@ -0,0 +1,49 @@
...
+        res = super().write(values)
+        style_fields = {
+            "external_report_layout_id",
+            "font",
+            "report_font_size",
+            "primary_color",
+            "secondary_color",
+        }
+        if not style_fields.isdisjoint(values):
+            self._update_asset_style()
```
rrebollo:
```suggestion
        if self._should_update_styles([values]):
            self._update_asset_style()
```

rrebollo (changes requested):
CR. Thank you for your contribution. Please consider my suggestions. Specially the DRY ones.

web_font_size_report_layout/models/base_document_layout.py
adasatorres: @rrebollo  I don't see the point of inheriting _compute_preview if you won't add any logic to the method — we only need it to run when report_font_size changes. Moreover, declaring a new @api.depends on the overriding method replaces the parent's depends list, so the final implementation's @api.depends is the one the ORM uses; therefore you would need to include all fields that other addons might have added.

web_font_size_report_layout/models/res_company.py
adasatorres: @rrebollo I agree.

rrebollo (approved):
Code Review. LGTM!

web_font_size_report_layout/models/base_document_layout.py
rrebollo: @adasatorres I wasn't aware that we need to list all dependencies from other addons. My understanding is this hasn't been consistent across Odoo versions, and the source code shows mixed approaches (see examples below).

While in many OCA repos you will fin a lot of overrides trying to list all dependencies, I've personally extended `_compute` methods without considering other addons' depends. If required, this would seriously discourage such extensions.

Let's verify our assumptions. My original suggestion came from caution around `onchange` methods.

**Examples:**
- https://github.com/odoo/odoo/blob/083d53c688a0d18a1f4594b9fcbbfa738aa5e86d/addons/account/models/account_move.py#L1780-L1781
- https://github.com/odoo/odoo/blob/083d53c688a0d18a1f4594b9fcbbfa738aa5e86d/addons/account_edi/models/account_move.py#L113-L114

adasatorres: @rrebollo I take it back. I had read on a forum that it only affected the last “depends”, but I’ve been testing it and there’s no problem. I'm sorry.

web_font_size_report_layout/readme/INSTALL.md
```diff
@@ -0,0 +1,3 @@
+1. Copy the `web_font_size_report_layout` directory to your Odoo addons folder
```
bealdav, line 1: Thanks a lot for your module. Here is a suggestion
```suggestion
1. Link the `web_font_size_report_layout` directory to your Odoo addons path
```

edescalona (approved):
LGTM

ivs-cetmix:
Hey @adasatorres there other PRs for the same module already:
17.0 https://github.com/OCA/web/pull/3371
16.0 https://github.com/OCA/web/pull/3271

Looks like the 17.0 is the most up-to-date for now. So let's merge one and then migrate to other versions.

flotho:
Hi @ivs-cetmix , the v17 module is merged, what's the process to have it in v18 ?

ivs-cetmix:
Hi @flotho we should open a migration PR so the commit history stays preserved. This one should be closed after that.

DorianMAG:
@adasatorres 
Hi, can you rebase plz.
Runboat is not load.
Thx

DorianMAG:
@rrebollo 
@flotho 
I propose migration from 17.0 based on your work @adasatorres 
https://github.com/OCA/web/pull/3661
Regards

PR #3391, opened 2025-12-15
[18.0][IMP] web_form_banner: adj to resolve eslint warnings

AungKoKoLin1997:
This is the warnings before this PR.
image

This is after the PR, all warning will be gone.
@qrtl QT6169

### DIFF
--- a/web_form_banner/static/src/js/web_form_banner.esm.js
+++ b/web_form_banner/static/src/js/web_form_banner.esm.js
@@ -1,17 +1,18 @@
-/** @odoo-module **/
 // Copyright 2025 Quartile (https://www.quartile.co)
 // License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-import {patch} from "@web/core/utils/patch";
 import {onMounted, onWillUnmount} from "@odoo/owl";
 import {FormController} from "@web/views/form/form_controller";
 import {Record} from "@web/model/relational_model/record";
+import {patch} from "@web/core/utils/patch";
 
 const recRoot = (c) => (c && c.model && c.model.root) || null;
 const childSpan = (el) => {
     try {
         return (el && el.querySelector(":scope > span")) || null;
-    } catch {}
+    } catch {
+        // Ignore
+    }
     const f = el && el.firstElementChild;
     return f && f.tagName === "SPAN" ? f : null;
 };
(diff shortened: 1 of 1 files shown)


yostashiro (approved):
LGTM.

nobuQuartile (approved):
LGTM

m-deura (approved):
LGTM!

AungKoKoLin1997:
@oca/web-maintainers 
Could we get this minor fix PR merged?

web_form_banner/static/src/js/web_form_banner.esm.js
etobella: In 18 this is not necessary and you would remove the last warning 😉

web_form_banner/static/src/js/web_form_banner.esm.js
```diff
@@ -22,10 +24,13 @@ const setHtml = (el, html) => {
...
     try {
         return await fn();
-    } catch {}
+    } catch {
+        // Ignore
+    }
     return fb;
 };
 
+/* eslint-disable no-inline-comments */
```
etobella: Why don't you move the comments to the line before. This way you are not disabling a lint condition

web_form_banner/static/src/js/web_form_banner.esm.js
AungKoKoLin1997: > I didn’t fix the /** @odoo-module **/ warnings to avoid future mess. While it’s optional for files under static/src (they’re already transpiled as Odoo modules) per the documentation, keeping the header is harmless.
> https://www.odoo.com/documentation/18.0/developer/reference/frontend/javascript_modules.html#native-javascript-modules

This was the reason I didn't remove as I explained in description text.

web_form_banner/static/src/js/web_form_banner.esm.js
AungKoKoLin1997: I want to reduce the line of codes as much as possible and the comments are very short enough to use as a inline comment in my opinion.

web_form_banner/static/src/js/web_form_banner.esm.js
etobella: Yes, but in 18, by default, everything is `@odoo-module` If you want to avoid it, you need to use `/** @odoo-module ignore **/`. This is not the case, so you don't need it :wink:

web_form_banner/static/src/js/web_form_banner.esm.js
etobella: I know that the diff will be longer, however, in a migration it is expected to have this kind of changes. I would prefer to avoid this kind of disables

web_form_banner/static/src/js/web_form_banner.esm.js
AungKoKoLin1997: Got it. Thank you, Enric. I will remove that.

yostashiro (approved):
Code review. Thanks @etobella for your advice!

etobella (approved):
Technical review

ivs-cetmix:
Hey @AungKoKoLin1997 , thank you for your contribution!
/ocabot merge patch

Issue #3234, opened 2025-07-28
web_theme_classic: as user setting, instead of global on/off?


smaddlsoft:
**Is your feature request related to a problem?**
Some people prefer the new Odoo style, others the old, especially those with restrictions in their eyesight.

**Describe the solution you'd like**
How about making the "classic theme" a setting on the user preference page, just like web_chatter_position

**Describe alternatives you've considered**
Using an accessiblity module like https://apps.odoo.com/apps/modules/18.0/odoo-allinoneaccessibility,
but no OCA solution exists currently on this?

**Additional context**
/

PR #3394, opened 2025-12-16
[18.0][IMP] web_theme_classic Add ability to toggle theme

ljmnoonan:
Now you can either enable by user with Classic Theme Persistent setting, or toggle it on and off like dark mode. It uses a cookie to remember your preference between reloads.

<details>
<summary>Screenshots</summary>

image

image

image

Toggle Hidden because Classic Theme Persistent turned on

image

</details


ljmnoonan:
Closes #3234

web_theme_classic/models/res_users.py
```diff
@@ -0,0 +1,29 @@
...
+# © 2025 Liam Noonan - Pyxiris
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    persistent_classic_theme = fields.Boolean(
```
legalsylvain, line 11:
```suggestion
    persistent_classic_theme = fields.Boolean(
    défaut=True,
```

legalsylvain (changes requested):
Feature looks disabled by défaut. Or did i missed something ?

ljmnoonan:
@legalsylvain yes, I left the persistent option disabled by default so that the toggle on the user menu would be visible and individual users can make a choice whether to turn it on or not, and what mode to use. 
I can change it though if you think it makes more sense to enable persistent mode for all users at install by default

legalsylvain:
> @legalsylvain yes, I left the persistent option disabled by default so that the toggle on the user menu would be visible and individual users can make a choice whether to turn it on or not, and what mode to use.

Well, If someone install this module, he hopes that the feature will be enabled by default. Otherwise, it forces each user to enable it manually. So, yes, please put the value enabled by default. If user want to disable it, he can.
Don't you think ?

ljmnoonan:
@legalsylvain Yes, I see what you mean. I will change it

ljmnoonan:
@legalsylvain All set!

legalsylvain (approved):
thanks !

ljmnoonan:
While working in `web_dark_mode`, I discovered that `res_users_settings.py` is already loaded into the `user.settings` object client side, so we can avoid an unnecessary orm call by using it. This means that `persistent_classic_theme` is just a related field in `res_users`. 

I also updated the tests to HttpCase, which makes a lot more sense in this context

web_theme_classic/readme/CONFIGURE.md
```diff
@@ -0,0 +1,7 @@
+This module allows each user to choose whether they would like input fields to be displayed the "classic" way or the new, standard way (as if this module were not installed)
+
+To do this you can either:
++ Check "Classic Theme Persistent" in user preferences. This will enable the classic theme for that user across all devices.
++ Check the "Classic Theme" toggle in the popover menu triggered bu clicking on the user icon in the navbar. This toggle is only visible when "Classic Theme Persistent" is disabled.
+
```
legalsylvain, line 6: nitpicking : this section should be more in a CONFIGURE.md file IMO.

(not a big deal)

ljmnoonan: Ahh, yes, I see! I will get it fixed tomorrow!
Thanks for catching that!

legalsylvain:
> While working in web_dark_mode, I discovered that res_users_settings.py is already loaded into the user.settings object client side

thanks. good to know.

ljmnoonan:
@jappi00 if you get some time, would you mind reviewing this PR?

jappi00:
@ljmnoonan functional test looks good. Will look into the code now.

web_theme_classic/readme/CONFIGURE.md
```diff
@@ -0,0 +1,7 @@
+This module allows each user to choose whether they would like input fields to be displayed the "classic" way or the new, standard way (as if this module were not installed)
+
+To do this you can either:
++ Check "Classic Theme Persistent" in user preferences. This will enable the classic theme for that user across all devices.
```
jappi00, line 4: I think you should write "Uncheck" since you enabled the theme by default if the conversation in the pr is right?

web_theme_classic/models/res_users_settings.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class ResUsersSettings(models.Model):
+    _inherit = "res.users.settings"
+
+    # These fields should be here in order to be accessible via in js
+    # as user.settings.persistent_classic_theme
+    persistent_classic_theme = fields.Boolean(default=True)
```
jappi00, line 12: nitpick: maybe add help for the note from the configure.md, something like: "When switching the theme all users need to reload the page to have a effect."

web_theme_classic/readme/CONFIGURE.md
jappi00: ofc: nitpick

jappi00 (approved):
Looks good. Two little nitpicks 💃

web_theme_classic/readme/CONFIGURE.md
ljmnoonan: I think it should stay "check" as writing "uncheck" would necessitate inverting the whole sentence. It's kind of odd describing a feature by what it doesn't do when it is turned off. The next line describes behavior when unchecked too.

web_theme_classic/models/res_users_settings.py
ljmnoonan: There is already a pretty verbose tooltip defined in [res_users.py](https://github.com/OCA/web/pull/3394/changes#diff-4177f5648058372482a5c2140f7cac6ec153785c725869671b2eedd6c60fbd6bR15), although it does not mention the need to reload. I'm hesitant to add this though as pressing "Save" from the `view_users_form_simple_modif` modal actually triggers a reload, so it is only in the context of `view_users_form` that this is relevant. Also, I think it is something that should be rather obvious to any user savvy enough to enable debug mode, which is necessary to see this option in the context of `view_users_form` in the first place.
If @legalsylvain thinks it is necessary too, I will add it

image

ljmnoonan:
@jappi00 Thank you for the review!

web_theme_classic/models/res_users_settings.py
jappi00: @ljmnoonan oh, clear!

ljmnoonan:
@legalsylvain this has been approved for a while. Would you mind merging it?

PR #3396, opened 2025-12-16
[17.0][ADD] bus_record_events: New Module(s)

zamberjo:
# 📝 TL;DR 

Real-time UI synchronization for Odoo 17
Say goodbye to manual F5 and the legacy patterns of web_refresher. This module introduces a clean, bus-based approach to keep views updated across different sessions automatically. By using a backend mixin and a dedicated OWL hook (useRecordStream), developers can now ensure users always see the latest data without the overhead of constant polling or manual reloads. A simple way to make Odoo feel more collaborative and reactive.

[Grabación de pantalla desde 2025-12-19 20-38-17.webm](https://github.com/user-attachments/assets/c2c53358-c9c4-4829-b3fe-d28ed6435c55)

----

This PR introduces three new modules designed to bring real-time reactivity to Odoo views via the Bus system:

`bus_record_events`: The core module. It provides a mixin (**bus.record.event.mixin**) that broadcasts CRUD operations (create, write, unlink) to the Odoo Bus. It also includes specialized JS classes for views (Form, Kanban, List, etc.) that listen to these events and update the interface automatically.
`bus_record_events_all`: An extension that automatically applies the mixin to all models in the system and injects the reactive JS classes into views globally.
`bus_record_events_demo`: Contains demo data and examples to showcase the functionality.

## Testing Instructions:

To verify the functionality using the demo module (`bus_record_events_demo`):

1. Ensure the `bus_record_events_demo` module is installed.
2. Open two different browsers (or use an incognito window) to simulate two concurrent sessions.
3. Log in with the provided demo users:
- User 1: Login: demo_bus_1 / Password: demo_bus_1
- User 2: Login: demo_bus_2 / Password: demo_bus_2
4. Navigate to the Bus Event Demo menu.
5. Interact with the records (Create, Edit, Delete) in one session and observe the real-time updates in the other session (e.g., open a Form view in one and modify it in the other, or watch the Kanban view update automatically).

## JavaScript API:

This module exposes tools for developers to build reactive OWL components easily:

`bus_record_event_service`: A new service that manages the subscription to record_events channels. It handles the underlying bus connection and dispatches notifications to subscribers.
`useRecordStream`: A custom OWL hook designed to simplify component integration. It automatically handles the lifecycle of the subscription (adding channels on `onWillStart`, unsubscribing on `onWillUnmount`) and provides callbacks for common actions:
- `onReload`: Triggered when a record changes (useful for refreshing views).
- `isDirty`: Checks if the current view has unsaved changes before acting.
- `onUpdate`: Allows custom handling of the event payload.

### Example Usage

```python
// So simple it feels native
useRecordStream("res.partner", {
    id: this.props.resId,
    onReload: async () => await this.model.load(),
});
```

## Design Decision regarding `bus_record_events_all`:

We considered including the global broadcasting functionality directly within the core `bus_record_events` module, controlled by an `ir.config_parameter` (defaulting to `False`). However, this approach would require checking the system parameter during every single CRUD operation across the entire system, which raised concerns about potential performance overhead.

To ensure the core module remains lightweight and performant, we decided to isolate this functionality into a separate module (`bus_record_events_all`). This approach avoids unnecessary checks in the CRUD methods and gives administrators explicit control over enabling this resource-intensive feature simply by installing or uninstalling the module.


ivs-cetmix:
Hey @zamberjo thank you for your contribution! Some notes:
- There should be no performance issues with queuing config param because the _get_param function is [cached](https://github.com/odoo/odoo/blob/50308d7d69af6848c2aa2cca8e5dbfff1491c983/odoo/addons/base/models/ir_config_parameter.py#L73).
- Have you seen this PR https://github.com/OCA/web/pull/3363 ? Isn't it something same or similar and if no, what's the main difference?

zamberjo:
Hi @ivs-cetmix thanks for the comment!

Just to clarify an important point: PR #3396 is not equivalent to #3363, even though both deal with UI updates from backend changes.

#3363 (web_refresh_from_backend) focuses on forcing view reloads from the backend and requires manual integration in the backend code where a refresh is needed.

#3396 (bus_record_events) introduces an event-driven approach based on the bus, providing mixins, services, and OWL hooks so the UI can react to record changes. It enables automatic view refresh out of the box and allows building more fine-grained, near real-time reactive views beyond full reloads.

zamberjo:
@ivs-cetmix hi! Just a friendly ping to check whether you’ve had a chance to review this PR. Any feedback would be very welcome. Thanks!

ivs-cetmix:
Hi @zamberjo , I did some check, however I'm not skilled enough on JS part to do a technical review. Let's ping @OCA/web-maintainers , probably they can have a look at it.
And please have a look at the codecove report, would be really good to increase the test coverage to have those checks green as well.

O-Mutt:
Le sigh, at face value this seems awesome but without some traction it is doa

zamberjo:
> Le sigh, at face value this seems awesome but without some traction it is doa

We’re gradually starting to roll this out in production. Happy to share feedback and real usage insights as it progresses.

maurochip (approved):

sersanchus (approved):

bus_record_events/models/ir_websocket.py
```diff
@@ -0,0 +1,42 @@
...
+            return False
+
+        model_name = parts[1]
+        if model_name not in self.env:
+            return False
+
+        if len(parts) == 2:  # record_events:model
+            # Check model read access
+            try:
+                self.env[model_name].check_access_rights("read")
```
etobella, line 29: This is quite dangerous. Imagine I am a portal user, I should be able to watch only my own sale orders, however, If I ask for record_events:sale.order, I will receive notifications of all sale orders!

zamberjo: First of all, I’m sorry it’s taken me so long to reply...

You're right! A portal user can subscribe to the model-level channel and receives `create`, `unlink` notifications for ALL sale orders — leaking record existence and IDs they shouldn't know about

I have just restricted channel subscriptions at the model level to internal users (base.group_user). Portal or public users can only subscribe to channels at the record level where the check_access_rule is applied.

PR #3397, opened 2025-12-16
[16.0][FIX] web_group_by_percentage: Added exclusion of non-field columns

Honeyxilia:
In some list/tree views, there are non-field columns, for instance buttons to handle an action on a group of lines. Consequently, with this module installed, it's impossible to group lines in certain views, if the grouped view has a button, as using isNumericColumn, which tries to access the column in the view's fields, throws an Exception.

As such, we exclude non-field columns from the computation of grouped columns percentages.


fcayre (approved):
Code review + functional test OK

petrus-v (approved):
code looks good to me

Hugo-Trentesaux (approved):

tarteo:
Makes sense, tests error is not related to this PR

/ocabot merge patch

PR #3405, opened 2026-01-09
DRAFT: Allow M2M items to appear multiple times in timeline

### LINKED ISSUE #3341: web_timeline: display an item multiple times when grouping by many2many
currently, when grouping by a many2many field, if it contains multiple values (for example: a task assigned to more than one person), the item will only be displayed in the first group.

this is a known limitation listed in the readme.

the goal of this issue is to discuss about how to fix this. could the author or maintainer give some information about the reasons that it is not yet done, what would be needed to do it and the possible issues?

i think we should consider this a bug, as the list and kanban views correctly display the item multiple times (because they make one `web_search_read` request per group).

satiowadahc:
Allows events to appear multiple times in web timelines. Open to Conversation here, as I modify the ID to be <id>_<group_id> to allow unique item ids to keep the library happy, then on events parse it back down to <id>.

Going to keep testing, opening PR for conversation.

I believe this would close #3341

image

### DIFF
--- a/web_timeline/static/src/views/timeline/timeline_controller.esm.js
+++ b/web_timeline/static/src/views/timeline/timeline_controller.esm.js
@@ -100,7 +100,6 @@ export class TimelineController extends Component {
      * @returns {Object}
      */
     _onUpdate(item) {
-
         const item_id =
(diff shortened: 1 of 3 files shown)


satiowadahc:
javascript is not my native language... will try to clean up the linting

PR #3407, opened 2026-01-12
[18.0][FIX] web_notify_channel_message

DavidJForgeFlow:
If other user than the admin is trying to send a message to a channel it will show an error message. Sending it as sudo solves the problem and makes sense that in this case you will always want to send the notification.


AaronHForgeFlow:
The code seems different from the comment

DavidJForgeFlow:
Features already in v18.0 hehe, my fault

PR #3410, opened 2026-01-13
[19.0][MIG] web_dark_mode: Migration to 19.0

ljmnoonan:
Odoo upped the contrast on a lot of things by making, for example, text `$gray-900` instead of `$gray-800`. I did not dial this down as I did not think it was worth it to fight the trend and introduce hidden inconsistencies. 

This PR, besides migration, also includes a fix to the device dependent dark mode setting. I feel that this is justified seeing that the new logic is a direct consequence of changes in 19.0, namely, that the decision on which asset bundle to load is made not by checking the `color_scheme` cookie, but rather by calling the `color_scheme()` method in `ir_http`. A full explanation of the new logic can be found in the last commit message.


ljmnoonan:
/ocabot migration web_dark_mode

I don't know if I can actually trigger this, but it's interesting to try :)

pedrobaeza:
You can become the maintainer of the module for triggering that bot commands.

/ocabot migration web_dark_mode

ljmnoonan:
@pedrobaeza, 
If you and and any other interested parties are ok with me having this role despite my low level of experience, I would be happy to be the maintainer for this module as it is something I will continue to work on and migrate. 
I was unable to find a PSC specifically for this repo, so who makes the decision on maintainers?

pedrobaeza:
The PSC would be @OCA/web-maintainers

ljmnoonan:
@pedrobaeza 
Ahh, I figured that there would be a PSC team separate from the maintainers team. thank you for pinging!

alexey-pelykh (approved):
Thanks for this migration, Liam -- solid work overall.

The `ir.http` rework is well thought out. Moving from a cookie-only check to the `color_scheme()` method override aligns properly with how 19.0 decides which asset bundle to load. Using `Sec-CH-Prefers-Color-Scheme` for server-side device-dependent detection (with the JS fallback for Firefox/Safari) is a clean approach and nicely documented in the commit message.

Test coverage is thorough -- 9 test cases covering portal users, cookie/preference combinations, device-dependent mode, and the Vary/Accept-CH headers. The SCSS work is carefully scoped with good comments explaining the rationale behind individual overrides.

One minor thought (non-blocking): in `_get_color_scheme`, when `user_device_dependant_scheme` is True and neither browser header nor cookie mismatch applies, the method returns `(None, existing_scheme_cookie)`. The caller `color_scheme()` then falls through to `super().color_scheme()`. This is correct behavior (lets the JS handle it), just wanted to confirm that was intentional -- and from the test_06 case and your commit message it clearly is.

LGTM.

ljmnoonan:
@alexey-pelykh  Hello, and thanks for the review. However, your review does appear to be AI generated and you have been asked already to stop this campaign of AI reviews. Please join the discussion in the contributors mailing list.

On the comment. `color_scheme()` only falls through to `super().color_scheme()` if the browser returns no hint and the `color_scheme` cookie does not exist. This is intended behavior. After loading, js will check the preferred color scheme and, trigger a reload if it is not "light". However, this only happens once, as now that the cookie is set to "dark", `color_scheme()` will always return "dark".

len-foss (approved):
LGTM :-)

ljmnoonan:
Thank you @len-foss and @pedrobaeza !

PR #3413, opened 2026-01-15
[18.0][FIX] web_widget_x2many_2d_matrix: fix total updates

RicardCForgeFlow:
Totals were only recomputed on initial render or after saving because the renderer cached matrix/rows/columns and only rebuilt them on props updates.

https://github.com/user-attachments/assets/1decd60a-33e2-4ec9-87c0-334726ac7566

Compute matrix data from reactive list.records on each render so row/column/grand totals update immediately when a cell value changes.

https://github.com/user-attachments/assets/a9d30d2e-39f9-4f61-bad5-2a4d2c68d290

### DIFF
--- a/web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm.js
+++ b/web_widget_x2many_2d_matrix/static/src/components/x2many_2d_matrix_renderer/x2many_2d_matrix_renderer.esm.js
@@ -1,19 +1,16 @@
-import {Component, onWillUpdateProps} from "@odoo/owl";
 import {evaluateBooleanExpr, evaluateExpr} from "@web/core/py_js/py";
+import {Component, onWillRender} from "@odoo/owl";
 import {Domain} from "@web/core/domain";
 import {Record} from "@web/model/relational_model/record";
 import {getFieldContext} from "@web/model/relational_model/utils";
 
 export class X2Many2DMatrixRenderer extends Component {
     setup() {
-        this.columns = this._getColumns();
-        this.rows = this._getRows();
-        this.matrix = this._getMatrix();
-
-        onWillUpdateProps((newProps) => {
-            this.columns = this._getColumns(newProps.list.records);
-            this.rows = this._getRows(newProps.list.records);
-            this.matrix = this._getMatrix(newProps.list.records);
+        onWillRender(() => {
+            const records = this.list.records;
+            this.columns = this._getColumns(records);
+            this.rows = this._getRows(records);
+            this.matrix = this._getMatrix(records);
         });
     }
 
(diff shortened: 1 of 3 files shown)


JoanSForgeFlow (approved):

JasminSForgeFlow (approved):
Code Review 👍

hbrunn (changes requested):
don't do this. The widget as is is already pretty expensive computationally, and this PR makes it much worse.

Please replace all of this with

```javascript
        onWillRender(() => {
            this.columns = this._getColumns();
            this.rows = this._getRows();
            this.matrix = this._getMatrix();
        });
```

in `setup()` instead of `onWillUpdateProps`, this should get the caching right and won't add several times o(n^2) complexity.

Also consider expanding the [test](https://github.com/OCA/web/blob/18.0/web_widget_x2many_2d_matrix/static/tests/web_widget_x2many_2d_matrix.test.js#L83) to edit a field several times and check if the totals are updated correctly

RicardCForgeFlow:
> don't do this. The widget as is is already pretty expensive computationally, and this PR makes it much worse.
> 
> Please replace all of this with
> 
> ```js
>         onWillRender(() => {
>             this.columns = this._getColumns();
>             this.rows = this._getRows();
>             this.matrix = this._getMatrix();
>         });
> ```
> 
> in `setup()` instead of `onWillUpdateProps`, this should get the caching right and won't add several times o(n^2) complexity.
> 
> Also consider expanding the [test](https://github.com/OCA/web/blob/18.0/web_widget_x2many_2d_matrix/static/tests/web_widget_x2many_2d_matrix.test.js#L83) to edit a field several times and check if the totals are updated correctly

Okay!

RicardCForgeFlow:
> don't do this. The widget as is is already pretty expensive computationally, and this PR makes it much worse.
> 
> Please replace all of this with
> 
> ```js
>         onWillRender(() => {
>             this.columns = this._getColumns();
>             this.rows = this._getRows();
>             this.matrix = this._getMatrix();
>         });
> ```
> 
> in `setup()` instead of `onWillUpdateProps`, this should get the caching right and won't add several times o(n^2) complexity.
> 
> Also consider expanding the [test](https://github.com/OCA/web/blob/18.0/web_widget_x2many_2d_matrix/static/tests/web_widget_x2many_2d_matrix.test.js#L83) to edit a field several times and check if the totals are updated correctly

Hi @hbrunn I changed the  `onWillUpdateProps` for your code. Also added a test with the goal of replicating the problem. Please tell me if you see everything correct. Thank you very much!

hbrunn (commented):
thanks. Do you agree we don't need the refactoring for passing rows, columns to `_getPointOfRecord` and `_getMatrix` with this?

RicardCForgeFlow:
> thanks. Do you agree we don't need the refactoring for passing rows, columns to `_getPointOfRecord` and `_getMatrix` with this?

Totally agree, much more simple. 

So the only change is replacing  '_onWillUpdateProps_' for '_onWillRender_'.

hbrunn (approved):
thanks

/ocabot merge patch

PR #3418, opened 2026-01-23
[18.0][IMP] web_widget_x2many_2d_matrix: sort rows alphabetically

RicardCForgeFlow:


hbrunn:
I don't think it's a good idea to try to add this client side. hr_timesheet_sheet just needs to impose an order on the underlying model, this is much more flexible and less code

RicardCForgeFlow:
> I don't think it's a good idea to try to add this client side. hr_timesheet_sheet just needs to impose an order on the underlying model, this is much more flexible and less code

Okay! Sorry :sweat_smile:

PR #3421, opened 2026-01-29
[MIG] web_widget_char_size: Migration to 18.0

SirPyTech:
Migration from `16.0` (https://github.com/OCA/web/tree/d27d8cc50de0c117b84dbeb4aa0da1575938bb41/web_widget_char_size).

The migration includes a fix to the contributors file name, backported to affected versions with https://github.com/OCA/web/pull/3422 and https://github.com/OCA/web/pull/3423.


web_widget_char_size/i18n/it.po
```diff
@@ -0,0 +1,14 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0\n"
```
quirino95, line 6: **Suggestion (non-blocking)**: should this be "Odoo Server 18.0"?

web_widget_char_size/i18n/web_widget_char_size.pot
```diff
@@ -0,0 +1,13 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0\n"
```
quirino95, line 6: Same here

quirino95 (approved):
Code and functional review: LGTM!

web_widget_char_size/i18n/it.po
sergiocorato: Translation are not managed in migration PR.

web_widget_char_size/static/src/js/char_widget.esm.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+// Let the CharField component know that there is a new prop
+CharField.props = {
+    ...CharField.props,
+    maxLength: {type: Number, optional: true},
+};
+
+// Use the new prop in CharField component's maxLength getter
+const _original_getter = Object.getOwnPropertyDescriptor(
```
StefanRijnhart, line 18: This seems very low level. Can't you just override the getter as is done for example in https://github.com/OCA/web/blob/18.0/web_calendar_slot_duration/static/src/js/calendar_common_renderer.esm.js#L7-L9?

web_widget_char_size/i18n/it.po
sergiocorato: As discussed in other chat, this is not mentioned in wiki but should be done in migration, as it is not changed from normal translation process.

web_widget_char_size/static/src/js/char_widget.esm.js
```diff
@@ -1,10 +1,26 @@
...
+// Extract the new prop from field options
+const _extractProps = charField.extractProps;
+charField.extractProps = ({attrs, options}) => {
+    return Object.assign(_extractProps({attrs, options}), {
+        maxLength: options.size,
     });
 };
+
+// Let the CharField component know that there is a new prop
+CharField.props = {
```
anusriNPS, line 12: **Suggestion(non-blocking):** As this getter and prop addition is added as a new feature, can it  be moved to seperate commit as they are not part of migration?

anusriNPS (approved):
Code and Functional Review: LGTM

web_widget_char_size/static/src/js/char_widget.esm.js
StefanRijnhart: If you consider this change as not a part of the migration, then you are sensing a way to migrate the module without overriding the getter. Can you explain how you would do that?

StefanRijnhart (changes requested):

StefanRijnhart:
Just updating my review status here as per my earlier comment about overriding the getter on a higher level.

web_widget_char_size/static/src/js/char_widget.esm.js
anusriNPS: Apologies for wrong suggestion. I misunderstood `getter` as a additional enhancement apart from introducing `size` option while comparing to `16.0` and `18.0` differences. I agree `getter` would require for completing this functionality as part of migration itself as way of handling introducing `size` option is different between `16.0` and `18.0`. Hence, the confusion. So, removing the suggestion

web_widget_char_size/static/src/js/char_widget.esm.js
SirPyTech: Hi,
I'm sorry I still haven't managed to look into this but I'm definitely willing to.
I remember I hadn't found a better way to do this, but I'm not that strong as a frontend developer so I'm curious to look into your suggestion.

Thanks for having a look!

SirPyTech: I implemented your suggestion and seems to work great!
Please update your review

StefanRijnhart (approved):
Thanks!

StefanRijnhart:
Three approvals and other comments also resolved, so

/ocabot merge nobump

sergiocorato:
@StefanRijnhart could you check the failed merge?

StefanRijnhart:
@SirPyTech thanks for the rebase!

/ocabot merge nobump

SirPyTech:
Thank you for merging! This PR was ~300 commits behind, so a rebase was due.

And also grazie a @sergiocorato for reminding me of this one!

PR #3424, opened 2026-01-29
[18.0][FIX] web_timeline: Render the SVG correctly to draw dependencies properly

### LINKED ISSUE #3420: [18.0] web_timeline : Dependency arrows are not shown anymore
Since [this commit](https://github.com/OCA/project/commit/b38074cb6dc921d40010042ed8c388d9690d36a4) in v15, we have the arrows displayed for dependant tasks in projects.

In v18, the arrows disappeared...

Steps to Reproduce : 
on runboat v18 (project repo) got to All Tasks and select the timeline view.
The tasks of Research and Development Project have some dependant tasks but the arrows are not visible.
on runboat v15-v17, it's ok

carlos-lopez-tecnativa:
This issue was introduced during the migration to v18 when converting old jQuery code to native JavaScript in this commit: https://github.com/OCA/web/commit/258a5a92dc85fa5cc6dd9181ca444e3b4b290d0f

Fixes https://github.com/OCA/web/issues/3420

@Tecnativa @CarlosRoca13 @pedrobaeza @cvinh could you please review this?

image

### DIFF
--- a/web_timeline/static/src/views/timeline/timeline_canvas.esm.js
+++ b/web_timeline/static/src/views/timeline/timeline_canvas.esm.js
@@ -10,20 +10,14 @@ export class TimelineCanvas {
     constructor(canvas_ref) {
         this.canvas_ref = canvas_ref;
     }
+
     /**
      * Clears all drawings (svg elements) from the canvas.
      */
     clear() {
-        if (this.canvas_ref) {
-            const tempElement = document.createElement("div");
-            tempElement.innerHTML = this.canvas_ref;
-            Array.from(tempElement.children).forEach((child) => {
-                if (child.tagName.toLowerCase() !== "defs") {
-                    child.remove();
-                }
-            });
-            this.canvas_ref = tempElement.innerHTML;
-        }
+        Array.from(this.canvas_ref.children)
+            .filter((el) => el.tagName.toLowerCase() !== "defs")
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Tested that works in runboat.

One side question, as you mention SVG. Is there a way to export the full SVG to a file?

carlos-lopez-tecnativa:
> Tested that works in runboat.
> 
> One side question, as you mention SVG. Is there a way to export the full SVG to a file?

Sorry, but I do not understand what you mean. This SVG is a template, and on top of it the arrows are drawn (it is a workaround). Even if we add this SVG as a file (as I understand you are referring to), I do not see what we would gain from that. We need to render the SVG dynamically.

pedrobaeza:
No, sorry, I'm talking for having an extra feature, which is to save all the "kanban" as an SVG, for inserting the timeline in a document, etc, like other planning softwares allow.

carlos-lopez-tecnativa:
> No, sorry, I'm talking for having an extra feature, which is to save all the "kanban" as an SVG, for inserting the timeline in a document, etc, like other planning softwares allow.

Ah, OK. I do not know if it is possible to export the timeline; I have not investigated it. If you need it, I can look into it, but could you please open a new task or issue for this?

cvinh (approved):
Tested on my environment

PR #3427, opened 2026-02-02
[18.0][IMP] web_company_color: Compile SCSS

SirPyTech:
Otherwise function calls are not evaluated in the created attachment.

Right now it is not needed because the created attachment does not use any functions, but if any were added, that would not be evaluated.

For instance, try editing the created CSS by adding a call to the `desaturate` function like:
```scss
.app-menu-container {
    background: url('/web_responsive/static/src/img/home-menu-bg-overlay.svg'),
        linear-gradient(
            to bottom,
            %(color_navbar_bg)s,
            desaturate(lighten(%(color_navbar_bg)s, 20%%), 15)
        );
}
```
that would be sent to the client as-is resulting in something like:
```scss
.app-menu-container {
    background: url('/web_responsive/static/src/img/home-menu-bg-overlay.svg'),
        linear-gradient(
            to bottom,
            #123456,
            desaturate(lighten(#123456, 20%), 15)
        );
}
```
that is not acceptable CSS, and thus is not evaluated.

Other SCSS assets are automatically compiled in their bundle, but this one skips that procedure because of the overrides for the `web_company_color.company_color_assets` bundle.


quirino95 (approved):
Code and functional review LGTM

CarlosRoca13 (approved):

PR #3310, opened 2025-10-08
[19.0][MIG] web_widget_x2many_2d_matrix: Migration to 19.0

bizzappdev:

PR #3438, opened 2026-02-17
[19.0][MIG] web_widget_x2many_2d_matrix: Migration to 19.0

ferran-S73:
Migration from https://github.com/OCA/web/pull/3310


alexey-pelykh (changes requested):
Thanks for picking up this migration, @ferran-S73!

I maintain this module, so I went through the changes carefully. The many2one tuple-to-object migration (`value[0]` → `value.id`, `value[1]` → `value.display_name`) and the `users` → `user_ids` rename in the demo view are both correct for 19.0. Manifest version bump looks good too.

However, I have one concern that I believe needs to be addressed before merging:

### Renderer setup: `onWillUpdateProps` vs `onWillRender`

The 18.0 version uses `onWillRender` to recalculate `columns`, `rows`, and `matrix` on **every** render cycle. The 19.0 PR replaces this with `setup()` + `onWillUpdateProps`. The problem is that `onWillUpdateProps` only fires when the **parent passes new props** — it does not fire when the user edits a value cell inline. Inline edits mutate the records within `this.props.list` without the parent re-passing props to the renderer, so the matrix/totals would not update.

The 18.0 branch includes a test that validates exactly this scenario ("matrix totals are updated correctly after several edits"), which was dropped in this PR. I suspect the test was removed because it fails with the current `onWillUpdateProps` approach.

**Suggested fix:** keep using `onWillRender` as in 18.0 — it's still available in OWL 3 and correctly handles all reactivity scenarios:

```js
setup() {
    onWillRender(() => {
        const records = this.list.records;
        this.columns = this._getColumns(records);
        this.rows = this._getRows(records);
        this.matrix = this._getMatrix(records);
    });
}
```

And please restore the dropped test from 18.0 as well:

```js
test("matrix totals are updated correctly after several edits", async () => { ... });
```

(along with the `contains` and `queryAll` imports it needs).

Everything else looks clean — nice work on the migration overall.

pedrobaeza:
Superseded by #3578

PR #3444, opened 2026-02-20
[18.0][ADD] web_toggle_chatter

LixFerox:
Adds a toggle button in backend form views to show/hide the chatter panel, giving users more screen space.
image


arch-fan:
Maybe should be renamed to `web_toggle_chatter`?

arch-fan:
It looks like this 
output

arch-fan:
Well, I can see a responsive issue. Responsive breaks here 
image

flotho (commented):
Hi,
Thanks for this contribution.
IMHO this module duplicates a behaviour already available with the https://odoo-community.org/shop/chatter-position-10837?search=chatter+position#attr=942727 module.
especially in v16 

Image

LixFerox:
image

That option is no longer available in Odoo 18.

pedrobaeza:
This is not duplicated IMO, as they serve different purposes. The old one places the chatter on the bottom, maybe including that extra feature, but this is exclusively for collapsing/expanding the chatter without moving it to bottom.

flotho:
> This is not duplicated IMO, as they serve different purposes. The old one places the chatter on the bottom, maybe including that extra feature, but this is exclusively for collapsing/expanding the chatter without moving it to bottom.

Hummm. I understand.
Maybe we could be more consistent and consider that chatter position could also deal with the module chatter_position? Instead of having multiple modules to manager chatter behavior ?

pedrobaeza:
I don't want to install `web_chatter_position` to have this feature. I consider that one not valid, as it's favoring a bad practice. This one is saner, simply letting to toggle the chatter when you require more space.

arch-fan:
also that `web_chatter_position` doesn't make reference to a collapsible chatter, it's difficult to find the functionality, when sometimes you just want to have more space on screen while you edit and invoice or whatever operation

flotho:
> I don't want to install `web_chatter_position` to have this feature. I consider that one not valid, as it's favoring a bad practice. This one is saner, simply letting to toggle the chatter when you require more space.

Thanks for your answer. Maybe I'm not understanding properly why you're considering that moving chatter to the bottom is bad practice. Could you elaborate ?
Regards

pedrobaeza:
The chatter is a fundamental point of information that gathers all the record activity, communications, and other tools like attachment view, activities, etc. Moving it to the bottom, decreases drastically its usefulness, as well as makes users to not be aware of certain important elements like pending activities or latest communications, only being available doing a lot of scroll.

That's why I don't like that module.

About this one, what I expect is that punctually, you can collapse it for doing an operation, but returning to be unfolded on each record visited (I haven't tested yet to know if it works this way), for not losing the goal of it.

arch-fan:
Yep, it works like you spect.

We have a case where employees work with long lines, and chatter makes smaller the main form, so they want to "hide" it temporally. Not as an option, just as a single click temporarily

pedrobaeza:
In general, the design principles for one2many fields is to minimize the number of columns they have. A lot of columns in a screen, so that it requires to have the chatter in the bottom, means a bad UX design IMO.

An exception is to have specific cases if the problem is not to have a lot of columns, but some long texts, which makes this tool acceptable.

flotho:
> The chatter is a fundamental point of information that gathers all the record activity, communications, and other tools like attachment view, activities, etc. Moving it to the bottom, decreases drastically its usefulness, as well as makes users to not be aware of certain important elements like pending activities or latest communications, only being available doing a lot of scroll.
> 
> That's why I don't like that module.
> 
> About this one, what I expect is that punctually, you can collapse it for doing an operation, but returning to be unfolded on each record visited (I haven't tested yet to know if it works this way), for not losing the goal of it.

Thanks for exposing your PoV, really interesting.
That will change some of my deployments strategies.

Regards

flotho (approved):
tested on runboat, do whats expected.
LGTM
Regards

arch-fan:
Well, all discovered bugs have been addressed. We think PR is ready for merge. Actually using in prod.

simahawk (approved):
Nice addon, thanks!
Tested on Runboat.

One question: any plan to make it persistent?
Eg: via cookie or via user settings?

arch-fan:
Well, for persistency `web_chatter_position` covers this case with hidding. We agree @pedrobaeza, chatter is fundamental so it shouldn't be hide, or at least by default.

In the case the feature is requested, I suggest adding a config parameter which maintains persistence between menus using a cookie.

web_toggle_chatter/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Vortex Dimensión Digital
```
pedrobaeza: You have to put here your name as well, Jorge.

pedrobaeza (commented):
Tested on runboat (uninstalling `web_chatter_position`), and works perfectly. Can you please squash together the commit history?

arch-fan:
Done, squashed, added contributors & added a gif to readme hehe

pedrobaeza (approved):
/ocabot merge nobump

PR #3446, opened 2026-02-21
[IMP] web_timeline: export timeline as SVG

cyrilfassiau:
## [IMP] web_timeline: export timeline as SVG

This PR adds an **Export SVG** feature to `web_timeline`, so users can save the currently displayed timeline as an SVG for documentation/reporting purposes (as requested in the related issue discussion).

### What’s inside
- Add an **Export SVG** button in the timeline renderer toolbar.
- Small toolbar layout tweak: wrap the existing buttons in an extra `<div>` and use flex layout so the new export button is visually separated and easier to find.
- Export implementation:
  - Capture the rendered timeline DOM (`vis-timeline`)
  - Embed it into an SVG using a `foreignObject`
  - Inline computed styles to keep the exported SVG visually consistent outside Odoo

### Notes / limitations
- The exported SVG relies on `foreignObject` (compatibility may vary depending on the SVG viewer/editor).
- Tested on a local Odoo instance in modern browsers.

### Feedback request
Does this approach match the expected feature described in the issue?  
I’m happy to adjust the implementation if another export strategy is preferred.


web_timeline/static/src/views/timeline/timeline_renderer.esm.js
```diff
@@ -19,8 +19,62 @@ import {useService} from "@web/core/utils/hooks";
...
+        const cs = window.getComputedStyle(src);
+        const style = props.map((p) => `${p}:${cs.getPropertyValue(p)};`).join("");
+        dst.setAttribute("style", style);
+    }
+}
+
 export class TimelineRenderer extends Component {
     setup() {
+        super.setup?.();
+        this.timelineRef = useRef("timelineRoot");
```
carlos-lopez-tecnativa, line 76: The `timelineRef` variable is unused, so it is not necessary; please remove it.

carlos-lopez-tecnativa (changes requested):
I’m not sure if this is due to an incorrect HTML or SVG structure, but I tested it locally with the `project_timeline` module. When I export the SVG from `All Tasks` and open it, I get this error.

Image

Image

Additionally, please squash the commits.

PR #3452, opened 2026-02-25
[18.0][ADD] web_chatter_position_toggle: add toggle button for chatter position

JoanSForgeFlow:
After installing the module, a toggle button (↔) will appear in the control panel of form views that include a chatter.

Click the button to switch the chatter between the side and bottom positions.

The toggle applies per form view and resets when navigating away.


LauraCForgeFlow (approved):
Functional review, works like a charm!

DavidJForgeFlow (changes requested):
When the chatter position is at bottom, the button does not work:
image

I think the main issue here is that you are not really using the base module and implementing this button on top using the previous logic. You are creating this new logic apart from the web_chatter_position logic. Maybe you should use the base module and implement the button from that logic.

Maybe it's better to FP this PR from 16.0: https://github.com/OCA/web/pull/3040

PR #3453, opened 2026-02-25
[18.0][IMP] web_chatter_position: add toggle button for chatter position

JoanSForgeFlow:


LauraCForgeFlow (approved):

web_chatter_position/__manifest__.py
```diff
@@ -1,11 +1,12 @@
 # Copyright 2022 Hynsys Technologies
 # Copyright 2024 Alitec Pte Ltd
+# Copyright 2026 ForgeFlow S.L. (https://www.forgeflow.com)
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 {
     "name": "Chatter Position",
     "summary": "Add an option to change the chatter position",
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.1.0",
```
AaronHForgeFlow: Shouldn't the bot do this?

AaronHForgeFlow (commented):
It works well, awesome feature. Nitpicking alert: I would add the button on the right part of the screen, having it next to the "New" button feels rare.

JoanSForgeFlow:
Hi @AaronHForgeFlow,

The button was there in previous versions, 15.0 and earlier.

AaronHForgeFlow (approved):
Ok! Thanks for the clarification! LGTM.

arch-fan:
This is now implemented at https://github.com/OCA/web/pull/3444

alexey-pelykh (approved):
Thanks for this, Joan -- a toggle button is a very welcome addition.

Code looks solid overall. A few observations:

1. **Compiler vs runtime interaction**: The existing module positions the chatter at compile time via `FormCompiler.compile()`. This PR adds runtime repositioning via DOM manipulation in `FormController`. When the user preference is `"auto"` (compiler does nothing special), the toggle still appears and switches to `"bottom"` on first click, which works because `_moveChatter` handles the class/DOM changes. Worth noting this dual approach in case of future refactors to the compiler patch -- the two mechanisms need to stay in sync.

2. **Toggle state is ephemeral**: The toggled position lives only in component state and resets on navigation. If that's intentional (per-session override), maybe worth a brief note in the README. If persistence is desired eventually, an RPC to update the user preference would be the way to go.

3. **Minor**: `moveChatter()` is called from both `onMounted` and `onPatched`, which is correct to handle initial render and subsequent OWL re-renders. The resize listener for `"bottom"` is a nice touch for responsive behavior.

LGTM -- nice feature.

JoanSForgeFlow:
Hi @arch-fan,

From my point of view, this PR is not duplicating `web_toggle_chatter`.

The goal here is to restore the possibility of moving the chatter to the bottom, which existed in previous versions of `web_chatter_position` but was lost during the migrations. This PR simply reintroduces that functionality so it is not lost.

Also, the behavior is different: `web_toggle_chatter` temporarily collapses the chatter, while this module allows moving the chatter between the side and the bottom. Because of that, both features can actually be complementary.

If there are concerns about the visibility of the button or the UX, we could also make the button optional, for example by showing it only when enabled through a configuration option

Thank you

arch-fan:
Ahh just for moving the chatter ATM. I thought it was just for collapsing. Ok sorry, great PR 🙏 GJ

JoanSForgeFlow:
Hi @etobella 

Could you take a look when you get a chance?

It would be great if you could merge it if everything looks good to you.

Thank you!

JoanSForgeFlow:
Hi @etobella,  friendly ping 🙂
  Would you have a moment to review/merge it? Thanks!

tarteo (approved):
Image

Tested on runboat, I would have put the button somewhere else, but that is up for debate.

PR #3455, opened 2026-02-26
[18.0] [IMP] web_timeline: Add multi group_by support

paradoxxxzero:
Adds to the web_timeline the ability  to handle multiple group_by with `nestedGroups`:

image

This changes the group record view action from single click to double click to avoid conflict with group expand/collapse.


pedrobaeza:
Isn't this the same as #3006 ?

paradoxxxzero:
It is similar and I missed this PR, this is a forward port of an unpublished fork I did on the 14.0 version.

A forward port of #3006 should be better, closing for now.

Sorry for the noise.

pedrobaeza:
Thanks. Can you please do the fw-port of it?

paradoxxxzero:
I will if I find the time

PR #3458, opened 2026-02-26
[FIX] web_dark_mode: always use dark mode on new tabs

len-foss:
Suppose you have your Odoo running at http://localhost:8069 (or https://domain.tld, doesn't matter)
Set your scheme to dark mode. Odoo switches to dark mode.
Open a new tab http://localhost:8069, it uses the light theme. Refresh the tab, it now uses the dark mode.
If you open your tab at http://localhost:8069/odoo (or any non-empty backend path) it would properly use dark mode. However, browser auto-complete defaults to the root address.

### DIFF
--- a/web_dark_mode/models/ir_http.py
+++ b/web_dark_mode/models/ir_http.py
@@ -9,15 +9,13 @@ class IrHttp(models.AbstractModel):
     _inherit = "ir.http"
 
     @classmethod
-    def _set_color_scheme(cls, response):
-        scheme = request.httprequest.cookies.get("color_scheme")
+    def _sanitize_cookies(cls, cookies):
+        """Set the color_scheme cookie before template rendering."""
+        super()._sanitize_cookies(cookies)
         user = request.env.user
+        if getattr(user, "dark_mode_device_dependent", None):
+            return
         user_scheme = "dark" if getattr(user, "dark_mode", None) else "light"
-        device_dependent = getattr(user, "dark_mode_device_dependent", None)
-        if (not device_dependent) and scheme != user_scheme:
-            response.set_cookie("color_scheme", user_scheme)
-
-    @classmethod
-    def _post_dispatch(cls, response):
-        cls._set_color_scheme(response)
-        return super()._post_dispatch(response)
+        if cookies.get("color_scheme") != user_scheme:
+            cookies["color_scheme"] = user_scheme
+            request.future_response.set_cookie("color_scheme", user_scheme)


len-foss:
@ljmnoonan could you confirm this? I suspect it also affects https://github.com/OCA/web/pull/3410 ?

As a side-note, the tests fail but they are testing how the code was implemented, not whether it works or not.

ljmnoonan:
@len-foss I'll take a look later, but just to verify, are you using the dark mode toggle or the device dependant dark mode? If it is the latter, then I know it is broken and has been that way since the  beginning. Basically, js was only checking devices preference if the color_scheme cookie didn't exist. I fixed in https://github.com/OCA/web/commit/5ef5802c7c2d41660e58057def531a7ea65f4416

len-foss:
@ljmnoonan no, I'm talking about the regular dark mode toggle. The bug is that in the redirect dispatch the template is rendered before the color scheme is applied. If my fix is indeed the proper solution to the bug, you would probably have to rewrite just a few lines of your fix.

ljmnoonan:
@len-foss , ok, I got some time to test it out and I see what the problem is. This was fixed in odoo 19.0 by https://github.com/odoo/odoo/commit/2120e4c6539b0ad8fadaf249c883b0529ad613f4 which 
> ... sets the `color_scheme` in the QWeb rendering context
instead of directly access the cookie

so it does not affect #3410. 

Your fix does indeed work, but the choice here is do we want to use your solution to patch the old logic, or effectively backport the odoo changes into this model (which would involve inheriting a very low priority view)? If you think the latter option is better I would also be able to backport https://github.com/OCA/web/pull/3410/commits/993e2d3b59e78fdd5b2b0889ba95b4e7688296d7 which fixes device dependent dark mode and implements server side asset bundling based on the Sec-CH-Prefers-Color-Scheme hint from chromium browsers.
I also realized that the reason I have never noticed this is that I normally have `website` installed, which prevents root domain from redirecting to a backend path, so this bug is limited to installations that are using the backend only.

len-foss:
@ljmnoonan sorry about the delay, and thank you for your reply. The backport is probably the best way? If only for ease of maintenance. I'm not sure of whether it requires much adaptations or not. Can you prepare a branch? If there's some problem to fix with it I can fix it, test it on my end and give a review.

ljmnoonan:
@len-foss sorry about the delay from me as well! When I get a chance I will create a backport and tag you in the PR

PR #3462, opened 2026-03-02
[18.0] [FIX] web_timeline: Migrate check_access_rights to has_access following deprecation

paradoxxxzero:
`check_access_rights` is deprecated in 18.0, replace it with `has_access`


alexey-pelykh (approved):
Thanks for this fix, Florian.

Verified that `check_access_rights` is indeed deprecated since Odoo 18.0 (emits `DeprecationWarning`) and that `has_access` is the correct boolean replacement.

The call signature change from `check_access_rights([operation, false])` to `has_access([[], operation])` correctly matches the new API — empty recordset + operation string. All three call sites (write, unlink, create) are consistently updated, and the test mock follows.

CI is green, LGTM.

PR #3464, opened 2026-03-03
[19.0][MIG] web_tree_many2one_clickable: Migration to 19.0

namtnt:
**Limitation:**
The web_tree_many2one_clickable module injects the button by patching a single base template `web.ListRenderer.RecordRow`. However, Odoo 19's Account, Sale, and Purchase modules use specialized list view renderers (`account.SectionAndNoteListRenderer.RecordRow`, `sale.ListRenderer.RecordRow`) for their order line fields, which are compiled separately from the base. So the button will be missing from the list view of these form views.


alexey-pelykh (approved):
Thanks for the migration!

The code is identical to 18.0 and all CI checks pass -- clean port. The `patch(ListRenderer.prototype, ...)` pattern, `useService` hook, and the `web.ListRenderer.RecordRow` template extension all work correctly in 19.0.

One minor thing (pre-existing, not introduced by this PR): the `doAction` call uses `target: "target"`, which is not a recognized `ir.actions.act_window` target value (`current`, `new`, `inline`, `fullscreen`, `main`). It works because Odoo silently falls back to the default, but it would be cleaner as `target: "current"`. Not blocking -- can be addressed separately if desired.

LGTM.

cparadis-impressfoods (approved):

pedrobaeza (approved):
Tested on runboat

/ocabot merge nobump

cuongnmtm:
We should have web_tree_many2one_clickable_sale and web_tree_many2one_clickable_account to handle the limitation mentioned in the description, with auto_install=True

@pedrobaeza what do you think?

pedrobaeza:
This is due to the special widget, that is the same for both, so maybe only one glue module is needed. Anyway, that widget makes the product clickable, so not sure it's needed at all.

cuongnmtm:
> This is due to the special widget, that is the same for both, so maybe only one glue module is needed. Anyway, that widget makes the product clickable, so not sure it's needed at all.

The product field is clickable, but the other Many2one fields are not. But let's wait to see if anyone raises any concerns about this.

PR #3465, opened 2026-03-03
[15.0][FIX] web_pivot_computed_measure: Error with pivots added to dashboard with computed measures

CarlosRoca13:
Before these changes, when a pivot view that included computed measures was added to a dashboard, an error was thrown when trying to access that dashboard.

The code related to the legacy pivot view model has been restored so that it loads correctly. Specifically, the necessary part of the code that had been removed during the migration has been reinstated to prevent the view from failing to load.

cc @Tecnativa TT61301


carlosdauden (approved):
Functional review

PR #3469, opened 2026-03-09
[17.0][MIG] web refresh from backend

ivs-cetmix:
Standard migration from 16.0.
Added some improvements in a separate commit - let me know if needed to be squashed 😉


ivs-cetmix:
Dear @etobella and anyone who knows OWL really well (eg @binhexsystems ) - would appreciate your review so we could proceed further with the migration to 18.0 😄

PR #3484, opened 2026-03-17
[18.0][IMP] web_company_color: allow to set bottom border color of navbar

hbrunn:
Odoo adds a 1px border below the nav bar, which is not visible below navbar buttons. Usually it's hard to notice because the colors are darkish, but when using web_company_color to set a bright color, it looks like this on current runbot:

image

but with this patch and setting the border color, it looks like this:

image

I took the liberty to simplify the code I had to touch anyways

### DIFF
--- a/web_company_color/models/res_company.py
+++ b/web_company_color/models/res_company.py
@@ -165,22 +165,15 @@ class ResCompany(models.Model):
         return super().unlink()
 
     def write(self, values):
-        if not self.env.context.get("ignore_company_color", False):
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Thanks for this. I think the first commit should be `[REF]`, not `[FIX]`.

hbrunn:
@pedrobaeza thanks for the headsup, fixed

pedrobaeza (approved):
/ocabot merge patch

PR #3487, opened 2026-03-26
[18.0][FIX] web_m2x_options_manager: fix create edit options

siemenv:
Odoo 18 renamed the field-level option keys and inverted their boolean sense:
- old: create=False  → new: no_quick_create=True
- old: create_edit=False → new: no_create_edit=True


omar7r (approved):
This correction is necessary; otherwise, the module won't function as designed.
Thanks LGTM

siemenv:
Can this PR be merged? Otherwise the module won't work.

PR #3488, opened 2026-03-26
[FIX] web_company_color: colors not changed

deeniiz:
Migration of this PR #3196


deeniiz:
@eantones, can you please review it?

PR #3495, opened 2026-03-27
[18.0][IMP] web_menu_reorder

mathben:
Reorder top-level menus alphabetically or by custom label list or by default (sequence)


web_menu_reorder/models/res_config_settings.py
```diff
@@ -0,0 +1,125 @@
...
+            .get_param("menu_reorder_settings.menu_order_labels", default="")
+        )
+        return res
+
+    def set_values(self):
+        super().set_values()
+        self.env["ir.config_parameter"].sudo().set_param(
+            "menu_reorder_settings.menu_order_labels",
+            self.menu_order_labels or "",
+        )
```
legalsylvain, line 67: you could remove this section, and just use the nativ configuration parameter ? don't you think ?

legalsylvain (changes requested):
thanks for your contribution.

PR #3501, opened 2026-04-01
[18.0][FIX] web_pwa_customize: adjust PWA manifest and browser theme-color for Odoo 18

javierjosecarpio:
Upgraded the manifest override to use the Odoo 18 _get_webmanifest() extension point. Added support for dynamic meta tag theme-color using System Parameters. Ensured manifest values are correctly pulled from ir.config_parameter.


web_pwa_customize/__manifest__.py
```diff
@@ -4,11 +4,14 @@
     "name": "Web Pwa Customize",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/web",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
```
victoralmau:
```suggestion
```
This change is not necessary.

web_pwa_customize/views/pwa_assets.xml
```diff
@@ -0,0 +1,32 @@
...
+<odoo>
+    <template
+        id="webclient_bootstrap_pwa"
+        inherit_id="web.webclient_bootstrap"
+        priority="99"
+    >
+        <xpath expr="//meta[@name='theme-color']" position="attributes">
+            <attribute
+                name="t-att-content"
+            >request.env['ir.config_parameter'].sudo().get_param('pwa.manifest.theme_color', '#71639e')</attribute>
```
victoralmau, line 11: IMO, the best approach would be to use the https://github.com/odoo/odoo/blob/512cacba2398832da7e8a9069df1441fe96b8fab/addons/web/models/ir_http.py#L73 method (I think that's the right one) to define whatever we need there (manifest_theme_color, manifest_short_name, etc.) and use it in the view.

web_pwa_customize/views/pwa_assets.xml
```diff
@@ -0,0 +1,32 @@
...
+        </xpath>
+        <xpath expr="//meta[@name='theme-color']" position="after">
+            <meta name="apple-mobile-web-app-capable" content="yes" />
+            <meta name="apple-mobile-web-app-status-bar-style" content="default" />
+            <meta
+                name="apple-mobile-web-app-title"
+                t-att-content="request.env['ir.config_parameter'].sudo().get_param('pwa.manifest.short_name', 'Odoo')"
+            />
+        </xpath>
+        <xpath expr="//link[@rel='apple-touch-icon']" position="replace">
```
victoralmau, line 21:
```suggestion
        <xpath expr="//link[@rel='apple-touch-icon']" position="attributes">
```
We should avoid using `replace`, and I think we can do that by using only `position=attributes` and whatever is defined in `webclient_rendering_context()`.

victoralmau (changes requested):

jans23:
@javierjosecarpio are you going to address Victor's comments and move this PR further?

PR #3507, opened 2026-04-13
[18.0][FIX] checklog-odoo.cfg: ignore Killing chrome descendants-or-self warning

StefanRijnhart:
Warning introduced in https://github.com/odoo/odoo/commit/d22e0b4e181de0940810b0f0a760bcfe288042a7. The previous warning apparently did not occur, as it was not featured in checklog-odoo.cfg.

```
2026-04-13 03:20:41,702 367 WARNING odoo odoo.addons.web_disable_export_group.tests.test_tour.TestTour.test_admin: Killing chrome descendants-or-self of 382: 6 remaining
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
```

### DIFF
--- a/checklog-odoo.cfg
+++ b/checklog-odoo.cfg
@@ -1,4 +1,5 @@
 [checklog-odoo]
 ignore=
     WARNING.* 0 failed, 0 error\(s\).*
+    WARNING .* Killing chrome descendants-or-self .*
     Missing widget: res_partner_many2one for field of type many2one


pedrobaeza:
Shouldn't this be added to the copier template, as it can happen in other repos?

pedrobaeza (approved):

StefanRijnhart:
@pedrobaeza There aren't that many repositories with javascript tests, so the change is only required on a small number of repos. Also, it sounds like a proper solution is still envisioned upstream (https://github.com/odoo/odoo/pull/258062) so hopefully this is a temporary requirement.

pedrobaeza:
OK, I understand this will happen if you have JS tours, so that's more frequent than thought initially. Anyway, if it's a temporary one, let's see if this scales.

pedrobaeza:
We have another case here: https://github.com/OCA/rma/pull/575

pedrobaeza:
And another in OCA/mail: https://github.com/OCA/social/pull/1851

etobella:
I will add it to the copier template. With Hoot, a lot of extra tests with OWL are appearing (Sign, account reconcile, automation, web...). We should be prepared.

PR #3301, opened 2025-10-02
[FP][17.0][IMP] web_timeline - support multiple group-by levels and allow to group by m2m field

cuongnmtm:
Forward-port of #3006

Test cases:

- [x]  group by one field (group by m2o, selection, integer, m2m, date, ...)
- [x]  group by multiple fields
- [x]  group by 0 field: back to default group by
- [x]  create a new record
- [x]  drag and drop record to update
- [x]  click on the group by to view the group by record detail
- [x]  when group by m2m field, record is displayed multiple times if linked to multiple m2m records
- [x]  when drag and drop, the value of the m2m field should not be updated as there is no best way to handle this
- [x]  when creating a new record, set the default value for all grouped fields
  - [x] m2m field value will also be set, starting from 17.0
- [x]  handle group by non-relation fields: click on the group of selection or char field do not popup record form view to avoid error.
- [x]  dependency_arrow between multiple collapse/expand groups
- [x] group by create_date year|month|day, try colapse the group by (year) in both project and task timelines

PR #3518, opened 2026-04-21
[18.0] [FWPORT] web_timeline: Multi group by

paradoxxxzero:
This is a fw port of #3301 with some arch parser fixes from #3455

### DIFF
--- a/web_timeline/README.rst
+++ b/web_timeline/README.rst
@@ -113,12 +113,12 @@ render the timeline items. You have to name the template
 'timeline-item'. These are the variables available in template
 rendering:
 
-- ``record``: to access the fields values selected in the timeline
-  definition.
-- ``formatters``: used to format values (see available functions in
-  ``@web/views/fields/formatters``).
-- ``parsers``: used to parse values (see available functions in
-  ``@web/views/fields/parsers``).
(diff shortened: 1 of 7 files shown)


cuongnmtm:
Thanks @paradoxxxzero  ,

You may want to review the checklist I used for previous PRs.

- [ ] group by one field (group by m2o, selection, integer, m2m, date, ...)
- [ ]  group by multiple fields
- [ ]  group by 0 field: back to default group by
- [ ]  create a new record
- [ ]  drag and drop record to update
- [ ]  click on the group by to view the group by record detail
- [ ]  when group by m2m field, record is displayed multiple times if linked to multiple m2m records
- [ ]  when drag and drop, the value of the m2m field should not be updated as there is no best way to handle this
- [ ]  when creating a new record, set the default value for all grouped fields
  - [ ]  m2m field value will also be set, starting from 17.0
- [ ]  handle group by non-relation fields: click on the group of selection or char field do not popup record form view to avoid error.
- [ ]  dependency_arrow between multiple collapse/expand groups
- [ ]  group by create_date year|month|day, try colapse the group by (year) in both project and task timelines

I added an additional check whenever an issue is reported.

flotho:
ping @DorianMAG

PR #3522, opened 2026-04-23
[ADD] web_list_record_popup: new module

rvalyi:
Extracted from what we used in the Brazilian localization in version 16.0

Recent Odoo versions allows quick edition of order lines with inline editable list.
But sometimes the user wants to read or edit more details on the lines through the Form popup. This modules allows the best of the two edition modes: you can still have quick edition via the inline list edition, but you now have a Javascript button you can click on to open the Form view popup using the view defined inline in the One2many field if any.

**Features:**

- Adds a popup button to editable lists in form views
- Opens records in a form dialog instead of inline editing
- Works with both existing and new (NewID) records
- Optional activation - no hard dependencies

**Usage:**

### A) To use it, adding a dependency to the web_list_record_popup module:

1. Make your model inherit from ``web_list_record_popup.mixin``
2. Define ``_popup_button_xpaths`` class attribute with xpath expressions where buttons should be injected

Example:

```python
    class MyModel(models.Model):
        _inherit = "web_list_record_popup.mixin"

        _popup_button_xpaths = [
            ("//field[@name='line_ids']/list/field[@name='product_id']", "before"),
        ]
```

### B) If you prefer use it WITHOUT adding a hard dependency to the web_list_record_popup module instead:

Then don't add the module dependency and don't inherit from ``web_list_record_popup.mixin`` but override ``_get_view`` instead.

Example:

```python
    class MyModel(models.Model):
        _popup_button_xpaths = [
            ("//field[@name='line_ids']/list/field[@name='product_id']", "before"),
        ]

    @api.model
    def _get_view(self, view_id=None, view_type="form", **options):
        arch, view = super()._get_view(view_id, view_type, **options)
        if view_type == "form" and "web_list_record_popup.mixin" in self.env:
            arch = self.env["web_list_record_popup.mixin"]._inject_popup_buttons(
                arch,
                self._popup_button_xpaths,
            )
        return arch, view

```

Thus, the popup button will only be injected if the ``web_list_record_popup`` module is installed.

I'm not sure why we cannot run the Hoot tests on Runboat, but here is a print of them passing locally:


CristianoMafraJunior:
@rvalyi draft?

PR #3527, opened 2026-04-29
[ADD] Introduce x2many_searchable module for inline quick search in embedded lists

drkpkg:
This module adds a new search bar that provides an optional inline quick search feature for x2many embedded lists in form views. The module includes necessary files such as the manifest, README, and JavaScript functionality to enable searching within these lists. Additionally, it includes tests to ensure the search functionality works as expected.

Screenshots

---

---

---


pedrobaeza:
Very useful one! Seeing the screenshots, I would say to reduce at maximum the padding for not losing too many vertical space, specially for small screens.

drkpkg:
> Very useful one! Seeing the screenshots, I would say to reduce at maximum the padding for not losing too many vertical space, specially for small screens.

I did an adjustment in `.o_x2m_search_panel` class to prevent the overlap and have a better UI.

Issue #3289, opened 2025-09-20
[Web favicon] gives 500 internal server error when restoring from backup -> (AttributeError: 'res.company' object has no attribute '_get_favicon').


Dream1975:
## Module
web_favicon

## Describe the bug
I installed the module to test and since then I can't use the gui anymore as it gives a "500: internal server error". Did make a backup but that seems to be corrupted also.

## To Reproduce
**Affected versions**: 18.0

1. Install the module
2. Add a favicon in both the favicon field in company and in website settings
3. Make a backup of the database (zip)
4. Restore the database (in a clean odoo environment, off course with all addons in the addon directory that were alsop present when making the backup
5. Select the database in database manager
6. You'll see the 500 page and below error appears in the log

**Expected behavior**
No Internal server error

**Additional context**
I tried adding a favicon column in table res_company (directly in the database) but this didn't solve it, unfortunately. As I can't get in the gui I also can't take a look in the technical part of views/models etc where the field is mentioned.

So I have 2 questions:
1. Does anyone know where to look so the 500 doesn't appear instantly when selecting the database in the database manager? Either in changing something directly in the database or CLI?
2.  Might this be a bug in the module (I think so as I tested with an old backup and the same happened 

**Error traceback**
### > Traceback (most recent call last):
>   File "<2011>", line 1931, in template_2011
>   File "<2011>", line 1309, in template_2011_content
>   File "/usr/lib/python3/dist-packages/odoo/addons/base/models/ir_qweb.py", line 2533, in _load_values
>     return get_value()
>            ^^^^^^^^^^^
>   File "<2011>", line 1299, in template_2011_t_cache_0_cache
>   File "<2011>", line 565, in template_2011_t_cache_0
> AttributeError: 'res.company' object has no attribute '_get_favicon'
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 2554, in __call__
>     response = request._serve_db()
>                ^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 2081, in _serve_db
>     return self._transactioning(
>            ^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3/dist-packages/odoo/http.py", line 2144, in _transactioning
>     return service_model.retrying(func, env=self.env)
[...]

PR #3528, opened 2026-04-30
[18.0][FIX] web_favicon: guard template _get_favicon call during DB restore

fsmw:
Fix for: #3289

**Problem:**
When restoring a database backup, the web.layout template renders before web_favicon has fully loaded its models. The template calls env['res.company'].sudo()._get_favicon(), but since the method doesn't exist yet during this transient state, it causes an AttributeError and a 500 error on every page.

**Fix:**
Use getattr() in the QWeb template so it safely falls back to the default favicon when _get_favicon is not yet available, instead of hard-calling a method that may not exist.

<attribute name="t-att-href">getattr(env['res.company'].sudo(), '_get_favicon', lambda: False)() or '/web/static/img/favicon.ico'</attribute>

This only adds a getattr wrapper around the existing call. When the module is loaded normally, it behaves identically. During module loading/DB restore, it returns False and the existing or clause falls back to the default favicon.

### DIFF
--- a/web_favicon/views/templates.xml
+++ b/web_favicon/views/templates.xml
@@ -6,7 +6,7 @@
         <xpath expr="//head/link[@rel='shortcut icon']" position="attributes">
             <attribute
                 name="t-att-href"
-            >env['res.company'].sudo()._get_favicon() or '/web/static/img/favicon.ico'</attribute>
+            >getattr(env['res.company'].sudo(), '_get_favicon', lambda: False)() or '/web/static/img/favicon.ico'</attribute>
         </xpath>
     </template>
 </odoo>


web_favicon/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 
 {
     "name": "Custom shortcut icon",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
```
pedrobaeza: This is done by the bot, so please don't do it yourself to cause conflicts.

web_favicon/models/res_company.py
```diff
@@ -23,10 +23,6 @@ def _get_default_favicon(self, original=False):
         with tools.file_open(img_path, "rb") as f:
             if original:
                 return base64.b64encode(f.read())
-            # Modify the source image to add a colored bar on the bottom
```
pedrobaeza, line 26: Don't remove any of the existing comments

pedrobaeza (changes requested):
I don't get the problem, as "restoring a DB" is not performing anything of you mention. Maybe it's a problem of the way you do that restore?

fsmw:
Adjusted the branch to keep the diff limited to the runtime guard only:

- reverted the manual manifest version bump
- restored the existing comments
- kept only the request.httprequest guard in _get_favicon()

The issue this guards is when _get_favicon() is evaluated without an active HTTP request object. In that state, the previous code dereferences request.httprequest.cookies directly and fails before it can fall back to the default favicon. The change keeps the existing behavior when cookies are available and returns False when there is no request object, preserving the template fallback.

fsmw:
Resolved the merge conflict against current 18.0.

The branch now keeps upstream’s request.cookies API and only adds a RuntimeError guard for calls without a bound HTTP request. The manual manifest bump is gone and existing comments are preserved.

Local check run: pre-commit run --files web_favicon/models/res_company.py.

web_favicon/__manifest__.py
fsmw: Fixed in the current branch. The manifest version bump was removed, so the PR no longer changes `web_favicon/__manifest__.py`.

web_favicon/models/res_company.py
fsmw: Fixed in the current branch. The existing comments in `_get_default_favicon()` are restored; the current diff only changes `_get_favicon()`.

fsmw:
I narrowed the PR to the actual runtime guard and updated the title to avoid implying that the SQL restore itself is the direct cause.

Current diff only protects `request.cookies.get("cids")` in `_get_favicon()`. In contexts without a bound HTTP request, accessing `request.cookies` raises `RuntimeError`; returning `False` lets the existing template fallback handle the default favicon.

The branch no longer changes the manifest version and no longer removes existing comments. All GitHub tests and pre-commit are green; only Codecov remains red.

fsmw:
@pedrobaeza thanks for the review. The guard is needed because  may be  in non-web contexts (e.g. shell scripts or during DB restore) where  exists but is not a regular HTTP request. The fix only changes behavior when  is absent; when cookies are present it continues to work exactly as before. Could you re-check when possible?

fsmw:
@pedrobaeza thanks for the review. The guard is needed because request.httprequest may be missing in non-web contexts (e.g. shell scripts or during DB restore) where request exists but is not a regular HTTP request. The fix only changes behavior when httprequest is absent; when cookies are present it continues to work exactly as before. Could you re-check when possible?

pedrobaeza:
I insist that the method you are using to restore the DB should be the problem, as the Odoo server is loaded normally and not half when you do it properly.

fsmw:
@pedrobaeza thank you for the review — you were absolutely right. The RuntimeError guard in _get_favicon() was not the right fix for the AttributeError: 'res.company' object has no attribute '_get_favicon' shown in the traceback. 

The actual issue is that the QWeb template calls _get_favicon() directly, and during DB restore the method may not yet exist on res.company. I've updated the PR to remove the Python change entirely and instead use getattr() in the template as the single-line safeguard. The diff is now only web_favicon/views/templates.xml.

Could you please re-review when possible?

fsmw:
@pedrobaeza — the fix has been corrected. The PR now only changes the template to use getattr(), and the Python model is back to the original upstream code. All checks are green. Could you please re-review when you have a moment?

pedrobaeza:
My answer is still the same, acting on one side or the other, the problem is the method you are using for such restore, that shouldn't happen with Odoo runnning, and that's the problem.

OdyX:
For the record (and to avoid running in that issue again in the future): we had this error when restoring a DB against a codebase that did _not_ have the `web_favicon` addon either available or in the addons-path (via `.odoorc`). When we ensured the addon files were correctly in the addons-path, this problem vanished.

pedrobaeza:
@OdyX but in that case, the problem is due to the dangling view. Remove `web_favicon.favicon_icon`. But this problem is not for the module itself, but this can happen in any missing module in your addons path. You uninstall the module if restoring in an environment without it, or you assure you have all the installed modules in your environment.

It's totally logical and the minimum to ask for.

OdyX:
Totally agreed. I was just trying to enhance the data for future encounters of that error message.

> Dear future me, this error message is _not_ bug that needs solving. This error message is a symptom of an Odoo code + addons' setup issue (and or broken database restore) rather than any web_favicon code issue.

PR #3529, opened 2026-05-01
[18.0][ADD] web_m2x_dialog_no_create

yostashiro:
This module hides the New button in many2one and many2many search dialogs
by default, preventing users from creating records directly from the dialog.

An allowlist of models can be configured per company in
Settings > General Settings > Search Dialog, so the New button is shown
only for selected models.

@qrtl QT6709

image


SanamiK (approved):
Functional test. LGTM.

AungKoKoLin1997 (approved):
Functional and code review: LGTM

PR #3546, opened 2026-05-12
[ADD] web_searchpanel_toggle_all

Hugo-Trentesaux:
This module simply adds a way to toggle all at once the checkboxes of a searchpanel "multi" filter by clicking the section title, the same way that the "clear all" button works in mobile view, but now available in desktop view.

(available as a standalone module here https://sources.le-filament.com/lefilament/web_searchpanel_toggle_all/)


Hugo-Trentesaux:
I had to fix precommit. Once it's fixed on 18.0, I can rebase on that, or we can merge as is.

PR #3558, opened 2026-05-25
Update copier template

SirPyTech:
I have also updated the rebel modules configuration that had been written directly in `test.yaml` (https://github.com/OCA/web/commit/0b5aa6775dfcc46a183e142d3dfa7755b43fbd37), @StefanRijnhart since you committed that change, you might want to have a look.

The order of the configurations has changed because in the template rebel modules are included first (https://github.com/OCA/oca-addons-repo-template/blob/ed7ccbe08991f72f71cfa260e98c9f9902636977/src/.github/workflows/test.yml.jinja#L88), I prefer to change the order so that we are aligned to the template and we will have fewer conflicts in the future.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -16,7 +16,8 @@ odoo_test_flavor: Both
 odoo_version: 18.0
 org_name: Odoo Community Association (OCA)
 org_slug: OCA
-rebel_module_groups: []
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

SirPyTech:
Thanks :pray:

PR #3561, opened 2026-05-29
[16.0][IMP] web_refresher: hotkey Alt R

DiegoParadeda:
Adds a simple hotkey to quick refresh the page with Alt + R

image

### DIFF
--- a/web_refresher/static/src/xml/refresher.xml
+++ b/web_refresher/static/src/xml/refresher.xml
@@ -14,6 +14,7 @@
                 aria-label="Refresh"
                 t-on-click="onClickRefresh"
                 title="Refresh"
+                data-hotkey="r"
                 tabindex="-1"
             />
         </nav>


legalsylvain (approved):

pedrobaeza:
Is there the possibility that the S key is assigned by the Odoo automatic hotkey assigner? If so, you should blacklist it.

DiegoParadeda:
> Is there the possibility that the S key is assigned by the Odoo automatic hotkey assigner? If so, you should blacklist it.

Good point. I investigated this in more detail.

In v16 it was technically possible to blacklist a hotkey, as the automatic access key assignment system used `_getAllUsedAccessKeys` and `knownUnusableAccessKeys` in `addons/web/static/src/legacy/frontend/keyboard_navigation_mixin.js`.

However, those legacy implementations were not carried forward. In v18, there no longer seems to be an automatic hotkey attribution system nor a centralized whitelist/blacklist mechanism.

To verify this, I tested overriding the well-known Alt+S behavior and was able to do so without any conflict (not even a warning in the logs):

image

So your concern is definitely valid, but from what I could find there does not appear to be a structured way to reserve or blacklist access keys in the current implementation.

That said, if there is another hotkey management mechanism in v18 that I missed, I'd be happy to adapt the implementation accordingly.

pedrobaeza:
But this is 16, so you can definitively blacklist it, isn't it?

DiegoParadeda:
> But this is 16, so you can definitively blacklist it, isn't it?

I investigated this further and found that my initial assumption was based on the legacy `KeyboardNavigationMixin`, which indeed contains a reserved-key list.

However, that code is only loaded through `web.frontend_legacy` and is not part of `web.assets_backend`. `web_refresher` runs exclusively in the backend, where hotkeys are handled by the OWL `hotkeyService` via `data-hotkey` attributes.

As far as I could verify, there is no backend auto-assignment mechanism that could dynamically assign `R` to another button, so there is no corresponding blacklist to update in this context.

Unless I'm overlooking another backend hotkey assigner, I believe `data-hotkey="r"` is sufficient here.

pedrobaeza (approved):
OK, thanks for the investigation.

/ocabot merge patch

Please fw-port it to upper versions

PR #3569, opened 2026-06-04
[18.0][IMP] web_responsive: Move AppsMenuPreferences to XML

tarteo:
This change makes it easier to change the button by inheritance

### DIFF
--- a/web_responsive/static/src/components/apps_menu/apps_menu.xml
+++ b/web_responsive/static/src/components/apps_menu/apps_menu.xml
@@ -89,4 +89,18 @@
             </button>
         </xpath>
     </t>
+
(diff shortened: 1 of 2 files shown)


tarteo:
@OCA/web-maintainers This is a simple change, can I fast forward this?

pedrobaeza:
Is this the drop icon menu? It would be good to remove it and move the preferences to the user config dialog.

tarteo:
Correct! This one:
image

pedrobaeza:
Then can you please do that option instead for fixing this once for all. I understand that if you want this to be inheritable, if for hiding it, so let's move it once for all into the user dropdown.

tarteo:
I don't have time for that atm.

PR #3570, opened 2026-06-06
[FIX][18.0] web_theme_classic: don't double-tint required monetary fields

Rad0van:
The required-field background sets `--o-input-background-color` on every `.o_input` inside an `o_required_modifier`. A monetary field renders an absolutely-positioned overlay span (which also carries `.o_input`) on top of the real input, so the semi-transparent tint was applied to both and the overlap rendered as a darker band over the value.

Keep the overlay background transparent so only the real input carries the required tint, in both form and list views.

Before: 
image

After:
image

This is from delivery.carrier configuration where something like this is used:
```
<label for="free_over" invisible="delivery_type == 'base_on_rule'"/>
<div name="free_over_amount" invisible="delivery_type == 'base_on_rule'">
    <field name="free_over"/>
    <field name="amount" widget="monetary" class="oe_inline" invisible="not free_over" required="free_over"/>
</div>

```

### DIFF
--- a/web_theme_classic/static/src/scss/web_theme_classic.scss
+++ b/web_theme_classic/static/src/scss/web_theme_classic.scss
@@ -100,6 +100,10 @@ body.classic-theme {
             /* Prevent having double border for monetary fields */
             span.o_input:has(~ input.o_input) {
                 border: $input-border-width solid transparent !important;
+                /* The overlay sits on top of the real input; keep it
+                   transparent so the required-field background is not applied
+                   twice (which renders as a darker band over the value). */
+                background-color: transparent !important;
             }
 
             /* Keep the monetary symbol away from the border when it is outside the border */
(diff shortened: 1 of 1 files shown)


legalsylvain:
Not sure to understand what you want to fix. could you provide screenshots before / after ? 
thanks !

Rad0van:
@legalsylvain sorry - my bad. Updated the initial cooment with screenshots and code snippet.

web_theme_classic/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     "name": "Web Theme Classic",
     "summary": "Contrasted style on fields to improve the UI.",
-    "version": "18.0.1.2.0",
+    "version": "18.0.1.2.1",
```
legalsylvain: That is not necessary.

legalsylvain:
thanks.

/ocabot merge patch

PR #3581, opened 2026-06-13
[19.0][MIG] web_send_message_popup: Migration to 19.0

dnplkndll:
Port of [web_send_message_popup](https://github.com/OCA/web/tree/18.0/web_send_message_popup) from 18.0 to 19.0.


web_send_message_popup/static/src/chatter/web/chatter_patch.esm.js
```diff
@@ -0,0 +1,202 @@
...
+                    (attachment) => attachment.id
+                ),
+                default_body,
+                default_email_add_signature: false,
+                default_model: this.state.thread.model,
+                default_partner_ids: allRecipients.map(
+                    (recipient) => recipient.partner_id
+                ),
+                default_res_ids: [this.state.thread.id],
+                default_subtype_xmlid: "mail.mt_comment",
```
gaikaz, line 113: `clicked_on_full_composer: true` is missing from this context. It's [new in 19.0](https://github.com/odoo/odoo/blob/ccd167e23ea80545e4f09e2302c1b2489e8c42d8/addons/mail/static/src/core/common/composer.js#L616) and the composer view uses it to determine how to print `partner_ids` field ([see here](https://github.com/odoo/odoo/blob/ccd167e23ea80545e4f09e2302c1b2489e8c42d8/addons/mail/wizard/mail_compose_message_views.xml#L45-L53))

* the "Followers only" `partner_ids` widget is `invisible` when the key is absent
* the forward-style "Add recipients…" widget becomes visible **and `required`**

So the popup this module opens shows a different recipients field than the standard full composer, and refuses to send until the user manually adds a recipient.

web_send_message_popup/static/src/chatter/web/chatter_patch.esm.js
```diff
@@ -0,0 +1,202 @@
...
+                            if (!isEmpty) {
+                                this.saveContent();
+                                this.restoreContent();
+                            }
+                        },
+                    });
+                } else {
+                    this.clear();
+                }
+                this.onCloseFullComposerCallback();
```
gaikaz: This callback [gained an argument in 19.0](https://github.com/odoo/odoo/blob/ccd167e23ea80545e4f09e2302c1b2489e8c42d8/addons/mail/static/src/chatter/web/chatter_patch.js#L370):

```js
onCloseFullComposerCallback(isDiscard) {
    this.toggleComposer();
    super.onCloseFullComposerCallback();
    if (!isDiscard) {
        this.reloadParentView();
    }
},
```

`isDiscard` is computed on line 106 but never passed on, so it arrives as `undefined`, `!isDiscard` is always true, and `reloadParentView()` runs on every close - including discard. That method starts with `await this.props.saveRecord?.()`, so discarding the message silently saves the underlying form record.

```suggestion
                this.onCloseFullComposerCallback(isDiscard);
```

web_send_message_popup/static/src/chatter/web/chatter_patch.esm.js
```diff
@@ -0,0 +1,202 @@
...
+            context: {
+                default_attachment_ids: this.state.thread.composer.attachments.map(
+                    (attachment) => attachment.id
+                ),
+                default_body,
+                default_email_add_signature: false,
+                default_model: this.state.thread.model,
+                default_partner_ids: allRecipients.map(
+                    (recipient) => recipient.partner_id
+                ),
```
gaikaz: Core [filters here](https://github.com/odoo/odoo/blob/ccd167e23ea80545e4f09e2302c1b2489e8c42d8/addons/mail/static/src/core/common/composer.js#L608-L613):

```js
allRecipients
    .filter((recipient) => recipient.partner_id)
    .map((recipient) => recipient.partner_id),
```

And for a good reason. `partner_id` can legitimately still be unset in some edge-cases. Best to keep it aligned with standard.

Also, inconsistent as well: line 65 adds an `if (recipient)` that core does *not* have, while this filter that core *does* have is gone.

web_send_message_popup/static/src/chatter/web/chatter_patch.esm.js
```diff
@@ -0,0 +1,202 @@
...
+                default_res_ids: [this.state.thread.id],
+                default_subtype_xmlid: "mail.mt_comment",
+                mail_post_autofollow: this.state.thread.hasWriteAccess,
```
gaikaz: Should this still be here? 19.0 dropped `mail_post_autofollow` from the full-composer context deliberately with the comment ([see here](https://github.com/odoo/odoo/commit/f839d846114143f504ef74b18e9ea77f6e94fd64#diff-6f878c7731d007474053f87b56afc58ee0898b13ba23d1e618e772ad7f9472cfR543)):

> Changed in 18.2+: finally get rid of autofollow, following should be done manually

The server still honours the key, so keeping it means this module's popup auto-subscribes recipients as followers while the standard 19.0 popup does not - difference from core Odoo which is not what this module is meant to do.

web_send_message_popup/static/src/chatter/web/chatter_patch.esm.js
```diff
@@ -0,0 +1,202 @@
...
+            const signatureEl = renderToElement("html_editor.Signature", {
+                signature,
+                signatureClass: "o-signature-container",
+            });
+            fragment.append(signatureEl);
+        }
+        const container = document.createElement("DIV");
+        container.append(...childNodes(fragment));
+        wrapInlinesInBlocks(container, {baseContainerNodeName: "DIV"});
+        return container.innerHTML;
```
gaikaz: This is still the 18.0 copy; core's version changed ([see here](https://github.com/odoo/odoo/blob/ccd167e23ea80545e4f09e2302c1b2489e8c42d8/addons/mail/static/src/core/web/composer_patch.js#L20-L37)):

1. **Missing `<BR>` before the signature.**
Odoo now appends a `BR` *before* `signatureEl` (new in 19.0). Without it the signature is glued onto the last line of the body.
2. `parseHTML(document, defaultBody)`
Odoo uses `createDocumentFragmentFromContent(defaultBody).body`, which runs `htmlEscape()` first.
3. `return container.innerHTML`
Odoo returns `markup(container.innerHTML)`. Harmless across JSON-RPC, but it breaks what generally 19.0 assumes.

web_send_message_popup/static/tests/chatter_patch.test.js
```diff
@@ -0,0 +1,64 @@
...
+                expect(action.context.default_model).toBe("res.fake");
+                expect(action.context.default_subtype_xmlid).toBe("mail.mt_comment");
+                // Asserting the dispatched action is enough.
+                return Promise.resolve();
+            },
+        });
+        await start();
+        await openFormView("res.fake", fakeId);
+        await contains("button", {text: "Send message"});
+        await chatter.openFullComposer();
```
gaikaz: This bypasses the thing the module actually does.

`openFullComposer()` is a new method added by this module - nothing in core calls it - so `toggleComposer("message")`, which *is* the feature, is never exercised. A test that goes through the button would also catch the patch failing to apply or being ordered wrong, which is the realistic breakage mode for a `patch(Chatter.prototype, …)` module.

Odoo's own chatter tests do exactly that:. `addons/mail/static/tests/chatter/web/chatter.test.js:104`:

```js
await click("button", { text: "Send message" });
```

The assertions could also go further (could check for):
- recipients → `default_partner_ids`
- `composerHtml` → `default_body`
- signature handling (isn't covered).

gaikaz (commented):
Thanks for picking this one up.
We are now migrating our system, so had to review for personal benefit 😄

web_send_message_popup/static/src/chatter/web/chatter_patch.esm.js
```diff
@@ -0,0 +1,210 @@
...
+            this.closeSearch();
+            this.state.composerType = false;
+            if (!this.state.thread.id) {
+                this.props.saveRecord?.();
+            }
+            this.openFullComposer();
+            return;
+        }
+        return super.toggleComposer(...arguments);
+    },
```
gaikaz, line 59: On `mode === "message"` we here will return before calling `super`. Because of it, we are not calling `updateRecipients` as [`super` method does](https://github.com/odoo/odoo/blob/ccd167e23ea80545e4f09e2302c1b2489e8c42d8/addons/mail/static/src/chatter/web/chatter_patch.js#L467-L485) 

Keep in mind that the missing method is async.

gaikaz (commented):
That's all, I think.

gaikaz (approved):

leemannd (approved):

BhaveshHeliconia (approved):
Code review LGTM!

PR #3584, opened 2026-06-13
[19.0][MIG] web_tree_dynamic_colored_field: Migration to 19.0

dnplkndll:
Port of [web_tree_dynamic_colored_field](https://github.com/OCA/web/tree/18.0/web_tree_dynamic_colored_field) from 18.0 to 19.0.


legalsylvain (changes requested):
What a IA slope !

dnplkndll:
> What a IA slope !

apologies this batch was indeed intended to start the review process on my fork not the upstream. would it be best to draft until ready and then un-draft or delete and re-create?

dnplkndll:
would it be safe to assume is the view changed the match error should be swallowed?

PR #3586, opened 2026-06-14
[18.0][FIX] web_chatter_position: prevent horizontal overflow in bottom chatter

sergio-teruel:
@Tecnativa

## Description

When `web_chatter_position` is configured to display the chatter at the bottom, form views may become wider than the browser viewport.

This is especially visible on wide forms, such as Sales Orders with many order line columns, or when the browser window is resized to a narrower desktop width. The page gets a large horizontal scrollbar instead of keeping the form constrained to the available viewport width.

The issue is caused by the bottom chatter/form layout keeping some sizing behavior from the aside layout and by the form renderer/sheet being allowed to grow according to its internal content.

## Before this PR

With the chatter forced to the bottom:

- the chatter container could create extra horizontal space;
- the form renderer could grow beyond the browser viewport;
- a large horizontal scrollbar appeared at the page level;
- the bottom chatter background did not match the regular aside chatter style.

Example:

image

## After this PR

With the chatter forced to the bottom:

- the form view is constrained to the browser viewport width;
- horizontal overflow is avoided at the page/form container level;
- wide inner content, such as order lines, scrolls inside the form sheet instead of expanding the whole page;
- the bottom chatter keeps the same grey background style as the aside chatter;
- the runtime toggle between side and bottom positions keeps the expected classes and overflow behavior.

Example:

image

ping @CarlosRoca13 @carlos-lopez-tecnativa @carlosdauden


CarlosRoca13 (approved):
LGTM 👍

BhaveshHeliconia (approved):
LGTM!

Issue #3112, opened 2025-03-04
[17.0]web_remember_tree_column_width: can't remember change width column


hamzaezzamane:
Column width changes are not saved immediately and sometimes not applied
Module

## Module
web_remember_tree_column_width

## Describe the bug
When resizing a column in a tree view, the new width is only saved in local storage after clicking outside the table. If the user does not click outside, the width is not remembered. Additionally, in some cases, even after clicking outside, the new width is stored but not actually applied when reloading the page.

## To Reproduce
Odoo 17

1.Resize a column in a tree view.
2.Do not click outside the table and refresh the page.
3.The column width is not remembered.
4.Resize a column and then click outside the table.
5.Refresh the page. Sometimes, the column width in local storage is updated, but the change is not applied in the UI.

**Expected behavior**
-The column width should be saved immediately when resized, without requiring an extra click outside the table.
-The stored column width should always be correctly applied when reloading the page.

**Additional context**
Odoo version: 17
Operating System: windonws 11
Browser: CHROME

![Image](https://github.com/user-attachments/assets/d117e59f-61e8-46fd-9328-65e72e1fe6f7)

Paulius11:
Any update on this? Still reproducible in Odoo 17.

SaumilLapis:
Still reproducible in odoo 17 at this moment.

Any updates here?

Larsq1:
Columns that contain a stored field do remember the column width, columns that are non-stored do not

Larsq1:
I made a fix for this: https://github.com/OCA/web/pull/3588

@SaumilLapis @Paulius11 @hamzaezzamane would any of you like to review?

PR #2720, opened 2024-01-12
[16.0][IMP] web_remember_tree_column_width: adjust SCSS selector

cuongnmtm:
This PR fixes the issue https://github.com/OCA/web/issues/2717

Technical note:
- Previously, only columns with the class o_column_sortable had the custom SCSS applied to them. As a result, columns like HTML were not sortable and didn't benefit from the custom SCSS.
- This pull request made changes to the SCSS selector to include all table columns in the list view. However, the record selector (the first checkbox column) and column optional (the last column) that have the class o_list_controller were excluded.

PR #3588, opened 2026-06-15
[FIX] web_remember_tree_column_width: Include non-stored columns to remember width

Larsq1:
Fixes https://github.com/OCA/web/issues/3112

### DIFF
--- a/web_remember_tree_column_width/__manifest__.py
+++ b/web_remember_tree_column_width/__manifest__.py
@@ -5,7 +5,7 @@
     "website": "https://github.com/OCA/web",
     "license": "LGPL-3",
     "category": "Extra Tools",
-    "version": "17.0.1.0.0",
+    "version": "17.0.1.0.1",
     "maintainers": [
         "frahikLV",
         "luisg123v",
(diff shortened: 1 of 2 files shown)


cuongnmtm:
I implemented a similar fix for 16.0 here: https://github.com/OCA/web/pull/2720.

It's a bit more complicated because the first and last columns need to be excluded. Do you think we still need to do that in 17.0?

sanderlienaerts (approved):

Larsq1:
@cuongnmtm Why do the first and last column need to be excluded?

cuongnmtm:
> @cuongnmtm Why do the first and last column need to be excluded?

@Larsq1 Because the first column is the record selection checkbox, and the last column is the optional fields column. However, your approach of checking data-name may already exclude those columns so that it might be a better solution.

I confirmed that the first and last columns have no data-name attribute.

cuongnmtm (approved):

web_remember_tree_column_width/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "website": "https://github.com/OCA/web",
     "license": "LGPL-3",
     "category": "Extra Tools",
-    "version": "17.0.1.0.0",
+    "version": "17.0.1.0.1",
```
BhaveshHeliconia, line 8: I don’t think we need to bump the version manually, as it should be updated automatically when the merge command is applied

PR #3591, opened 2026-06-17
[19.0][ADD] web_dark_mode_responsive: dark mode styles for responsive apps menu

gerard-domatix:
Adds a glue module between 'web_dark_mode' and 'web_responsive' so each module stays independent for OCA.

Fixes dark mode styling of the responsive full-screen apps menu: background, search field contrast and app icons.

Only SCSS assets in `web.assets_web_dark`; no Python logic changes.

Screenshots before/after attached (apps menu with dark mode enabled).

## Screenshots

Same view in all cases: responsive apps menu open, dark mode enabled.

### 1. Without `web_dark_mode_responsive`
imagen
 
### 2. With this PR (current `web_dark_mode` palette)
imagen
 
### 3. With this PR + blue-dark palette
imagen

- **1:** `web_dark_mode` + `web_responsive` only — broken/light apps menu in dark mode.
- **2:** this PR installed — menu aligned with dark mode (current gray palette).
- **3:** preview if the blue-dark palette PR is merged — not part of this PR.


pedrobaeza:
Why not depending `web_dark_mode` on `web_responsive`? At the end, enterprise users have already dark mode.

gerard-domatix:
> Why not depending `web_dark_mode` on `web_responsive`? At the end, enterprise users have already dark mode.

We preferred not to add web_responsive as a dependency of web_dark_mode because many setups use dark mode without the responsive apps menu, and we didn't want to force that module on everyone. The glue module keeps the integration optional.

pedrobaeza:
Which setups? Can you be more specific?

gerard-domatix:
> Which setups? Can you be more specific?

You're right, I didn't express myself well. I wasn't referring to specific setups we can list.

What we had in mind is that someone might use dark mode without web_responsive, and we didn't want to force that module for a visual fix only. That's why we went with a glue module.

That said, if you think it fits better somewhere else (e.g. in web_responsive or as a direct dependency), that works for us too, might even be simpler.

pedrobaeza:
Yeah, I think we should seek simplicity whenever possible, and having that enterprise users have already dark mode, I can't figure out now any setup wanting dark mode, but not responsive features, so I think we can depend on `web_responsive` for `web_dark_mode`.

gerard-domatix:
> Yeah, I think we should seek simplicity whenever possible, and having that enterprise users have already dark mode, I can't figure out now any setup wanting dark mode, but not responsive features, so I think we can depend on `web_responsive` for `web_dark_mode`.

Sounds good to me, simplicity makes sense here. I'll close this PR and move the integration into web_dark_mode with a web_responsive dependency instead.

Issue #3387, opened 2025-12-09
web_responsive overhaul feeler


ljmnoonan:
While working on trying to make my [web_dark_mode](#3324) work with `web_responsive` I have realized that it may be time for a bit of an overhaul.

However, before spending any time doing so I'd like to get the consensus of the users and maintainers of this module to see what I should and shouldn't do. Here are the main points:

## Lack of Direction

Two themes for the app screen as well as 3 different search modes.  I have read all the discussion on #2684 where this was introduced and feel like there should have been an agreement reached before going forward rather than just keeping all options. The two themes, especially, make implementing dark mode confusing and inconvenient. 

As far as search modes go, my personal preference would be odoo command palette, but keep the search bar. I could possibly modify the css to line it up with the search bar in order to seem more coherent. In any case, I would not get rid of the search bar as it is the only thing allowing searches from mobile.

Getting rid of these options would also eliminate the need for the little drop icon in the navbar. If we cannot agree on a single mode + theme, I would like to also suggest moving this to the user preferences section as I don't see any real reason for it to be in the navbar.

## Outdated Code

In 18.0 list view headers are sticky by default and in 19.0 form status bars are sticky too. Maybe other features too?

## File Structure

Not really a big thing, but the file structure does not follow `web`. `primary_varables.scss` for instance are tucked away in a `legacy` folder. Also, everything to do with the home menu would be under `webclient` and probably not split between `apps_menu` and `apps_menu_item`. If there are no objections, I would like to reorganize this. 

## Redirect to Home

I like this feature, but it seems a bit out of place. Could it be incorporated into `web_quick_start_screen`? I would also like to add a little home button to the navbar and could have sworn that I have actually seen that done already somewhere. Anyway, not really the subject of this issue.

<br>
<br>

Lastly, the big question: should I start these modifications on 19.0 of 18.0 and would you prefer it be broken up into multiple PRs, or just one?

Thank you all!

pedrobaeza:
As the "native" search can be triggered with Ctrl+K, and then typing `/`, I would reduce the menu search to the fuzz one.

I think the changes can be done in 19.0, that is still penetrating in the instances.

ljmnoonan:
@pedrobaeza does the fuzz search offer any advantages over the native search? My understanding is that it allows for misspellings by using a less exact match. I do not have enough experience with it to know whether this makes a practical difference though. 
[...]

PR #3592, opened 2026-06-17
[19.0][IMP] web_dark_mode: dark mode styles for responsive apps menu

gerard-domatix:
When 'web_dark_mode' and 'web_responsive' are used together, the full-screen apps menu still uses light-theme colors: pale background, dark search text, and icons that don't match the dark UI.

This PR adds `web_responsive` as a dependency and includes dark styles for the apps menu so it stays consistent with the rest of the backend in dark mode.

Change is limited to `__manifest__.py` and a new `apps_menu.dark.scss` — no palette changes.

## Screenshots

Same view in all cases: responsive apps menu open, dark mode enabled.

### 1. Before this PR

imagen

### 2. After this PR

imagen

### 3. After this PR (blue-dark palette preview, PR #3590)

imagen

- **1:** `web_dark_mode` + `web_responsive` without these changes — apps menu stays on light colors in dark mode.
- **2:** with this PR — apps menu matches the dark UI (current gray palette).
- **3:** same fix with a blue-dark palette (PR #3590) — illustrative preview only, not included here.

### DIFF
--- a/web_dark_mode/__manifest__.py
+++ b/web_dark_mode/__manifest__.py
@@ -8,7 +8,7 @@
     "version": "19.0.1.0.0",
     "website": "https://github.com/OCA/web",
     "author": "initOS GmbH, Odoo Community Association (OCA)",
-    "depends": ["web"],
+    "depends": ["web", "web_responsive"],
     "excludes": ["web_enterprise"],
     "installable": True,
     "assets": {
--- a/web_dark_mode/static/src/web_responsive/apps_menu.dark.scss
+++ b/web_dark_mode/static/src/web_responsive/apps_menu.dark.scss
@@ -1,0 +1,98 @@
+// Copyright 2026 Domatix
+// License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+// Dark overrides for the web_responsive apps menu.
+// Use :root prefix to match web_responsive specificity and win by load order.
+
+$app-menu-overlay: url("/web_responsive/static/src/img/home-menu-bg-overlay.svg");
+
+// Teal glow (#122e3a), radial focus from bottom-center
+// to soften the transition with the navbar above.
+$app-menu-glow: radial-gradient(
+    120% 95% at 50% 108%,
+    rgba(28, 74, 95, 0.6),
+    rgba(28, 74, 95, 0) 64%
+);
+
+:root .o_grid_apps_menu[data-theme="milk"] {
+    --app-menu-background:
+        #{$app-menu-glow}, #{$app-menu-overlay},
+        linear-gradient(to bottom, #{$o-gray-200}, #{$o-gray-300});
+    --app-menu-text-color: #{$o-gray-900};
+    --app-menu-text-shadow: none;
+    --app-menu-hover-background: #{rgba($o-black, 0.08)};
+    --apps-menu-scrollbar-background: #{$o-gray-500};
+    --apps-menu-empty-search-color: #{$o-gray-900};
+}
+
+:root .o_grid_apps_menu[data-theme="community"] {
+    --app-menu-background:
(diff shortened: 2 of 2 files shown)


alvaro-domatix (approved):

web_dark_mode/__manifest__.py
```diff
@@ -5,10 +5,10 @@
     "name": "Dark Mode",
     "summary": "Enabled Dark Mode for the Odoo Backend",
     "license": "AGPL-3",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.0.1",
```
pedrobaeza: No need to manually change the version. In fact, this will conflict if we merge the other PR.

pedrobaeza (approved):

BhaveshHeliconia (approved):
LGTM!

web_dark_mode/__manifest__.py
gerard-domatix: Done, thanks for the review!

web_dark_mode/static/src/web_responsive/apps_menu.dark.scss
ljmnoonan, line 1: This should be in `web_responsive` itself.

web_dark_mode/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     "version": "19.0.1.0.0",
     "website": "https://github.com/OCA/web",
     "author": "initOS GmbH, Odoo Community Association (OCA)",
-    "depends": ["web"],
+    "depends": ["web", "web_responsive"],
```
ljmnoonan, line 11: Even if you were to keep this in `web_dark_mode` you still would not need to have this dependency. You are not calling on any scss variables that are only defined in in `web_responsive`, so this css will just get loaded and do nothing

ljmnoonan (changes requested):
Thanks for the contribution! This has been a problem for a while now.

I like the style, however it should be in `web_responsive`, not `web_dark_mode`, and it should certainly not make `web_dark_mode` dependent on `web_responsive`

It is as simple as as putting this css file anywhere in `web_responsive` and adding it to `web.assets_web_dark` in the manifest.

I am also working on a solution for this as part of the project I started with #3387. It is a much bigger project though, and may wind up getting rejected due to its scope. You can see it [here](https://github.com/Pyxiris/web/tree/19.0-web_responsive_cleanup)

In the meantime, this fixes the dark menu issue nicely.

len-foss:
The change is a welcome one :-)

It makes as much sense to put this here or to put it in web_responsive, doesn't it?
If it is preferred to to make a bridge module (which is totally understandable) it would indeed be important to not make one depend other the other (especially since there are bugs or unintended behaviours with `web_responsive`), so I think it would be enough to simply explain the situation in line comments/README.

ljmnoonan:
@len-foss 
> It makes as much sense to put this here or to put it in web_responsive, doesn't it?

From a purely technical standpoint, yet, this is true. It will be added to the minimized css file either way.

However, from a code convention and organization point of view, every other module in odoo contains its own dark mode css assets suffixed with `dark.scss` so I don't see why this module should break this convention, especially considering how easy it is to avoid breaking convention

len-foss:
> However, from a code convention and organization point of view, every other module in odoo contains its own dark mode css assets suffixed with `dark.scss` so I don't see why this module should break this convention, especially considering how easy it is to avoid breaking convention

:+1: makes total sense, thank you for the clarification.

pedrobaeza:
@gerard-domatix are you going to do the changes? There's a similar change in #3631

pedrobaeza:
Superseded by #3631, as this one didn't attend the requested changes, and the color scheme is controversial.

PR #3597, opened 2026-06-23
[18.0][ADD] web_filter_paste_multiline: creation of module

guewen:
This module brings a usability improvement when using custom filters.

Usually, when you paste a multi line text in a filter using the "is in" operator, it is interpreted as a single value:

image

When this module is installed, each value is expanded as one value per line:

image

Very useful when searching values from a spreadsheet, csv file, ...

https://github.com/user-attachments/assets/bf76e443-d2ca-4033-82af-4d15ea33ac50


mtriki (approved):
Thank you! Very useful indeed 🔥

WentworthJ (approved):
Functionally tested on Runboat. When adding a custom filter, each line of pasted text is correctly added as a separate value.

BhaveshHeliconia (approved):

PR #3600, opened 2026-06-30
[FIX] web_responsive: replace this.ui.isSmall with env.isSmall for Odoo

volkantasci:
Odoo 18 replaced this.ui.isSmall with env.isSmall in the navbar
template. The xpath expression and template condition were using the
old property name, causing an OwlError because the target element
could not be located in the element tree.


pedrobaeza:
It's the contrary: #3536

Please update your whole code, including Odoo.

PR #3604, opened 2026-07-10
19.0 upgrade web notify

3bdalla3adil:
use the existing bus simple notification service instead of the patch


pedrobaeza:
Clearly incorrect PR, so closing.

PR #3608, opened 2026-07-17
[18.0][ADD] web_field_tooltip_copy: web field tooltip copy

polqubiq:
## Web Field Tooltip Copy

When writing a task (ticket, PR, or spec) you almost always need to quote **technical field names**. In **developer mode**, Odoo shows them in the `?` next to the field label… but you still have to copy the field label by hand.

**This module fixes it with one click:**

- Enable **developer mode** and hover the `?` on a field
- The popover **stays open** while you hover it
- Click any technical value → **copied to the clipboard** (field name, model, widget, domain, selection…)

Less friction when documenting, estimating, or debugging. A small daily win for Odoo developers.

Preview:
image


polqubiq:
Hi @pedrobaeza @Tardo @SplashS I've added a couple of modules that I think could be useful to the community. They're the kind of small things we do in our day-to-day work that can save us a bit of time.
One is this one, and the other is here: https://github.com/OCA/web/pull/3609

PR #3623, opened 2026-07-27
[18.0][FIX] web_calendar_slot_duration, web_refresher: remove broken pt_BR.po stubs

Guilherme-Eller:
While auditing all `pt_BR.po` files in this repo with `msgfmt --check` (gettext 0.26), two files fail with a fatal error:

- `web_calendar_slot_duration/i18n/pt_BR.po` — **empty file (0 bytes)**
- `web_refresher/i18n/pt_BR.po` — **no PO header** (`PO file header missing or invalid`), and both entries have empty `msgstr`

Neither file carries any actual translation, so removing them loses nothing and unbreaks downstream tooling that validates .po files. Weblate will regenerate valid files when pt_BR translations arrive.

One commit per module.


web_calendar_slot_duration/i18n/pt_BR.po
pedrobaeza, line 1: You should remove this one.

web_refresher/i18n/pt_BR.po
```diff
@@ -1,13 +0,0 @@
-#. module: web_refresher
```
pedrobaeza, line 1: Why this one is being removed?

yajo (approved):
Comment 1 explains everything and is reasonable.

/ocabot merge nobump

pedrobaeza:
@yajo I have aborted. The patch is incorrect. How adding an empty file is correct? And removing the entire .po because there's no header instead of adding the header is not the correct patch as well.

yajo:
Ah sorry I got confused! I thought he was removing both files, not adding one! Thanks for aborting.

Regarding the removal, it might not be the best solution, but it's also true that there are no translations there, so it's a solution at least.

pedrobaeza:
OK about the second one to be removed not having translations, but the first one should be removed (so the patch is only to remove the second file).

PR #3630, opened 2026-07-31
[16.0][ADD] web_required_field_indicator: initial version

corredato:
Odoo's "Invalid fields:" notification is misleading for an empty required field: it is not invalid, it is simply not filled in. This module renames the notification title accordingly.

It also highlights, in red, the notebook tab of any page that contains such a field, so it does not have to be hunted down page by page on forms with many notebook pages.


corredato:
@sysadminmatmoz @rvalyi @simahawk @sbidoul @dreispt @gurneyalex 

review pls

PR #3631, opened 2026-08-03
[19.0][IMP] web_responsive: add dark mode support for the apps menu

volkantasci:
When the user enables dark mode (e.g. with the `web_dark_mode` addon), the
grid apps menu rendered by `web_responsive` kept its light background image,
light gradient and light colors, making the menu inconsistent with the rest
of the backend and hard to read.

## What this PR does

- Adds `web_responsive/static/src/components/apps_menu/apps_menu.dark.scss`,
  registered in the `web.assets_web_dark` bundle so it is only loaded when
  dark mode is active.
- Restyles in dark mode:
  - the apps menu grid background (uses `$o-gray-300` instead of the light
    `$app-menu-background-color` gradient and overlay image)
  - menu item text color, text shadow and hover background
  - the search box (background, icon, input and placeholder)
  - the canonical search results and their highlight/hover state
  - the scrollbar and "empty search" colors
- Adds a `("remove", "web_responsive/static/src/**/*.dark.scss")` directive
  to `web.assets_backend` so dark mode files never leak into light mode,
  mirroring the pattern used by the `web` addon itself.

All colors reuse the dark mode variables (`$o-gray-*`) so the menu follows
the same palette as the rest of `web_dark_mode`.

## How to test

1. Install `web_responsive` and `web_dark_mode`.
2. Enable dark mode from the user menu (or set the user preference).
3. Open the apps menu grid and search: everything is now dark-themed.
4. Switch back to light mode: no visual change from before.

## Affected files

- `web_responsive/__manifest__.py`
- `web_responsive/static/src/components/apps_menu/apps_menu.dark.scss` (new)

### DIFF
--- a/web_responsive/static/src/components/apps_menu/apps_menu.dark.scss
+++ b/web_responsive/static/src/components/apps_menu/apps_menu.dark.scss
@@ -1,0 +1,67 @@
+/* Copyright 2023 Taras Shabaranskyi
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). */
+
+:root {
+    .o_grid_apps_menu[data-theme="milk"],
+    .o_grid_apps_menu[data-theme="community"] {
+        --app-menu-background: linear-gradient(
+            to bottom,
+            #{$o-gray-300},
+            #{desaturate(darken($o-gray-300, 5%), 15)}
+        );
+        --app-menu-text-color: #{$o-gray-900};
+        --app-menu-text-shadow: 1px 1px 1px #{rgba(black, 0.4)};
+        --app-menu-hover-background: #{rgba(white, 0.08)};
+        --apps-menu-scrollbar-background: #{$o-gray-400};
+        --apps-menu-empty-search-color: #{$o-gray-900};
+    }
+}
+
+.o_grid_apps_menu {
+    .o-app-menu-item {
+        &__icon {
+            background-color: $o-gray-400;
+        }
+
+        &__active {
+            background-color: $o-gray-900;
+            border-radius: 50%;
+            color: $o-gray-100;
+            text-shadow: none;
+        }
+    }
+
+    .search-container {
+        // Search box
+        .search-input {
+            background-color: $o-gray-300;
+
+            .search-icon {
+                color: $o-gray-900;
+            }
+
+            .form-control {
+                color: $o-gray-900;
+
+                &::placeholder {
+                    color: $o-gray-700;
+                }
+            }
+        }
+
+        // Search results (canonical searchbar)
+        .search-item {
+            background-color: $o-gray-300;
+
+            &__name {
+                color: $o-gray-900;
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Isn't this the same as #3592 ?

ljmnoonan:
@pedrobaeza, practically, yes, although this PR does correctly add the dark mode assets to `web_responsive` instead of `web_dark_mode`.
I tested on runboat and it does appear to work, but I am not a huge fan of the white icon backgrounds
image

pedrobaeza:
Yes, please use another color like the other PR.

volkantasci:
Hi @pedrobaeza, @ljmnoonan, thanks for the review!

To clarify the relation with #3592: this PR adds the dark assets to `web_responsive`, which owns the apps menu components, while #3592 adds them to `web_dark_mode` and makes it depend on `web_responsive`.

I just pushed an update addressing the feedback:
- app icons now use `$o-gray-400` instead of the white background
- fixed the prettier formatting issue that failed pre-commit

Thanks for testing on runboat @ljmnoonan! If maintainers prefer one of the two approaches, happy to align with it.

web_responsive/static/src/components/apps_menu/apps_menu.dark.scss
```diff
@@ -0,0 +1,58 @@
...
+        --app-menu-hover-background: #{rgba(white, 0.08)};
+        --apps-menu-scrollbar-background: #{$o-gray-400};
+        --apps-menu-empty-search-color: #{$o-gray-900};
+    }
+}
+
+.o_grid_apps_menu {
+    .o-app-menu-item__icon {
+        background-color: $o-gray-400;
+    }
```
ljmnoonan: The "Active" bubble should probably contrast:

```suggestion
    .o-app-menu-item {
        &__icon { 
            background-color: $o-gray-400;
        }
        &__active {
            background-color: $black
            // The shadow needs to change as well as the text color.
            // I have not played with it enough though
        }
    }
```

ljmnoonan (commented):
Very good. I think the active bubble should probably contrast. This is not a big deal though

volkantasci:
Thanks @ljmnoonan for the review and the suggestion on [the active bubble](https://github.com/OCA/web/pull/3631#discussion_r3713839574)! Applied in the updated commit: it now gets a `$o-gray-900` circular background with a light (`$o-gray-100`) check and no text shadow, so it contrasts properly against the `$o-gray-400` icon.

ljmnoonan (approved):
I was thinking that the active bubble would be all light, but I suppose this works too:

Image

Also, although it may be out of scope for this PR, if you wanted to tackle the message menu highlighting in dark mode that would be awesome

Image

Thank you for your contribution!

pedrobaeza (approved):
Let's go with this one then.

It would be good if the other problem is solved as well in a new PR.

/ocabot merge minor

PR #3641, opened 2026-08-13
[19.0][ADD] web_grid_view: add community grid view type

alvaro-domatix:
# web_grid_view

Adds a **grid view** type to Odoo.

A grid view renders records as a 2-dimensional matrix: rows group records by a
field (typically a many2one, with an optional section level), columns represent
consecutive time periods, and each cell aggregates a measure of the records that
fall in that row and column. Useful for timesheets, planning, forecasting and any
data that fits a row/column/date layout.

## Features

- Row grouping with an optional section level
- Date columns with selectable ranges: day, week, month, year
- Configurable measure field
- Inline cell editing through `grid_update_cell`
- Section, column and grand totals
- Magnifier to open the records behind a cell
- Sticky headers and row titles
- Includes a demo grid view on `ir.cron`

## Usage

```xml
<record id="view_my_model_grid" model="ir.ui.view">
    <field name="name">my.model.grid</field>
        <grid_view string="My Grid" editable="1">
            <field name="project_id" type="row" section="1" />
            <field name="user_id" type="row" />
            <field name="date" type="col">
                <range
                    name="week"
                    string="Week"
                    span="week"
                    step="day"
                    default="1"
                />
            </field>
            <field name="unit_amount" type="measure" widget="float_time" />
        </grid_view>
    </field>
</record>
```

Add `grid_view` to the action's `view_mode` to use it.


pedrobaeza:
Can you include demo data with an example over `ir.cron` for example?

web_grid_view/models/demo.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2026 Domatix
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class GridDemoEntry(models.Model):
+    _name = "grid.demo.entry"
```
pedrobaeza: This model shouldn't be here, as it's loaded no matter if demo or not.

alvaro-domatix:
Done! @pedrobaeza

pedrobaeza:
Thanks, I will review next week and tell you back.

PR #2976, opened 2024-10-29
[16.0][ADD] web_form_view_resizable

sbejaoui:
This module allows users to resize the form view, enhancing usability by providing additional space for better display. It's especially practical when more room is needed to view attachments in the chatter or to expand the form view for improved readability.

web_form_view_resizable (1)

PR #3642, opened 2026-08-18
[19.0][MIG] web_form_view_resizable

samirGuesmi:
migration from **v16** from this PR: https://github.com/OCA/web/pull/2976


petrus-v:
Code review and tested on runboat, that looks pretty good.

It's a shame that we haven't an utility method such the one provided by jquery it makes code much harder to maintains... we should probably add some test somehow :thinking: 

Probably because of https://github.com/OCA/web/pull/2976#issuecomment-5327399918 I think we use to support resizing PDF preview, should we support it in v19 too ?

AnizR:
Tested on runboat, with and without the PDF preview but, it doesn't seem to work

PR #3645, opened 2026-08-28
[19.0][FIX] web_widget_one2many_tree_line_duplicate: keep relational values on the clone

CarlosRoca13:
Extend the snapshot we re-apply after the onchanges to the relational types. Many2ones are copied as a new value object so both records don't share it, and many2manys through a SET command (and not LINK) so the copy ends up with exactly the same records as the source: this way we also drop the ones the onchange added on its own. One2manys are still left out, as the core does.

cc @Tecnativa TT64309

ping @victoralmau @pilarvargas-tecnativa

### DIFF
--- a/web_widget_one2many_tree_line_duplicate/static/src/list/list_renderer.esm.js
+++ b/web_widget_one2many_tree_line_duplicate/static/src/list/list_renderer.esm.js
@@ -5,6 +5,20 @@ import {ListRenderer} from "@web/views/list/list_renderer";
 import {patch} from "@web/core/utils/patch";
 import {exprToBoolean} from "@web/core/utils/strings";
 import {browser} from "@web/core/browser/browser";
+import {x2ManyCommands} from "@web/core/orm_service";
+
+const SCALAR_TYPES = [
+    "integer",
+    "float",
+    "monetary",
+    "char",
+    "text",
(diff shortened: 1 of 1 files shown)


web_widget_one2many_tree_line_duplicate/static/src/list/list_renderer.esm.js
```diff
@@ -69,17 +70,26 @@ patch(ListRenderer.prototype, {
...
+            } else if (relationalTypes.includes(field.type)) {
+                // Copy the value object so both records don't share it.
+                snapshot[name] = value && Object.assign({}, value);
+            } else if (field.type === "many2many") {
+                // SET (and not LINK) so the copy ends up with exactly the same
+                // records as the source: this way we also drop the ones the
+                // onchange added on its own.
+                snapshot[name] = [x2ManyCommands.set([...value.currentIds])];
             }
+            // One2many is left out: like the core does when duplicating, we
```
pedrobaeza: We need to copy one2many records, like with `sale_variant_configurator`. Or that's another thing?

pedrobaeza (approved):
Tested and working in a real instance, so let's fast-track it to fix the problem.

/ocabot merge patch

PR #3654, opened 2026-09-02
[18.0][FIX] web_responsive: prevent duplicate navigation on app menu click

### LINKED ISSUE #3653: [18.0] web_responsive: AppMenuItem bypasses unsaved changes check
## Module
web_responsive

## Describe the bug
`AppMenuItem.onClick()` does not call `ev.preventDefault()` on the `<a>` tag click. This causes the browser to follow the href directly, bypassing Odoo's `clearUncommittedChanges` check in the action manager.

Users can then navigate away from dirty forms with missing required fields without any warning, the form's `beforeLeave`/auto-save mechanism is never triggered.

This does not happen with standard navbar submenu items because `DropdownItem.onClick()` calls `ev.preventDefault()` before firing `onSelected`.

## To Reproduce
**Affected versions**:

1. Open any form view with required fields
2. Edit a field (make the form dirty) but leave required fields empty
3. Click the apps menu grid icon
4. Click any app tile
5. App freezes, full browser windows needed, unsaved changes are lost

**Expected behavior**
5. "Invalid fields" notification, navigation blocked.

Optionally, also check `clearUncommittedChanges` in `AppsMenu.onMenuClick` before opening the overlay, so the menu doesn't open at all when the form can't be saved.

Pepo48:
Closes: #3653

### DIFF
--- a/web_responsive/static/src/components/apps_menu_item/apps_menu_item.esm.js
+++ b/web_responsive/static/src/components/apps_menu_item/apps_menu_item.esm.js
@@ -31,7 +31,8 @@ export class AppMenuItem extends Component {
         this.webIconData = getWebIconData(nextProps.app);
     }
 
-    onClick() {
+    onClick(ev) {
+        ev.preventDefault();
         if (typeof this.props.onClick === "function") {
             this.props.onClick(this.props.app);
         }


pedrobaeza:
This was done on purpose, as we don't want to block the navigation to the apps menu if there are missing required fields or to auto-save. Only when navigating to another menu, this will be performed.

Pepo48:
@pedrobaeza thanks for the prompt reaction! 

As for the `clearUncommittedChanges` check, I acknowledged it the issue description that it is a UX preference. I'm fine with dropping that part, although I would still argue that it makes the behaviour unintuitive.

But the `ev.preventDefault()` in `AppMenuItem.onClick` is a separate issue. Without it, clicking an app tile triggers navigation twice, once from the click handler, once from the browser following the link. Normally that's harmless, but when the current form has unsaved changes that can't be saved, the two paths conflict and the app freezes. Only a full page refresh recovers it.

The fix is just preventing the browser from following the link, which is standard for links handled by JavaScript. It's the same pattern Odoo uses for its own menu items.

pedrobaeza:
OK, please keep only this second commit, rephrase the commit message accordingly, and remove `[18.0]` from it.

pedrobaeza (approved):
Thanks. Please add `[FIX]` to the commit message:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

Pepo48:
@pedrobaeza done.

tarteo (approved):
Confirmed, I was able to reproduce the bug and this fixes the issue 👍

pedrobaeza:
/ocabot merge patch

Please fw-port it to 19 if needed

PR #3663, opened 2026-09-12
[19.0][IMP] web_pwa_customize: apple-touch-icon + settings visibility

LadyHwesta:
## What's missing today

`web_pwa_customize` lets you configure a short name, colors, and icon for Odoo's "Install app" prompt - but only for `/web/manifest.webmanifest`, which Chrome, Edge, and Android read. iOS Safari's "Add to Home Screen" reads a separate `apple-touch-icon` link tag instead, which this module never touches - so an icon configured here never shows up there at all.

Its own settings also inherit an existing restriction: the `pwa_settings` block (which also holds core's own `web.web_app_name` field, itself not surfaced anywhere else) is hidden behind `base.group_no_one`, i.e. invisible unless developer mode is on.

## What this PR does

- Adds a `/web_pwa_customize/apple_touch_icon` route that finds the best available configured icon - preferring the 192x192 PNG variant this module already generates, falling through the other sizes, then an SVG upload, then Odoo's own stock artwork if nothing is configured - and points the `apple-touch-icon` tag at it via a one-attribute template inherit on `web.webclient_bootstrap`.
- Drops the `pwa_settings` block's `groups="base.group_no_one"` restriction. General Settings itself already requires admin access, so this doesn't expose anything that wasn't already admin-only - it just stops requiring developer mode on top of that.

No new dependencies, no changes to how the icon/colors are stored - purely additive on top of the existing `res.config.settings` fields and `ir.attachment`-based icon storage.

## Testing

5 new tests (apple-touch-icon fallback with nothing configured, redirect to a configured PNG, redirect to a configured SVG, the link tag itself, and settings-block visibility for a plain admin without `base.group_no_one`), all green alongside the 3 existing ones. Ran the full pre-commit suite (ruff, prettier, pylint-odoo, README regeneration) clean.

Happy to adjust the approach if maintainers would rather see this split differently (e.g. a separate module) or handled another way.


web_pwa_customize/views/res_config_settings_views.xml
```diff
@@ -5,34 +5,46 @@
...
-                        />
-                    </div>
-                </setting>
-                <setting string="Icon">
-                    <div class="w-50 row">
-                        <field name="pwa_icon" widget="image" class="oe_avatar" />
-                    </div>
-                </setting>
-            </block>
+            <data>
```
pedrobaeza: Why adding `<data>`?

PR #3664, opened 2026-09-14
[19.0][MIG] web_notify_upgrade: Migration to 19.0

AnizR:
Migration of web_notify_upgrade to version 19.0

Another PR for migrating this module exists but isn't working: https://github.com/OCA/web/pull/3441#discussion_r3966217413

# Changes
The notification is now using an action instead of a button (otherwise it didn't work):
image


web_notify_upgrade/models/ir_model.py
```diff
@@ -3,9 +3,9 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import logging
 
-from odoo import _, api, models
```
camille-spiritus-acs, line 6: You're using _ in the return of _get_upgrade_notification_params :)

camille-spiritus-acs (changes requested):
I only looked at your commits and apart from my comment it LGTM. Looks like you'll have to add a few test if I get the pipeline error correctly (never seen that one)
Good luck 💪

web_notify_upgrade/models/ir_model.py
AnizR: Yes, but I am using self.env._ :wink: 
So, the import can be removed

camille-spiritus-acs: my bad :woman_facepalming: sorry!

tobiaszehntner (approved):
LGTM

CristianoMafraJunior (approved):
LGTM

BhaveshHeliconia (approved):
LGTM!

lmignon (approved):
🥇

sbidoul:
@paradoxxxzero @bealdav good for you?

bealdav (approved):
Thanks a lot

PR #3665, opened 2026-09-14
[16.0][FIX] web_dialog_size: fetch the size configuration once per session

baguenth:
## Problem

Every `ExpandButton` — i.e. every dialog instance — queried
`get_web_dialog_size_config` over RPC in `setup()`. Flows that open many
dialogs (wizards, record pickers, chained forms) issue the same request again
and again.

## Fix

The configuration does usually not change during a session, so keep one shared
promise for all dialogs. A failed request resets the promise so the next
dialog retries.

Once configuration changes, it will be refetched on a browser refresh.


baguenth:
Hey @pedrobaeza,

kindly asking for a review :)

QuatschSalat (approved):

pedrobaeza (approved):
/ocabot merge patch

Issue #3334, opened 2025-10-17
[18.0][BUG] web_responsive pdf preview glitch on the side


david-justel:
## The bug is in the last important up
web_responsive.

## Describe the bug
With an updated Odoo 18 Community core (OCB) and only account module installed, if you attach a PDF to a supplier invoice, this PDF is preview displayed on the side when you enter in the invoice. But, in the moment, you install web_responsive module and re-enter the invoice, the preview is no longer displayed on the side.   
## To Reproduce

**Affected versions**:
web_responsive: 18.0.1.0.2

1. With an update and clean core from Odoo 18 Community (OCB) install account module
2. Create a supplier invoice and attach a PDF, the PDF is preview display in the left side os the screen 
3. Install web_responsive module
4. Re-enter the invoice created previously, the PDF preview is no longer displayed on the side.

pedrobaeza:
This is on purpose, as we don't want such automatic display that moves your chatter to down and that most of the times you open the bill, you don't want to preview the PDF. On contrary, if you want, you are only 2 clics away from previewing, which is far more convenient.

Closing as no bug, but intended behavior.

smaddlsoft:
Could we make this optional via a setting, as some (most?) of our customers really prefer the auto-opening of the invoice?

pedrobaeza:
If anyone is willing to contribute with the code to set that setting and to make it work, for me it's OK and welcome.

vancouver29:
@pedrobaeza @smaddlsoft FYI 

https://github.com/OCA/web/pull/3666

PR #3666, opened 2026-09-14
[18.0][IMP] add opt-in setting for automatic PDF invoice preview opening

vancouver29:
Improve the PDF preview in `web_responsive`Module

Improvements #3334 
See [this comment](https://github.com/OCA/web/issues/3334#issuecomment-3803640110) for context.


pedrobaeza (commented):
I think the option should be set per user, not per system.

web_responsive/__manifest__.py
```diff
@@ -16,7 +16,7 @@
     "Odoo Community Association (OCA)",
     "license": "LGPL-3",
     "installable": True,
-    "depends": ["web", "web_tour", "mail"],
+    "depends": ["web", "web_tour", "mail", "account"],
```
pedrobaeza: This change is a total blocker. You can't depend on account.

pedrobaeza (changes requested):

smaddlsoft:
> I think the option should be set per user, not per system.

I was agnostic about this and could see advantage for both, but lets go with your suggestion.
@vancouver29 kindly adjust

web_responsive/__manifest__.py
vancouver29: its changed!

smaddlsoft:
@vancouver29 I cant see this setting in the users preferences
image

Also, the screenshot in the readme is not working.

Best
Jan

vancouver29:
> @vancouver29 I cant see this setting in the users preferences image
> 
> Also, the screenshot in the readme is not working.
> 
> Best Jan

Hi Jan @smaddlsoft ,

you should change that option in settings module, not in profile references. Here is the click path: 

Settings (Module/App)  -> Users & Companies -> Users -> choose the User -> Preferences -> check/uncheck the checkbox in the File Viewer. 
I also documented the click path in ‘Usage.md’.

best regards

web_responsive/readme/USAGE.md
```diff
@@ -3,3 +3,13 @@ The following keyboard shortcuts are implemented:
 - Navigate app search results - Arrow keys
 - Choose app result - `Enter`
 - `Esc` to close app drawer
+
+## Auto-open invoice preview
+
+Enable *Auto-open Invoice Preview* in your user preferences
+(*Your Profile → Preferences*). When enabled, opening an invoice
```
smaddlsoft, line 10: currently its not in the users preferences which are accessible directly by the user, but only in the settings->users, which obviously is not accesible by each user. 

Kindly move to the user-accsessible user preference window, where also the chatter position is maintained, which is a logical place to put also this opt-in, because both are somewhat related.

Image

smaddlsoft (changes requested):
kindly move the setting to the user preferences accessible by each user, not via settings->users, which most users wont be able to access.

web_responsive/readme/USAGE.md
vancouver29: done!

jmveddeler (approved):
Looks good, from functional use perspective now. @pedrobaeza can you double check from code side and then we can merge? Thanks!

Image

web_responsive/models/res_users.py
```diff
@@ -32,6 +32,13 @@ class ResUsers(models.Model):
         readonly=False,
     )
 
+    is_auto_open_invoice_preview = fields.Boolean(
+        string="Auto-open Invoice Preview",
+        help="Automatically open the PDF preview of the invoice attachment "
+        "when the invoice is opened.",
+        default=True,
```
pedrobaeza: This should be defaulted to False for keeping current behavior.

pedrobaeza (commented):
I have tried on runboat, and it seems to not be working:

Image

pedrobaeza:
And please remove `[18.0]` from the commit message.

smaddlsoft:
@pedrobaeza @vancouver29 I had checked Vendor Bills, there it was working, maybe its only picking for Vendor Bills not Customer Invoices. Should work, imho, the same for both.

web_responsive/models/res_users.py
vancouver29: thanks for reviewing! Its changed back to False now.

vancouver29:
> And please remove `[18.0]` from the commit message.



> @pedrobaeza @vancouver29 I had checked Vendor Bills, there it was working, maybe its only picking for Vendor Bills not Customer Invoices. Should work, imho, the same for both.

I could not reproduce the bug at my side locally. @smaddlsoft could you ?

PR #3669, opened 2026-09-15
[19.0][FIX] web_dialog_size: preserve standard Dialog sizes

Raul-S73:
### Changes proposed in this PR:

**Preserve Odoo 19 standard `Dialog` sizes in `web_dialog_size`.**

The affected consumer is the Odoo Enterprise `mrp_workorder` module (MRP II),
specifically its `MrpWorksheetDialog` used by MRP Shop Floor. Its XML uses
the standard Odoo 19 value `size="fs"`.

The module currently replaces `Dialog.props.size.validate` with a validator
that only accepts `sm`, `md`, `lg`, `xl` and `dialog_full_screen`. Odoo 19
also defines `fs` and `fullscreen` as valid standard dialog sizes.

This causes an Owl validation error when the Enterprise worksheet is opened in
MRP Shop Floor with `debug=assets`:

```text
Invalid props for component 'Dialog': 'size' is not valid
```

The fix is intentionally made in `web_dialog_size`, not in the Enterprise
`mrp_workorder` consumer. It keeps Odoo's original validator and adds the custom
`dialog_full_screen` value provided by this module. This preserves both the
standard Odoo 19 contract and the module's maximize/restore behavior.

### How to test the changes

1. Install or upgrade `web_dialog_size` in an Odoo 19 test database with the
   Enterprise `mrp_workorder` module enabled.
2. Open MRP Shop Floor with `?debug=assets`.
3. Open a work order worksheet.
4. Verify that the worksheet dialog opens without an Owl props validation
   error, keeps its full-width layout, displays the worksheet content and can
   be closed normally.
5. Repeat the test without `debug=assets`.
6. Verify the standard dialog sizes `sm`, `md`, `lg`, `xl`, `fs` and
   `fullscreen`, as well as the module-specific `dialog_full_screen` value
   through the maximize/restore control.
7. Check standard dialogs such as form and select/create dialogs to ensure
   that their existing sizes remain valid.

### Technical notes

The validator is shared by all `Dialog` instances. Removing `size` from the
Enterprise worksheet XML or changing `fs` to `dialog_full_screen` would hide
the symptom but would make the standard MRP dialog depend on the OCA-specific
value and could change its standard `modal-fs` styling. The `fs` value belongs
to Odoo 19's standard `Dialog` contract; `mrp_workorder` is the affected
consumer, not the source of the incompatibility.

The functional tests above have not been run against production. A static
JavaScript syntax check was completed locally.

### Risks

The change affects the global `Dialog` prop validator, as the current module
already does. The implementation deliberately delegates to Odoo's original
validator so that future standard Odoo sizes are not duplicated and do not
become stale in this module.


pedrobaeza:
Testing in runboat, restoring the dialog to their previous size doesn't work (not sure if this was before this patch).

Reyes4711-S73 (approved):

Raul-S73:
@pedrobaeza Thanks for testing. 
The main purpose of this PR is to update the validation of the Dialog size values. 
The maximize/restore logic hasn’t been changed, so the restore issue seems unrelated to this PR and should be handled separately.

pedrobaeza (approved):
Indeed the problem is already present in the module before this PR.

Let's merge it, but I would appreciate that you check the other problem, as the module is working half.

/ocabot merge patch

PR #3672, opened 2026-09-17
[19.0][ADD] web_x2many_list_scroll: lists inside forms scroll within a set number of rows

alan196:
New module `web_x2many_list_scroll`.

A list of lines inside a form (order lines, invoice lines...) stretches the page as it grows: past a screen of rows the column titles and the row of totals are out of sight. With this module the list scrolls by itself within a set number of rows, in both directions, with its header and its totals staying in sight. The pager of the list keeps working as before.

- **Setting** in *General Settings ▸ Lists in Forms*: the visible rows (system parameter `web_x2many_list_scroll.rows`, `0` = off), read once per session.
- **Per view**: `options="{'scroll_rows': 15}"` on the x2many field wins over the setting; `0` switches it off for that list.
- The height is the number of rows times the row height and the header height **measured on the rendered list**, so it follows the font, the density and the device (rows are taller on touch screens).
- Inside a form Odoo flattens the list header to `z-index: 0`; the module lifts header and footer above the widgets of the rows rolling under them (tags, badges, toggles), tunable through `--o-x2many-list-scroll-zindex`.
- Tests: hoot (setting, per-view override, off with zero, measured heights, sticky header) and Python (parameter to session).

https://claude.ai/code/session_01U2vJWgfX8LRMaFnszPEhC5


UsmanGhias (commented):
Clean patch for OCA/web.

Observations:
- Code structure follows standard OCA guidelines and cleanly isolates the extension logic.

Ready for testing on standard environments.

Best,
Usman

